PR #52, opened 2015-07-18
Vc2010

alfredh:


alfredh:
thanks Victor for the nice patch, I have merged parts of it to baresip/master

- The copy of windows .h header files is not included
- no need to do ifdef MSC_VER for module init, better to use a portable style all over the place
- running ccheck.py should give Zero warnings:

PR #55, opened 2015-07-20
--add hold for previous call command(too hold active call when we hav…

AlexMarlo:
add hold for previous call command (too hold active call when we have incoming call)

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -286,7 +286,6 @@ static int options_command(struct re_printf *pf, void *arg)
 	return err;
 }
 
-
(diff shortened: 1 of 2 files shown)


alfredh:
thank you,

committed with some minor additional fixes:

fix compile warning:

```
src/ua.c:933:2: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
```


fix ccheck warning:

```
$ ccheck.py 
./src/ua.c:934: no single space after operator 'if()'
./src/ua.c:938: no single space after operator 'if()'
```

PR #58, opened 2015-08-01
Modules mk

alfredh:


alfredh:
adding support for $(MOD)_CFLAGS is something that I wanted to do for many years.

Please have a look, review and testing is welcome :)

Issue #65, opened 2015-08-14
Baresip build on Debian Wheezy fails | GTK related


github-ernst:
Relevant build log section:

 ```CC [M]  build-x86_64/modules/gtk/gtk_mod.o
modules/gtk/gtk_mod.c: In function ‘notify_incoming_call’:
modules/gtk/gtk_mod.c:271:2: error: unknown type name ‘GNotification’
modules/gtk/gtk_mod.c:278:2: warning: implicit declaration of function ‘g_notification_new’ [-Wimplicit-function-declaration]
modules/gtk/gtk_mod.c:278:2: warning: nested extern declaration of ‘g_notification_new’ [-Wnested-externs]
modules/gtk/gtk_mod.c:278:15: warning: assignment makes pointer from integer without a cast [enabled by default]
modules/gtk/gtk_mod.c:284:2: warning: implicit declaration of function ‘g_notification_set_urgent’ [-Wimplicit-function-declaration]
modules/gtk/gtk_mod.c:284:2: warning: nested extern declaration of ‘g_notification_set_urgent’ [-Wnested-externs]
modules/gtk/gtk_mod.c:288:2: warning: implicit declaration of function ‘g_notification_set_body’ [-Wimplicit-function-declaration]
modules/gtk/gtk_mod.c:288:2: warning: nested extern declaration of ‘g_notification_set_body’ [-Wnested-externs]
modules/gtk/gtk_mod.c:289:2: warning: implicit declaration of function ‘g_notification_add_button_with_target_value’ [-Wimplicit-function-declaration]
modules/gtk/gtk_mod.c:289:2: warning: nested extern declaration of ‘g_notification_add_button_with_target_value’ [-Wnested-externs]
modules/gtk/gtk_mod.c:293:2: warning: implicit declaration of function ‘g_application_send_notification’ [-Wimplicit-function-declaration]
modules/gtk/gtk_mod.c:293:2: warning: nested extern declaration of ‘g_application_send_notification’ [-Wnested-externs]
modules/gtk/gtk_mod.c: In function ‘denotify_incoming_call’:
modules/gtk/gtk_mod.c:305:2: warning: implicit declaration of function ‘g_application_withdraw_notification’ [-Wimplicit-function-declaration]
modules/gtk/gtk_mod.c:305:2: warning: nested extern declaration of ‘g_application_withdraw_notification’ [-Wnested-externs]
modules/gtk/gtk_mod.c: In function ‘message_handler’:
[...]

PR #73, opened 2015-09-20
More ways to answer calls in GTK+

clehner:
This adds support for old versions of glib that don't have have GNotification (#65), by using libnotify as a fallback. However, libnotify notifications can't have buttons and so can't be used for accepting and rejecting incoming calls. So a new submenu is added here for each current incoming call, so that the user can accept or reject a call using the menu instead of by using the notification.

incoming call submenu

### DIFF
--- a/mk/modules.mk
+++ b/mk/modules.mk
@@ -225,6 +225,10 @@ ifeq ($(OS),freebsd)
 MODULES   += dtmfio
 endif
 
+ifneq ($(USE_GTK),)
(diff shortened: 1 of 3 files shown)


alfredh:
looks good!

tested on Debian 7.8 and Testing

```
                  libgtk2.0    libglib2.0     libnotify

Debian 5.0        2.12.12      2.16.6         0.4.4

Debian 7.8        2.24.10-2    2.33.12        0.7.5

Debian Testing    2.24.28      2.44.1         0.7.6-2

```

PR #74, opened 2015-09-27
daala: upstream doesn't have ogg_packet structure any more

czarkoff:
Use `daala_packet` structure instead.


alfredh:
Many thanks :)

PR #80, opened 2015-10-09
gtk: don't run pkg-config on every operation

czarkoff:
Thankfully GNU make has `!=` operator.


alfredh:
thanks, merged!

what is the purpose of the != operator ? there are also other places in baresip makefiles
that call pkg-config, perhaps they should be updated to ?

czarkoff:
With `!=` you execute command only once and substitute its output afterwards. I'll check other `pkg-config` invocations.

alfredh:
sorry I had to revert this patch, it was causing build problems for many users

it does not seem to be working for Debian 7.0 ?

PR #81, opened 2015-10-10
Get rid of VAR += `pkg-config ...` idiom

czarkoff:
Output of `pkg-config` shouldn't change during build, so it is sufficient to run it once per check.

This pull request is follow-up to #80.

### DIFF
--- a/modules/directfb/module.mk
+++ b/modules/directfb/module.mk
@@ -7,7 +7,7 @@
 
 MOD                := directfb
 $(MOD)_SRCS        += directfb.c
-$(MOD)_LFLAGS      += `pkg-config --libs directfb `
-CFLAGS             += `pkg-config --cflags directfb `
+$(MOD)_LFLAGS      != pkg-config --libs directfb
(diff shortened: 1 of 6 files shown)


alfredh:
thanks a lot Dmitrij :)

alfredh:
this patch seems to break the build for at least Debian 7.0 and OSX 10.9
(GNU make 3.81)

czarkoff:
Sorry, I don't have such an old version of GNU make around. Can you show some output?

alfredh:
I tried a simple != snippet with GNU make 3.81 and it was not working.

when using ```remake``` I get this error:

```
$ remake
modules/rst/module.mk:11: *** missing separator.  Stop.
Command-line invocation:
	" none
```


is this an alternative:

https://www.gnu.org/software/make/manual/html_node/Setting.html

```
hash := $(shell printf '\043')
var := $(shell find . -name "*.c")
```

czarkoff:
This is an alternative, albeit ugly.

PR #83, opened 2015-10-16
build infrastructure: get rid of " = `pkg-config ...`" idiom

czarkoff:
Followup to #80 and #81. This time using `shell` macro.


czarkoff:
FWIW I tested on Wheezy. Builds. I am still not sure whether supporting debian oldstable is worth having these ugly `shell` macros.

czarkoff:
Added another commit because testing on Wheezy revealed that current header test is not sufficient. Also, `pkg-config --exists ...` is more readable then bunch of `[ -f ... ]`.

alfredh:
thank you :)

PR #95, opened 2015-12-13
Multi Call Echo Module - Experimental

sreimers:
Hi Alfred,

this module is a proof of concept for making concurrent and independent echo calls possible. Very helpful for testing. At the moment I have no way to test vidbridge, but this could be implemented really fast. Its a modified version of the b2bua module and requires aubridge.


alfredh:
thank you very much, Sebastian :)

PR #98, opened 2016-01-09
enable audiounit as default for OSX 10.11 and later

alfredh:


alfredh:
more testing and feedback from people is needed

PR #101, opened 2016-01-14
sndio: test whether par was allocated before freeing it

czarkoff:
I hit this bug when `sndiod` didn't start after system update, and my hardware rejected requested parameters. Most users should never need it, and this is not really a fix to the real issue, but IMO if someone tries baresip without `sndiod` on unfortunate hardware, it is better to report failure then to dump core.

### DIFF
--- a/modules/sndio/sndio.c
+++ b/modules/sndio/sndio.c
@@ -268,7 +268,9 @@ static int play_alloc(struct auplay_st **stp, const struct auplay *ap,
 		st->run = false;
 
  out:
-	mem_deref(par);
+	if (par)
+		mem_deref(par);
+
 	if (err)
 		mem_deref(st);
 	else


alfredh:
checking the parameter to mem_deref should not be necessary, it is already doing that
inside mem_deref.

I think this bug hints at a memory corruption, can you either try to run baresip with valgrind
or perhaps add many asserts to check for valid memory objects ?

czarkoff:
@alfredh:
1. For me baresip reliably crashed with sndio first time it tried to emit any sound.
2. Gdb shows that issue is with free() on `par`, which is filled with junk by free().
3. This patch reliably fixes the issue.
4. It shouldn't, and I am puzzled. I'll figure out and write back. Closing this PR for now.

alfredh:
I just reviewed the source code in sndio.c

the "struct sio_par" pointer declared in
https://github.com/alfredh/baresip/blob/master/modules/sndio/sndio.c#L147
is not initialized. 

if the case fails before prm is set, in:
https://github.com/alfredh/baresip/blob/master/modules/sndio/sndio.c#L167

then mem_deref will try to destroy an un-initialized pointer.


A solution would be to initialize "struct sio_par *prm = NULL", this applies for both
source and playback.

czarkoff:
This matches my observations. New PR will come up shortly.

PR #103, opened 2016-01-15
sndio: initialize "par" to NULL

czarkoff:
As discussed in #101, uninitialized use of `par` causes issues.


czarkoff:
OK, I've got it all wrong this time...

PR #109, opened 2016-01-21
menu: add switch audio device during call

AlexMarlo:


modules/menu/menu.c
```diff
@@ -544,6 +545,40 @@ static int hold_prev_call(struct re_printf *pf, void *unused)
...
+{
+	const struct cmd_arg *carg = arg;
+	static bool switch_aud_inprogress;
+	char *mod;
+	char *device;
+	int err = 0;
+
+	if (!switch_aud_inprogress && !carg->complete) {
+		statmode = STATMODE_OFF;
+		re_hprintf(pf, "\rPlease enter transfer target SIP uri:\n");
```
alfredh: the text does not match the expected user input, it should be something
like: "Please enter new audio device in format driver,device"

modules/menu/menu.c
```diff
@@ -544,6 +545,40 @@ static int hold_prev_call(struct re_printf *pf, void *unused)
...
+	}
+
+	switch_aud_inprogress = true;
```
alfredh: this variable is only set to true, and never set to false. It should be set
to false when the user input is completed.

modules/menu/menu.c
```diff
@@ -544,6 +545,40 @@ static int hold_prev_call(struct re_printf *pf, void *unused)
...
+static int switch_audio_dev(struct re_printf *pf, void *arg)
+{
+	const struct cmd_arg *carg = arg;
+	static bool switch_aud_inprogress;
+	char *mod;
+	char *device;
+	int err = 0;
+
+	if (!switch_aud_inprogress && !carg->complete) {
+		statmode = STATMODE_OFF;
```
alfredh: statmode is never updated after the user input is complete.

modules/menu/menu.c
```diff
@@ -544,6 +545,40 @@ static int hold_prev_call(struct re_printf *pf, void *unused)
...
+		re_hprintf(pf, "\rPlease enter transfer target SIP uri:\n");
+	}
+
+	switch_aud_inprogress = true;
+
+	if (carg->complete) {
+
+		mod = strtok( carg->prm, ",");
+		device = strtok( NULL, ",");
```
alfredh: as Dmitrij correctly pointed out, strtok() is not thread-safe and should
not be used in baresip. In baresip we prefer to use re_regex, example:

```
re_regex(carg->prm, str_len(carg->prm), "[^,]+,[~]*", &pl1, &pl2);
```

and then convert the pl1/pl2 to string.

modules/menu/menu.c
```diff
@@ -544,6 +545,40 @@ static int hold_prev_call(struct re_printf *pf, void *unused)
...
+		if( mod == NULL || device == NULL){
+			re_hprintf(pf, "\rFormat should be: alsa,default\n");
+			return 0;
+		}
+
+		err = audio_set_player(call_audio(ua_call(uag_cur())), mod, device);
+		if( err)
+			return err;
+
+		err = audio_set_source(call_audio(ua_call(uag_cur())), mod, device);
```
alfredh: instead of fetching the struct audio from uag_cur twice, you can move that
operation to the top so that it is only done once (minor detail)

modules/menu/menu.c
```diff
@@ -602,6 +637,7 @@ static const struct cmd callcmdv[] = {
 	{'r', CMD_IPRM,"Transfer call",       call_xfer             },
 	{'x',       0, "Call hold",           call_holdresume       },
 	{'H',       0, "Hold previous call",  hold_prev_call        },
+	{'A', CMD_IPRM,"Switch audio device", switch_audio_dev      },
```
alfredh: I think we have run out of letters now :)

the auloop module is using 'a' and 'A'...

as Doug has pointed out we should now add support for an additional
set of commands with long strings, in addition to the current one.

alfredh:
one more comment about source code formatting, baresip is using the tool ccheck.py to
verify that source code format:

it should give Zero warnings and errors.

with this patch, I get some warnings:


```
./modules/menu/menu.c:567: no single space after operator 'if()'
./modules/menu/menu.c:572: line is too wide (84 - max 79)
./modules/menu/menu.c:573: no single space after operator 'if()'
./modules/menu/menu.c:576: line is too wide (84 - max 79)
./modules/menu/menu.c:577: no single space after operator 'if()'
```

AlexMarlo:
I made some fixes and waiting for multichar commands support

alfredh:
Hi AlexMarlo,

thanks for your patch, I have made some minor improvements and commited it here:

https://github.com/alfredh/baresip/commit/ee1719f51d31eecf061b7b126782963fb5e020ca

PR #125, opened 2016-04-23
Added support for MPA coding

choene:
I need it for interop testing with old broadcast audiocodecs.


modules/mpa/decode.c
```diff
@@ -0,0 +1,182 @@
...
+
+
+int mpa_decode_frm(struct audec_state *ads, int16_t *sampv, size_t *sampc,
+		    const uint8_t *buf, size_t len)
+{
+	int mpaerr, channels, encoding, i;
+	long samplerate,res;
+	size_t n;
+	uint32_t header;
+	static int16_t ds[2304];
```
alfredh: the ds buffer is static, meaning multiple instances of mpa will not work. You should either declare
it on the stack or move it to the audec_state.

modules/mpa/encode.c
```diff
@@ -0,0 +1,161 @@
...
+	debug("mpa: encoder destroyed\n");
+#endif
+}
+
+int mpa_encode_update(struct auenc_state **aesp, const struct aucodec *ac,
+		       struct auenc_param *param, const char *fmtp)
+{
+	struct auenc_state *aes;
+	struct mpa_param prm;
+	const struct aucodec *auc = aucodec_find("MPA", 90000, 1);
```
alfredh: this should not be necessary, the same object is available in the ``ac`` pointer.

modules/mpa/encode.c
```diff
@@ -0,0 +1,161 @@
...
+						prm.mode == JOINT_STEREO ? TWOLAME_JOINT_STEREO :
+						prm.mode == STEREO ? TWOLAME_STEREO : TWOLAME_AUTO_MODE);
+	mpares |= twolame_set_version(aes->enc, prm.samplerate < 32000 ? TWOLAME_MPEG2 : TWOLAME_MPEG1);
+	mpares |= twolame_set_bitrate(aes->enc, prm.bitrate/1000);
+	mpares |= twolame_set_in_samplerate(aes->enc, prm.samplerate);
+	mpares |= twolame_set_out_samplerate(aes->enc, prm.samplerate);
+	mpares |= twolame_set_num_channels(aes->enc, 2);
+	if(mpares!=0) {
+		error("mpa: encoder set failed\n");
+		return EINVAL;
```
alfredh: if this fails it will lead to a memory leak of the state object. You should instead
set ``err`` to the error value and then goto out. In the ``out`` label you must handle
both success and error.

(same with other return statements further below).

modules/mpa/mpa.c
```diff
@@ -0,0 +1,194 @@
...
+   Published specification: RFC 3551
+
+   Applications which use this media type:
+        Audio and video streaming and conferencing tools.
+
+*/
+
+
+static struct aucodec mpa = {
+	.pt        = NULL,
```
alfredh: I guess we can use static payload type 14 here?

https://tools.ietf.org/html/rfc3551#section-6

modules/mpa/mpa.h
```diff
@@ -0,0 +1,32 @@
...
+int mpa_encode_frm(struct auenc_state *aes, uint8_t *buf, size_t *len,
+		    const int16_t *sampv, size_t sampc);
+
+
+/* Decode */
+int mpa_decode_update(struct audec_state **adsp, const struct aucodec *ac,
+		       const char *fmtp);
+int mpa_decode_frm(struct audec_state *ads, int16_t *sampv, size_t *sampc,
+		    const uint8_t *buf, size_t len);
+int mpa_decode_pkloss(struct audec_state *st, int16_t *sampv, size_t *sampc);
```
alfredh: is the "pkloss" prototype used?

alfredh:
Thank you for this very nice patch :)

I tested the mpa module briefly on Debian Testing locally and it works pretty nicely.
Both encoder and decoder.

could you please run ``ccheck.py`` on the source code and make sure it gives zero warnings:

choene:
Thank you for the fast code review...

alfredh:
Many thanks for your contributions. The patch was merged to master.

PR #126, opened 2016-05-01
aucodec: split srate into srate and crate (Clock Rate)

alfredh:
All, @choene please review and test this patch.

The patch splits the "srate" parameter of struct aucodec into 2 parts; srate and crate.
The srate is the Audio sample-rate for the DSP, the crate is the RTP Clock rate.

Most audio codecs will use exactly the same value here. The exception is e.g G.722 and MPA.
For example G.722 has 8000 Hz and 16000 Hz:

https://tools.ietf.org/html/rfc3551#section-4.5.2

```
   Even though the actual sampling rate for G.722 audio is 16,000 Hz,
   the RTP clock rate for the G722 payload format is 8,000 Hz because
   that value was erroneously assigned in RFC 1890 and must remain
   unchanged for backward compatibility.  The octet rate or sample-pair
   rate is 8,000 Hz.
```

the goal is to make audio.c agnostic to these codec details.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -737,7 +737,8 @@ struct aucodec {
 	struct le le;
 	const char *pt;
 	const char *name;
-	uint32_t srate;
+	uint32_t srate;             /* Audio samplerate */
+	uint32_t crate;             /* RTP Clock rate   */
 	uint8_t ch;
 	const char *fmtp;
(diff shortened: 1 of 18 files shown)


choene:
Looks good.
Tomorrow, I will test it.

PR #127, opened 2016-05-06
Some minor issues

choene:


alfredh:
Hey, this patch is based on a branch which is not "master"..

could you please submit a new PR that is based on "master" ... ?

czarkoff:
FWIW `result2` doesn't look right.  Maybe checking `result` every time it is set would be a better idea?

choene:
Yeah,

 

actually, it would be better to do the result checking first, and then in a second step the resampling.

 

I’m working on it…

 

 

Von: Dmitrij D. Czarkoff [mailto:notifications@github.com] 
Gesendet: Dienstag, 10. Mai 2016 08:57
An: alfredh/baresip <baresip@noreply.github.com>
Cc: choene <christian.hoene@symonics.com>; Author <author@noreply.github.com>
Betreff: Re: [alfredh/baresip] Some minor issues (#127)

 

FWIW result2 doesn't look right. Maybe checking result every time it is set would be a better idea?

Issue #122, opened 2016-04-11
 gtk module compile error on Centos 6


sebd77:
Hi! Regarding the issue "gtk module compile error on OpenBSD 5.7 #53 ", I do have a very similar issue over Centos6. I could install properly re-0.4.15 and rem-0.4.7, however when I try to install baresip-0.4.18, the make returns the following error:
[root@t420sdlinux baresip-0.4.18]# make
CC [M] build-i686/modules/gtk/gtk_mod.o
modules/gtk/gtk_mod.c:71: error: expected â=â, â,â, â;â, âasmâ or âattributeâ before âapp_entriesâ
modules/gtk/gtk_mod.c:74: warning: ISO C does not allow extra â;â outside of a function
modules/gtk/gtk_mod.c: In function âget_call_from_gvariantâ:
modules/gtk/gtk_mod.c:84: warning: cast to pointer from integer of different size
modules/gtk/gtk_mod.c: In function ânotify_incoming_callâ:
modules/gtk/gtk_mod.c:314: error: too few arguments to function ânotify_notification_newâ
modules/gtk/gtk_mod.c: In function âmessage_handlerâ:
modules/gtk/gtk_mod.c:575: error: too few arguments to function ânotify_notification_newâ
modules/gtk/gtk_mod.c: In function âgtk_threadâ:
modules/gtk/gtk_mod.c:830: warning: implicit declaration of function âg_action_map_add_action_entriesâ
modules/gtk/gtk_mod.c:830: warning: nested extern declaration of âg_action_map_add_action_entriesâ
modules/gtk/gtk_mod.c:830: warning: implicit declaration of function âG_ACTION_MAPâ
modules/gtk/gtk_mod.c:830: warning: nested extern declaration of âG_ACTION_MAPâ
modules/gtk/gtk_mod.c:831: error: âapp_entriesâ undeclared (first use in this function)
modules/gtk/gtk_mod.c:831: error: (Each undeclared identifier is reported only once
modules/gtk/gtk_mod.c:831: error: for each function it appears in.)
make: *** [build-i686/modules/gtk/gtk_mod.o] Error 1
[root@t420sdlinux baresip-0.4.18]#
would you have idea? Not sure which package is missing? Thanks!
@sebd77

Attach files by dragging & dropping or selecting them.
Styling with Markdown is supported

clehner:
[...]

PR #128, opened 2016-05-09
Build gtk module only with new enough glib

clehner:
GLib 2.32 is needed for some things: https://developer.gnome.org/gio/stable/api-index-2-32.html

#122


alfredh:
thanks, I tested this quickly on CentOS 6 and the gtk.so module was not compiled (as expected.)

PR #129, opened 2016-05-12
Fixing some issues with the MPA codec

choene:
I improve the MPA codec module.

Fixes where:
1) Error in SDP handling
2) Memory leaks
3) Sending of empty RTP packets

In addition, I did some cleanups.

One issue remains, the RTP timestamp are not set correctly (they are incremented by 20ms), which is not correct. But I do not have an idea on how to change the code to query the MPA module about the actual and current frame duration

### DIFF
--- a/modules/mpa/decode.c
+++ b/modules/mpa/decode.c
@@ -24,7 +24,7 @@ static void destructor(void *arg)
 {
 	struct audec_state *ads = arg;
 
-	if(ads->resampler)
+	if (ads->resampler)
 		speex_resampler_destroy(ads->resampler);
 
 	mpg123_close(ads->dec);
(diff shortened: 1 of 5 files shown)


alfredh:
check the encode_rtp_send function in audio.c:

```
	/* Convert from audio samplerate to RTP clockrate */
	sampc_rtp = sampc * tx->ac->crate / tx->ac->srate;

	/* The RTP clock rate used for generating the RTP timestamp is
	 * independent of the number of channels and the encoding
	 */
	frame_size = sampc_rtp / get_ch(tx->ac);

	tx->ts += (uint32_t)frame_size;
```

for MPA is 1 or 2 channels no? perhaps the get_ch() logic needs review.
At least the sampc_rtc is calculated from ac->crate which is 90000
for MPA.

PS: This PR has conflicts, can you try to rebase from master@HEAD ?

PR #131, opened 2016-05-19
Bug at stereo wih 16 and 24k

choene:
one buffer was too small


alfredh:
some new ccheck warnings must be fixed before merging:

```
$ ccheck.py 
./modules/mpa/encode.c:162: has trailing space(s)
./modules/mpa/encode.c:163: has trailing space(s)
./modules/mpa/encode.c:183: line is too wide (83 - max 79)
```

apart from that, it looks good.

PR #137, opened 2016-06-03
Use PATH_MAX for all buffers that hold filesystem paths

czarkoff:
Mere numbers in buffer definitions don't communicate the purpose of the buffers to readers of the source code.  Also, this would help in weird environments where certain paths may happen to be longer or shorter then expected.

As a bit of context:  I would like to introduce more macros for specific limits baresip has to deal with.  I think that readability, correctness and maintainability are improved when similar limits are set centrally, as opposed to `buf[number]` form.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,5 +14,5 @@ install:
     - wget "https://github.com/alfredh/pytools/raw/master/ccheck.py"
 
 script: 
-    - make test
+    - make V=1 test
     - python2 ccheck.py
--- a/Makefile
+++ b/Makefile
@@ -221,7 +221,7 @@ $(TEST_BIN):	$(STATICLIB) $(TEST_OBJS)
 
 $(BUILD)/%.o: %.c $(BUILD) Makefile $(APP_MK)
 	@echo "  CC      $@"
-	$(HIDE)$(CC) $(CFLAGS) -c $< -o $@ $(DFLAGS)
+	$(HIDE)$(CC) $(CPPFLAGS) $(CFLAGS) -c $< -o $@ $(DFLAGS)
 
 $(BUILD)/%.o: %.cpp $(BUILD) Makefile $(APP_MK)
 	@echo "  CXX     $@"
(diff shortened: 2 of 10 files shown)


alfredh:
I tried to compile this PR on Linux (Debian Testing) but it fails:

```
$ make test
  CC      build-x86_64/src/account.o
  CC      build-x86_64/src/aucodec.o
  CC      build-x86_64/src/audio.o
  CC      build-x86_64/src/aufilt.o
  CC      build-x86_64/src/auplay.o
  CC      build-x86_64/src/ausrc.o
  CC      build-x86_64/src/call.o
  CC      build-x86_64/src/cmd.o
  CC      build-x86_64/src/conf.o
  CC      build-x86_64/src/config.o
  CC      build-x86_64/src/contact.o
  CC      build-x86_64/src/log.o
  CC      build-x86_64/src/menc.o
  CC      build-x86_64/src/message.o
  CC      build-x86_64/src/metric.o
  CC      build-x86_64/src/mnat.o
  CC      build-x86_64/src/module.o
src/module.c: In function ‘load_module’:
src/module.c:58:12: error: ‘PATH_MAX’ undeclared (first use in this function)
  char file[PATH_MAX];
            ^
src/module.c:58:12: note: each undeclared identifier is reported only once for each function it appears in
src/module.c:58:7: warning: unused variable ‘file’ [-Wunused-variable]
  char file[PATH_MAX];
       ^
Makefile:221: recipe for target 'build-x86_64/src/module.o' failed
make: *** [build-x86_64/src/module.o] Error 1
```

is PATH_MAX part of the Posix API and is it always defined by including <limits.h> ?


http://stackoverflow.com/questions/9449241/where-is-path-max-defined-in-linux

czarkoff:
Well, according to POSIX `PATH_MAX` should be defined in `<limits.h>`.  I guess there is some convoluted way to force GNU libc respect POSIX in this case as well, but I have no machines with GNU libc around to find it out, and GNU documentation quality is...

I've seen people working this around like this:
```
#ifndef PATH_MAX
#define PATH_MAX _POSIX_PATH_MAX
#endif
```
Which may make sense because `_POSIX_PATH_MAX` is what every POSIX-compliant system must provide.  This should fit into PATH_MAX on every platform, so functions like `realpath(3)` won't blow up if fed with such buffer.

czarkoff:
The question is where to put it.  Including this in every file with path buffer does not seem reasonable, and baresip does not have "util.h"-style header.

alfredh:
the core.h header file can be used for this, and is included by all .c files in baresip core.
but no modules can access this header file.

I found this code in libapr which looks nice:

```
#if defined(PATH_MAX)
#define APR_PATH_MAX       PATH_MAX
#elif defined(_POSIX_PATH_MAX)
#define APR_PATH_MAX       _POSIX_PATH_MAX
#else
#error no decision has been made on APR_PATH_MAX for your platform
#endif
```

czarkoff:
I think this is an example of how not to do it.  Compare:
```
src/conf.c:158:11: error: no decision has been made on APR_PATH_MAX for your platform
```
and
```
src/conf.c:158:11: error: ‘PATH_MAX’ undeclared (first use in this function)
```
The former says nothing about the problem.  What decision has not been made?  Why?  How it was supposed to be made?  The latter provides just enough information to be able to find the problem quickly.

I think the proper solution here is to use `_POSIX_PATH_MAX` directly.  Does GNU libc define `_POSIX_PATH_MAX` in `limits.h`?

PR #143, opened 2016-06-12
Use _POSIX_PATH_MAX for all buffers that hold filesystem paths

czarkoff:
`_POSIX_PATH_MAX` is supposed to be supported over all filesystems on
POSIX-compliant system.  That should include network-attached filesystems and
other locations that are may not respect host OS views on path limitations.

This is a prettified version of PR #137.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,9 @@ CXXFLAGS  += -I$(LIBREM_PATH)/include
 CXXFLAGS  += -I$(SYSROOT)/local/include/rem -I$(SYSROOT)/include/rem
 CXXFLAGS  += $(EXTRA_CXXFLAGS)
 
+# XXX: GNU libc
+CPPFLAGS += -D_XOPEN_SOURCE=500
+
 # XXX: common for C/C++
 CPPFLAGS += -DHAVE_INTTYPES_H
 
--- a/Makefile
+++ b/Makefile
@@ -219,7 +219,7 @@ $(TEST_BIN):	$(STATICLIB) $(TEST_OBJS)
 
 $(BUILD)/%.o: %.c $(BUILD) Makefile $(APP_MK)
 	@echo "  CC      $@"
-	$(HIDE)$(CC) $(CFLAGS) -c $< -o $@ $(DFLAGS)
+	$(HIDE)$(CC) $(CPPFLAGS) $(CFLAGS) -c $< -o $@ $(DFLAGS)
 
 $(BUILD)/%.o: %.cpp $(BUILD) Makefile $(APP_MK)
 	@echo "  CXX     $@"
@@ -227,11 +227,11 @@ $(BUILD)/%.o: %.cpp $(BUILD) Makefile $(APP_MK)
 
 $(BUILD)/%.o: %.m $(BUILD) Makefile $(APP_MK)
 	@echo "  OC      $@"
-	$(HIDE)$(CC) $(CFLAGS) $(OBJCFLAGS) -c $< -o $@ $(DFLAGS)
+	$(HIDE)$(CC) $(CPPFLAGS) $(CFLAGS) $(OBJCFLAGS) -c $< -o $@ $(DFLAGS)
 
 $(BUILD)/%.o: %.S $(BUILD) Makefile $(APP_MK)
 	@echo "  AS      $@"
-	$(HIDE)$(CC) $(CFLAGS) -c $< -o $@ $(DFLAGS)
+	$(HIDE)$(CC) $(CPPFLAGS) $(CFLAGS) -c $< -o $@ $(DFLAGS)
 
 $(BUILD): Makefile
 	@mkdir -p $(BUILD)/src $(MOD_BLD) $(BUILD)/test/mock $(BUILD)/test/sip
--- a/mk/mod.mk
+++ b/mk/mod.mk
@@ -33,36 +33,36 @@ ifeq ($(STATIC),)
 
 $(MOD)$(MOD_SUFFIX): $($(MOD)_OBJS)
 	@echo "  LD [M]  $@"
-	$(HIDE)$(LD) $(LFLAGS) $(SH_LFLAGS) $(MOD_LFLAGS) \
-		$($(basename $@)_OBJS) \
-		$($(basename $@)_LFLAGS) -L$(LIBRE_SO) -lre -o $@
+	$(HIDE)$(LD) $(LFLAGS) $(SH_LFLAGS) $(MOD_LFLAGS) $($(MOD)_OBJS) \
+		$($(MOD)_LFLAGS) -L$(LIBRE_SO) -lre -o $@
 
 $(BUILD)/modules/$(MOD)/%.o: modules/$(MOD)/%.c $(BUILD) Makefile mk/mod.mk \
 				modules/$(MOD)/module.mk mk/modules.mk
 	@echo "  CC [M]  $@"
 	@mkdir -p $(dir $@)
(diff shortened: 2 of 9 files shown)


mk/mod.mk
```diff
@@ -33,36 +33,36 @@ ifeq ($(STATIC),)
 
 $(MOD)$(MOD_SUFFIX): $($(MOD)_OBJS)
 	@echo "  LD [M]  $@"
-	$(HIDE)$(LD) $(LFLAGS) $(SH_LFLAGS) $(MOD_LFLAGS) \
-		$($(basename $@)_OBJS) \
-		$($(basename $@)_LFLAGS) -L$(LIBRE_SO) -lre -o $@
+	$(HIDE)$(LD) $(LFLAGS) $(SH_LFLAGS) $(MOD_LFLAGS) $($(MOD)_OBJS) \
+		$($(MOD)_LFLAGS) -L$(LIBRE_SO) -lre -o $@
```
alfredh: what is the reason for changing the `basename $@` parts here, is it related to this patch ?

mk/mod.mk
```diff
@@ -33,36 +33,36 @@ ifeq ($(STATIC),)
...
 
 $(BUILD)/modules/$(MOD)/%.o: modules/$(MOD)/%.c $(BUILD) Makefile mk/mod.mk \
 				modules/$(MOD)/module.mk mk/modules.mk
 	@echo "  CC [M]  $@"
 	@mkdir -p $(dir $@)
-	$(HIDE)$(CC) $(CFLAGS) $($(call modulename,$@)_CFLAGS) \
+	$(HIDE)$(CC) $(CPPFLAGS) $(CFLAGS) $($(MOD)_CFLAGS) \
```
alfredh: what is the reason for changing `call modulename` ?

Makefile
```diff
@@ -55,6 +55,9 @@ CXXFLAGS  += -I$(LIBREM_PATH)/include
 CXXFLAGS  += -I$(SYSROOT)/local/include/rem -I$(SYSROOT)/include/rem
 CXXFLAGS  += $(EXTRA_CXXFLAGS)
 
+# XXX: GNU libc
+CPPFLAGS += -D_XOPEN_SOURCE=500
+
```
alfredh: I am not sure if I like this, it is quite an intrusive change? basically it means that
all .c files are now compiled with different flags.

Would it be possible to #define this before including limits.h instead?

alfredh:
many thanks for your patch.

I tried to compile this on OSX and it fails:

```
make test

  CC      build-x86_64/src/realtime.o
In file included from src/realtime.c:10:
In file included from /usr/include/sys/sysctl.h:83:
/usr/include/sys/ucred.h:91:2: error: unknown type name 'u_long'
        u_long  cr_ref;                 /* reference count */
        ^
/usr/include/sys/ucred.h:133:9: error: unknown type name 'u_int'
        u_int   cr_version;             /* structure layout version */
        ^
In file included from src/realtime.c:10:
In file included from /usr/include/sys/sysctl.h:84:
/usr/include/sys/proc.h:111:2: error: unknown type name 'u_int'
        u_int   p_estcpu;        /* Time averaged value of p_cpticks. */
        ^
/usr/include/sys/proc.h:116:2: error: unknown type name 'u_int'
        u_int   p_swtime;        /* Time swapped in or out. */
        ^
/usr/include/sys/proc.h:117:2: error: unknown type name 'u_int'
        u_int   p_slptime;       /* Time since last blocked. */
        ^
/usr/include/sys/proc.h:131:2: error: unknown type name 'u_char'; did you mean 'char'?
        u_char  p_priority;     /* Process priority. */
        ^
/usr/include/sys/proc.h:132:2: error: unknown type name 'u_char'; did you mean 'char'?
        u_char  p_usrpri;       /* User-priority based on p_cpu and p_nice. */
        ^
/usr/include/sys/proc.h:137:2: error: unknown type name 'u_short'; did you mean 'short'?
        u_short p_xstat;        /* Exit status for wait; also stop signal. */
        ^
/usr/include/sys/proc.h:138:2: error: unknown type name 'u_short'; did you mean 'short'?
        u_short p_acflag;       /* Accounting flags. */
        ^
9 errors generated.
make: *** [build-x86_64/src/realtime.o] Error 1

```

```
make
  CC [M]  build-x86_64/modules/stun/stun.o
  LD [M]  stun.so
clang: error: no such file or directory: 'build-x86_64/modules/x11grab/x11grab.o'
make: *** [stun.so] Error 1

```

my system:

```
OSX 10.9

GNU Make 3.81

Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 6.0 (clang-600.0.56) (based on LLVM 3.5svn)
Target: x86_64-apple-darwin13.4.0
Thread model: posix
```


A bit question comes up; what happens on platforms that does not define
_POSIX_PATH_MAX? Baresip should be portable and work on most systems
with posix api.



I also tried to compile it using Mingw32:

```
  CC      build-i386/src/conf.o
src/conf.c: In function `conf_path_get':
src/conf.c:158: error: `_POSIX_PATH_MAX' undeclared (first use in this function)
src/conf.c:158: error: (Each undeclared identifier is reported only once
src/conf.c:158: error: for each function it appears in.)
src/conf.c:158: warning: unused variable `buf'
src/conf.c: In function `conf_configure':
src/conf.c:299: error: `_POSIX_PATH_MAX' undeclared (first use in this function)
src/conf.c:299: warning: unused variable `path'
src/conf.c:299: warning: unused variable `file'
make[1]: *** [build-i386/src/conf.o] Error 1
make: *** [baresip] Error 2
```

czarkoff:
POSIX requires `_POSIX_PATH_MAX`, so it must be present everywhere.  I am not sure about MinGW, it might require some work.  I have no access to Windows machines ATM, and MinGW's documentation is lacking at best...

Provided that baresip targets POSIX, I think that `_XOPEN_SOURCE` should be defined globally, probably in `re/mk/re.mk`.  Apparently on OS X `_DARWIN_C_SOURCE` is required after `_XOPEN_SOURCE`.

mk/mod.mk
czarkoff: I was trying to fit `${CPPFLAGS}` there, and it didn't go nicely.  I agree, this is unrelated change. but both `$(call modulename,$@)` and `$(MOD)` resolved to the same construct, and latter is both more readable and less unportable.

alfredh:
hey @czarkoff 

how should we proceed on this one ?


I have defined a value in core.h that we can use for path names
(only in core for now, lets do the modules later):

```
/* max bytes in pathname */
#define FS_PATH_MAX 256
```



what do you think about using your PATH_MAX logic here ?


then we only need to change the logic in one place, and it would be something
like:

```
ifdef PATH_MAX
define FS_PATH_MAX PATH_MAX
elif defined POSIX_PATH_MAX
define FS_PATH_MAX POSIX_PATH_MAX
else
define FS_PATH_MAX 512
endif

```

if you agree with this kind of logic, can you please make a new PR for this?
I can help with testing.

czarkoff:
I would rather prefer to do it the other way around:
```
#ifndef PATH_MAX
# if defined(_POSIX_PATH_MAX)
#  define PATH_MAX _POSIX_PATH_MAX
# elif defined(_XOPEN_PATH_MAX)
#  define PATH_MAX _XOPEN_PATH_MAX
# else
#  define PATH_MAX 512
#endif
```
The benefit of this solution is that arrays take form of well-known `PATH_MAX`, so people don't need to navigate through headers to find the actual value.

That said, I am absolutely fine with your approach, so if you find it preferable for some reason, I'll use it.

PR #153, opened 2016-09-25
sndio: use driver-suggested buffer size

czarkoff:
Sndio provides value of buffer size (`par->bufsz`) that is equal to audio delay
in audio driver.  Using buffer under this size causes underruns without reducing delay.  As I get it, using `par->bufsz` for buffer size ensures minimal delay while saving audio driver from additional buffer management.


alfredh:
thank you very much :)

PR #157, opened 2016-10-14
avformat network initialization

koman-maciej:
Due to warning message:

`Using network protocols without global network initialization. Please use avformat_network_init(), this will become mandatory later.`

(I've got following warning when I use avformat library as video RTSP source - SIP call initialization)


alfredh:
many thanks for your patch. Could you please add an #if version check for the new
call to avformat_network_inti() ?

This is nice so that we can build baresip with older versions of ffmpeg.

koman-maciej:
Yes, you're right. I see now, the `avformat_network_deinit()` should be also invoked

PR #165, opened 2016-10-30
Use PATH_MAX for sizes of buffers that are supposed to hold paths

czarkoff:
This is an implementation of my approach.  After some thinking over this issue I decided that it makes sense to define a set of macros that baresip modules and downstream projects may rely.  As of this PR this set only includes `PATH_MAX`, although in future appropriate macros for URI length, error message strings length, etc. should be added.

I included `_XOPEN_SOURCE` and `_DARWIN_C_SOURCE` macros in files that end up using `PATH_MAX` in order to unhide `_POSIX_PATH_MAX` on Linux and macOS, although I still believe that `_XOPEN_SOURCE` should be defined via `CFLAGS` for all targets.  I guess it is a subject for further investigation.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1111,6 +1111,22 @@ struct commands *baresip_commands(void);
 struct player *baresip_player(void);
 
 
+/*
+ * Macros for sizes of buffers holding specific types of data
+ */
+
+/* PATH_MAX */
+#ifndef PATH_MAX
+# if defined(_POSIX_PATH_MAX)
+#  define PATH_MAX _POSIX_PATH_MAX
+# elif defined(_XOPEN_PATH_MAX)
+#  define PATH_MAX _XOPEN_PATH_MAX
+# else
+#  define PATH_MAX 512
+# endif
+#endif
+
+
 #ifdef __cplusplus
 }
 #endif
--- a/modules/account/account.c
+++ b/modules/account/account.c
@@ -3,6 +3,9 @@
  *
  * Copyright (C) 2010 - 2015 Creytiv.com
  */
+#define _XOPEN_SOURCE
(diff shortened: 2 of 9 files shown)


alfredh:
many thanks @czarkoff 

I would like to test out these changes on baresip "core" first, i.e. all .c files
that are under src/.

(let us do the modules later, once we know that which approach works).

I have made a small patch that only updates the FS_PATH_MAX value
in `core.h`:

https://github.com/alfredh/baresip/tree/path_max

what do you think?

czarkoff:
I think this should work.  You might want to add another check for `_XOPEN_PATH_MAX` (PATH_MAX in X/Open System Interfaces context), which seems to be a trendy thing in POSIX.  It is also supposed to be defined in `limits.h`.  I am not sure that there is any practical benefit from adding that though: to my knowledge all reasonably XSI-compliant systems provide `_POSIX_*` interfaces.

But anyway, as it is in this branch, this should work just fine.

Now, I am not sure how to make my point without sounding pressing, but I indeed think that ensuring availability of `PATH_MAX` to modules and making it a part of baresip's public interface is a better idea, as well as using well-known macro instead of local one.  I understand that it is not my call, but in my opinion my changes will be more beneficial to baresip in a long run.

alfredh:
I have merged the branch to master. The next step should be testing on known platforms.

the macro `FS_PATH_MAX` is available for baresip core only for now (defined in core.h)
and once we have proven that this approach works and is portable, we can consider
moving the macro to the public interface in `baresip.h`.

alfredh:
I will close this one for now, let us do some more testing for a couple of weeks.
if it works well, I am happy to commit a new PR to extend this principle (small patches please)

Thanks :)

Issue #173, opened 2016-11-26
Release baresip version 0.5.0


alfredh:
Documentation:
* update ChangeLog: Done

Brief testing so far:

* Android NDK r13: OK
* Arch Linux: OK
* CentOS 7: OK
* Debian 6.0 (i386): OK
* Debian 8.0 (i386): OK
* Fedora 24: OK
* Mac OSX 10.6: OK
* Minwgw: OK
* OpenWRT CHAOS CALMER (15.05.1, r48532): OK
* Ubuntu 14.04: OK
* Ubuntu 16.04: OK

czarkoff:
@alfredh, looks like I'll open a couple of pull request that I would prefer to be reviewed before 0.5.0 is cut.

czarkoff:
Looks like I am done with testing.  With an exception below everything seems fine on OpenBSD-current.

I experience an issue when using libv4l2: baresip picks `YU12` (`V4L2_PIX_FMT_YUV420`), and my video is distorted (topmost line is just repeated over and over).  I blame libv4l2 which is very buggy here (eg. it hides all native formats of my webcam), but I would appreciate if you could make a test build of baresip which forces `YU12` and see whether your image is distorted.

alfredh:
version 0.5.0 is now released:

https://github.com/alfredh/baresip/releases/tag/v0.5.0

Announcement email:

http://lists.creytiv.com/pipermail/re-devel/2016-December/001154.html

PR #175, opened 2016-11-27
config: list several modules that were missing

czarkoff:
Add missing modules to configuration file generator.  I did not add deprecated and obscure modules.

(*Referencing PR #173 because it would make sense to merge this before release.*)

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -658,7 +658,6 @@ int config_write_template(const char *file, const struct config *cfg)
 	(void)re_fprintf(f, "#module\t\t\t" MOD_PRE "vp8" MOD_EXT "\n");
 	(void)re_fprintf(f, "#module\t\t\t" MOD_PRE "vp9" MOD_EXT "\n");
 	(void)re_fprintf(f, "#module\t\t\t" MOD_PRE "h265" MOD_EXT "\n");
-	(void)re_fprintf(f, "#module\t\t\t" MOD_PRE "daala" MOD_EXT "\n");
 
 	(void)re_fprintf(f, "\n# Video filter Modules (in encoding order)\n");
 	(void)re_fprintf(f, "#module\t\t\t" MOD_PRE "selfview" MOD_EXT "\n");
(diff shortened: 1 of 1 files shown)


src/config.c
```diff
@@ -651,9 +656,14 @@ int config_write_template(const char *file, const struct config *cfg)
 	(void)re_fprintf(f, "#module\t\t\t" MOD_PRE "avcodec" MOD_EXT "\n");
 #endif
 	(void)re_fprintf(f, "#module\t\t\t" MOD_PRE "vp8" MOD_EXT "\n");
+	(void)re_fprintf(f, "#module\t\t\t" MOD_PRE "vp9" MOD_EXT "\n");
+	(void)re_fprintf(f, "#module\t\t\t" MOD_PRE "h265" MOD_EXT "\n");
+	(void)re_fprintf(f, "#module\t\t\t" MOD_PRE "daala" MOD_EXT "\n");
```
alfredh: daala is a very experimental codec, and it is now deprecated in favour of "AV1" codec.
I dont think we should include it in the default config

src/config.c
```diff
@@ -715,6 +735,9 @@ int config_write_template(const char *file, const struct config *cfg)
 	(void)re_fprintf(f, "module_app\t\t" MOD_PRE "auloop"MOD_EXT"\n");
 	(void)re_fprintf(f, "module_app\t\t"  MOD_PRE "contact"MOD_EXT"\n");
 	(void)re_fprintf(f, "module_app\t\t"  MOD_PRE "debug_cmd"MOD_EXT"\n");
+	(void)re_fprintf(f, "module_app\t\t"  MOD_PRE "dtmfio"MOD_EXT"\n");
+	(void)re_fprintf(f, "module_app\t\t"  MOD_PRE "echo"MOD_EXT"\n");
```
alfredh: the echo module is not compiled by default, perhaps it should be commented out ?

alfredh:
many thanks for your patch, it was very useful to clean up the config


I tested with your default config, and get a couple of warnings:

```
dl: mod: ./dtmfio.so (dlopen(./dtmfio.so, 6): image not found)
module dtmfio.so: No such file or directory
dl: mod: ./echo.so (dlopen(./echo.so, 6): image not found)
module echo.so: No such file or directory
```


these modules are not built by default, so I think they should be commented out
in default config.

czarkoff:
From my reading of `mk/modules.mk` module `dtmfio` is enabled by default on Linux and disabled elsewhere.  Thus I've wrapped it in `#ifdef LINUX`.  Is it fine?

alfredh:
thank you @czarkoff  :)

PR #193, opened 2016-12-26
my attempt to specify different sndfile directory. 

jungle-boogie:


jungle-boogie:
my attempt to fix https://github.com/alfredh/baresip/pull/193

PR #197, opened 2016-12-30
prepare nvenc_h264 code for ffmpeg >= 3.2.x

hargut:
ffmpeg 3.2 renames nvenc_h264 to h264_nvenc & drops a deprecated warning when nvenc_h264 is used.
The patch updates the nvenc_h264 section to set the encoder options on both possible selections with ffmpeg >=3.2. Additionally the debug messages are more general.


alfredh:
the patch looks nice, there is a small ccheck warning could you please fix that
and commit it (in the same branch) ?

then I will merge it

PR #203, opened 2017-01-11
[RFC] swscale avformat resize video output

hargut:
Dear Alfred,

below pull request is a "draft" which works, for enabling resizing the video within avformat, using swscale & video_size in the config.
The PR is at this stage to be considered as draft. There are too many unclear questions how to handle things the way to suite baresips overall design.

Points are the following:
- Currently there is only video_size in the config. The PR uses video_size for setting the "output" (SIP transfered) video_size. The same value is also used across modules for setting the video input size. How should be distinguish between sz_src (size source/input) & sz_dst (destination/output) in future? Do we want separate config values, or eventually introduce a swscale / aformat parameters? The PR uses video_size for sz_dst.
- Is there a way to check, except evaluating the config file parameters, if a module was successfully loaded?

### DIFF
--- a/modules/avcodec/encode.c
+++ b/modules/avcodec/encode.c
@@ -36,6 +36,7 @@ struct picsz {
 	uint8_t mpi;        /**< Minimum Picture Interval (1-32) */
 };
 
+static struct vidsz sz_dst = {0, 0};
 
 struct videnc_state {
 	AVCodec *codec;
@@ -162,6 +163,10 @@ static int init_encoder(struct videnc_state *st)
 	if (!st->codec)
 		return ENOENT;
 
+	/* TODO: Only set sz_dst when swscale is active.
+	   Is there a way to check if swscale is loaded? */
+	(void)conf_get_vidsz(conf_cur(), "video_size", &sz_dst);
+
 	return 0;
 }
 
@@ -205,8 +210,16 @@ static int open_encoder(struct videnc_state *st,
 	av_opt_set_defaults(st->ctx);
 
 	st->ctx->bit_rate  = prm->bitrate;
-	st->ctx->width     = size->w;
-	st->ctx->height    = size->h;
+
+	if (sz_dst.w <= 0) {;
+		sz_dst.w = size->w;
+	}
+	if (sz_dst.h <= 0) {
+		sz_dst.h = size->h;
+	}
+
+	st->ctx->width     = sz_dst.w;
+	st->ctx->height    = sz_dst.h;
 	st->ctx->gop_size  = DEFAULT_GOP_SIZE;
 	st->ctx->pix_fmt   = pix_fmt;
 	st->ctx->time_base.num = 1;
--- a/modules/avcodec/avcodec.c
+++ b/modules/avcodec/avcodec.c
@@ -48,8 +48,8 @@
 
 
 const uint8_t h264_level_idc = 0x0c;
-AVCodec *avcodec_h264enc;             /* optinal; specified H.264 encoder */
-AVCodec *avcodec_h264dec;             /* optinal; specified H.264 decoder */
+AVCodec *avcodec_h264enc;             /* optional; specified H.264 encoder */
+AVCodec *avcodec_h264dec;             /* optional; specified H.264 decoder */
 
 
 int avcodec_resolve_codecid(const char *s)
--- a/modules/swscale/swscale.c
+++ b/modules/swscale/swscale.c
@@ -9,13 +9,14 @@
 #include <libswscale/swscale.h>
 
 
+static struct vidsz sz_dst = {0, 0};
+
 struct swscale_enc {
 	struct vidfilt_enc_st vf;   /**< Inheritance           */
 
 	struct SwsContext *sws;
 	struct vidframe *frame;
 };
-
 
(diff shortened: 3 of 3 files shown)


modules/avcodec/encode.c
```diff
@@ -162,6 +163,10 @@ static int init_encoder(struct videnc_state *st)
 	if (!st->codec)
 		return ENOENT;
 
+	/* TODO: Only set sz_dst when swscale is active.
+	   Is there a way to check if swscale is loaded? */
+	(void)conf_get_vidsz(conf_cur(), "video_size", &sz_dst);
```
alfredh: after baresip has started and parsed the config file, the same config is available
to read from `struct config_video`. Use `conf_config()` to fetch the global config.

alfredh:
I will need to think about more about this and come back with better answers later.

but for now:

- The video size used to display incoming frames, is the same size as the
incoming stream. This is a very simple logic. We might add a new config option
to force a specific resolution, in this case we also need to consider keeping
the aspect ration between incoming frame and the configured vidisp output.

- One module cannot check if another module is loaded. A given module should
not have to _know_ if another module is loaded or not. The baresip module
architecture is very simple; baresip.h defines and interface, the baresip core is using
that interface, and a number of modules are _implementing_ that interface.

modules/avcodec/encode.c
```diff
@@ -205,8 +210,16 @@ static int open_encoder(struct videnc_state *st,
...
+
+	if (sz_dst.w <= 0) {;
+		sz_dst.w = size->w;
+	}
+	if (sz_dst.h <= 0) {
+		sz_dst.h = size->h;
+	}
+
+	st->ctx->width     = sz_dst.w;
+	st->ctx->height    = sz_dst.h;
```
alfredh: not sure if this is needed. Each input frame to the encoder contains the size of the image,
and the size might also change in a stream.

modules/avcodec/encode.c
hargut: Sounds like a good start to check if the config contains the swscale module, and otherwise issue a warning. The code does work without swscale, the visible effect is that video parts are missing.

E.g. configuring rescaling from 1280x720 to 640x360 would then display only the "1st" quater of the picture in the output stream. So it is not "critical", but the user should be informed that his configuration set will lead to a "loss" of video stream information in the output.

modules/avcodec/encode.c
hargut: During preparing the patch I didn't check the details of avcodec if it will be required or not. This code part was only updated to use sz_dst instead of size->w, as used before.

Having a quick look at the FFmpeg documentation it must be set during the encoding process:
```
int AVCodecContext::width
- encoding: MUST be set by user.
- decoding: May be set by the user before opening the decoder if known e.g. from the container. Some decoders will require the dimensions to be set by the caller. During decoding, the decoder may overwrite those values as required while parsing the data.
```
https://ffmpeg.org/doxygen/3.2/structAVCodecContext.html#a0d8f46461754e8abea0847dcbc41b956

hargut:
Dear Alfred,
thank you for your comments & feedback.

```The video size used to display incoming frames, is the same size as the
incoming stream. This is a very simple logic. We might add a new config option
to force a specific resolution, in this case we also need to consider keeping
the aspect ration between incoming frame and the configured vidisp output.```

I'd like to have an option to resize/rescale a video stream on the output side in order to have a possibility reducing video stream size for hardware phones. Some of them (e.g. Polycom VVX600, Grandstream GXV3140, Yealink VP530, ...) do only support CIF/QCIF or similar resolutions. My intention is not related to local display output or baresip display output, since nearly any nowadays used PC would easily handle the streams without rescaling.

For avformat that topic is not too important, since we are dealing with input formats where I have not encountered any situation needing to specify the input video size. This question came up thinking on input devices like V4L2 where it does make sense to specify the input resolution.

I'm just wondering if the aspect ratio should be our concern, or swscales and the users. Will test how the patch does when feeding values which are not the same aspect ratio. I missed that point, and think a warning could be sufficient in such case (depending how swscale does with unequal aspect ratios on input/output).

**Comment by hargut** [comment]
Another completely different approach would be to set the output size parameters during codec negotiation evaluating the offered h264 profile-level-id from the counterpart.

**Comment by alfredh** [comment]
I have misunderstood and thought that you wanted to change the VIDISP resolution.

now that I understand your usecase, I can only say that it is a valid usecase and
a very nice improvement.

A similar problem would also be solved with this; for example the H.263 encoder
supports only certain resolutions. If the `vidsrc` cannot provide this exact resolution,
then the video encoding pipeline will fail.

An example video encoding pipeline:

```
.--------------.
|    videnc    |
|              |
| (avcodec.so) |
'--------------'
     /|\
      |   352x288 (a.k.a. CIF)
      |
.--------------.
|   vidfilt    |
|              |
| (swscale.so) |
'--------------'
     /|\
      |   320x240
      |
.--------------.
|    vidsrc    |
|              |
|  (e.g V4L2)  |
'--------------'
```

the swscale destination parameter can either be read from `video_size`
or we can add a new config `swscale_encode_size` (not sure what is best)

hargut:
I've just tested 546456b which does the job nicely. :)

Issue #201, opened 2017-01-10
vidloop & sdl2 display fails


hargut:
Dear Alfred,

I've discovered a segfault with sdl2 & "-e/vidloop h264" on my test setup:
```
hargut@benchvice:~/baresip_alfredh/baresip$ LANG=C gdb --args ./baresip -e"/vidloop h264"   
GNU gdb (Ubuntu 7.11.90.20161005-0ubuntu1) 7.11.90.20161005-git
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./baresip...done.
(gdb) run
Starting program: /home/hargut/baresip_alfredh/baresip/baresip -e/vidloop\ h264
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
baresip v0.5.0 Copyright (C) 2010 - 2017 Alfred E. Heggestad et al.
Local network address:  IPv4=enp2s0:10.0.0.11 
aucodec: PCMU/8000/1
aucodec: PCMA/8000/1
aufilt: vumeter
avcodec: using h264 decoder by name (h264_cuvid)
vidcodec: H264
vidcodec: H263
vidcodec: MP4V-ES
avcodec: using h264 encoder by name (h264_nvenc)
vidsrc: avformat
vidisp: x11
vidisp: sdl2
medianat: stun
medianat: turn
medianat: ice
Populated 1 account
Populated 3 contacts
Populated 2 audio codecs
Populated 1 audio filter
Populated 3 video codecs
Populated 0 video filters
baresip is ready.
/vidloop h264
Enable video-loop on avformat,/home/hargut/dwhelper/st-pl.mp4: 1280 x 720
vidloop: avformat,/home/hargut/dwhelper/st-pl.mp4: open video source: 1280 x 720 at 500 fps
[...]

PR #205, opened 2017-01-13
enhance sdl2 pixelformat selection

### LINKED ISSUE #202: sdl2 does not convert pixel formats properly?
Dear Alfred,

I've just tried to play around a little with nvenc/cuvid & avformat and stumbled across another issue/lack with sdl2. When using cuvid to decode the pixel format seems to be wrong. The picture looks similar to what I've seen during implementing nv12_to_rgb32.

Attached please find a screenshot of the sdl2 displayed picture when decoded with h264_cuvid.
baresip_sdl2_cuvid

Setup is as follows: baresip (avformat,file & avcodec_h264enc h264_nvenc) <-> baresip (avcodec_h264dec h264_cuvid / sdl2)

hargut:
fixes #202 and #201

### DIFF
--- a/modules/sdl2/sdl.c
+++ b/modules/sdl2/sdl.c
@@ -30,6 +30,25 @@ struct vidisp_st {
 static struct vidisp *vid;
 
 
+static enum vidfmt match_fmt(u_int32_t fmt)
+{
+	switch (fmt) {
+	case VID_FMT_YUV420P:	return SDL_PIXELFORMAT_IYUV;
+	case VID_FMT_YUYV422:	return SDL_PIXELFORMAT_YUY2;
+	case VID_FMT_UYVY422:	return SDL_PIXELFORMAT_UYVY;
+	case VID_FMT_RGB32:	return SDL_PIXELFORMAT_RGBA8888;
+	case VID_FMT_ARGB:	return SDL_PIXELFORMAT_ARGB8888;
+	case VID_FMT_RGB565:	return SDL_PIXELFORMAT_RGB565;
+	case VID_FMT_RGB555:	return SDL_PIXELFORMAT_RGB555;
+	case VID_FMT_NV12:	return SDL_PIXELFORMAT_NV12;
+	case VID_FMT_NV21:	return SDL_PIXELFORMAT_NV21;
+	case VID_FMT_YUV444P:
+	case VID_FMT_N:
+	default:		return SDL_PIXELFORMAT_UNKNOWN;
+	}
+}
+
+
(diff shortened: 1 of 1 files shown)


modules/sdl2/sdl.c
```diff
@@ -30,6 +30,25 @@ struct vidisp_st {
...
+	case VID_FMT_UYVY422:	return SDL_PIXELFORMAT_UYVY;
+	case VID_FMT_RGB32:	return SDL_PIXELFORMAT_RGBA8888;
+	case VID_FMT_ARGB:	return SDL_PIXELFORMAT_ARGB8888;
+	case VID_FMT_RGB565:	return SDL_PIXELFORMAT_RGB565;
+	case VID_FMT_RGB555:	return SDL_PIXELFORMAT_RGB555;
+	case VID_FMT_NV12:	return SDL_PIXELFORMAT_NV12;
+	case VID_FMT_NV21:	return SDL_PIXELFORMAT_NV21;
+	case VID_FMT_YUV444P:
+	case VID_FMT_N:
+	default:		return SDL_PIXELFORMAT_UNKNOWN;
```
alfredh: have you tested that all these pixel formats are working?

we dont have to add all formats now, I think it is better to just add a couple of
formats that is tested and working, such as YUV420P and NV12.

modules/sdl2/sdl.c
```diff
@@ -30,6 +30,25 @@ struct vidisp_st {
 static struct vidisp *vid;
 
 
+static enum vidfmt match_fmt(u_int32_t fmt)
```
alfredh: the input parameter type should be: `enum vidfmt`

the return type should be `Uint32` (same as input to `SDL_CreateTexture`).

alfredh:
this is a nice improvement. I think the long list of pixel formats can be shorter.

we also need to update the code that copies pixels from vidframe into the SDL texture:

https://github.com/alfredh/baresip/blob/master/modules/sdl2/sdl.c#L174

the code assumes YUV420P. it needs to handle that `data[i]` and `linesize[i]`
is zero for some formats.

for example when I tested your patch locally, an input frame with NV12
causes a crash in the pixel copying code.

hargut:
Dear Alfred,

to be honest, in this shape this shouldn't have appeared as pull request. The patch wasn't even working.

I've tried to fix the issue in a rush, and have overlooked that h264_decoder was not set properly to h264_cuvid. Therefore I thought everything was working during the test, but it simply wasn't as there where no NV12 frames input to the sdl2 module.

Thank you for your efforts to fix it!

PR #206, opened 2017-01-16
Configure Path where soundfiles are dumped to

elektm93:
Dear Alfred,

I had the problem to localize my snd-dumps, when running baresip in deamon mode. Due to lack of permissions, the got not created or dropped somewhere I couldn't find.

## Solution
- path where the soundfiles are dropped is now configurable (like in other modules)

The way how the destination-filename is constructed may not be an advanced one. Feel free to edit!

Best Regards,
Martin

### DIFF
--- a/modules/sndfile/sndfile.c
+++ b/modules/sndfile/sndfile.c
@@ -80,10 +80,11 @@ static SNDFILE *openfile(const struct aufilt_prm *prm, bool enc)
 	SNDFILE *sf;
 
 	(void)re_snprintf(filename, sizeof(filename),
-			  "/dump-%H-%s.wav",
+			  "dump-%H-%s.wav",
 			  timestamp_print, tm, enc ? "enc" : "dec");
 
 	strcpy(filepath, file_path.p);
+	strcat(filepath, '/');
 	strcat(filepath, filename);
 	strcpy(filename, filepath);
 
--- a/modules/sndfile/sndfile.c
+++ b/modules/sndfile/sndfile.c
@@ -17,7 +17,6 @@
  * Example Configuration:
  \verbatim
   snd_path 					/tmp/
-	snd_file_prefix		dump
  \endverbatim
  */
 
@@ -32,9 +31,7 @@ struct sndfile_dec {
 	SNDFILE *dec;
 };
 
-static struct pl file_path = {
-	PL("./")
-};
+static char file_path[256] = ".";
 
 
 static int timestamp_print(struct re_printf *pf, const struct tm *tm)
@@ -80,13 +77,9 @@ static SNDFILE *openfile(const struct aufilt_prm *prm, bool enc)
 	SNDFILE *sf;
 
 	(void)re_snprintf(filename, sizeof(filename),
-			  "dump-%H-%s.wav",
+			  "%s/dump-%H-%s.wav",
+				file_path,
 			  timestamp_print, tm, enc ? "enc" : "dec");
-
-  pl_strcpy(&file_path, filepath, 256);
-	strcat(filepath, "/");
-	strcat(filepath, filename);
-	strcpy(filename, filepath);
 
 	sfinfo.samplerate = prm->srate;
 	sfinfo.channels   = prm->ch;
@@ -185,9 +178,9 @@ static int module_init(void)
 {
 	aufilt_register(&sndfile);
 
-	if (conf_get(conf_cur(), "snd_path", &file_path)) {
-		pl_set_str(&file_path, "./");
-	}
+	// TODO: conf_get_str and variable instead of struct
+
+	conf_get_str(conf_cur(), "snd_path", &file_path, 256);
 
 	info("sndfile: saving files in %s\n", file_path);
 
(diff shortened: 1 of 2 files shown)


elektm93:
sorry - there was a segmentation fault in there. fixed it.

alfredh:
many thanks for your patch.

I get some errors when compiling (OSX and clang):

```
  CC [M]  build-x86_64/modules/sdl2/sdl.o
  LD [M]  sdl2.so
  CC [M]  build-x86_64/modules/sndfile/sndfile.o
modules/sndfile/sndfile.c:36:2: warning: excess elements in scalar initializer
        PL("./")
        ^~~~~~~~
../re/include/re_fmt.h:22:21: note: expanded from macro 'PL'
#define PL(s) {(s), sizeof((s))-1}
                    ^~~~~~~~~~~~~
```

modules/sndfile/sndfile.c
```diff
@@ -13,6 +13,12 @@
  * @defgroup sndfile sndfile
  *
  * Audio filter that writes audio samples to WAV-file
+ *
+ * Example Configuration:
+ \verbatim
+  snd_path 					/tmp/
+	snd_file_prefix		dump
```
alfredh: snd_file_prefix is not used ?

modules/sndfile/sndfile.c
```diff
@@ -168,6 +184,13 @@ static struct aufilt sndfile = {
 static int module_init(void)
 {
 	aufilt_register(&sndfile);
+
+	if (conf_get(conf_cur(), "snd_path", &file_path)) {
```
alfredh: as a general rule in baresip/libre, when using the `struct pl` type you should use the value
in the same function, because the data might be gone when the function returns.

please use a string instead and the function `conf_get_str` to read the config. The return value
can be ignored.

modules/sndfile/sndfile.c
```diff
@@ -26,6 +32,10 @@ struct sndfile_dec {
 	SNDFILE *dec;
 };
 
+static struct pl file_path = {
+	PL("./")
+};
+
```
alfredh: please use a string here (static char file_path[256] = ".")

modules/sndfile/sndfile.c
```diff
@@ -72,6 +83,11 @@ static SNDFILE *openfile(const struct aufilt_prm *prm, bool enc)
 			  "dump-%H-%s.wav",
 			  timestamp_print, tm, enc ? "enc" : "dec");
 
+  pl_strcpy(&file_path, filepath, 256);
+	strcat(filepath, "/");
+	strcat(filepath, filename);
+	strcpy(filename, filepath);
+
```
alfredh: I think it is better to modify the code that creates the path instead, and always insert
the path in the beginning. for example:

```
	r = re_snprintf(filename, sizeof(filename),
			  "%s/dump-%H-%s.wav",
			  path,
			  timestamp_print, tm, enc ? "enc" : "dec");
	if (r <= 0) return NULL;
```

elektm93:
Hi Alfred,

thank you for you comments! I've implemented everything as suggested.

* removed the commend for snd_file_prefix. I intended to implement further ideas, but I reassessed that thought after writing that comment.
* changed the static struct pl file_path to an string, hence also the compiling error for OSX and clang should be gone.
* the approach of of creating the filename is implemented as suggested

Sorry for the inconvinience and thank you for the creation of this awesome project!! 

Best Regards,
Martin

modules/sndfile/sndfile.c
```diff
@@ -63,13 +70,15 @@ static void dec_destructor(void *arg)
 static SNDFILE *openfile(const struct aufilt_prm *prm, bool enc)
 {
 	char filename[128];
+	char filepath[256];
```
alfredh: this variable is unused and can be removed.

modules/sndfile/sndfile.c
```diff
@@ -168,6 +177,11 @@ static struct aufilt sndfile = {
 static int module_init(void)
 {
 	aufilt_register(&sndfile);
+
+	conf_get_str(conf_cur(), "snd_path", &file_path, 256);
```
alfredh: the last two parameters here should be:

`, file_path, sizeof(file_path)`

alfredh:
I am still getting some compile warning on OSX:

```
 make sndfile.so
  CC [M]  build-x86_64/modules/sndfile/sndfile.o
modules/sndfile/sndfile.c:73:7: warning: unused variable 'filepath' [-Wunused-variable]
        char filepath[256];
             ^
modules/sndfile/sndfile.c:181:39: warning: incompatible pointer types passing 'char (*)[256]' to parameter of type 'char *' [-Wincompatible-pointer-types]
        conf_get_str(conf_cur(), "snd_path", &file_path, 256);
                                             ^~~~~~~~~~
../re/include/re_conf.h:15:67: note: passing argument to parameter 'str' here
int conf_get_str(const struct conf *conf, const char *name, char *str,
                                                                  ^
2 warnings generated.
  LD [M]  sndfile.so
```

alfredh:
fixed in https://github.com/alfredh/baresip/commit/903ce9869c933e4f13995cd9f10c1498665cfd58

PR #213, opened 2017-02-22
enabled long commands on httpd module

elektm93:
changed the parsing of the command that an '%20' in an url is equivalent
to an whitespace.

### DIFF
--- a/modules/httpd/httpd.c
+++ b/modules/httpd/httpd.c
@@ -37,16 +37,16 @@ static int html_print_head(struct re_printf *pf, void *unused)
 }
 
 
-static int html_print_cmd(struct re_printf *pf, const struct http_msg *req)
+static int html_print_cmd(struct re_printf *pf, const struct pl *prm)
 {
 	struct pl params;
 
-	if (!pf || !req)
+	if (!pf || !prm)
 		return EINVAL;
 
-	if (pl_isset(&req->prm)) {
-		params.p = req->prm.p + 1;
-		params.l = req->prm.l - 1;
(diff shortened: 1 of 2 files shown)


alfredh:
thanks for your patch

the cmd subsystem is not the right place to do this. It should be done in the
httpd module instead.

please use the function `uri_header_unescape` to unescape any escaped URL
parameters.

example:

in:  "foo%20bar"
out: "foo bar"

example usage:

```
char *buf;

re_sdprintf(&buf, "%H", uri_header_unescape, &req->prm);

mem_deref(buf);
```

elektm93:
Thanks for your feedback @alfredh !!
I implemented your suggestions as described and long commands work through the httpd module.

But I still have problems using commands with an ',', like the auplay command.
`uric: unescape: illegal ',' in ?/auplay%20alsa,default`

Do you have any suggestions for me how/where to tackle the problem?

alfredh:
this looks much better.

perhaps comma is a legal character in http uri, the uri module in libre is mainly
written for SIP. you can try to escape it (%2c)

you need to take a copy of the "struct pl" from http msg. then do the unescape,
and update pl again. you also need to check the return value from re_sdprintf.


/Alfred

elektm93:
Thanks for your help, @alfredh !
I implemented your suggestions but I also had to change the `html_print_raw` and `html_print_cmd`. Hope it is ok like that.

Martin

alfredh:
many thanks @elektm93 I have reviewed the code and it looks good. I have also tested
it and it works just fine :)

PR #218, opened 2017-03-01
Video device selection bug in v4l2_codecs

ripleyXLR8:
Correcting a bug making impossible to select the video device in v4l2_codecs


alfredh:
that was a good catch, well spotted. And thanks for the nice and non-intrusive patch :)

PR #228, opened 2017-03-27
Video Display Support for Raspberry Pi

strfry:
As previously announced on the Mailing list, this patch introduces support for hardware-accelerated scaled video playback on the Raspberry Pi. It is not very well tested, and breaks for some resolutions, but otherwise running stable.
The module is named 'omx', due to the "OpenMAX Integration Layer".
In the future, it might support the native video codec (h.264), and video capture (Pi Camera).
It is probably not compatible with OpenMAX on other platforms like Android, without further changes.

Cheers,
Jonathan

### DIFF
--- a/mk/modules.mk
+++ b/mk/modules.mk
@@ -201,6 +201,10 @@ USE_VPX  := $(shell [ -f $(SYSROOT)/include/vpx/vp8.h ] \
 	|| [ -f $(SYSROOT)/local/include/vpx/vp8.h ] \
 	|| [ -f $(SYSROOT_ALT)/include/vpx/vp8.h ] \
 	&& echo "yes")
+USE_OMX := $(shell [ -f /opt/vc/include/bcm_host.h ] || \
+	[ -f $(SYSROOT)/include/bcm_host.h ] || \
+	[ -f $(SYSROOT_ALT)/include/bcm_host.h ] \
+	&& echo "yes")
 else
 # Windows.
 # Accounts for mingw with Windows SDK (formerly known as Platform SDK)
@@ -412,6 +416,9 @@ MODULES   += v4l
 endif
 ifneq ($(USE_V4L2),)
 MODULES   += v4l2 v4l2_codec
+endif
+ifneq ($(USE_OMX),)
+MODULES   += omx
 endif
 ifneq ($(USE_VPX),)
 MODULES   += vp8
--- a/mk/modules.mk
+++ b/mk/modules.mk
@@ -203,7 +203,8 @@ USE_VPX  := $(shell [ -f $(SYSROOT)/include/vpx/vp8.h ] \
 	&& echo "yes")
 USE_OMX := $(shell [ -f /opt/vc/include/bcm_host.h ] || \
 	[ -f $(SYSROOT)/include/bcm_host.h ] || \
-	[ -f $(SYSROOT_ALT)/include/bcm_host.h ] \
+	[ -f $(SYSROOT_ALT)/include/bcm_host.h ] || \
+	[ -f $(SYSROOT)/include/OMX_Video.h ] \
 	&& echo "yes")
 else
 # Windows.
--- a/mk/modules.mk
+++ b/mk/modules.mk
@@ -32,6 +32,9 @@
 #   USE_L16           L16 audio codec
 #   USE_MPA           MPA audo codec
 #   USE_MPG123        Use mpg123
+#   USE_OMX_RPI       RaspberryPi VideoCore display driver
+#   USE_OMX_BELLAGIO  libomxil-bellagio xvideosink driver
+#   USE_OPUS          Opus audio codec
 #   USE_OPUS          Opus audio codec
 #   USE_OSS           OSS audio driver
 #   USE_PLC           Packet Loss Concealment
@@ -201,10 +204,13 @@ USE_VPX  := $(shell [ -f $(SYSROOT)/include/vpx/vp8.h ] \
 	|| [ -f $(SYSROOT)/local/include/vpx/vp8.h ] \
 	|| [ -f $(SYSROOT_ALT)/include/vpx/vp8.h ] \
 	&& echo "yes")
-USE_OMX := $(shell [ -f /opt/vc/include/bcm_host.h ] || \
-	[ -f $(SYSROOT)/include/bcm_host.h ] || \
-	[ -f $(SYSROOT_ALT)/include/bcm_host.h ] || \
-	[ -f $(SYSROOT)/include/OMX_Video.h ] \
+USE_OMX_RPI := $(shell [ -f /opt/vc/include/bcm_host.h ] || \
+	[ -f $(SYSROOT)/include/bcm_host.h ] \
+	|| [ -f $(SYSROOT_ALT)/include/bcm_host.h ] \
+	&& echo "yes")
+USE_OMX_BELLAGIO := $(shell [ -f /usr/include/OMX_Core.h ] \
+	|| [ -f $(SYSROOT)/include/OMX_Core.h ] \
+	|| [ -f $(SYSROOT_ALT)/include/OMX_Core.h ] \
 	&& echo "yes")
 else
 # Windows.
@@ -418,7 +424,10 @@ endif
 ifneq ($(USE_V4L2),)
 MODULES   += v4l2 v4l2_codec
 endif
-ifneq ($(USE_OMX),)
+ifneq ($(USE_OMX_RPI),)
+MODULES   += omx
+endif
+ifneq ($(USE_OMX_BELLAGIO),)
 MODULES   += omx
 endif
 ifneq ($(USE_VPX),)
--- a/modules/omx/README
+++ b/modules/omx/README
@@ -1,0 +1,17 @@
+README
+------
+
+This module implements support for the VideoCore4 of
+the Raspberry Pi A/B/2/3.
+Currently it only does video playback.
+
+EXAMPLE CONFIG
+--------------
+
+# Video
+video_display            omx,nil
+
+# Video codec Modules (in order)
+module                  omx.so
+
+
--- a/modules/omx/module.c
+++ b/modules/omx/module.c
@@ -1,0 +1,145 @@
+/**
+ * @file omx/module.c     Raspberry Pi VideoCoreIV OpenMAX interface
+ *
+ * Copyright (C) 2016 - 2017 Creytiv.com
+ * Copyright (C) 2016 - 2017 Jonathan Sieber
+ */
+
+
+#include "omx.h"
+
+#include <stdlib.h>
+
+#include <re/re.h>
+#include <rem/rem.h>
+#include <baresip.h>
+
+int omx_vidisp_alloc(struct vidisp_st **vp, const struct vidisp* vd,
+	struct vidisp_prm *prm, const char *dev, vidisp_resize_h *resizeh,
+	void *arg);
+int omx_vidisp_display(struct vidisp_st *st, const char *title,
+	const struct vidframe *frame);
+
+struct vidisp_st {
+	const struct vidisp *vd;  /* inheritance */
+	struct vidsz size;
+	struct omx_state* omx;
+};
+
+static struct vidisp* vid;
+
+static struct omx_state omx;
+
+static void destructor(void *arg)
+{
+	struct vidisp_st *st = arg;
+	omx_display_disable(st->omx);
+}
+
+int omx_vidisp_alloc(struct vidisp_st **vp, const struct vidisp* vd,
+	struct vidisp_prm *prm,	const char *dev, vidisp_resize_h *resizeh,
+	void *arg)
+{
+	struct vidisp_st *st;
+
+	/* Not used by OMX */
+	(void) prm;
+	(void) dev;
+	(void) resizeh;
+	(void) arg;
+
+	info("omx: vidisp_alloc\n");
+
+	st = mem_zalloc(sizeof(*st), destructor);
+	if (!st)
+		return ENOMEM;
+
+	st->vd = vd;
+	*vp = st;
+
+	st->omx = &omx;
+
+	return 0;
+}
+
+
+int omx_vidisp_display(struct vidisp_st *st, const char *title,
+		   const struct vidframe *frame)
+{
+	int i;
+	void* buf;
+	uint32_t len;
+	uint32_t offset = 0;
+
+	size_t plane_bytes;
+	void* dest;
+
+	int err = 0;
+
+	if (frame->fmt != VID_FMT_YUV420P) {
+		return EINVAL;
+	}
+
+	if (!vidsz_cmp(&st->size, &frame->size)) {
+		info("omx: new frame size: w=%d h=%d\n",
+			frame->size.w, frame->size.h);
+		info("omx: linesize[0]=%d\tlinesize[1]=%d\tlinesize[2]=%d\n",
+			frame->linesize[0], frame->linesize[1],
+			frame->linesize[2]);
+		err = omx_display_enable(st->omx,
+			frame->size.w, frame->size.h, frame->linesize[0]);
+		if (err) {
+			error("omx_display_enable failed");
+			return err;
+		}
+		st->size = frame->size;
+		memcpy(&st->size, &frame->size, sizeof(struct vidsz));
+	}
+
+	/* Get Buffer Pointer */
+	omx_display_input_buffer(st->omx, &buf, &len);
+
+	for (i = 0; i < 3; i++) {
+		plane_bytes = frame->linesize[i] *
+			frame->size.h / (i > 0 ? 2 : 1);
+		dest = buf + offset;
+
+		if (offset + plane_bytes > len) {
+			warn("Too large frame for OMX Buffer size. Expected"
+				" %d, got %d\n", len, offset + plane_bytes);
+			return ENOMEM;
+		}
+
+		memcpy(dest, frame->data[i], plane_bytes);
+		offset += plane_bytes;
+	}
+	omx_display_flush_buffer(st->omx);
+	return 0;
+}
+
+static int module_init(void)
+{
+	if (!omx_init(&omx)) {
+		error("Could not initialize OpenMAX");
+		return ENODEV;
+	}
+
+	return vidisp_register(&vid, "omx",
+		omx_vidisp_alloc, NULL, omx_vidisp_display, NULL);
+}
+
+static int module_close(void)
+{
+	/* HACK: not deinitializing OMX because of a hangup */
+	/* omx_deinit(&omx) */
+	vid = mem_deref(vid);
+	return 0;
+}
+
+
+EXPORT_SYM const struct mod_export DECL_EXPORTS(omx) = {
+	"omx",
+	"vidisp",
+	module_init,
+	module_close
+};
--- a/modules/omx/module.c
+++ b/modules/omx/module.c
@@ -66,15 +66,13 @@ int omx_vidisp_alloc(struct vidisp_st **vp, const struct vidisp* vd,
 int omx_vidisp_display(struct vidisp_st *st, const char *title,
 		   const struct vidframe *frame)
 {
-	int i;
+	int err = 0;
 	void* buf;
 	uint32_t len;
-	uint32_t offset = 0;
 
-	size_t plane_bytes;
-	void* dest;
+	struct vidframe omx_frame;
 
-	int err = 0;
+	(void)title;
 
 	if (frame->fmt != VID_FMT_YUV420P) {
 		return EINVAL;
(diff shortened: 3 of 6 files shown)


mk/modules.mk
```diff
@@ -413,6 +417,10 @@ endif
 ifneq ($(USE_V4L2),)
 MODULES   += v4l2 v4l2_codec
 endif
+ifneq ($(USE_OMX),)
+MODULES   += omx 
+#rpicam
```
alfredh: the commented out line can be removed.

modules/omx/README
```diff
@@ -0,0 +1,17 @@
...
+
+This module implements support for the VideoCore4 of
+the Raspberry Pi A/B/2/3.
+Currently it only does video playback.
+
+EXAMPLE CONFIG
+--------------
+
+# Video
+video_source            omx,nil
```
alfredh: does it do video playback or video source  ?

modules/omx/make_test.sh
```diff
@@ -0,0 +1,6 @@
+gcc -g -o test_display test.c omx.c \
+	-DOMX_SKIP64BIT \
+	-I /usr/local/include -I/usr/local/include/interface/vcos/pthreads/ \
+	-I/usr/local/include/interface/vmcs_host/linux/ \
+	-L/usr/local/lib \
+	-lopenmaxil -lbcm_host -lvcos -lrem -lpthread
```
alfredh: this file is not needed, can be removed.

modules/omx/module.c
```diff
@@ -0,0 +1,149 @@
+#include "omx.h"
```
alfredh: a doxygen header is needed. you can use the same style as other module.
please keep the copyright creytiv line, you can also add your own name if you wish.

modules/omx/module.c
```diff
@@ -0,0 +1,149 @@
...
+
+#include <stdlib.h>
+
+#include <re/re.h>
+#include <rem/rem.h>
+#include <baresip.h>
+
+/////// baresip Compatibility Layer
```
alfredh: please dont use C++ comments. use `/* comment */` instead.

modules/omx/module.c
```diff
@@ -0,0 +1,149 @@
...
+
+struct vidisp_st {
+	const struct vidisp *vd;  /* inheritance */	
+	struct vidsz size;             /**< Current size          */
+	struct omx_state* omx;
+};
+
+static struct vidisp* vid; // ???
+
+static struct omx_state omx;
```
alfredh: this looks like a global state. is it possible to avoid this?

modules/omx/module.c
```diff
@@ -0,0 +1,149 @@
...
+	// GHOST omx_copy_vidframe_to_buf()
+	
+	// FUNCTION check_resize
+	
+	if (memcmp(&st->size, &frame->size, sizeof(struct vidsz))) {
+		info("new frame size: w=%d h=%d\n", frame->size.w, frame->size.h);
+		info("linesize[0]=%d\tlinesize[1]=%d\tlinesize[2]=%d\n", frame->linesize[0], frame->linesize[1], frame->linesize[2]);
+		omx_display_enable(st->omx, frame->size.w, frame->size.h, frame->linesize[0]);
+		memcpy(&st->size, &frame->size, sizeof(struct vidsz));
+	}
```
alfredh: you can use `vidsz_cmp` instead which is typesafe.

modules/omx/module.c
```diff
@@ -0,0 +1,149 @@
...
+	warn("omx_vidisp_update\n");
+	printf("fullscreen=%d	orient=%d	window=%p\n", fullscreen, orient, window);
+	return -EINVAL; // NOT IMPLEMENTED
+}
+
+
+static void omx_vidisp_hide(struct vidisp_st *st)
+{
+	error("omx_vidisp_hide not implemented\n");
+}
```
alfredh: update-handler and hide-handler is optional, you can send in NULL instead of vidisp_register.

modules/omx/module.c
```diff
@@ -0,0 +1,149 @@
...
+
+static int module_init(void)
+{
+	omx_init(&omx);
```
alfredh: please check return value, and return in case of errors.

modules/omx/omx.c
```diff
@@ -0,0 +1,322 @@
+/**
+ * @file vc4.c     Raspberry Pi VideoCoreIV OpenMAX interface
```
alfredh: the name does not match the filename.

modules/omx/omx.c
```diff
@@ -0,0 +1,322 @@
...
+	&EventHandler,
+	&EmptyBufferDone,
+	&FillBufferDone
+};
+
+  
+int omx_init(struct omx_state* st)  
+{
+	OMX_ERRORTYPE err;
+	(void)st;
```
alfredh: this line is not needed.

modules/omx/omx.c
```diff
@@ -0,0 +1,322 @@
...
+
+	bcm_host_init();
```
alfredh: no error checking ?

modules/omx/omx.c
```diff
@@ -0,0 +1,322 @@
...
+	
+	st->buffers = NULL;
+   
+   	pthread_mutex_init(&st->omx_mutex, 0);
+	pthread_mutex_lock(&st->omx_mutex);	
+
+	err = OMX_Init();
+	err |= OMX_GetHandle(&st->video_render, "OMX.broadcom.video_render", 0, &callbacks);
+	assert (err == 0);
```
alfredh: please avoid assert in the code. It is better to return gracefully.

modules/omx/omx.c
```diff
@@ -0,0 +1,322 @@
...
+		if((r = OMX_GetParameter(hComponent, OMX_IndexParamPortDefinition, &portdef)) != OMX_ErrorNone) {
+			error("block_until_port_changed: Failed to get port definition. Result=%d\n", r);
+		}
+		if(portdef.bEnabled != bEnabled) {
+			usleep(10000);
+		}
+	}
+}
+
+void omx_display_enable(struct omx_state* st, int width, int height, int stride) {
```
alfredh: this function should probably return `int` and an errorcode.

modules/omx/omx.c
```diff
@@ -0,0 +1,322 @@
...
+    // TODO: rotate the buffers now.
+    
+    return 0;
+}
+
+//int omx_send_frame(void* luma, void* chroma_u, void* chroma_v, uint32_t len)
+//int omx_send_frame(void* data, uint32_t len)
+/*
+int omx_send_frame(struct vidframe* frame)
+{
```
alfredh: this function is commented out, it can probably be removed.

alfredh:
hi @strfry many thanks for this nice work

couple of review comments inline, couple of high-level comments here:

* the PR has more than 50 commits. Could you please collapse them all into a nice
 single commit ?

* the source code must be formatted properly, please use the tool ccheck.py
  to verify this:

modules/omx/module.c
strfry: I found that all the display modules seem to have this global pointer to the vidisp struct? So i assumed that all these modules have some kind of global state.
In this case, this thing exists because OMX needs to be initialized from the main thread context, from module_init().
This might be possible to code this in a nicer way, but that's beyond what i can do tonight in a quick code cleanup, without the test platform. ;)

modules/omx/omx.c
strfry: No, it returns void, AFAIR

strfry:
Fixed all the code style issues.
The global state problem is not so easily fixable without further tesitng.

modules/omx/module.c
alfredh: yes, you are right this is no problem. I was misreading the code.

alfredh:
Thank you, I will do some more review.

modules/omx/module.c
```diff
@@ -0,0 +1,144 @@
...
+#include <baresip.h>
+
+int omx_vidisp_alloc(struct vidisp_st **vp, const struct vidisp* vd,
+	struct vidisp_prm *prm, const char *dev, vidisp_resize_h *resizeh,
+	void *arg);
+int omx_vidisp_display(struct vidisp_st *st, const char *title,
+	const struct vidframe *frame);
+int omx_vidisp_update(struct vidisp_st *st, bool fullscreen, int orient,
+	const struct vidrect* window);
+void omv_vidisp_hide(struct vidisp_st *st);
```
alfredh: the prototypes for update and hide are not used any more and can be removed.

modules/omx/module.c
```diff
@@ -0,0 +1,144 @@
...
+{
+	struct vidisp_st *st = arg;
+	omx_display_disable(st->omx);
+}
+
+int omx_vidisp_alloc(struct vidisp_st **vp, const struct vidisp* vd,
+	struct vidisp_prm *prm,	const char *dev, vidisp_resize_h *resizeh,
+	void *arg)
+{
+	info("vidisp_alloc\n");
```
alfredh: add "omx: " prefix to the logline

modules/omx/module.c
```diff
@@ -0,0 +1,144 @@
...
+
+	struct vidisp_st *st;
```
alfredh: variables must be declared first in functions (C89 style).

modules/omx/module.c
```diff
@@ -0,0 +1,144 @@
...
+int omx_vidisp_display(struct vidisp_st *st, const char *title,
+		   const struct vidframe *frame)
+{
+	int err = 0;
+
+	if (frame->fmt != VID_FMT_YUV420P) {
+		return EINVAL;
+	}
+
+	if (vidsz_cmp(&st->size, &frame->size)) {
```
alfredh: this condition must be inverted (does this code work)?

modules/omx/module.c
```diff
@@ -0,0 +1,144 @@
...
+		info("linesize[0]=%d\tlinesize[1]=%d\tlinesize[2]=%d\n",
+			frame->linesize[0], frame->linesize[1],
+			frame->linesize[2]);
+		err = omx_display_enable(st->omx,
+			frame->size.w, frame->size.h, frame->linesize[0]);
+		if (err) {
+			error("omx_display_enable failed");
+			return err;
+		}
+		memcpy(&st->size, &frame->size, sizeof(struct vidsz));
```
alfredh: you can assign the variable directly:

`st->size = frame->size`

modules/omx/module.c
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	/* Get Buffer Pointer */
+	int i;
+	void* buf;
+	uint32_t len;
+	uint32_t offset = 0;
```
alfredh: move declarations to the top

modules/omx/omx.c
```diff
@@ -0,0 +1,300 @@
...
+#include "omx.h"
+
+#include <re/re.h>
+#include <rem/rem.h>
+#include <baresip.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <assert.h>
```
alfredh: no need to include assert.h as the code should not use assert()
(only in extremely rare cases).

modules/omx/omx.c
```diff
@@ -0,0 +1,300 @@
...
+static const int VIDEO_RENDER_PORT = 90;
+
+static int EventHandler(OMX_HANDLETYPE hComponent, void* pAppData,
+	OMX_EVENTTYPE eEvent, OMX_U32 nData1, OMX_U32 nData2,
+	void* pEventData)
+{
+	(void) hComponent;
+	switch (eEvent) {
+		case OMX_EventCmdComplete:
+			printf("omx.EventHandler: Previous command completed\n"
```
alfredh: please dont use `printf` which only prints to stdout. Better to use `info` or `debug` which prints
to the logging system with multiple backends.

(in this case perhaps `debug` is best for low-level details)

modules/omx/omx.c
```diff
@@ -0,0 +1,300 @@
...
+	return 0;
+}
+
+static OMX_ERRORTYPE FillBufferDone(OMX_HANDLETYPE hComponent,
+	OMX_PTR pAppData, OMX_BUFFERHEADERTYPE* pBuffer)
+{
+	(void) hComponent;
+	(void) pAppData;
+	(void) pBuffer;
+	puts("FillBufferDone2");
```
alfredh: please dont use `puts` either. Use `debug` for this.

modules/omx/omx.c
```diff
@@ -0,0 +1,300 @@
...
+		"OMX.broadcom.video_render", 0, &callbacks);
+
+	if (err != OMX_ERROR_NONE) {
+		error("Failed to create OMX video_render component");
+	}
+	else {
+		info("created video_render component");
+	}
+
+	pthread_mutex_unlock(&st->omx_mutex);
```
alfredh: I dont think mutex lock/unlock is needed here. Can any other parts of the code
use the mutex until this function has returned ?

modules/omx/omx.c
```diff
@@ -0,0 +1,300 @@
...
+				st, portdef.nBufferSize);
+			if (err) {
+				error("OMX_AllocateBuffer failed: %d\n", err);
+				err = ENOMEM;
+				goto exit;
+			}
+		}
+	}
+
+	puts("omx_update_size: send to execute state");
```
alfredh: use `debug` instead

modules/omx/omx.h
```diff
@@ -0,0 +1,30 @@
+#include <IL/OMX_Core.h>
```
alfredh: add a doxygen header here

modules/omx/omx.h
```diff
@@ -0,0 +1,30 @@
...
+#include <IL/OMX_Video.h>
+#include <IL/OMX_Broadcom.h>
+#include <pthread.h>
+
+
+struct vidframe* frame;
```
alfredh: this one looks strange .. is it actually used ?

modules/omx/omx.h
```diff
@@ -0,0 +1,30 @@
...
+
+
+struct omx_state {
+	pthread_mutex_t omx_mutex;
+
+	OMX_HANDLETYPE video_render;
+	OMX_BUFFERHEADERTYPE** buffers;
+	int num_buffers;
+	int current_buffer;
```
alfredh: `current_buffer` is only set to 0 but never read. Intentional ?

alfredh:
I added some more review comments.

is it possible for me to compile this module, even if I dont own a Raspberry Pi ?
i.e. can I download the headers from somewhere ?

or perhaps someone else with a Raspberry can try to build it ?

there are a couple of doxygen warnings that needs to be fixed.
using doxygen version 1.8.13

```
$ make dox
/home/alfredh/tmp/baresip/modules/omx/omx.c:25: warning: missing title after \defgroup omx
/home/alfredh/tmp/baresip/modules/omx/module.c:2: warning: the name `module.c' supplied as the second argument in the \file statement matches the following input files:
   /home/alfredh/tmp/baresip/modules/omx/module.c
   /home/alfredh/tmp/baresip/src/module.c
Please use a more specific name by including a (larger) part of the path!
/home/alfredh/tmp/baresip/src/module.c:2: warning: the name `module.c' supplied as the second argument in the \file statement matches the following input files:
   /home/alfredh/tmp/baresip/modules/omx/module.c
   /home/alfredh/tmp/baresip/src/module.c
Please use a more specific name by including a (larger) part of the path!

```

modules/omx/module.c
strfry: Should i do this for all kinds of log messages (info/warn/error)?

modules/omx/module.c
strfry: Thanks, i changed this without testing.

modules/omx/omx.c
strfry: You are right. I'm not even sure why the locking code is there. Probably it's just a relict from the previous attempt to do the OMX calls in a separate thread (which might be necessary for implementation of codec or camera capture).
I guess it's safe to assume that baresip's viddisp callbacks don't overlap/multithread?
Later today i will try to remove the locks entirely and test it.

modules/omx/omx.h
strfry: Need to do more research on that. OpenMAX returns a certain amount of buffers for a given port definition, probably to do some kind of double/triple-buffering.
But it worked fine with just using one buffer, so i didn't bother to implement that yet. Might be necessary to support the bigger resolutions, where one frame spans across multiple OMX buffers.

strfry:
I added a new commit. The code now should work on Desktop Linux with libomxil-bellagio.
But the build system needs attention to handle the 2 different cases.
Can you compile the code now?

alfredh:
nice work @strfry :)

on my Debian Testing I installed libomxil-bellagio:

```
apt-get install libomxil-bellagio-dev
apt-get install libomxil-bellagio0-components-xvideo
```

and compiled it:

```
modules/omx/omx.c:98:2: warning: initialization from incompatible pointer type [-Wincompatible-pointer-types]
  &EventHandler,
  ^
modules/omx/omx.c:98:2: note: (near initialization for ‘callbacks.EventHandler’)
modules/omx/omx.c:99:2: warning: initialization from incompatible pointer type [-Wincompatible-pointer-types]
  &EmptyBufferDone,
  ^
modules/omx/omx.c:99:2: note: (near initialization for ‘callbacks.EmptyBufferDone’)
modules/omx/omx.c: In function ‘block_until_state_changed’:
modules/omx/omx.c:141:4: warning: implicit declaration of function ‘usleep’ [-Wimplicit-function-declaration]
    usleep(10000);
    ^~~~~~
modules/omx/omx.c:141:4: warning: nested extern declaration of ‘usleep’ [-Wnested-externs]
modules/omx/omx.c: In function ‘omx_display_disable’:
modules/omx/omx.c:179:9: warning: extra tokens at end of #endif directive [-Wendif-labels]
  #endif RASPBERRY_PI
         ^~~~~~~~~~~~
  CC [M]  build-x86_64/modules/omx/module.o
modules/omx/module.c: In function ‘omx_vidisp_display’:
modules/omx/module.c:96:3: warning: implicit declaration of function ‘memcpy’ [-Wimplicit-function-declaration]
   memcpy(&st->size, &frame->size, sizeof(struct vidsz));
   ^~~~~~
modules/omx/module.c:96:3: warning: incompatible implicit declaration of built-in function ‘memcpy’
modules/omx/module.c:96:3: note: include ‘<string.h>’ or provide a declaration of ‘memcpy’
modules/omx/module.c:105:14: warning: pointer of type ‘void *’ used in arithmetic [-Wpointer-arith]
   dest = buf + offset;
              ^
modules/omx/module.c:108:4: warning: implicit declaration of function ‘warn’ [-Wimplicit-function-declaration]
    warn("Too large frame for OMX Buffer size. Expected"
    ^~~~
modules/omx/module.c:108:4: warning: nested extern declaration of ‘warn’ [-Wnested-externs]
modules/omx/module.c:113:3: warning: incompatible implicit declaration of built-in function ‘memcpy’
   memcpy(dest, frame->data[i], plane_bytes);
   ^~~~~~
modules/omx/module.c:113:3: note: include ‘<string.h>’ or provide a declaration of ‘memcpy’
```

local testing with vidloop works:

```
$ ./baresip -mcairo.so -momx.so -e/vidloop
```

the patch is almost ready to be merge now. Just please fix the above warnings
and I will merge it.

strfry:
Just pushed fixes for the last warnings.
Unfortunately, i noticed that it breaks with -e "/vidloop vp8".
At least i'm able to debug this one without a Raspi now :)

alfredh:
that is great, should I merge it now ?

strfry:
I'm inclined to say yes. But maybe i should fix the build system first?
It needs some attention to automatically select between PC and Raspberry Pi Build.

modules/omx/omx.h
```diff
@@ -5,11 +5,26 @@
...
+#define OMX_VERSION 0x01010101
+#define OMX_ERROR_NONE 0
+#endif
+
 #include <pthread.h>
+#include <stdint.h>
+#include <string.h>
 
+/* Needed for usleep to appear */
+#define _BSD_SOURCE
```
alfredh: you can also use sleep functions from re_sys.h:

```
void sys_usleep(unsigned int us);
void sys_msleep(unsigned int ms)
```

modules/omx/omx.c
```diff
@@ -0,0 +1,336 @@
...
+	return err;
+}
+
+
+int omx_display_input_buffer(struct omx_state* st,
+	void** pbuf, uint32_t* plen)
+{
+	pthread_mutex_lock(&st->omx_mutex);
+
+	if (!st->buffers) return EINVAL;
```
alfredh: returning here will leave the omx_mutex in a locked state.

strfry:
I removed all pthread usages. There is no multithreading happening in this code anyway.

alfredh:
hi @strfry I have merged the PR to master.

many thanks for your contribution, this is a piece of very good work.

You are welcome to submit more PRs for the OMX module (or other things).
even small patches are welcome :)

PR #231, opened 2017-04-04
Add gapless repeat for tone playback

janh:
This allows to play call progress tones having defined intervals without introducing redundancy in the audio file. This is especially useful for a continuous dial tone or busy tones with a short pause between tones.

### DIFF
--- a/src/play.c
+++ b/src/play.c
@@ -86,7 +86,7 @@ static void write_handler(int16_t *sampv, size_t sampc, void *arg)
 		left = mbuf_get_left(play->mb);
 		count = (left > sz - pos) ? sz - pos : left;
 
-		(void)mbuf_read_mem(play->mb, (void *)(sampv + pos), count);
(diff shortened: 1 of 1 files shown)


alfredh:
I have a couple of observations:

- the current code has a gap of 2 seconds between repeated tones. This was added
intentionally. if some tones should be played with 0seconds gap and other tones with
2 seconds gap, should we extend the api with a new parameter ?

- while testing some tones in the share/ folder with alsa driver, baresip crashed
  inside the play.c file. I can fetch the backtrace later, but can you please try to run your
 code with valgrind and check for memory corruptions ?



/Alfred

janh:
I have fixed the memory corruption which was due to incorrect pointer arithmetic.

As for an additional parameter: I think being able to specify the length of the pause could be useful if it is counted from the actual end of the tone. As I understand the current implementation, the gap length can vary between 1 and 2 seconds (depending on the length of the tone) as it uses polling.

alfredh:
Thank you, this is a very nice improvement.

PR #235, opened 2017-04-16
Update for re/rem/baresip windows projects 

mikhail-barg:
I'm planning to use baresip under windows, so I've updated the MSVS projects in whole baresip/rem/re combo to vs2015 (and to build at all). Also I've changed project filters to reflect the source structure.

The project is set up so that it does not require anything external, except of the following files, that should be placed in the `../misc` folder:
* unistd.h: http://stackoverflow.com/questions/341817/is-there-a-replacement-for-unistd-h-for-windows-visual-c
* getopt.h/.cpp: https://gist.github.com/ashelly/7776712
* qedit.h: https://github.com/thegrandpoobah/dsgrab/blob/master/qedit.h

I've decided to keep `dirent` dependency, so it should be checked out to `../dirent` folder from https://github.com/tronkko/dirent

To avoid external dependencies I've disabled a number of modules, including `avcodec`, which requires `ffmpeg`.

see also:
* https://github.com/alfredh/baresip/issues/212
* https://github.com/creytiv/re/pull/55
* https://github.com/creytiv/rem/pull/11


mk/win32/static.c
```diff
@@ -7,7 +7,7 @@ extern const struct mod_export exports_wincons;
 extern const struct mod_export exports_g711;
 extern const struct mod_export exports_winwave;
 extern const struct mod_export exports_dshow;
-extern const struct mod_export exports_avcodec;
+//extern const struct mod_export exports_avcodec;
```
alfredh: please delete the whole line, instead of commenting it out.

mk/win32/static.c
```diff
@@ -22,7 +22,7 @@ extern const struct mod_export exports_vumeter;
 
 const struct mod_export *mod_table[] = {
 	&exports_wincons,
-	&exports_avcodec,
+//	&exports_avcodec,
```
alfredh: please delete the whole line, instead of commenting it out.

modules/dshow/dshow.cpp
```diff
@@ -43,10 +45,11 @@ IID IID_ISampleGrabberCB = {
 
 #include "qedit.h"
 
+/*
 const CLSID CLSID_SampleGrabber = { 0xc1f400a0, 0x3f08, 0x11d3,
   { 0x9f, 0x0b, 0x00, 0x60, 0x08, 0x03, 0x9e, 0x37 }
 };
-
+*/
```
alfredh: this was added in commit 2076043d899a575c524a9c7132226d4461836aaf

is it safe to delete it ?

alfredh:
many thanks for this PR :)

a couple of nits, apart from that it looks good.

modules/dshow/dshow.cpp
mikhail-barg: Really not sure on this one. 
I am using `qedit.h` from here: https://github.com/thegrandpoobah/dsgrab/blob/master/qedit.h
and it has `CLSID_SampleGrabber` defined. The thing is that even though this is an extract from DirectX SDK, in this repo it has GPL preamble in the file. So I'd prefer not to use it at all.

As far as I understand `qedit.h` was a part of Windows SDK/ DirectX SDK, but now it's removed (see https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/2ab5c212-5824-419d-b5d9-7f5db82f57cd/qedith-missing-in-current-windows-sdk-v70?forum=windowsdirectshowdevelopment or https://blogs.msdn.microsoft.com/karinm/2010/01/15/where-is-qedit-h/)

Do you think we could incorporate required CLSIDs and other stuff directly into `dshow.cpp` or new header file in the project?

mk/win32/static.c
mikhail-barg: done

mikhail-barg:
removed comments, an unused .lib.
Now what's remain is to decide what to do with `qedit.h`

alfredh:
the baresip project should not contain copies of platform header files,
unless it is absolutely necessary and the file has a liberal license (BSD).

distributing copies of platform files quickly becomes a maintainance burden :)

PR #241, opened 2017-04-29
Set correct clock rate for telephone events

janh:
For telephone events the same clock rate as for audio has to be used, as both share the timestamp base.

This depends on the corresponding pull request for libre.


GGGO:
That could explains why sometimes my DTMF don't work.

alfredh:
which SIP endpoints have you tested this against?

GGGO:
Using baresip on windows with asterisk.

PR #266, opened 2017-06-06
added support for UPDATE method

juha-h:
Another try.  Let's see if Travis likes it.

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -1858,7 +1858,7 @@ struct list *uag_list(void)
 const char *uag_allowed_methods(void)
 {
 	return "INVITE,ACK,BYE,CANCEL,OPTIONS,REFER,"
-		"NOTIFY,SUBSCRIBE,INFO,MESSAGE";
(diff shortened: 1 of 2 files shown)


src/ua.c
```diff
@@ -1858,7 +1858,7 @@ struct list *uag_list(void)
 const char *uag_allowed_methods(void)
 {
 	return "INVITE,ACK,BYE,CANCEL,OPTIONS,REFER,"
-		"NOTIFY,SUBSCRIBE,INFO,MESSAGE";
+		"NOTIFY,SUBSCRIBE,INFO,MESSAGE.UPDATE";
```
alfredh: small nit; should be comma instead of dot :)

juha-h:
Should be fixed now.

alfredh:
Hi @juha-h 

the patch for adding UPDATE in libre was closed:

https://github.com/creytiv/re/pull/58

so I will close this one also.

We are still interested in adding support for SIP UPDATE at some point,
but unfortunately we have no time to implement it right now ..

Alfred

Issue #262, opened 2017-05-30
ZRTP Error after a while


MobiSciLab:
When I make a video call, the call works for about 2 minutes then, It cannot transmit data anymore and ZRTP reports errors:

05-30 09:49:13.795 19845-20022/mobisci-lab W/CMV: zrtp: zrtp_process_srtp: 4
05-30 09:49:13.795 19845-20022/mobisci-lab I/CMV:  [zrtp protoco]: ERROR! Decrypt failed. ID=20:MULTI s=SRTP authentication failure (RTP size=1044 ssrc=1874372 seq=1040/1040 pt=0)
05-30 09:49:13.795 19845-20022/mobisci-lab W/CMV: zrtp: zrtp_process_srtp: 4
05-30 09:49:13.795 19845-20022/mobisci-lab I/CMV:  [zrtp protoco]: ERROR! Decrypt failed. ID=20:MULTI s=SRTP authentication failure (RTP size=680 ssrc=1874372 seq=676/676 pt=0)
05-30 09:49:13.795 19845-20022/mobisci-lab W/CMV: zrtp: zrtp_process_srtp: 4
05-30 09:49:13.795 19845-20022/mobisci-lab I/CMV:  [zrtp protoco]: ERROR! Decrypt failed. ID=19:DH s=SRTP authentication failure (RTP size=180 ssrc=510657 seq=176/176 pt=0)
05-30 09:49:13.795 19845-20022/mobisci-lab W/CMV: zrtp: zrtp_process_srtp: 4
05-30 09:49:13.795 19845-20022/mobisci-lab I/CMV:  [zrtp protoco]: ERROR! Decrypt failed. ID=19:DH s=SRTP authentication failure (RTP size=180 ssrc=510817 seq=176/176 pt=0)
05-30 09:49:13.795 19845-20022/mobisci-lab W/CMV: zrtp: zrtp_process_srtp: 4

juha-h:
MobiSciLab writes:

> When I make a video call, the call works for about 2 minutes then, It
> cannot transmit data anymore and ZRTP reports errors:

Does the error only appear in video calls?  I tried with audio only
calls and didn't get it.

-- Juha

MobiSciLab:
Yes, it happens on audio calls also. We are using android devices.

juha-h:
MobiSciLab writes:

> Yes, it happens on audio calls also. We are using android devices.

I don't have such software to try.  ZRTP audio calls work fine in Debian
Linux.

[...]

PR #267, opened 2017-06-08
zrtp: check for RTP packet in send handler

alfredh:
non-RTP packets such as STUN, TURN, ICE might be shared on the
same UDP socket, and these packets should not be processed by ZRTP.

ref #262 

@juha-h please review and merge to master if OK.

### DIFF
--- a/modules/zrtp/zrtp.c
+++ b/modules/zrtp/zrtp.c
@@ -50,6 +50,19 @@ static zrtp_config_t zrtp_config;
 static zrtp_zid_t zid;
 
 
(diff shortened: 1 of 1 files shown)


juha-h:
Alfred E. Heggestad writes:
> non-RTP packets such as STUN, TURN, ICE might be shared on the
> same UDP socket, and these packets should not be processed by ZRTP.
> 
> ref #262 
> 
> @juha-h please review and merge to master if OK.

I'll merge after @edgeworks-vucao has tested the patch.

-- Juha

edgeworks-vucao:
@juha-h , I tested the patch, it works well.

PR #276, opened 2017-07-02
Presence: Handle contacts added at run-time

strfry:
Preliminary work for avahi/Zeroconf support

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -283,18 +283,26 @@ enum presence_status {
 	PRESENCE_BUSY
 };
 
+
+struct contact;
+typedef void (contact_update_h)(struct contact *c, bool remove, void *arg);
+
 struct contacts {
 	struct list cl;
 	struct hash *cht;
+
+	contact_update_h *handler;
+	void* handler_arg;
 };
 
-struct contact;
 
 int  contact_init(struct contacts *contacts);
 void contact_close(struct contacts *contacts);
 int  contact_add(struct contacts *contacts,
 		 struct contact **contactp, const struct pl *addr);
 void contact_remove(struct contacts *contacts, struct contact *c);
+void contact_set_update_handler(struct contacts *contacs,
+				contact_update_h *updateh, void *arg);
 int  contacts_print(struct re_printf *pf, const struct contacts *contacts);
 enum presence_status contact_presence(const struct contact *c);
 void contact_set_presence(struct contact *c, enum presence_status status);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -295,6 +295,7 @@ void contact_close(struct contacts *contacts);
 int  contact_add(struct contacts *contacts,
 		 struct contact **contactp, const struct pl *addr);
 int  contacts_print(struct re_printf *pf, const struct contacts *contacts);
+enum presence_status contact_presence(const struct contact *c);
 void contact_set_presence(struct contact *c, enum presence_status status);
 bool contact_block_access(const struct contacts *contacts, const char *uri);
 struct contact  *contact_find(const struct contacts *contacts,
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -294,6 +294,7 @@ int  contact_init(struct contacts *contacts);
 void contact_close(struct contacts *contacts);
 int  contact_add(struct contacts *contacts,
 		 struct contact **contactp, const struct pl *addr);
+void contact_remove(struct contacts *contacts, struct contact *c);
 int  contacts_print(struct re_printf *pf, const struct contacts *contacts);
 enum presence_status contact_presence(const struct contact *c);
 void contact_set_presence(struct contact *c, enum presence_status status);
--- a/modules/presence/subscriber.c
+++ b/modules/presence/subscriber.c
@@ -280,6 +280,43 @@ static int presence_alloc(struct contact *contact)
 	return 0;
 }
 
+static void contact_handler(struct contact *contact,
+				bool removed, void *arg)
+{
+	struct le *le;
+	struct pl val;
+	struct presence *pres = NULL;
+	struct sip_addr *addr = contact_addr(contact);
+	(void)arg;
+
+	if (0 == msg_param_decode(&addr->params, "presence", &val) &&
+				0 == pl_strcasecmp(&val, "p2p")) {
+		if (!removed) {
+			if (presence_alloc(contact) != 0) {
+				error("presence: presence_alloc failed\n");
+				return;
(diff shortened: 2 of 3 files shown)


include/baresip.h
```diff
@@ -283,18 +283,29 @@ enum presence_status {
...
 int  contact_init(struct contacts *contacts);
 void contact_close(struct contacts *contacts);
 int  contact_add(struct contacts *contacts,
 		 struct contact **contactp, const struct pl *addr);
+int contact_remove(struct contacts* contacts, struct contact* c);
+
+void contact_set_update_handler(struct contacts* contacs,
+				contact_update_h* updateh, void* arg);
 int  contacts_print(struct re_printf *pf, const struct contacts *contacts);
+enum presence_status contact_presence(struct contact* c);
```
alfredh: the input parameter `struct contact *c` can be made `const` , as the function
does not modify state.

modules/presence/subscriber.c
```diff
@@ -280,6 +280,45 @@ static int presence_alloc(struct contact *contact)
 	return 0;
 }
 
+static void presence_dealloc(struct presence* pres)
+{
+	list_unlink(&pres->le);
+	tmr_cancel(&pres->tmr);
+	mem_deref(pres->contact);
+}
```
alfredh: this looks a bit suspicious. the pointer to `pres->contact` must be set to `NULL`
after `mem_deref`.

perhaps it is better to deref the whole `struct presence` object ?

modules/presence/subscriber.c
```diff
@@ -280,6 +280,45 @@ static int presence_alloc(struct contact *contact)
...
+	struct le* le;
+	struct presence *pres;
+
+	if (!remove) {
+		struct sip_addr *addr = contact_addr(contact);
+		struct pl val;
+
+		if (0 == msg_param_decode(&addr->params, "presence", &val) &&
+		0 == pl_strcasecmp(&val, "p2p")) {
+			presence_alloc(contact);
```
alfredh: return value from `presence_alloc` must be checked and handled.

modules/presence/subscriber.c
```diff
@@ -302,6 +341,8 @@ int subscriber_init(void)
 
 	info("Subscribing to %u contacts\n", list_count(&presencel));
 
+	contact_set_update_handler(contacts, contact_handler, (void*)0x1337);
+
```
alfredh: this function must also be called with a NULL handler from `subscriber_close`.

otherwise, after the module has been unloaded, the `contact.c` may still call
a function that does not exist (will lead to a crash).

src/contact.c
```diff
@@ -103,6 +106,36 @@ int contact_add(struct contacts *contacts,
...
+	if (!contacts || !contact)
+		return EINVAL;
+
+	if (contacts->handler)
+		contacts->handler(contact, true, contacts->handler_arg);
+
+	hash_unlink(&contact->he);
+	list_unlink(&contact->le);
+
+	return 0;
```
alfredh: I think it is better that this function returns `void` because it cannot fail.

src/contact.c
```diff
@@ -103,6 +106,36 @@ int contact_add(struct contacts *contacts,
...
+}
+
+void contact_set_update_handler(struct contacts* contacts,
+				contact_update_h* updateh, void* arg)
+{
```
alfredh: check `contacts`, if NULL return

src/contact.c
```diff
@@ -159,6 +192,13 @@ void contact_set_presence(struct contact *c, enum presence_status status)
 	c->status = status;
 }
 
+enum presence_status contact_presence(struct contact* c)
+{
+	if (!c)
+		return -1;
```
alfredh: return `PRESENCE_UNKNOWN` instead of -1.

alfredh:
@strfry thanks for this nice patch

please see inline for some review comments.

I tested the patch locally. After adding and removing a couple of contacts,
baresip crashes in the presence module. Can you try to do the same and
see if you get the crash?

there are also some ccheck errors that must be fixed:

```
ccheck.py 
./modules/presence/subscriber.c:307: else: ending if bracket should be on previous line
./modules/presence/subscriber.c:319: else: ending if bracket should be on previous line
./modules/presence/subscriber.c:320: line is too wide (80 - max 79)
```

Thanks,

/Alfred

modules/presence/subscriber.c
strfry: You are right, the presence object must be deref'd as well.
But this is still necessary, since `pres->contact` is `mem_ref`'d in `presence_alloc`.

strfry:
All the issues from review should be fixed now.
How did you add and remove contacts? I don't have a test case for that, only the experimental avahi module, and couldn't reproduce the issue with one other peer connecting and disconnecting.
Another crash during shutdown was fixed by setting the contact_handler to NULL.

include/baresip.h
```diff
@@ -283,18 +283,28 @@ enum presence_status {
...
+	void* handler_arg;
 };
 
-struct contact;
 
 int  contact_init(struct contacts *contacts);
 void contact_close(struct contacts *contacts);
 int  contact_add(struct contacts *contacts,
 		 struct contact **contactp, const struct pl *addr);
+void contact_remove(struct contacts* contacts, struct contact* c);
```
alfredh: please note that the pointer star `*` should be oriented to the right, just like
the rest of the source code. (multiple places).

modules/presence/subscriber.c
```diff
@@ -280,6 +280,52 @@ static int presence_alloc(struct contact *contact)
...
+		for (le = list_head(&presencel); le; le = le->next) {
+			pres = (struct presence*)le->data;
+			if (pres->contact == contact) {
+				break;
+			}
+			pres = NULL;
+		}
+
+		if (pres) {
+			presence_dealloc(pres);
```
alfredh: perhaps it is simpler to just do:

```
    mem_deref(pres);
```

strfry: Now i got it, didn't see that `destructor`, thanks! ;)

strfry:
I fixed a potential problem in the presence `contact_handler` (check for presence=p2p not handled consistently), and a memory leak in `contact_remove`.
Can you check again?
Hopefully i found all the *alignment issues.  ;)

Thanks,
strfry

alfredh:
it looks good now.

there are still a couple of ccheck warnings, please fix them and I will merge.



/Alfred

alfredh:
@strfry merged to master. Thanks Jonathan :)

PR #279, opened 2017-07-09
Make sdl2 Fullscreen Mode configurable

strfry:
Hello Alfred,

In my application, the video output must always be fullscreen, so i added a global configuration option to specify that as default. This is currently only obeyed by the sdl2 module, but could be implemented by others?

Another Commit changes the SDL flag to SDL_WINDOW_FULLSCREEN_DESKTOP, which means that SDL won't switch resolution, but create instead will create a borderless window with the size of the desktop.
I don't know if this is preferable to everyone, maybe it should be configurable in some way, too?

Thanks for your review,
Jonathan

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -776,7 +776,8 @@ struct vidisp_st;
 
 /** Video Display parameters */
 struct vidisp_prm {
-	void *view;  /**< Optional view (set by application or module) */
+	void *view;       /**< Optional view (set by application or module) */
+	bool fullscreen;  /**< Enable fullscreen display                    */
 };
 
 typedef void (vidisp_resize_h)(const struct vidsz *size, void *arg);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -216,6 +216,7 @@ struct config_video {
 	unsigned width, height; /**< Video resolution               */
 	uint32_t bitrate;       /**< Encoder bitrate in [bit/s]     */
 	uint32_t fps;           /**< Video framerate                */
+	bool fullscreen;        /**< Enable fullscreen display      */
 };
 #endif
 
--- a/modules/sdl2/sdl.c
+++ b/modules/sdl2/sdl.c
@@ -134,7 +134,6 @@ static int alloc(struct vidisp_st **stp, const struct vidisp *vd,
 	int err = 0;
 
 	/* Not used by SDL */
-	(void)prm;
 	(void)dev;
 	(void)resizeh;
 	(void)arg;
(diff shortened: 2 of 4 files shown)


alfredh:
Hi @strfry many thanks for this


in general this looks good. my only comment is that the sdl module should not
access the global config struct. it should rather have the relevant config "passed down"
from the core to the module via `vidisp` parameters. In order to facilitate for this,
I have added a fullscreen flag to vidisp:

https://github.com/alfredh/baresip/commit/49df75bec3fb305a4c2f50d966ea15c3f1dd52b9

```
/** Video Display parameters */
struct vidisp_prm {
	void *view;       /**< Optional view (set by application or module) */
	bool fullscreen;  /**< Enable fullscreen display                    */
};
```

the fullscreen flag should then be read from config_video (in core) and can
now be used by any vidisp module wanting to implement it.

also, please note that baresip supports multiple video-calls simultaneously.
then each call will have it's own video display window. that is why the `f` key
to toggle fullscreen is handled by SDL2 module, only the active window that
received the `f` event will be fullscreened.

adding the global `video_fullscreen        yes` for the initial state is a good idea, I think.

could you please rebase your patch on top of git@HEAD ?


/Alfred

strfry:
Good point about the multiple video streams.
I changed it back to keep that fullscreen flag in sdl2.c, added initialization of that setting in src/video.c:set_vidisp, and added a default option to the config template.

Also I noticed there already is a fullscreen flag `struct vrx`, along with the `int orient` does not seem to be used?

include/baresip.h
```diff
@@ -216,6 +216,7 @@ struct config_video {
 	unsigned width, height; /**< Video resolution               */
 	uint32_t bitrate;       /**< Encoder bitrate in [bit/s]     */
 	uint32_t fps;           /**< Video framerate                */
+	bool fullscreen;		/**< Enable fullscreen display      */
```
alfredh: the doxygen comments should be aligned.

modules/sdl2/sdl.c
```diff
@@ -24,7 +24,7 @@ struct vidisp_st {
 	SDL_Texture *texture;           /**< Texture for pixels    */
 	struct vidsz size;              /**< Current size          */
 	enum vidfmt fmt;                /**< Current pixel format  */
-	bool fullscreen;                /**< Fullscreen flag       */
+	bool fullscreen;
```
alfredh: this hunk removes the comment, please revert this part.

alfredh:
looks good! just a couple of nits, please fix and I will merge it.

PR #285, opened 2017-07-21
Hack to automatically enable PulseAudio's Echo Cancellation

strfry:
Hello Alfred,

This Pull Request is more a request for comments than a real solution.

I wanted to pull the attention to PulseAudios AEC capabilities: [Reddit Thread](https://www.reddit.com/r/linux/comments/2yqfqp/just_found_that_pulseaudio_have_noise/)

There are other ways to do that via the PulseAudio API, but that does not seem to be possible without moving away from the "Simple API", which basically means rewriting the entire module.

Something else that bugs me about this approach, is that the same environment variable would also work for the ALSA module, if it is routed through PulseAudio. 

Should this option generally be configurable?
My goal is simply to avoid external configuration of baresip behaviour through environment variables.

Thanks for your time,
Jonathan

### DIFF
--- a/modules/pulse/player.c
+++ b/modules/pulse/player.c
@@ -3,6 +3,10 @@
  *
  * Copyright (C) 2010 - 2016 Creytiv.com
  */
+
+/* Note: Macro needed for setenv() declaration to appear */
+#define _DEFAULT_SOURCE
+#include <stdlib.h>
 #include <pulse/pulseaudio.h>
 #include <pulse/simple.h>
 #include <pthread.h>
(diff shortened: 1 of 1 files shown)


alfredh:
I think a better solution is to create a new module e.g. webrtc_aec that is using the same
library that Pulse is using. Then it can be used with any audio driver.

Also, could you just start baresip like this:

```
$ PULSE_PROP=filter.want=echo-cancel baresip
```

strfry:
I agree, this would be the best solution. I started to work on this 2 months ago, but it doesn't properly work yet
https://github.com/strfry/baresip/commit/b2dff31db809846a71e10cf2e44e87278dbf37c8

PR #307, opened 2017-09-30
modules/zrtp: encrypt/decrypt RTCP packets

glenvt18:
Incoming RTCP packet's encrypt-flag is checked in libzrtp.

### DIFF
--- a/modules/zrtp/zrtp.c
+++ b/modules/zrtp/zrtp.c
@@ -7,6 +7,7 @@
 #include <baresip.h>
 #include <zrtp.h>
 
+#include <string.h>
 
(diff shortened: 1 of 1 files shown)


alfredh:
many thanks for the patch, it is very useful.

here is an unrelated change:

BIN  modules/zrtp/.zrtp.c.swp


/alfred

glenvt18:
@alfredh updated

glenvt18:
UPDATE. Better packet filtering, more informative logging on send/receive.
@alfredh mind a review?

alfredh:
many thanks :)

PR #311, opened 2017-10-11
modules/zrtp: add signaling hash support

glenvt18:
Use SDP attribute a=zrtp-hash to protect the Hello message (RFC 6189, section 8.1).
https://tools.ietf.org/html/rfc6189#section-8

This adds an extra level of protection especially when using a secure signaling channel. When an attack is detected, all incoming and outgoing UDP packets are being dropped, and the call is aborted.

If, for whatever reason, this feature is undesirable (for example, if one wants to hide the UA's ZRTP capability), it can be disabled by the configuration file option.

### DIFF
--- a/modules/zrtp/zrtp.c
+++ b/modules/zrtp/zrtp.c
@@ -26,6 +26,13 @@
  * Thanks:
  *
  *   Ingo Feinerer
+ *
+ * Configuration options:
+ *
+ \verbatim
+  zrtp_hash       {yes,no}   # Enable SDP zrtp-hash (recommended)
+ \endverbatim
+ *
  */
 
 
@@ -35,10 +42,14 @@ enum {
 
 struct menc_sess {
 	zrtp_session_t *zrtp_session;
+	menc_error_h *errorh;
+	void *arg;
+	struct tmr abort_timer;
+	int err;
 };
 
 struct menc_media {
-	const struct menc_sess *sess;
+	struct menc_sess *sess;
 	struct udp_helper *uh_rtp;
 	struct udp_helper *uh_rtcp;
 	struct sa raddr;
@@ -51,6 +62,10 @@ struct menc_media {
 static zrtp_global_t *zrtp_global;
 static zrtp_config_t zrtp_config;
 static zrtp_zid_t zid;
+
+
+/* RFC 6189, section 8.1. */
+static bool use_sig_hash = true;
 
 
 enum pkt_type {
@@ -93,6 +108,8 @@ static void session_destructor(void *arg)
 {
 	struct menc_sess *st = arg;
 
+	tmr_cancel(&st->abort_timer);
+
 	if (st->zrtp_session)
 		zrtp_session_down(st->zrtp_session);
 }
@@ -112,6 +129,32 @@ static void media_destructor(void *arg)
 }
 
 
+static void abort_timer_h(void *arg)
+{
+	struct menc_sess *sess = arg;
+
+	if (sess->errorh) {
+		sess->errorh(sess->err, sess->arg);
+		sess->errorh = NULL;
+	}
+}
+
+
+static void abort_call(struct menc_sess *sess)
+{
+	if (!sess->err) {
+		sess->err = EPIPE;
+		tmr_start(&sess->abort_timer, 0, abort_timer_h, sess);
+	}
+}
+
+
+static bool drop_packets(const struct menc_media *st)
+{
+	return (st)? st->sess->err != 0 : true;
+}
+
+
 static bool udp_helper_send(int *err, struct sa *dst,
 			    struct mbuf *mb, void *arg)
 {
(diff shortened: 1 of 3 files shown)


modules/zrtp/zrtp.c
```diff
@@ -35,10 +42,14 @@ enum {
 
 struct menc_sess {
 	zrtp_session_t *zrtp_session;
+	menc_error_h *errorh;
+	void *arg;
+	struct tmr abort_timer;
```
alfredh: you should call `tmr_cancel` on this new timer in the destructor, even if it is not fired.

modules/zrtp/zrtp.c
```diff
@@ -112,6 +127,31 @@ static void media_destructor(void *arg)
 }
 
 
+static void abort_timer_h(void *arg)
+{
+	const struct menc_sess *sess = arg;
+
+	if (sess->errorh) {
+		sess->errorh(sess->err, sess->arg);
+	}
```
alfredh: perhaps also set `errorh` to `NULL` so that the callback handler is only called once.

modules/zrtp/zrtp.c
```diff
@@ -198,15 +244,74 @@ static bool udp_helper_recv(struct sa *src, struct mbuf *mb, void *arg)
 }
 
 
+static void sig_hash_encode(struct zrtp_stream_t *stream,
+                         struct sdp_media *m)
```
alfredh: this function should return `int`

if the function is successful, it should return zero.
for any other errors, it should return `errno`.

modules/zrtp/zrtp.c
```diff
@@ -198,15 +244,74 @@ static bool udp_helper_recv(struct sa *src, struct mbuf *mb, void *arg)
...
+{
+	char buf[ZRTP_SIGN_ZRTP_HASH_LENGTH + 1];
+	zrtp_status_t s;
+	int err;
+
+	if (!use_sig_hash)
+		return;
```
alfredh: I think it is better to move this check to where the function is invoked.

then we also avoid the negated if statement

modules/zrtp/zrtp.c
```diff
@@ -198,15 +244,74 @@ static bool udp_helper_recv(struct sa *src, struct mbuf *mb, void *arg)
...
+
+	attr_val = sdp_media_rattr(m, "zrtp-hash");
+	if (!attr_val)
+		return;
+
+	err = re_regex(attr_val, strlen(attr_val),
+	               "[0-9]+.[0-9]2 [0-9a-fA-F]+",
```
alfredh: `re_regex` is case insensitive, so you can drop `A-F` if you want

alfredh:
Many thanks for this patch. The code is very high quality and the patch
is very easy to read.

Some review comments inline.

* is it possible to test this feature with any free SIP clients ?

* compiler warning on OSX 10.12:

```
  CC [M]  build-x86_64/modules/zrtp/zrtp.o
modules/zrtp/zrtp.c:302:51: warning: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
        s = zrtp_signaling_hash_set(stream, hash.p, hash.l);
            ~~~~~~~~~~~~~~~~~~~~~~~                 ~~~~~^
1 warning generated.
  LD [M]  zrtp.so
```

@juha-h do you have any comments about this patch?


on a side note I was thinking about adding a new callback handler to the "Mediaenc" API.
the callback handler can be some kind of status handler from the module to the
baresip core, that will be called when the encryption state changes. Some modules
can then call the handler, for example SDES module can say "MENC_ENCRYPTED"
and ZRTP can say "MENC_ENCRYPTED_VERIFIED".


/Alfred

glenvt18:
@alfredh Thanks for the review. I did forget to call tmr_cancel. I've added your changes and updated this PR.

This feature has been tested with baresip, linphone and sip:4444@sip2sip.info (supports zrtp and dtls). It's a part of ZRTP RFC, but is not mandatory. linphone doesn't send zrtp-hash if there is no zrtp-hash in the INVITE message. If it gets a corrupted hash, it doesn't send commit and keeps the call in the clear (not encrypted) state. I think aborting the call is more appropriate. You won't get a hash mismatch under normal circumstances. If you think it's a DoS attack, you can set `zrtp_hash no` in the config file.

The idea with a callback is nice. It would help to display encryption status of the current call and the calls in the list. It's especially useful when the screen is flooded with warnings about malformed  RTP packets.

juha-h:
glenvt18 writes:

> The idea with a callback is nice. It would help to display encryption
> status of the current call and the calls in the list. It's especially
> useful when the screen is flooded with warnings about malformed  RTP
> packets.

I recently added a console message that tells if the call is zrtp
verified when it starts, but displaying the status during the whole call
would be better.

Also, it would be good if user could verify the remote party zid with a
command, such as '/zrtp ok', that does not require copy/paste of the zid
from the console message.

-- Juha

PR #313, opened 2017-10-17
debian build now builds also libbaresip and libbaresip-dev packages

juha-h:
In this branch debian build creates also libbaresip and libbaresip-dev packages.  The former includes dynamic library libbaresip.so.  The latter depends on the former and itself includes baresip.h include file as well as static library libbaresip.a.

Is something missing or wrong?


alfredh:
thanks. I have tested this patch by creating Debian packages (on Debian Testing).

my only comment is that we should perhaps sync with the Debian guys, who I believe
are maintaining their own set of scripts.

juha-h:
Alfred E. Heggestad writes:

> thanks. I have tested this patch by creating Debian packages (on
> Debian Testing).

I tested it on Debian stable (Stretch).

> my only comment is that we should perhaps sync with the Debian guys,
> who I believe are maintaining their own set of scripts.

Fine if they come up with better solution.

-- Juha

PR #314, opened 2017-10-23
modules/gzrtp: new module using GNU ZRTP C++ library

glenvt18:
I created this module after having regular ZRTP issues with libzrtp and linphone. It uses Werner Dittmann's [ZRTPCCP](https://github.com/wernerd/ZRTPCPP) library as a backend.

Pros:
* maintained
* reviewed
* widely used (Twinkle, PJSIP, CSipSimple etc.)
* flexible
* works with linphone!

Cons:
* more glue code
* C++ (if it matters)

The module is statically linked with libzrtpcppcore.a which should be built from source using the following commands:
```sh
git clone https://github.com/wernerd/ZRTPCPP.git
cd ZRTPCPP
mkdir build
cd build
cmake -DCMAKE_POSITION_INDEPENDENT_CODE=1 -DCORE_LIB=1 -DSDES=1 -DBUILD_STATIC=1 ..
make
```

Then make baresip with
```
make USE_GZRTP=1 ZRTP=<ZRTPCPP_DIR> <OTHER OPTIONS>
```

And add `module gzrtp.so` to the configuration file.

This module has been tested with linphone and baresip (zrtp/gzrtp). Interoperability test results with other UAs are welcome.

Signaling hash is not supported yet.

There is an minor [issue](https://github.com/wernerd/ZRTPCPP/issues/39) with libzrtp which can be fixed in either library.

### DIFF
--- a/modules/gzrtp/gzrtp.cpp
+++ b/modules/gzrtp/gzrtp.cpp
@@ -1,0 +1,220 @@
+/**
+ * @file gzrtp.cpp  GNU ZRTP: Media Path Key Agreement for Unicast Secure RTP
+ *
+ * Copyright (C) 2010 - 2017 Creytiv.com
+ */
+#include <stdint.h>
+
+#include <re.h>
+#include <baresip.h>
+
+#include <string.h>
+
+#include <libzrtpcpp/ZRtp.h>
+
+#include "session.h"
+#include "stream.h"
+
+
+/**
+ * @defgroup gzrtp gzrtp
+ *
+ * ZRTP: Media Path Key Agreement for Unicast Secure RTP
+ *
+ * Experimental support for ZRTP
+ *
+ *     See http://tools.ietf.org/html/rfc6189
+ *
+ *
+ *     This module is using GNU ZRTP C++ library
+ *
+ *     https://github.com/wernerd/ZRTPCPP
+ *
+ * Configuration options:
+ *
+ \verbatim
+  zrtp_parallel   {yes,no}   # Start all streams at once
+ \endverbatim
+ *
+ */
+
+
+static ZRTPConfig *s_zrtp_config = NULL;
+
+
+struct menc_sess {
+	Session *session;
+};
+
+
+struct menc_media {
+	Stream *stream;
+};
+
+
+static void session_destructor(void *arg)
+{
+	struct menc_sess *st = (struct menc_sess *)arg;
+
+	delete st->session;
+}
+
+
+static void media_destructor(void *arg)
+{
+	struct menc_media *st = (struct menc_media *)arg;
+
+	delete st->stream;
+}
+
+
+static int session_alloc(struct menc_sess **sessp, struct sdp_session *sdp,
+                         bool offerer, menc_error_h *errorh, void *arg)
+{
+	struct menc_sess *st;
+	(void)offerer;
+	(void)errorh;
+	(void)arg;
+	int err = 0;
+
+	if (!sessp || !sdp)
+		return EINVAL;
+
+	st = (struct menc_sess *)mem_zalloc(sizeof(*st), session_destructor);
+	if (!st)
+		return ENOMEM;
+
+	st->session = new Session(*s_zrtp_config);
+	if (!st->session)
+		err = ENOMEM;
+
+	if (err)
+		mem_deref(st);
+	else
+		*sessp = st;
+
+	return err;
+}
+
+
+static int media_alloc(struct menc_media **stp, struct menc_sess *sess,
+                       struct rtp_sock *rtp,
+                       int proto, void *rtpsock, void *rtcpsock,
+                       struct sdp_media *sdpm)
+{
+	struct menc_media *st;
+	int err = 0;
+	StreamMediaType med_type;
+	const char *med_name;
+
+	if (!stp || !sess || !sess->session || proto != IPPROTO_UDP)
+		return EINVAL;
+
+	st = *stp;
+	if (st)
+		goto start;
+
+	st = (struct menc_media *)mem_zalloc(sizeof(*st), media_destructor);
+	if (!st)
+		return ENOMEM;
+
+	med_name = sdp_media_name(sdpm);
+	if (str_cmp(med_name, "audio") == 0)
+		med_type = MT_AUDIO;
+	else if (str_cmp(med_name, "video") == 0)
+		med_type = MT_VIDEO;
+	else if (str_cmp(med_name, "text") == 0)
+		med_type = MT_TEXT;
+	else if (str_cmp(med_name, "application") == 0)
+		med_type = MT_APPLICATION;
+	else if (str_cmp(med_name, "message") == 0)
+		med_type = MT_MESSAGE;
+	else
+		med_type = MT_UNKNOWN;
+
+	st->stream = sess->session->create_stream(
+	                       *s_zrtp_config,
+	                       (struct udp_sock *)rtpsock,
+	                       (struct udp_sock *)rtcpsock,
+	                       rtp_sess_ssrc(rtp), med_type);
+	if (!st->stream) {
+		err = ENOMEM;
+		goto out;
+	}
+
+	st->stream->sdp_encode(sdpm);
+
+ out:
+	if (err) {
+		mem_deref(st);
+		return err;
+	}
+	else
+		*stp = st;
+
+ start:
+	if (sa_isset(sdp_media_raddr(sdpm), SA_ALL)) {
+		st->stream->sdp_decode(sdpm);
+		err = sess->session->start_stream(st->stream);
+		if (err) {
+			warning("zrtp: stream start failed: %d\n", err);
+		}
+	}
+
+	return err;
+}
+
+
+static struct menc menc_zrtp = {
+	LE_INIT, "zrtp", "RTP/AVP", session_alloc, media_alloc
+};
+
+
+static const struct cmd cmdv[] = {
+	{"zrtp_verify", 0, CMD_PRM, "Verify ZRTP SAS <session ID>",
+	        Session::cmd_verify_sas },
+	{"zrtp_unverify", 0, CMD_PRM, "Unverify ZRTP SAS <session ID>",
+	        Session::cmd_unverify_sas },
+};
+
+
+static int module_init(void)
+{
+	char config_path[256];
+	int err = 0;
+
+	err = conf_path_get(config_path, sizeof(config_path));
+	if (err) {
+		warning("zrtp: could not get config path: %m\n", err);
+		return err;
+	}
+
+	s_zrtp_config = new ZRTPConfig(conf_cur(), config_path);
+	if (!s_zrtp_config)
+		return ENOMEM;
+
+	menc_register(baresip_mencl(), &menc_zrtp);
+
+	return cmd_register(baresip_commands(), cmdv, ARRAY_SIZE(cmdv));
+}
+
+
+static int module_close(void)
+{
+	delete s_zrtp_config;
+	s_zrtp_config = NULL;
+
+	cmd_unregister(baresip_commands(), cmdv);
+
+	menc_unregister(&menc_zrtp);
+
+	return 0;
+}
+
+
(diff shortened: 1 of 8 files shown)


alfredh:
Many thanks for this patch. I will start reviewing it in the next couple of days.

Alfred

juha-h:
Patch of mk/modules.mk does not include automatic detection of
USE_GZRTP.  It would be useful to have especially when Debian testing
includes libbzrtp.

-- Juha

juha-h:
I got confused.  Debian testing includes bzrtp library.  Still,
automatic detection similar to this would be useful:

USE_ZRTP := $(shell [ -f $(SYSROOT)/include/libzrtp/zrtp.h ] || \
	[ -f $(SYSROOT)/local/include/libzrtp/zrtp.h ] || \
	[ -f $(SYSROOT_ALT)/include/libzrtp/zrtp.h ] && echo "yes")

-- Juha

juha-h:
juha-h writes:

> I got confused.  Debian testing includes bzrtp library.  Still,
> automatic detection similar to this would be useful:
> 
> USE_ZRTP := $(shell [ -f $(SYSROOT)/include/libzrtp/zrtp.h ] || \
> 	[ -f $(SYSROOT)/local/include/libzrtp/zrtp.h ] || \
> 	[ -f $(SYSROOT_ALT)/include/libzrtp/zrtp.h ] && echo "yes")

Here is what I used:

USE_GZRTP := $(shell [ -f $(SYSROOT)/include/libzrtpcpp/ZrtpCodes.h ] || \
	[ -f $(SYSROOT)/local/include/libzrtpcpp/ZrtpCodes.h ] || \
	[ -f $(SYSROOT_ALT)/include/libzrtpcpp/ZrtpCodes.h ] && echo "yes")

After that I made some test calls and zrtp worked fine.

If this zrtp module is clearly better than the existing one, I would
prefer to replace the current one with it rather than having two for the
same purpose.

-- Juha

glenvt18:
@juha-h Unfortunately it's impossible to install libzrtpcppcore properly at the moment - some header files (ZRtp.h, for example) are missing. CMakeLists.txt and some other stuff have to be fixed.

Distributions have libzrtpcpp (not libzrtpcppcore), which is a ccRTP client depending on libccrtp, and that is just not our case. BTW Ubuntu doesn't supply libzrtpcpp starting from 16.10 and 16.04's version is 2.3.4 which is pretty old. So, if one wants to have ZRTP support, he (she) has to to build ZRTP library (either libzrtpcppcore or libzrtp) from source using the latest git version. There is no other way ATM. I don't see much sense in detecting USE_ZRTP or USE_GZRTP if one has to build ZRTP library intentionally beforehand.

Perhaps `modules/gzrtp/module.mk` should be improved to automate building of libzrtpcppcore. I just couldn't find an easy and elegant way of doing it. Another approach is to add all requiered zrtpcpp source files and some vital CXXFLAGS to the module the way it was done in PJSIP:
https://github.com/wernerd/ZRTP4PJ/blob/master/build/zsrtp/Makefile

But I think running cmake is safer and more portable.

This module needs some testing before it can replace zrtp completely. BTW zrtp module itself is a handy testing/debugging tool. It prints a lot of information on the protocol. The problem is that libzrtp has been abandoned for years. Also https://freeswitch.org/jira/browse/FS-9387.

juha-h:
glenvt18 writes:

> Distributions have libzrtpcpp (not libzrtpcppcore), which is a ccRTP
> client depending on libccrtp, and that is just not our case. BTW
> Ubuntu doesn't supply libzrtpcpp starting from 16.10 and 16.04's
> version is 2.3.4 which is pretty old. So, if one wants to have ZRTP
> support, he (she) has to to build ZRTP library (either libzrtpcppcore
> or libzrtp) from source using the latest git version. There is no
> other way ATM. I don't see much sense in detecting USE_ZRTP or
> USE_GZRTP if one has to build ZRTP library intentionally beforehand.

There is since it makes installation easier and there can be more than
one application that is using the lib.  After adding automatic detection
to modules.mk, I was able to build baresip debian package that includes
gzrtp.so like this:

git clone https://github.com/wernerd/ZRTPCPP.git
cd ZRTPCPP
mkdir build
cd build
cmake -DCMAKE_POSITION_INDEPENDENT_CODE=1 -DCORE_LIB=1 -DSDES=1 -DBUILD_STATIC=1 ..
make
sudo make install
cd <baresip source dir>
dpkg-buildpackage -rfakeroot -b -us -uc

-- Juha

glenvt18:
@juha-h Yes, libzrtpcppcore (-DCORE_LIB=1) can be installed with `make install`, but then I have
```
/usr/local/include/libzrtpcpp
/usr/local/include/libzrtpcpp/common
/usr/local/include/libzrtpcpp/common/osSpecifics.h
/usr/local/include/libzrtpcpp/ZrtpUserCallback.h
/usr/local/include/libzrtpcpp/ZrtpCWrapper.h
/usr/local/include/libzrtpcpp/ZrtpCallback.h
/usr/local/include/libzrtpcpp/ZrtpConfigure.h
/usr/local/include/libzrtpcpp/ZrtpCodes.h
```
and that's just not enough to use the C++ API and build the module. Perhaps your ZRTP variable points to ZRTPCPP source directory. Try removing it.

alfredh:
small compilation error on OSX 10.12

```
  CXX [M] build-x86_64/modules/gzrtp/stream.o
modules/gzrtp/stream.cpp:173:9: error: use of undeclared identifier 'pthread_mutexattr_init'
        err  = pthread_mutexattr_init(&attr);
               ^
modules/gzrtp/stream.cpp:174:42: error: use of undeclared identifier 'PTHREAD_MUTEX_ERRORCHECK'
        err |= pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_ERRORCHECK);
                                                ^
modules/gzrtp/stream.cpp:175:9: error: use of undeclared identifier 'pthread_mutex_init'
        err |= pthread_mutex_init(&m_zrtp_mutex, &attr);
               ^
modules/gzrtp/stream.cpp:229:2: error: use of undeclared identifier 'pthread_mutex_destroy'
        pthread_mutex_destroy(&m_zrtp_mutex);
        ^
modules/gzrtp/stream.cpp:795:2: error: use of undeclared identifier 'pthread_mutex_lock'
        pthread_mutex_lock(&m_zrtp_mutex);
        ^
modules/gzrtp/stream.cpp:801:2: error: use of undeclared identifier 'pthread_mutex_unlock'
        pthread_mutex_unlock(&m_zrtp_mutex);
        ^
6 errors generated.
make: *** [build-x86_64/modules/gzrtp/stream.o] Error 1
```

(probably a missing include pthread.h)

modules/gzrtp/gzrtp.cpp
```diff
@@ -0,0 +1,218 @@
...
+}
+
+
+static int module_close(void)
+{
+	delete s_zrtp_config;
+	s_zrtp_config = NULL;
+
+	menc_unregister(&menc_zrtp);
+
```
alfredh: you should also unregister the commands by calling `cmd_unregister`

alfredh:
I have tested the module on Debian Testing (x86_64):

```
baresip zrtp.so:

zrtp: Stream <audio> is encrypted (AES-256/EC25), SAS is [djo7] (NOT VERIFIED)
zrtp: SAS is not verified, type '/zrtp_verify 1' to verify
zrtp: All streams are encrypted (AES-256/EC25), SAS is [djo7] (NOT VERIFIED)


baresip gzrtp.so:

zrtp: Stream <audio> is encrypted (Twofish-256/EC25), SAS is [yxtt] (NOT VERIFIED)
zrtp: SAS is not verified, type '/zrtp_verify 1' to verify
zrtp: All streams are encrypted (Twofish-256/EC25), SAS is [yxtt] (NOT VERIFIED)


linphone:

zrtp: Stream <audio> is encrypted (AES-128/DH3k), SAS is [izgq] (NOT VERIFIED)
zrtp: SAS is not verified, type '/zrtp_verify 1' to verify
zrtp: All streams are encrypted (AES-128/DH3k), SAS is [izgq] (NOT VERIFIED)

```


apart from the minor nits I found this module is ready for merge to master.

The code is impressively clear and easy to read. This is a very nice contribution.

I think removing "zrtp.so" is too radical, there might be people using it.
I recommend adding gzrtp.so so that we have 2 ZRTP modules, which gives
people the freedom to choose.


Alfred

modules/gzrtp/module.mk
```diff
@@ -0,0 +1,34 @@
...
+#     cd ZRTPCPP
+#     mkdir build
+#     cd build
+#     cmake -DCMAKE_POSITION_INDEPENDENT_CODE=1 -DCORE_LIB=1 -DSDES=1 \
+#         -DBUILD_STATIC=1 ..
+#     make
+#
+
+# GNU ZRTP C++ library (ZRTPCPP) source directory
+ZRTP ?= ../ZRTPCPP
```
alfredh: nit: perhaps ZRTP could be renamed to `ZRTP_PATH`

glenvt18:
@alfredh Thanks for the review. Udated.

I've just copied `PRESZ = 36` from the zrtp module. Is it still needed by TURN (TODO in stream.cpp)?

It will be great to run interoperability tests with proprietary UAs like Zoiper (I don't have it).

BTW. Travis has to be updated.

alfredh:
if the TURN client is used to send the RTP stream via a TURN server,
the mbuf packets must have free headroom of 4 or 36 bytes. This and other
criterias such as permissions/channels must be met for the TURN server
to be used. The TURN client is prepending a header, so this logic was chosen
to avoid copying of the buffer.

it is not ideal that a "mediaenc" module should have to know details about the
RTP transport, but that is what we have at the moment.

glenvt18:
@alfredh Thanks for the explanation. I wanted to make sure it is valid. I've removed the comment.

BTW If you want me to squash commits, let me know when you're ready to merge.

alfredh:
it is ready for merge now :)

glenvt18:
@alfredh Should I squash them (rebase into one commit)?

alfredh:
yeah that would be nice

glenvt18:
Done.

juha-h:
glenvt18 writes:

> @juha-h Yes, libzrtpcppcore (-DCORE_LIB=1) can be installed with `make
> install`, but then I have 
> ```
> /usr/local/include/libzrtpcpp
> /usr/local/include/libzrtpcpp/common
> /usr/local/include/libzrtpcpp/common/osSpecifics.h
> /usr/local/include/libzrtpcpp/ZrtpUserCallback.h
> /usr/local/include/libzrtpcpp/ZrtpCWrapper.h
> /usr/local/include/libzrtpcpp/ZrtpCallback.h
> /usr/local/include/libzrtpcpp/ZrtpConfigure.h
> /usr/local/include/libzrtpcpp/ZrtpCodes.h
> ```
> and that's just not enough to use the C++ API and build the
> module. Perhaps your ZRTP variable points to ZRTPCPP source
> directory. Try removing it.

Yes, I had ZRTPCPP source directory in the same dir as baresip source
dir, which is the default in modules/gzrtp/module.mk:

ZRTP ?= ../ZRTPCPP

Why is it that 'make install' in ZRTPCPP source dir does not install
everything that is need to use the lib?

-- Juha

glenvt18:
@juha-h I don't know. ZRtp.h includes some headers from cryptcommon and zrtp/crypto. We also need headers from srtp directory. As you can see, there are three(!) include paths in the module.mk. If you just copy those files, you'll pollute /usr/include and still have to add two -I paths. So, that can't be easily fixed without breaking existing software. Definitely, the directory structure of ZRTPCPP should be better organized. Notice that no srtp headers are installed with -DSDES=1 but srtp suite is added to libzrtpcppcore.a.

The only thing you can do with libzrtpcppcore after 'make install' is to use the C wrapper (ZrtpCWrapper.h) for the key exchange part and libre to do srtp stuff. That should work (though I didn't test it), but ZRTPCPP/srtp has richer set of algorithms.

And it looks like no Linux distribution provides libzrtpcppcore library at the moment, only FreeBSD does.

PR #319, opened 2017-10-31
modules/gzrtp: move SRTP processing to a separate class

glenvt18:
And add an option to use libre SRTP processing stuff instead of the one provided by ZRTPCPP (-DGZRTP_USE_RE_SRTP).

This makes stream class code a bit cleaner. libre SRTP processing can be used as a fallback if one has issues with ZRTPCPP's one. It works fine but supports only a standard set of algorithms (AES).


alfredh:
Thank you :)

PR #322, opened 2017-11-10
audio: remove silence at the beginning of the RTP stream

glenvt18:
Having min_sz > tx->psize in aubuf_alloc() results in insertion of 16 packets of silence at the beginning of the RTP stream. This adds an unnecessary delay.

The delay (~ 1/3 sec) is clearly noticeable when calling localhost->localhost using baresip at both ends. Use hold/resume to hear the difference.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -1380,7 +1380,7 @@ static int start_source(struct autx *tx, struct audio *a)
 		tx->psize = 2 * calc_nsamp(prm.srate, prm.ch, prm.ptime);
 
 		if (!tx->aubuf) {
-			err = aubuf_alloc(&tx->aubuf, tx->psize * 2,
+			err = aubuf_alloc(&tx->aubuf, tx->psize,
 					  tx->psize * 30);
 			if (err)
 				return err;


alfredh:
Thank you,

the silence at the beginning of the audio stream is now significantly shorter.

actually there is not really a need to send silence in the beginning.

The audio transmit pacer has 3 different modes:

* Flush out all complete packets (done in the context of the audio-source)

* A separate thread that is sending out the packets

* A separate timer that is sending out the packets
  (in the context of the re_main thread)

Perhaps all these modes are not needed?

We could skip the aubuf for transmit and use mqueue instead, this would mean:

* The outgoing packet will always be sent from re_main context

* The audio driver just has to copy the samples to the mqueue,
  no extra processing is needed.

* The encoding can be done in the re_main thread.


/Alfred

glenvt18:
@alfredh Some thoughts.

I think AUDIO_MODE_POLL is OK provided tx path is thread safe (I have to fix gzrtp BTW).

AUDIO_MODE_THREAD_REALTIME and AUDIO_MODE_TMR might add a small amount of jitter.

Threaded modes (1 and 2) have an advantage of utilizing multiple CPU cores for encoding.

I think mqueue is an overkill. Also, transferring plain audio data through a pipe (on Windows) is less secure than just copying memory. As an option, push a single byte to mqueue and use aubuf to transfer data.

Do you want to use mqueue for video too? In this case encoding both audio and video on a single thread could be too much for a weak CPU core.

PR #323, opened 2017-11-11
pulse: remove garbage at the beginning of a recording

glenvt18:
I've found that sometimes pa_simple_read() gets some garbage from my USB camera before 'real' samples arrive. This introduces an extra audio delay when calling baresip->baresip because a batch of packets (with real timestamps) is inserted at the beginning of the RTP stream. It happens randomly, usually after fast re-dials, and only when the USB camera's microphone is used (built-in analog input works fine). ALSA is not affected by this issue.

I propose a simple workaround for such cases which measures the time spent in pa_simple_read() and compares it to ptime.

Though, it's more a pulseaudio/driver issue, this workaround can be very helpful. In my case it dropped up to 50 packets (= 1 sec delay).

### DIFF
--- a/modules/pulse/recorder.c
+++ b/modules/pulse/recorder.c
@@ -21,6 +21,7 @@ struct ausrc_st {
 	void *sampv;
 	size_t sampc;
 	size_t sampsz;
+	uint32_t ptime;
 	ausrc_read_h *rh;
 	void *arg;
 };
@@ -49,6 +50,16 @@ static void *read_thread(void *arg)
 	struct ausrc_st *st = arg;
 	const size_t num_bytes = st->sampc * st->sampsz;
 	int ret, pa_error = 0;
+	uint64_t now, last_read, diff;
+	unsigned dropped = 0;
+	bool init = true;
+
+	if (pa_simple_flush(st->s, &pa_error)) {
+		warning("pulse: pa_simple_flush error (%s)\n",
+		        pa_strerror(pa_error));
+	}
+
+	last_read = tmr_jiffies();
 
 	while (st->run) {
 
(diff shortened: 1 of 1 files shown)


alfredh:
the audio that is coming in the beginning, do you know if it contains silence
or garbage audio?

glenvt18:
I don't know actually, and I don't care. What I see is a huge number of samples (20-40 ptime frames) being read in a very short interval (1-2 ms) right after the read thread starts. This results in a batch of RTP packets being sent almost instantly (can be seen in wireshark). After that, frames are being read at regular (~ptime) intervals. As the result, the audio is played by the receiver with a huge (up to 1 sec) delay. If I cut the initial burst out, the audio is played with a minimal delay.

As there is no noticeable delay between pulse_recorder_alloc() and the first pa_simple_read(), you are not expected to get 1 sec of audio at the first pa_simple_read() call. So, I consider it a pulseaudio/driver/device bug. It doesn't happen often, but the result (huge delay) is really nasty.

There is a chance that some 'real' samples could be thrown away by this method. I don't mind loosing several milliseconds at the beginning of the conversation. After all, pa_simple_new() takes some time (with my camera - up to 1 sec after being idle for a while).

The 'garbage' data are not really garbage. They look more like an low-level audio signal/noise, maybe leftovers from the previous session. If so, it's also a security issue:)

alfredh:
Thank you :)

glenvt18:
@alfredh How about moving pa_simple_new to read_thread to eliminate main loop stalls when the initialization takes a long time?

alfredh:
The old ALSA module was opening the device from the context of the worker thread,
for the same reason. But we found some problems when using USB soundcards, where
sometimes it would not open the device successfully. So we changed it to always open
the device from the main thread (the current code).

However, if your suggested approach works with Pulse I am happy to test it out.

Alfred

PR #324, opened 2017-11-12
modules/gzrtp: make send handler thread-safe

glenvt18:
It can be called from another thread (audio/video read).


alfredh:
Many thanks,

I tested this with audio+video, and both using audio mode "poll" and "thread".

Alfred

glenvt18:
@alfredh gzrtp module sends ZRTP packets from the main thread while RTP/RTCP packets can be sent from the audio/video worker thread. That means some UDP send handlers (like TURN) called in `udp_send_internal` (libre) might be called from two different threads. Is it safe?

PR #326, opened 2017-11-17
Max skew is calculated during runtime and not a defined fixed value.

Ola-Palm:
Perhaps should be done once when RX stream is set up, and stored in
struct audio.
/ Ola Palm

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -984,8 +984,9 @@ static void stream_recv_handler(const struct rtp_header *hdr,
 		 * negative skew:  the packet arrived too early
 		 * Max skew is calulated from rx ptime * Jb_max / 1000
 		*/
-		/*  this might be better if calculated once when  it
-		is set up and we know the RX ptime, --> store in struct Audio */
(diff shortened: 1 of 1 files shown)


src/audio.c
```diff
@@ -857,12 +855,14 @@ static void stream_recv_handler(const struct rtp_header *hdr,
...
 	struct audio *a = arg;
 	struct aurx *rx = &a->rx;
+	const struct config_avt *avt = &a->strm->cfg;
 	bool discard = false;
 	size_t i;
 	int wrap;
 	double wall_clock=0;
 	double dur=0;
 	double skew=0;
+  double max_skew=0;
```
alfredh: the indentation here should be 1 TAB.

alfredh:
hi @Ola-Palm 

many thanks for the patch, it looks good.

a minor ccheck error has to be fixed:

./src/audio.c:988: line is too wide (96 - max 79)

Ola-Palm:
Still not valid?

alfredh:
Thank you :)

PR #334, opened 2017-12-04
avcodec: change x264 rate control mode to ABR

strfry:
Just a very small fix: 
I found that the avcodec module with x264 won't obey the video_bitrate config option.
With a different rate control mode (ABR=Average-Bitrate instead of CQP=Constant-Quality), it will usually keep the bitrate around or below the configured value.

### DIFF
--- a/modules/avcodec/encode.c
+++ b/modules/avcodec/encode.c
@@ -429,7 +429,7 @@ static int open_encoder_x264(struct videnc_state *st, struct videnc_param *prm,
 	xprm.i_fps_num = prm->fps;
 	xprm.i_fps_den = 1;
 	xprm.rc.i_bitrate = prm->bitrate / 1000; /* kbit/s */
-	xprm.rc.i_rc_method = X264_RC_CQP;
+	xprm.rc.i_rc_method = X264_RC_ABR;
 	xprm.i_log_level = X264_LOG_WARNING;
 
 	/* ultrafast preset */


alfredh:
hi @strfry 

many thanks for the patch, I have tested it on Debian Testing (libx264 0.148.2795+gitaaa9aa8-1 )
and the Rate Control seems to work much better.

we should also consider using X264_RC_CRF, this is the default after
setting preset to "ultrafast", "zerolatency". 

there are some RC guidelines here:

http://slhck.info/video/2017/03/01/rate-control.html

BTW: which is the oldest x264 build that we should support, we have many
ifdefs that we should probably clean up:

```
#if X264_BUILD >= 59
	xprm.rc.i_aq_mode = 0;
#endif
	xprm.analyse.b_mixed_references = 0;
	xprm.analyse.i_trellis = 0;
#if X264_BUILD >= 63
	xprm.i_bframe_adaptive = X264_B_ADAPT_NONE;
#endif
#if X264_BUILD >= 70
	xprm.rc.b_mb_tree = 0;
#endif
```

strfry:
According to this guide, CRF might be the better choice. To enable rate control, the parameters xprm.rc.i_vbv_max_bitrate and xprm.rc.i_vbv_buffer_size need to be set.
The first one is obvious, the latter has some influence on the decoding buffer and must be between 10 and 14000 (in baseline profile). I've no idea what would be the ideal value in this application.

On the x264 versions: Even Debian oldstable has x264 version 123 these days.
I don't know if there's another benchmark for really old software still being shipped in the wild, but my guess is that these switches can be removed.

PR #337, opened 2017-12-12
fix #336

### LINKED ISSUE #336: ringback on sip response 180
When using same the output for player and alert, the ringback when sip 180 is received results in call not possible to establish. This due to the sound card is busy playing sound files in one format, and opus tries to set the card in another format. 
I'm not sure If problem exist if using G711 or other 8kHz codecs. But we are having problems when using opus.

Ola-Palm:
if value, ringback_disabled, is set to true in config file, there audio played when sip 180 is received is skipped. 
If ringback_disabled is not present or set to false, baresip will run as normal.

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -941,8 +941,6 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 	(void)prm;
 	(void)arg;
 
-	conf_get_bool(conf_cur(), "ringback_disabled", &menu.ringback_disabled);
-
 	switch (ev) {
 
 	case UA_EVENT_CALL_INCOMING:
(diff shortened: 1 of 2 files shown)


alfredh:
Hi @Ola-Palm 

it looks the patch is empty, it says:

"These merge commits were added into this branch cleanly."

perhaps you can make a new branch, add only 1 commit and then push
and make the PR from there ?

Alfred

modules/menu/menu.c
```diff
@@ -941,6 +941,8 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 	(void)prm;
 	(void)arg;
 
+	conf_get_bool(conf_cur(), "ringback_disabled", &menu.ringback_disabled);
+
```
alfredh: can you please move this to `module_init` ?

it is enough to read this config value once, during init.

Ola-Palm: will do.

Issue #196, opened 2016-12-28
please implement support for SCReAM algorithm


jonassmedegaard:

jonassmedegaard:
The SCReAM algorithm - https://github.com/EricssonResearch/scream - offers a way to efficiently compute dynamically the available bandwith for RTP streams.

Would be awesome for baresip to support optionally linking against SCReAM.

Even if over time it might turn out that the Google GCC algorithm or something else does a better job at bandwidth prediction, I imagine that work now on what seems to me the most efficient contender in the field likely aids also in hooking into alternative algorithms later.

alfredh:
many thanks for the input regarding congestion control. For now I have added it to
the roadmap:

https://github.com/alfredh/baresip/wiki/Roadmap


adding support for CC and testing it properly is quite a large task. Even if the 3rd party
library contains a working stack, it will still have to be integrated with baresip.

personally I have no time to work on this any time soon. If someone is interested
in implementing and testing this, please let me know.



/Alfred

PR #339, opened 2017-12-23
avcodec: Request Frame Refresh in more cases

strfry:
Hello Alfred,

Previously, when packet loss occured on a H264 stream, the image would get corrupted and not recover for many seconds, until the left-to-right-rolling periodic intra-refresh came by.

Today i did some tests with simulated packet loss, and noticed that in most cases, decode_h264() won't return an error, because the missing fragment is already noticed by h264_fu_hdr_decode(), and the subsequent code returns 0;

I changed these cases to return EPROTO, and now it recovers much faster.
Only one or a few frames of a destroyed image are shown, before the video recovers, which still is annoying and probably unnecessary, but the overall experience seems to already be much better.

I'd be happy if you have some time to look into this and comment, i'm very much interested in a more robust video-stream with baresip. :)

Best Regards,
Jonathan

### DIFF
--- a/modules/avcodec/decode.c
+++ b/modules/avcodec/decode.c
@@ -333,6 +333,7 @@ int decode_h264(struct viddec_state *st, struct vidframe *frame,
 				      " ignoring NAL\n");
 				fragment_rewind(st);
 				++st->stats.n_lost;
+				return EPROTO;
 			}
 
 			st->frag_start = st->mb->pos;
@@ -351,7 +352,7 @@ int decode_h264(struct viddec_state *st, struct vidframe *frame,
 			if (!st->frag) {
 				debug("avcodec: ignoring fragment\n");
 				++st->stats.n_lost;
-				return 0;
+				return EPROTO;
 			}
 
 			if (seq_diff(st->frag_seq, seq) != 1) {
@@ -359,7 +360,7 @@ int decode_h264(struct viddec_state *st, struct vidframe *frame,
 				fragment_rewind(st);
 				st->frag = false;
 				++st->stats.n_lost;
-				return 0;
+				return EPROTO;
 			}
 		}
 


alfredh:
Thank you, I will review and merge this after baresip 0.5.7 is released.

alfredh:
I have tested your patch and it is working better, as you describer.

however I think there are 2 things that we should consider first:

* ffmpeg libavcodec does not seem to handle missing packets
  very well. if there is a missing packet in the stream, it tries to repair (conceal) the
  image and still render it. it would be better if the broken video frame was dropped.
  are there special config options for libavcodec that we can tweak ?

* when the packet loss occurs, it is not always a wise idea to request a new
  full frame if the bandwidth is limited. Delta frames are relatively small, while
  full frames can be quite large. The full frames are normally fragmented to
  fit the MTU size. If one of those fragments are also lost, the whole full-frame
  will be lost.

I think we should also consider adding support for re-transmission of lost delta
packets (similar to how it is done with WebRTC stack), or possibly a FEC scheme.

Alfred

alfredh:
hi @strfry I was not sure about this one, but I decided to merge it.

lets do some more testing and see if it causes any problems.

one idea I had regarding libavcodec and packetloss:

1. if packet loss is detected, discard buffer and set got_keyframe=false

2. request a Full intra frame

3. in the meantime, any delta frames arriving will be discarded,
  as the got_keyframe flag is not set

4. when the Full intra frame arrives, pass it to libavcodec
  and set got_keyframe=true

if this scheme is working, we should not see any artifacts on the
decoded picture.

strfry:
Thanks Alfred,

Your proposed algorithm might lead to extended periods of a still image, when network congestion doesn't allow for a full intra-frame to arrive undamaged. This might or might not be an improvement, depending on the application and subjective preference.

I looked at bit into the more advanced options to improve this, e.g. the WebRTC FEC scheme. It is a bit demotivating, to see how difficult it is to keep up with the pace of developments in the WebRTC field ;)

As far as i understood, video FEC is supposed to be sent on a separate RTP channel.
This is also related to #196, since determining the FEC rate goes hand-in-hand with bandwidth adjustment.
In any case, it is could a bit of work to get something substantially better.

PR #346, opened 2018-01-14
ctrl_tcp module

jmillan:
This  PR corresponds to the **ctrl_tcp** module which has been briefly commented in the [mailing list](http://lists.creytiv.com/pipermail/re-devel/2017-December/001137.html).

It consist on two parts: the **ctrl_tcp** module itself and the **netstring** framing for TCP. I believe the **netstring** part could be added to **libre** so other modules could make use of it. 

The folder `netstring-c` is actually a very small public [library](https://github.com/PeterScott/netstring-c) which provides **netstring** encoding/decoding.

Few additions/modifications have been made to some parts of **baresip** like the addition of the `id` attribute on events of class `call` representing the session (dialog) call ID.


alfredh:
many thanks I will start to review it now.

src/call.c
```diff
@@ -901,6 +901,12 @@ int call_sdp_get(const struct call *call, struct mbuf **descp, bool offer)
 }
 
 
+const char *call_id(const struct call *call)
+{
+	return call ? sip_dialog_callid(sipsess_dialog(call->sess)) : NULL;
+}
+
+
```
alfredh: nice, I will add this one in a separate commit

src/event.c
```diff
@@ -79,6 +79,8 @@ static int add_rtcp_stats(struct odict *od_parent, const struct rtcp_stats *rs)
 
  out:
 	mem_deref(od);
+	mem_deref(tx);
+	mem_deref(rx);
```
alfredh: thanks for fixing this, I added it in a separate commit (67829d86c4c2d173807b532e97c09402415341c5)

modules/ctrl_tcp/ctrl_tcp.c
```diff
@@ -0,0 +1,358 @@
...
+	struct re_printf pf = {print_handler, resp};
+	struct odict *od = NULL;
+	char *buf = NULL;
+	int err;
+
+	err = odict_alloc(&od, 8);
+	if (err)
+		return err;
+
+	resp->pos = NETSRING_HEADER_SIZE;
```
alfredh: is there a typo in `NETSRING_HEADER_SIZE` ?

modules/ctrl_tcp/ctrl_tcp.c
```diff
@@ -0,0 +1,358 @@
...
+{
+	struct ctrl_st *st = arg;
+
+	(void)peer;
+
+	/* only one connection allowed */
+	st->tc = mem_deref(st->tc);
+
+	(void)tcp_accept(&st->tc, st->ts, NULL, NULL, tcp_close_handler, st);
+	(void)netstring_insert(&st->ns, st->tc, 0, command_handler, st);
```
alfredh: is there a memory leak here ?

if the TCP connection is closed, and a new one is created, will the `st->ns` be derefed ?

alfredh:
I am getting some compiler warnings on OSX 10.12:

```
$ make ctrl_tcp.so
  CC [M]  build-x86_64/modules/ctrl_tcp/ctrl_tcp.o
  CC [M]  build-x86_64/modules/ctrl_tcp/netstring.o
modules/ctrl_tcp/netstring.c:79:20: warning: cast from function call of type 'double' to non-matching type 'size_t' (aka 'unsigned long') [-Wbad-function-cast]
        num_len = (size_t)ceil(log10((double)mb->end + 1));
                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
  CC [M]  build-x86_64/modules/ctrl_tcp/./netstring-c/netstring.o
modules/ctrl_tcp/./netstring-c/netstring.c:52:17: warning: comparison of integers of different signs: 'int' and 'size_t' (aka 'unsigned long') [-Wsign-compare]
  for (i = 0; i < buffer_length && isdigit(buffer[i]); i++) {
              ~ ^ ~~~~~~~~~~~~~
modules/ctrl_tcp/./netstring-c/netstring.c:78:18: warning: cast from function call of type 'double' to non-matching type 'size_t' (aka 'unsigned long') [-Wbad-function-cast]
  return (size_t)ceil(log10((double)data_length + 1)) + data_length + 2;
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
modules/ctrl_tcp/./netstring-c/netstring.c:94:23: warning: cast from function call of type 'double' to non-matching type 'size_t' (aka 'unsigned long') [-Wbad-function-cast]
    num_len = (size_t)ceil(log10((double)len + 1));
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
3 warnings generated.
  LD [M]  ctrl_tcp.so
```


also there are some ccheck warnings that must be fixed:

```
$ ccheck.py 
./modules/ctrl_tcp/netstring-c/netstring.c:46: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:53: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:54: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:55: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:56: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:66: has trailing space(s)
./modules/ctrl_tcp/netstring-c/netstring.c:89: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:90: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:91: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:92: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:93: else: ending if bracket should be on previous line
./modules/ctrl_tcp/netstring-c/netstring.c:94: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:95: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:96: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:97: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:98: starts with 4 spaces, use tab instead
./modules/ctrl_tcp/netstring-c/netstring.c:104: has trailing space(s)
./modules/ctrl_tcp/netstring-c/testsuite.c:30: line is too wide (83 - max 79)
./modules/ctrl_tcp/netstring-c/testsuite.c:40: line is too wide (84 - max 79)
./modules/ctrl_tcp/netstring-c/testsuite.c:43: line is too wide (84 - max 79)
./modules/ctrl_tcp/netstring-c/testsuite.c:97: line is too wide (80 - max 79)
./modules/ctrl_tcp/netstring-c/testsuite.c:107: has trailing space(s)
./modules/ctrl_tcp/ctrl_tcp.c:139: line is too wide (80 - max 79)
./modules/ctrl_tcp/ctrl_tcp.c:161: line is too wide (80 - max 79)
./modules/ctrl_tcp/ctrl_tcp.c:272: line is too wide (84 - max 79)
./modules/ctrl_tcp/netstring.c:68: C++ comment, use C comments /* ... */ instead
./modules/ctrl_tcp/netstring.c:134: line is too wide (83 - max 79)
./modules/ctrl_tcp/netstring.c:139: line is too wide (82 - max 79)
./modules/ctrl_tcp/netstring.c:143: line is too wide (85 - max 79)
./modules/ctrl_tcp/netstring.c:226: line is too wide (84 - max 79)
Statistics:
~~~~~~~~~~~
Number of files processed:     c: 240  cpp: 7  h: 48  mk: 103  m4: 0  py: 0  m: 4  s: 0  java: 0 
Number of lines with errors:   30

```

src/call.c
jmillan: 👍

src/event.c
jmillan: 👍

modules/ctrl_tcp/ctrl_tcp.c
jmillan: Indeed :-). I'll fix it.

modules/ctrl_tcp/ctrl_tcp.c
jmillan: true. I'll fix it.

jmillan:
> modules/ctrl_tcp/netstring.c:79:20: warning: cast from function call of type 'double' to non-matching type 'size_t' (aka 'unsigned long') [-Wbad-function-cast]
        num_len = (size_t)ceil(log10((double)mb->end + 1));
                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

I need to get the number of digits of a size_t in a different way, yes. This is how `cstring-c` does.

> also there are some ccheck warnings that must be fixed:

This is the `cstring-c` code itself. I think I'll pick the useful parts and get rid of it. Of course referencing the original project.

jmillan:
New 'event' tests failing because the `type` event JSON attribute is asserted.

I'll change the `type` attribute of the messages in `ctrl_tcp` module. Any suggestion @alfredh?

The JSON sent by `ctrl_tcp` contains a `type` attribute which value can be: `response` or `event`.

https://github.com/jmillan/baresip/blob/ctrl_tcp/modules/ctrl_tcp/ctrl_tcp.c#L37

alfredh:
could you please rebase the PR from master ? (or merge from master) ?

not sure about the event "type".

Regarding the source code of netstring, I dont think that we should include all
the source code verbatim. There are some unused files there, for example "makefile".
It would be better if we include only files that are used by ctrl_tcp module.

jmillan:
Hi @alfred,

There is no collision with the 'type' event attribute anymore. It's not used in `ctrl_tcp` messages.

Few thoughts regarding to code location. This is IMHO how code from this module should be located:

## libre

* `netstring` (current ctrl_tcp/netstring-c/)
* `tcp_netstring` (current ctrl_tcp/netstring)

## baresip

* `ctrl_tcp` (current ctrl_tcp/ctrl_tcp.h/c)


The rationale behind is that both `netstring` and `tcp_netstring` can be used by other modules independently.

If that's OK I can make the cleanup myself.

alfred:
Looks good to me.

But @alfredh might want to take another look. 😃

jmillan:
> Looks good to me.

hahaha

> But @alfredh might want to take another look. 😃

Absolutely agree ;-)

alfredh:
I think the netstring code is not generic enough to be located in libre, yet.
Currently the ctrl_ctp module is the only known usecase.

Instead we could put netstring in librew, which is a more experimental library:
https://github.com/alfredh/rew

so I think that leaves us with 2 options:

1. add a new "netstring" module to librew, and use it from ctrl_tcp module

2. keep the netstring code in baresip/modules/ctrl_tcp as it is now
  (The generic code can always be relocated later, should there be a need for it)

I will leave it up to you @jmillan to choose which option is best

jmillan:
> so I think that leaves us with 2 options:

> add a new "netstring" module to librew, and use it from ctrl_tcp module

> keep the netstring code in baresip/modules/ctrl_tcp as it is now
(The generic code can always be relocated later, should there be a need for it)


I would leave it inside the module folder then.

I've made some cleanups which I found convenient. Please let me know your opinion if you disagree with any or would prefer other.

Also, something that I still find a bit uncomfortable is having to rewrite `buf->pos` to `NETSTRING_HEADER_SIZE` all the time in `ctrl_tcp.c`. I'll re-check that.

src/call.c
```diff
@@ -1299,6 +1299,8 @@ static void sipsess_close_handler(int err, const struct sip_msg *msg,
 	if (err) {
 		info("%s: session closed: %m\n", call->peer_uri, err);
 
+		(void)re_snprintf(reason, sizeof(reason), "%m", err);
+
```
alfredh: nice fix, I cherry-picked it into master

src/event.c
```diff
@@ -116,6 +116,7 @@ int event_encode_dict(struct odict *od, struct ua *ua, enum ua_event ev,
 		err |= odict_entry_add(od, "direction", ODICT_STRING, dir);
 		err |= odict_entry_add(od, "peeruri",
 				       ODICT_STRING, call_peeruri(call));
+		err |= odict_entry_add(od, "id", ODICT_STRING, call_id(call));
```
alfredh: nice, added to master as a separate commit

src/ua.c
```diff
@@ -835,9 +835,10 @@ void ua_hangup(struct ua *ua, struct call *call,
 			return;
 	}
 
-	(void)call_hangup(call, scode, reason);
+	ua_event(ua, UA_EVENT_CALL_CLOSED, call,
+	         reason ? reason : "Connection reset by user");
 
-	ua_event(ua, UA_EVENT_CALL_CLOSED, call, reason);
+	(void)call_hangup(call, scode, reason);
```
alfredh: is this change needed, what is the reason for it ?

modules/ctrl_tcp/netstring/makefile
```diff
@@ -0,0 +1,11 @@
...
+LDFLAGS= -lm
+
+.PHONY: clean
+
+all: testsuite
+
+testsuite: testsuite.c netstring.o
+
+clean:
+	rm -f *.o testsuite
```
alfredh: this file is not used, it can be removed.

modules/ctrl_tcp/netstring/testsuite.c
```diff
@@ -0,0 +1,133 @@
+#include <stdio.h>
```
alfredh: this file is not used, it can be removed.

modules/ctrl_tcp/tcp_netstring.c
```diff
@@ -0,0 +1,215 @@
+
+#include <math.h>
```
alfredh: copyright header is missing ?

modules/ctrl_tcp/tcp_netstring.h
```diff
@@ -0,0 +1,12 @@
+
+
```
alfredh: copyright header is missing ?

modules/ctrl_tcp/tcp_netstring.c
```diff
@@ -0,0 +1,215 @@
...
+	/* handle re-assembly */
+	if (!netstring->mb) {
+		netstring->mb = mbuf_alloc(1024);
+		if (!netstring->mb) {
+			*errp = ENOMEM;
+			return true;
+		}
+	}
+
+	size_t pos = netstring->mb->pos;
```
alfredh: the declaration of `pos` should be moved to the beginning of the function

alfredh:
Great work! it looks much better now.

I think the unused netstring files can be deleted.

src/ua.c
jmillan: aef93f0a2535948606b04ffd311c863f6da4422d

````
    ua: emit 'call_hangup' event before calling the function

    otherwise the call session is dereferenced and call_id inaccesible.
````

`call_hangup()` releases `call->sess`, which is required for `call_id()` on the later `UA_EVENT_CALL_CLOSED` event generation.

https://github.com/jmillan/baresip/blob/ctrl_tcp/src/call.c#L759

modules/ctrl_tcp/tcp_netstring.c
jmillan: True, let me fix that.

src/ua.c
alfredh: I fixed it in a different way, by saving the callid string. Here is the commit

9d0b0fec87691312ade86b4d330265b5ca59c997

can you please revert this change ?

alfredh:
this PR should be ready for a merge now.

Please rebase/merge from master. After this is done I will have another look, and then merge.

jmillan:
Hi @alfredh,

I think it's ready.

alfredh:
Hi @jmillan 

Many thanks for your contribution :)

Alfred

jmillan:
You're welcome @alfredh,

Thanks to you.

PR #368, opened 2018-03-12
Vidsrc add timestamp

alfredh:
This PR adds a new _timestamp_ field to the vidsrc frame handler.

Background: The video encoding pipeline currently does not use the timestamp
from the video source at all. Instead the video codecs have a frame counter,
that is used together with `fps` to calculate the RTP timestamp.
This solution is not accurate, and I want to change it so that the exact timestamp
for each video frame (coming from the source) will be transported all the way
through the video pipeline, which currently looks like this:

1. Video source
2. Video filters
3. Video encoder and packetizer
4. RTP sending

For timestamp format the following units was considered:

* double float (seconds)
* uint64_t timestamp (milliseconds)
* uint64_t timestamp (microseconds)
* int64_t pts, Presentation Timestamp in timebase units
* uint64_t rtp_ts (RTP Timestamp in 90000Hz units)
* struct timestamp with value, duration and timebase units

For now the timestamp is `uint64_t timestamp` and the unit is defined
in the `VIDEO_TIMEBASE` macro (currently set to 1 million which means
in units of 1 microseconds). This type is subject to discussion.

| Vidsrc module | Timestamp units                |
| ------------- | ------------------------------ |
| avcapture     | double float seconds           |
| avformat      | int64_t pts in time_base units |
| cairo         | uint64_t milliseconds          |
| dshow         | ?                              |
| fakevideo     | uint64_t milliseconds          |
| qtcapture     |                                |
| rst           | uint64_t milliseconds          |
| v4l           | ?                              |
| v4l2          | struct timeval                 |
| vidbridge     | opaque, same as vidisp         |
| x11grab       | uint64_t milliseconds          |

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -21,6 +21,12 @@ extern "C" {
 #endif
 
 
+/*
+ * Clock-rate for video timestamp
+ */
+#define VIDEO_TIMEBASE 1000000U
+
+
 /* forward declarations */
 struct sa;
 struct sdp_media;
@@ -764,7 +770,8 @@ struct vidsrc_prm {
 	double fps;       /**< Wanted framerate                            */
 };
 
-typedef void (vidsrc_frame_h)(struct vidframe *frame, void *arg);
+typedef void (vidsrc_frame_h)(struct vidframe *frame, uint64_t timestamp,
+			      void *arg);
 typedef void (vidsrc_error_h)(int err, void *arg);
 
 typedef int  (vidsrc_alloc_h)(struct vidsrc_st **vsp, const struct vidsrc *vs,
@@ -1034,6 +1041,7 @@ int   video_debug(struct re_printf *pf, const struct video *v);
 uint64_t video_calc_rtp_timestamp(int64_t pts, double fps);
 double video_calc_seconds(uint64_t rtp_ts);
 struct stream *video_strm(const struct video *v);
+double video_timestamp_to_seconds(uint64_t timestamp);
 
 
 /*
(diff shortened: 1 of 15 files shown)


alfredh:
I am about the merge this PR to master very soon.

if someone would like to fix the `dshow` and `qtcapture` module that would be great :)

if you want to help with testing of this PR, you can build the branch
and run the following command:

```
$ ./baresip -mcairo.so -msdl2.so -e/vidloop
```

(replace `cairo.so` with a different video-source module).

Alfred

jmillan:
Hi,

This is the output of the command you provided in my machine (MacOS Sierra 10.12.6):

```
/vidloop
Enable video-loop on avcapture,nil: 64 x 64
vidloop: open video display (opengl.nil)
vidloop: avcapture,nil: open video source: 64 x 64 at 0 fps
vidloop: NOTE: pixel-format conversion needed: uyvy422  -->  yuv420p
status: [] []  intra=0  EFPS=14.9      4971 kbit/s
```

Some times I'm getting this message on baresip console. I don't know whether it can be related to this branch or not:

```
CoreAnimation: warning, deleted thread with uncommitted CATransaction; set CA_DEBUG_TRANSACTIONS=1 in environment to log backtraces, or set CA_ASSERT_MAIN_THREAD_TRANSACTIONS=1 to abort when an implicit transaction isn't created on a main thread.
```

alfredh:
@jmillan thanks for testing.

the warning about CoreAnimation is not related to the PR. it is a warning that the SDL2 window is created from a non-main thread. there are ways to fix it but I am not sure if we should do it.

PR #392, opened 2018-05-07
added account_mediaenc function

juha-h:
I was not able to figure out how to get account's media encoding and created a function for it.


alfredh:
many thanks :)

PR #394, opened 2018-05-08
added menc status handler

juha-h:
Added menc status handler.  Will provide a new pull request where it is used in zrtp module after this pull request has been accepted.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -99,6 +99,7 @@ enum call_event {
 	CALL_EVENT_CLOSED,
 	CALL_EVENT_TRANSFER,
 	CALL_EVENT_TRANSFER_FAILED,
+	CALL_EVENT_MENC,
 };
 
 struct call;
@@ -517,8 +518,11 @@ struct menc_media;
 
 typedef void (menc_error_h)(int err, void *arg);
 
+typedef void (menc_status_h)(const char *status, const char *prm, void *arg);
+
 typedef int  (menc_sess_h)(struct menc_sess **sessp, struct sdp_session *sdp,
-			   bool offerer, menc_error_h *errorh, void *arg);
+			   bool offerer, menc_status_h *statush,
+			   menc_error_h *errorh, void *arg);
 
 typedef int  (menc_media_h)(struct menc_media **mp, struct menc_sess *sess,
 			    struct rtp_sock *rtp, int proto,
@@ -599,6 +603,7 @@ enum ua_event {
 	UA_EVENT_CALL_DTMF_START,
 	UA_EVENT_CALL_DTMF_END,
 	UA_EVENT_CALL_RTCP,
+	UA_EVENT_CALL_MENC,
 
 	UA_EVENT_MAX,
 };
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -72,6 +72,7 @@ const char *account_outbound(const struct account *acc, unsigned ix);
 const char *account_stun_user(const struct account *acc);
 const char *account_stun_pass(const struct account *acc);
 const char *account_stun_host(const struct account *acc);
+const char *account_mediaenc(const struct account *acc);
 
 
 /*
--- a/modules/dtls_srtp/dtls_srtp.c
+++ b/modules/dtls_srtp/dtls_srtp.c
@@ -49,6 +49,7 @@
 struct menc_sess {
 	struct sdp_session *sdp;
 	bool offerer;
+	menc_status_h *statush;
 	menc_error_h *errorh;
 	void *arg;
 };
@@ -151,7 +152,8 @@ static bool verify_fingerprint(const struct sdp_session *sess,
 
 static int session_alloc(struct menc_sess **sessp,
 			 struct sdp_session *sdp, bool offerer,
-			 menc_error_h *errorh, void *arg)
+			 menc_status_h *statush, menc_error_h *errorh,
+			 void *arg)
 {
 	struct menc_sess *sess;
 	int err;
--- a/modules/dtls_srtp/dtls_srtp.c
+++ b/modules/dtls_srtp/dtls_srtp.c
@@ -167,6 +167,7 @@ static int session_alloc(struct menc_sess **sessp,
 
 	sess->sdp     = mem_ref(sdp);
 	sess->offerer = offerer;
+	sess->statush  = statush;
 	sess->errorh  = errorh;
 	sess->arg     = arg;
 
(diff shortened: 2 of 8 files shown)


alfredh:
hi @juha-h many thanks for the PR.

one thing that we should consider is if the status handler should be on
the media-level or the session-level.

for example a video-call has 2 media states (audio + video) and when using
e.g. DTLS-SRTP the media encryption will be established in parallel.

perhaps it is most flexible that each media stream has its own "encrypted" flag.

I guess a call should be considered "secure" if both audio and video is encrypted.


I have more comments but only this one for now :)

Alfred

alfredh:
I made a small comparison between all the Mediaenc modules:

|  Module   | Level  | Established      | Verified                |
|-----------|--------|------------------|-------------------------|
| dtls_srtp | Media  | DTLS complete    | SDP Fingerprint verified|
| gzrtp     | Session| ZRTP Exchange    | SAS Verified            |
| libsrtp   | Media  | SDP Offer/Answer | N/A                     |
| srtp      | Media  | SDP Offer/Answer | N/A                     |
| zrtp      | Session| ZRTP Exchange    | SAS Verified            |

With DTLS-SRTP, the media is secure when the DTLS handshake completes.

With SDES, the media is secure when the SDP Offer/Answer exchange
completes with a valid `a=crypto` attribute. Verification is not possible here.

I guess we can unify this with a status/secure handler on the session level.
DTLS-SRTP and SRTP modules can save the "is_secure" bit on the media object,
and then call the session level status/secure handler when all media streams
are secure.

Should we also add a `bool verified` flag to the handler ?

include/baresip.h
```diff
@@ -516,8 +518,11 @@ struct menc_media;
 
 typedef void (menc_error_h)(int err, void *arg);
 
+typedef void (menc_status_h)(const char *status, const char *prm, void *arg);
+
```
alfredh: is `menc_secure_h` a better and more self-describing name ?

are you planning to call this from:

```
static void on_zrtp_secure(zrtp_stream_t *stream)
{
	const struct menc_media *st = zrtp_stream_get_userdata(stream);
	const struct menc_sess *sess = st->sess;
	zrtp_session_info_t sess_info;

	zrtp_session_get(sess->zrtp_session, &sess_info);
	if (!sess_info.sas_is_verified && sess_info.sas_is_ready) {
		info("zrtp: verify SAS <%s> <%s> for remote peer %w"
		     " (type /zrtp %w to verify)\n",
		     sess_info.sas1.buffer,
		     sess_info.sas2.buffer,
		     sess_info.peer_zid.buffer,
		     (size_t)sess_info.peer_zid.length,
		     sess_info.peer_zid.buffer,
		     (size_t)sess_info.peer_zid.length);
	}
	else if (sess_info.sas_is_verified) {
		info("zrtp: secure session with verified remote peer %w\n",
		     sess_info.peer_zid.buffer,
		     (size_t)sess_info.peer_zid.length);
	}
}
```

juha-h:
Alfred E. Heggestad writes:

> +typedef void (menc_status_h)(const char *status, const char *prm, void *arg);
> +
> 
> is `menc_secure_h` a better and more self-describing name ?
> 
> are you planning to call this from:

As shown below, I'm currently calling the handler when peer of call
needs to be verified ("call verify" status value) and when peer of call
has been verified ("call verified" status value).

zrtp API includes also other callback events (protocol and security
related) that I'm not currently using or that are not currently
supported by zrtp module:

http://zfoneproject.com/docs/sdk/man/html/index.html

So perhaps 'menc_security_h' would be better than 'menc_secure_h' and
name of first param could be 'event' rather than 'status':

typedef void (menc_security_h)(const char *event, const char *prm, void *arg);


-- Juha

--------------------------------------------------------------------------

static void on_zrtp_secure(zrtp_stream_t *stream)
{
	const struct menc_media *st = zrtp_stream_get_userdata(stream);
	const struct menc_sess *sess = st->sess;
	zrtp_session_info_t sess_info;
	char buf[128] = "";

	zrtp_session_get(sess->zrtp_session, &sess_info);
	if (!sess_info.sas_is_verified && sess_info.sas_is_ready) {
		info("zrtp: verify SAS <%s> <%s> for remote peer %w"
		     " (type /zrtp_verify %w to verify)\n",
		     sess_info.sas1.buffer,
		     sess_info.sas2.buffer,
		     sess_info.peer_zid.buffer,
		     (size_t)sess_info.peer_zid.length,
		     sess_info.peer_zid.buffer,
		     (size_t)sess_info.peer_zid.length);
		if (sess->statush) {
			if (re_snprintf(buf, sizeof(buf), "%s,%s,%w",
					sess_info.sas1.buffer,
					sess_info.sas2.buffer,
					sess_info.peer_zid.buffer,
					(size_t)sess_info.peer_zid.length))
				(sess->statush)("call verify", buf, sess->arg);
		} else {
			warning("zrtp: failed to print verify arguments\n");
		}
	}
	else if (sess_info.sas_is_verified) {
		info("zrtp: secure session with verified remote peer %w\n",
		     sess_info.peer_zid.buffer,
		     (size_t)sess_info.peer_zid.length);
		if (sess->statush) {
			if (re_snprintf(buf, sizeof(buf), "%w",
					sess_info.peer_zid.buffer,
					(size_t)sess_info.peer_zid.length))
				(sess->statush)("call verified", buf,
						sess->arg);
		} else {
			warning("zrtp: failed to print verified argument\n");
		}
	}
}

juha-h:
Alfred E. Heggestad writes:

> Should we also add a `bool verified` flag to the handler ?

Perhaps not, since it does not apply to all events.

-- Juha

alfredh:
after thinking more about this, it looks like we need to signal these different
events from a Mediaenc module to the Core:

1. Media is secure
2. Verify request (ask the user to verify something)
3. Verified (peer is verified)

Number 2 and 3 would be used by ZRTP module (only).

Perhaps we should then define a generic event handler with an enum that
defines the event, for example:

```
enum menc_event {
    MENC_EVENT_SECURE,
    MENC_EVENT_VERIFY_REQUEST,
    MENC_EVENT_PEER_VERIFIED
};
```

and it would be used in the handler, instead of the string:

```
typedef void (menc_event_h)(enum menc_event event, const char *prm, void *arg);
```

perhaps something like this ?

juha-h:
Alfred,

What you suggest is fine with me.  Would you like me to create a pull
request based on the proposal?

-- Juha

-------------------------------------------------------------------------

juha-h:
Replaced by another pull request.

PR #410, opened 2018-05-22
added menc event handling and replaced zrtp module zrtp command with two commands

juha-h:
Based on Alfred's suggestion, added menc_event handler. Call event is still CALL_EVENT_MENC and the type is in the parameter.  Should there be separate CALL_EVENT_SECURE, CALL_EVENT_VERIFY_REQUEST, and CALL_EVENT_PEER_VERIFIED events instead?

-- Juha

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -58,6 +58,7 @@ int account_set_auth_user(struct account *acc, const char *user);
 int account_set_auth_pass(struct account *acc, const char *pass);
 int account_set_outbound(struct account *acc, const char *ob, unsigned ix);
 int account_set_display_name(struct account *acc, const char *dname);
+int account_set_regint(struct account *acc, uint32_t regint);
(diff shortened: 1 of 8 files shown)


alfredh:
Great, many thanks!

I think this is looking much better. It should be enough with one event handler
for the session level I think. Other protocols can wait for all media lines to be secure
and then call the session event handler.

If needed in the future, additional events can be added.

juha-h:
Alfred E. Heggestad writes:

> I think this is looking much better. It should be enough with one
> event handler for the session level I think. Other protocols can wait
> for all media lines to be secure and then call the session event
> handler.

I didn't quite understand the above.  What kind of change you propose to
the pull request?

-- Juha

juha-h:
Closing due to conflicts.  Will create a new one.

PR #411, opened 2018-05-22
added account_set_mediaenc API function

juha-h:


src/account.c
```diff
@@ -486,6 +486,20 @@ int account_set_regint(struct account *acc, uint32_t regint)
 }
 
 
+int account_set_mediaenc(struct account *acc, const char *mencid)
+{
+	if (!acc)
+		return EINVAL;
+
```
alfredh: I think we should check that `mencid` contains a valid mediaenc module,
by looking up the mencid in the list of mediaencs.

This code is done in `account_alloc`:

```
        if (acc->mencid) {
                acc->menc = menc_find(baresip_mencl(), acc->mencid);
                if (!acc->menc) {
                        warning("account: mediaenc not found: `%s'\n",
                                acc->mencid);
                }
        }
```

This makes the code more robust, in cases where the user tries
to set an invalid mediaenc.

PR #412, opened 2018-05-23
do not default auth_user to sip uri userpart, but leave unset instead

juha-h:
auth_user may not be needed at all or it may not be the same as sip uri
userpart.  for example, sip uri userpart may be telephone number and
auth_user may be userid.

### DIFF
--- a/src/account.c
+++ b/src/account.c
@@ -322,8 +322,6 @@ static int sip_params_decode(struct account *acc, const struct sip_addr *aor)
 
 	if (0 == msg_param_decode(&aor->params, "auth_user", &auth_user))
 		err |= pl_strdup(&acc->auth_user, &auth_user);
-	else
-		err |= pl_strdup(&acc->auth_user, &aor->uri.user);
 
 	if (pl_isset(&aor->dname))
 		err |= pl_strdup(&acc->dispname, &aor->dname);


alfredh:
BTW, also notice that the selftest breaks:

```
[ RUN      ] test_ua_register_auth
mock/sipsrv: user not found ()
reg: sip:alfredh@localhost: 401 Unauthorized (mock SIP server)
selftest: event handler error: Unknown error 217
selftest: ua_register test failed (Unknown error 217)
TEST_ERR: test/ua.c:449: (Unknown error 217)
test_ua_register_auth: test failed (Unknown error 217)
test failed (Unknown error 217)
```

The selftest code is there to verify that certain things are working,
it even has a small SIP proxy inside :)

The selftest code can be run with this command:

```
$ make test
```

juha-h:
Alfred E. Heggestad writes:

> suggested code:
> 
> ```
> if (auth_user)
> 	*username = mem_ref(acc->auth_user);
> else
> 	*username = mem_ref(acc->username);  
> ```

I tried like this:

int account_auth(const struct account *acc, char **username, char **password,
		 const char *realm)
{
	if (!acc)
		return EINVAL;

	(void)realm;

	if (acc->auth_user) {
		*username = mem_ref(acc->auth_user);
		return 0;
	} else {
		return pl_strcpy(&(acc->luri.user), *username,
				strlen(*username));
	}
}

but 'make test' given segmentation fault:

RUN      ] test_ua_register_auth
Makefile:216: recipe for target 'test' failed
make: *** [test] Segmentation fault

How can I make it dump core?

-- Juha

alfredh:
the returned "username" parameter must be a `mem_ref` able object.

it is probably better to use this one to copy the username string:

```
int      pl_strdup(char **dst, const struct pl *src);
```

juha-h:
Alfred E. Heggestad writes:

> the returned "username" parameter must be a `mem_ref` able object.
> 
> it is probably better to use this one to copy the username string:
> 
> ```
> int      pl_strdup(char **dst, const struct pl *src);
> ```

Tried like this:

	if (acc->auth_user) {
		*username = mem_ref(acc->auth_user);
		return 0;
	} else {
		return pl_strdup(username, &(acc->luri.user));
	}

Now 'make test' does not crash, but gives error when main loop is run
and gets some event:

[ RUN      ] test_ua_register_auth
reg: sip:alfredh@localhost: 401 Unauthorized (mock SIP server)
selftest: event handler error: Unknown error 217
selftest: ua_register test failed (Unknown error 217)

TEST_ERR: test/ua.c:449: (Unknown error 217)
test_ua_register_auth: test failed (Unknown error 217)
test failed (Unknown error 217)
re main loop:
  maxfds:  1024
  nfds:    24
  method:  3 (epoll)

-- Juha

alfredh:
perhaps it is better to add `int err` and set the return value, instead of returning
from the function in the middle of the function ?

are you sure that the password variable is set ?

juha-h:
Alfred E. Heggestad writes:

> perhaps it is better to add `int err` and set the return value,
> instead of returning from the function in the middle of the function ?

Did that.

> are you sure that the password variable is set ?

Should be, since there exist these in test/au.c:

#define USER   "alfredh"
#define PASS   "pass@word"
#define DOMAIN "localhost"

ans PASS is used in reg_auth function.

-- Juha

src/account.c
```diff
@@ -541,15 +539,20 @@ int account_set_display_name(struct account *acc, const char *dname)
...
 {
+	int err = 0;
+
 	if (!acc)
 		return EINVAL;
 
 	(void)realm;
 
-	*username = mem_ref(acc->auth_user);
-	*password = mem_ref(acc->auth_pass);
```
alfredh: you also removed setting of the `*password` :) this will break.

src/account.c
```diff
@@ -541,15 +539,20 @@ int account_set_display_name(struct account *acc, const char *dname)
...
+	if (acc->auth_user) {
+		*username = mem_ref(acc->auth_user);
+	} else {
```
alfredh: please move the else part to the next line:
```
}
else {
```

juha-h:
Alfred E. Heggestad writes:

> -	*username = mem_ref(acc->auth_user);
> -	*password = mem_ref(acc->auth_pass);
> 
> you also removed setting of the `*password` :) this will break.

Happened by accident. Thanks for noticing.

PR #413, opened 2018-05-23
added account_set_audio_codecs API function

juha-h:
for ease of use and implementation, audio codecs argument is a string
containing comma separated list of codecs like in accounts file

### DIFF
--- a/src/account.c
+++ b/src/account.c
@@ -231,6 +231,8 @@ static int audio_codecs_decode(struct account *acc, const struct pl *prm)
 			}
 
 			/* NOTE: static list with references to aucodec */
+                        acc->acv[i].prev = acc->acv[i].next = NULL;
+                        acc->acv[i].list = NULL;
 			list_append(&acc->aucodecl, &acc->acv[i++], ac);
 
 			if (i >= ARRAY_SIZE(acc->acv))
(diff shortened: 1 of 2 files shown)


src/account.c
```diff
@@ -506,6 +508,31 @@ int account_set_mediaenc(struct account *acc, const char *mencid)
...
+ *
+ * @param acc      User-Agent account
+ * @param codecs   Comma separed list of audio codecs (NULL to disable)
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int account_set_audio_codecs(struct account *acc, const char *codecs)
+{
+	char buf[256];
+	struct pl pl;
```
alfredh: the `acc` parameter must be checked, if null it should return `EINVAL`.

src/account.c
```diff
@@ -506,6 +508,31 @@ int account_set_mediaenc(struct account *acc, const char *mencid)
...
+
+	if (!codecs) {
+		list_clear(&acc->aucodecl);
```
alfredh: I believe the list can be cleared in all cases, and then we also dont need
to reset the list element (line 234).

something like this:

```
list_clear()

if (codecs) {
  audio_codecs_decode()
}
```

PR #419, opened 2018-05-27
zrtp: replaced zrtp command with zrtp_verify and zrtp_unverify commands

juha-h:


alfredh:
Thanks @juha-h 

FYI @glenvt18

glenvt18:
It makes sense. Now there is a way to reset a verified flag without cleaning the whole zrtp cache.

PR #424, opened 2018-05-31
added ua_update APU function

juha-h:
How about this as a solution to issue https://github.com/alfredh/baresip/issues/408?


juha-h:
Removing of possible outbound related extensions is still missing if sipnat is not outbound.

Issue #230, opened 2017-03-31
Is it possible to manipulate custom SIP headers with baresip?


mikhail-barg:
I'd like to be able to set custom headers for dials (INVITEs), and also I want to read incoming custom headers. I'm pretty much new to baresip, so I'm not used to the code yet. Could you give me some directions?

alfredh:
Please see here https://github.com/alfredh/baresip/blob/master/src/call.c#L1516

```
	err = sipsess_connect(&call->sess, uag_sipsess_sock(),
			      call->peer_uri,
			      call->local_name,
			      call->local_uri,
			      ua_cuser(call->ua),
			      routev[0] ? routev : NULL,
			      routev[0] ? 1 : 0,
			      "application/sdp", desc,
			      auth_handler, call->acc, true,
			      sipsess_offer_handler, sipsess_answer_handler,
			      sipsess_progr_handler, sipsess_estab_handler,
			      sipsess_info_handler, sipsess_refer_handler,
			      sipsess_close_handler, call,
			      "Allow: %s\r\n%H", uag_allowed_methods(),
			      ua_print_supported, call->ua);
	if (err) {
		warning("call: sipsess_connect: %m\n", err);
	}
```

additional custom headers can be added after the "Allow" string

PR #425, opened 2018-05-31
Sending and receiving custom headers

Encamy:
Hello Alfred
According to issue [#230](https://github.com/alfredh/baresip/issues/230) @mikhail-barg added the opportunity to send and to receive custom headers via INVITE request.
It can be useful in many situations when we need to send information to the server or to receive information from the server.

**Example of sending headers:**
at calling a peer:
```c++
int err = 0;
int some_id = 7;
struct custom_hdrs *custom_hdrs;

err = custom_hdrs_alloc(&custom_hdrs); //return 0 if ok, otherwise error code; err should be checked!
err = custom_hdrs_add(custom_hdrs, "X-CALL_ID", some_id); //also should be checked
err = custom_hdrs_add(custom_hdrs, "HEADER_NAME", "VALUE"); 
err = ua_connect(userAgent, &callp, NULL, number, NULL, VIDMODE_OFF, custom_hdrs); //and here too
mem_deref(custom_hdrs);
```
Packet captured with wireshark:
image

This headers can be parsed with a server (asterisk).

Because headers can't be passed from one peer to other we can catch them with a server and pass them to second peer with adding headers at server-side.
For a simpler example just pass some header (asterisk - lua pbx):
```lua
    ["_."] = function(context, extension)
      app.SIPAddHeader("X-CUSTOM_HEADER: value")
      app.Dial('SIP/'..extension);
    end;
``` 
image

**After this we can catch incoming headers.**
```c++
bool ExtractCustomHeadersFromPl(const struct pl *name, const struct pl *val)
{
    if (pl_strcasecmp(name, "X-CUSTOM_HEADER") == 0)
    {
        if (pl_strcasecmp(val, "value") == 0)
        {
            *dst = 1;    //mean that incoming header is "X-CUSTOM_HEADER:value"
        }
    }
    return false;
}

static int CustomHeaderExtractor(const struct pl *name, const struct pl *val, void *arg)
{
    (void *)arg;
    (void *)ExtractCustomHeadersFromPl(name, val);
    return 0;    //no error;
}
```
```c++
struct call *call; //pointer to incoming call

const struct custom_hdrs *hdrs = call->custom_hdrs;
custom_hdrs_apply(hdrs, CustomHeaderExtractor, NULL);
```

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -150,9 +150,25 @@ void call_set_current(struct list *calls, struct call *call);
 
 
 /*
+* Custom headers
+*/
+struct custom_hdrs;
+const struct custom_hdrs *call_get_custom_hdrs(const struct call *call);
+typedef int (custom_hdrs_h)(const struct pl *name, const struct pl *val,
+                            void *arg);     //returns error code if any
+int custom_hdrs_alloc(struct custom_hdrs **hdrs);
+int custom_hdrs_add(struct custom_hdrs *hdrs, const char *name,
+					const char *val);
+int custom_hdrs_add_pl(struct custom_hdrs *hdrs, const struct pl *name,
+					   const struct pl *val);
+int custom_hdrs_add_int(struct custom_hdrs *hdrs, const char *name, int val);
+int custom_hdrs_apply(const struct custom_hdrs *hdrs,
+					  custom_hdrs_h *h, void *arg);
+
+
+/*
  * Conf (utils)
  */
-
 
 /** Defines the configuration line handler */
 typedef int (confline_h)(const struct pl *addr, void *arg);
@@ -168,7 +184,6 @@ int  conf_get_sa(const struct conf *conf, const char *name, struct sa *sa);
 bool conf_fileexist(const char *path);
 void conf_close(void);
 struct conf *conf_cur(void);
-
 
 /*
  * Config (core configuration)
@@ -637,7 +652,8 @@ typedef void (ua_exit_h)(void *arg);
 int  ua_alloc(struct ua **uap, const char *aor);
 int  ua_connect(struct ua *ua, struct call **callp,
 		const char *from_uri, const char *uri,
-		const char *params, enum vidmode vmode);
+		const char *params, enum vidmode vmode,
+		struct custom_hdrs *custom_hdrs);
 void ua_hangup(struct ua *ua, struct call *call,
 	       uint16_t scode, const char *reason);
 int  ua_answer(struct ua *ua, struct call *call);
@@ -665,6 +681,7 @@ enum presence_status ua_presence_status(const struct ua *ua);
 void ua_presence_status_set(struct ua *ua, const enum presence_status status);
 void ua_set_media_af(struct ua *ua, int af_media);
 void ua_set_catchall(struct ua *ua, bool enabled);
+int ua_add_xhdr_filter(struct ua *ua, const char *hdr_name);
 
 
 /* One instance */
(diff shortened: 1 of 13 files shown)


mikhail-barg:
@alfredh Could you please have a look at this PR? We think that this feature would be a nice addition to baresip. It makes posible exchanging infromation between SIP clients and server, like passing some call-related meta-data to SIP client together with call.

We actually use similar feature from some other propietary lib, but we'd like to switch to baresip. So we've come up with this implementation, and would like to share it with community.

alfredh:
thanks for the PR. I will have a look at this soon.

Do you also need custom headers for other methods such as *REGISTER* ?

one thing I noticed is that your new type for storing customer headers is
a basic key=value type container. the Ordered Dictionary type `struct odict`
in libre provides a similar container. would it make sense to use this type ?

mikhail-barg:
> Do you also need custom headers for other methods such as REGISTER ?

That's quite interesting. Frankly speaking, I've thought that it's impossible to have custom headers in other messages, but it seems that it's not true. Thanks for pointing at this matter!

As for us, we are generally focused on handling INVITE message, because it seems that Asterisk is capable to handle custom header in this message only. I'll keep this matter in mind, maybe we could find some way to make use of this functional elsewhere.

> one thing I noticed is that your new type for storing customer headers is a basic key=value type container. the Ordered Dictionary type `struct odict` in libre provides a similar container. would it make sense to use this type ?

Thanks for the point again. It seems there's really no point in creating a separate entity for this task. We'll reqork the PR.

alfredh:
in general custom SIP headers can be appended to any SIP methods.
custom SIP headers that is not defined by IETF/IANA should be prefixed with `X-`

I would like to split this feature into two parts:

1. Sending custom headers
2. Receiving custom headers

Most of the SIP apis in libre support custom headers, as a simple string.
(defined in `fmt` as a formatted string).

Example:

```
int sipreg_register(struct sipreg **regp, struct sip *sip, const char *reg_uri,
		    const char *to_uri, const char *from_name,
		    const char *from_uri, uint32_t expires,
		    const char *cuser, const char *routev[], uint32_t routec,
		    int regid, sip_auth_h *authh, void *aarg, bool aref,
		    sip_resp_h *resph, void *arg,
		    const char *params, const char *fmt, ...);

int  sipsess_connect(struct sipsess **sessp, struct sipsess_sock *sock,
		     const char *to_uri, const char *from_name,
		     const char *from_uri, const char *cuser,
		     const char *routev[], uint32_t routec,
		     const char *ctype, struct mbuf *desc,
		     sip_auth_h *authh, void *aarg, bool aref,
		     sipsess_offer_h *offerh, sipsess_answer_h *answerh,
		     sipsess_progr_h *progrh, sipsess_estab_h *estabh,
		     sipsess_info_h *infoh, sipsess_refer_h *referh,
		     sipsess_close_h *closeh, void *arg, const char *fmt, ...);
```

I think for outgoing custom headers, there is no need to create a new type
for this. Instead we can just use a simple string that can contain 1 or more
custom headers.

For incoming custom headers, your patch is doing the right thing.
Extracting the custom headers from `struct sip_msg` and making them
available to the application.

mikhail-barg:
> I think for outgoing custom headers, there is no need to create a new type
for this. Instead we can just use a simple string that can contain 1 or more
custom headers.

> For incoming custom headers, your patch is doing the right thing.
Extracting the custom headers from struct sip_msg and making them
available to the application.

I got your point on not creating new type, actually we may just have a regular `list` instead of `custom_hdrs`. We'll update the code this way, thanks!

But regarding using a plain string for outgoing headers, I'd prefer to have some structured data instead of a plain string. The reason is that this way we can keep the actual headers in the call. Therefore handling of headers for incoming and outgoing calls would be the same, via `call->custom_hdrs`. I think this is a proper way to do this.

alfredh:
will you make a new PR for this work, or just modify this PR ?

if you have time please also add a new test in selftest for this usecase.
it can be a simple test that just sends a custom header from A to B.

Encamy:
Hello
We will make a new PR a bit later

PR #439, opened 2018-06-08
Dshow fixes

nicotyze:
This is a new fix for dshow module, related to pixel format:
- RGB32 pixel format is passed to vidframe_init_buf in order to be consistent with media subtype 
- By default in Dshow, RGB32 image orientation is bottom-up => a flipping loop is added in the capturing callback in order to provide a top-down oriented image buffer as expected for the RGB32 to YUV420 conversion.


modules/dshow/dshow.cpp
```diff
@@ -86,12 +86,24 @@ class Grabber : public ISampleGrabberCB {
...
+		uint32_t *buf_RGB32;
 		struct vidframe vidframe;
 		uint64_t timestamp = sample_time * VIDEO_TIMEBASE;
 
-		/* XXX: should be VID_FMT_BGR24 */
 		vidframe_init_buf(&vidframe, VID_FMT_RGB32, &src->size, buf);
 
+		//By default in Dshow RGB32 image orientation is bottom-up
+		buf_RGB32 = (uint32_t *)buf;
+		for (int i = buf_len_RGB32-1 ; i > buf_len_RGB32/2; i-=1) {
```
alfredh: the variable `i` is also declared in line 89

alfredh:
thanks for your patch. on which OS/HW was this patch tested ?
(I dont have access to any Windows machines).

nicotyze:
My testing environment;
- Binaries tested on Win10, HP laptop (core I5)
- Webcams: integrated laptop webcam, HP HD 4310 and logitech C920 => RGB32 capturing pixel format is  always supported with  bottom-up image orientation.

Without setting RGB32 pix format in vidframe_init_buf,  a vidloop craches (RGB24 => memory size mismatches because media type is set to RGB32 for the grabber).
Without the "flipping loop", images are upside down.

PR #445, opened 2018-06-11
SDL2 Win32 only workarounds:

nicotyze:
This is a proposal to fix SDL2 "unusability" on Win32:
- SDL2 window is not responsive: can't be moved, can't be  (un)changed  (from)to full screen.
- If a SDL2 display function is called (e.g. with /vidoop) then baresip crashes when exiting the application (q pressed)

Proposed fix:
- Move recursive call of event_handler from module_init (alloc function) to display function (without recursivity)
- Replace SDL_VideoQuit by SDL_Quit: this fixes applications crashes when quiting the application (video displayed with sdl2)

Testing environment:
- Binaries tested on Win10, HP laptop (core I5) 
- Webcams: integrated laptop webcam, HP HD 4310 and logitech C920

### DIFF
--- a/modules/sdl2/sdl.c
+++ b/modules/sdl2/sdl.c
@@ -86,7 +86,9 @@ static void event_handler(void *arg)
 	struct vidisp_st *st = arg;
 	SDL_Event event;
 
+#ifndef WIN32
 	tmr_start(&st->tmr, 100, event_handler, st);
+#endif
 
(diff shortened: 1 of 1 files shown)


alfredh:
can you make the sdl_quit change a separate PR?
the fixes should be generic for all platforms, i.e. no ifdef


polling events from display is ok, if it works with multithreading

PR #446, opened 2018-06-12
Add a specific tagging and processing for long commands passed throug httpd

nicotyze:
This is a proposal to avoid the 'cascading' effect of long command name printing with httpd  UI.
    
Example with /vidloop command (http://127.0.0.1:8000/?/vidloop):

"Cascading" effect

	/
	/v
	/vi
	/vid
	/vidl
	/vidlo
	/vidloo
	/vidloop
	Enable video-loop on ,: 352 x 288

With the proposed patch

	Enable video-loop on ,: 352 x 288

Tested on Ubuntu 16.04 (Firefox) and Win10 (Firefox, Chrome)

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -771,6 +771,7 @@ int  cmd_init(struct commands **commandsp);
 int  cmd_register(struct commands *commands,
 		  const struct cmd *cmdv, size_t cmdc);
 void cmd_unregister(struct commands *commands, const struct cmd *cmdv);
+int  cmd_fill_ctx(struct cmd_ctx **ctxp, char key);
 int  cmd_process(struct commands *commands, struct cmd_ctx **ctxp, char key,
 		 struct re_printf *pf, void *data);
 int  cmd_process_long(struct commands *commands, const char *str, size_t len,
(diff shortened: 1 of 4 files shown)


alfredh:
perhaps it is better to add echo on/off mode?

nicotyze:
Hi,
Ok, I understand the need for debugging purposes (?).
This is an idea :
http://127.0.0.1:8000/??/vidloop => echo on
http://127.0.0.1:8000/?/vidloop => echo off

What do you think about it ?

Nicolas

alfredh:
hi @nicotyze 

many thanks for this PR. I think this solution is a bit too complex, so
I am proposing something simpler.

I have added a new function:

`int ui_input_long_command(struct re_printf *pf, const struct pl *pl)`


this function can be used for sending long commands, with or without
the `/` prefix. This command does not echo the input characters.

So what I suggest is that you in the httpd module detect the first character
of the input string. If the first char is `/` you can call `ui_input_long_command`.
otherwise call `ui_input_pl` which handles both short and long commands.


Alfred

PR #449, opened 2018-06-13
notify application when DTLS-SRTP is complete

juha-h:
Should also media name, RTP|RTPC, and profile name be included as parameter?


alfredh:
the callback handler should also consider different media lines, such as audio
and video.

for RTP and RTCP ports, the RTP port is always used but the RTCP port is optional.
for example if rtp/rtcp multiplexing is enabled and negotiated, it only needs
one port (RTP) to multiple both RTP and RTCP over.

so this logic may get a bit complex ..

PR #451, opened 2018-06-14
SDL2 events polling

nicotyze:
Following https://github.com/alfredh/baresip/pull/445 PR, this is a new proposal (platform independent)  to move sdl2 events polling in the display function.

This is the only way have found to make it works on Win10.  
I've successfully tested it on Ubunt16.04 as well but I can not test it on Mac OS  so far...

### DIFF
--- a/modules/sdl2/sdl.c
+++ b/modules/sdl2/sdl.c
@@ -5,6 +5,7 @@
  */
 
 #include <SDL2/SDL.h>
+#include <pthread.h>
 #include <re.h>
 #include <rem.h>
 #include <baresip.h>
@@ -27,6 +28,9 @@ struct vidisp_st {
 	bool fullscreen;                /**< Fullscreen flag       */
 	struct tmr tmr;
 	Uint32 flags;
+	char *title ;
+	struct vidframe *frame ;
+	pthread_mutex_t frame_mutex;
 };
 
 
@@ -35,6 +39,7 @@ static struct vidisp *vid;
 
 static void event_handler(void *arg);
 
+static int display(struct vidisp_st *st);
 
 static uint32_t match_fmt(enum vidfmt fmt)
 {
@@ -86,6 +91,8 @@ static void event_handler(void *arg)
 	struct vidisp_st *st = arg;
 	SDL_Event event;
 
+	tmr_start(&st->tmr, 5, event_handler, st);
+
 	/* NOTE: events must be checked from main thread */
 	while (SDL_PollEvent(&event)) {
 
@@ -109,6 +116,13 @@ static void event_handler(void *arg)
 				SDL_SetWindowFullscreen(st->window, st->flags);
 				break;
 
+			case SDLK_F5:/*
+				/* A new frame is ready to be displayed */
+				pthread_mutex_lock(&st->frame_mutex);
+				display(st);
+				pthread_mutex_unlock(&st->frame_mutex);
+				break;
+
 			default:
 				break;
 			}
@@ -121,6 +135,8 @@ static void destructor(void *arg)
 {
 	struct vidisp_st *st = arg;
 
+	mem_deref(st->frame);
+	mem_deref(st->title);
 	tmr_cancel(&st->tmr);
 	sdl_reset(st);
 }
@@ -148,6 +164,9 @@ static int alloc(struct vidisp_st **stp, const struct vidisp *vd,
 	st->vd = vd;
 	st->fullscreen = prm ? prm->fullscreen : false;
 
+	pthread_mutex_init(&st->frame_mutex, NULL);
+	tmr_start(&st->tmr, 100, event_handler, st);
+
 	if (err)
 		mem_deref(st);
 	else
@@ -157,8 +176,7 @@ static int alloc(struct vidisp_st **stp, const struct vidisp *vd,
 }
 
 
-static int display(struct vidisp_st *st, const char *title,
-		   const struct vidframe *frame)
+static int display(struct vidisp_st *st)
 {
 	void *pixels;
 	uint8_t *d;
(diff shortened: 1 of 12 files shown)


alfredh:
thanks, I tested your PR on OSX 10.12 but it crashes:

```
2018-06-14 22:00:31.134249+0200 baresip[5593:7771188] [General] An uncaught exception was raised
2018-06-14 22:00:31.134283+0200 baresip[5593:7771188] [General] nextEventMatchingMask should only be called from the Main Thread!
2018-06-14 22:00:31.134341+0200 baresip[5593:7771188] [General] (
	0   CoreFoundation                      0x00007fffa93f82cb __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x00007fffbe21348d objc_exception_throw + 48
	2   AppKit                              0x00007fffa75e1e82 -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 4480
	3   libSDL2-2.0.0.dylib                 0x000000010389d80c Cocoa_PumpEvents + 179
	4   libSDL2-2.0.0.dylib                 0x000000010381c021 SDL_PumpEvents_REAL + 23
	5   libSDL2-2.0.0.dylib                 0x000000010381c089 SDL_WaitEventTimeout_REAL + 57
	6   sdl2.so                             0x0000000102fa7a1b event_handler + 43
	7   sdl2.so                             0x0000000102fa71dd display + 61
	8   baresip                             0x0000000100026a8f vidisp_display + 79
	9   vidloop.so                          0x000000010ca50387 display + 343
	10  vidloop.so                          0x000000010ca50685 vidsrc_frame_handler + 645
	11  cairo.so                            0x0000000100485697 process + 359
	12  cairo.so                            0x000000010048548e read_thread + 158
	13  libsystem_pthread.dylib             0x00007fffbed1293b _pthread_body + 180
	14  libsystem_pthread.dylib             0x00007fffbed12887 _pthread_body + 0
	15  libsystem_pthread.dylib             0x00007fffbed1208d thread_start + 13
```

the SDL documentation says this:

"As this function implicitly calls SDL_PumpEvents(), you can only call this function in the thread that set the video mode."

https://wiki.libsdl.org/SDL_PollEvent?highlight=%28%5CbCategoryAPI%5Cb%29%7C%28SDLFunctionTemplate%29


I think that SDL is supposed to be portable, i.e. the same code should work
on all supported platforms.

nicotyze:
Thanks for the test.

SDL documentation also says:
"You should not expect to be able to create a window, render, or receive
events on any thread other than the main one. For platform-specific
exceptions or complicated options ask on the forums/mailing list
<https://discourse.libsdl.org/>."

So, to fulfill this requirement I propose a new commit where all SDL2
functions are called from the main thread:
  - video frames are pushed through a new push_frame registered function
  - a pushed frame is stored in an intermediary vidframe structure before
being displayed
  - once the video frame is copied, a specific SDL event is pushed
  - video frame display is called as a specific SDL event callback

By doing so, the crash on Win10 with SDL_VideoQuit disappears (good point
for this approach).

Tested successfully on Ubuntu 16.04 as well but still needs to be tested on
OS X.


2018-06-14 22:08 GMT+02:00 Alfred E. Heggestad <notifications@github.com>:

alfredh:
really great, nice patch !

I tested your patch on OSX 10.13 and it works fine.

I agree that rendering/events should be done in the main thread. I have also
used ffplay.c from the FFmpeg project as a reference, they also do all rendering
from the main thread.

modules/sdl2/sdl.c
```diff
@@ -5,6 +5,7 @@
  */
 
 #include <SDL2/SDL.h>
+#include <pthread.h>
```
alfredh: should we use pthread mutex here or SDL mutex ?

modules/sdl2/sdl.c
```diff
@@ -111,6 +116,13 @@ static void event_handler(void *arg)
 				SDL_SetWindowFullscreen(st->window, st->flags);
 				break;
 
+			case SDLK_F5:/*
+				/* A new frame is ready to be displayed */
+				pthread_mutex_lock(&st->frame_mutex);
+				display(st);
+				pthread_mutex_unlock(&st->frame_mutex);
+				break;
+
```
alfredh: perhaps it is better to use `SDL_USEREVENT` for this ?

modules/sdl2/sdl.c
```diff
@@ -150,6 +164,7 @@ static int alloc(struct vidisp_st **stp, const struct vidisp *vd,
 	st->vd = vd;
 	st->fullscreen = prm ? prm->fullscreen : false;
 
+	pthread_mutex_init(&st->frame_mutex, NULL);
```
alfredh: you must also call `pthread_mutex_destroy` in the destructor.

modules/sdl2/sdl.c
```diff
@@ -161,32 +176,32 @@ static int alloc(struct vidisp_st **stp, const struct vidisp *vd,
 }
 
 
-static int display(struct vidisp_st *st, const char *title,
-		   const struct vidframe *frame)
+static int display(struct vidisp_st *st)
```
alfredh: I would suggest to keep the `frame` argument here. 

then you dont have to change all the lines from `frame` to `st->frame`
and hence will make the patch smaller

modules/sdl2/sdl.c
```diff
@@ -297,6 +315,38 @@ static int display(struct vidisp_st *st, const char *title,
...
+
+	/* Allocate st->frame and st->title from first frame info */
+	if (!st->frame) {
+
+		err = vidframe_alloc(&st->frame, frame->fmt,
+					 &frame->size);
+		if (err)
+			return err;
+
+		str_dup(&st->title, title);
```
alfredh: the return value (int err) must be checked.

modules/sdl2/sdl.c
```diff
@@ -329,7 +379,7 @@ static int module_close(void)
 {
 	vid = mem_deref(vid);
 
-	SDL_Quit();
+	SDL_VideoQuit();
```
alfredh: I think it is okay to just call `SDL_Quit` now ..

alfredh:
I get a compiler warning on OSX 10.13

```
make sdl2.so
  CC [M]  build-x86_64/modules/sdl2/sdl.o
modules/sdl2/sdl.c:120:5: warning: '/*' within block comment [-Wcomment]
                                /* A new frame is ready to be displayed */
                                ^
1 warning generated.
  LD [M]  sdl2.so
```

alfredh:
I tested vidloop again today with this configuration:

```
video_size              "1080x720"
video_fps               120.00
```

it is working, but the re_main loop stops responding to events.
for example the stdin is blocked, and baresip does not respond to
keys anymore.

this is probably because of the while loop here:

`while (SDL_PollEvent(&event)`

after the last frame is displayed, and the mutex is unlocked, the next frame
comes in very fast and does not give the while loop a chance to break.

since the timer is running with an interval of 5 ms (max refresh rate 200 Hz)
I guess it should be better to just check for any pending frames in that
timer handler. hence there is no need for a dedicated "refresh" event.

another thing that I thought about today; the regular video call (`video.c`)
will present frames in the main thread. it is only the vidloop module that will display
frames from a worker thread (e.g. cairo). perhaps we should move this logic
to the vidloop module instead ? I guess there are also other display modules
that require frames to be displayed from the main thread...

PR #462, opened 2018-06-21
video switch command added to menu module.

alishir:
Using vidsrc driver,device to change video source during call.


alfredh:
Many thanks :)

alishir:
My pleasure :)

PR #468, opened 2018-06-23
don't append domain if uri is IP address.

alishir:
If user use IP address of peer, uri_complete try to append domain to the uri and call doesn't established. In this pull request first we check peer uri and if it is IP address we don't append domain to it.

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -756,11 +756,14 @@ static int uri_complete(struct ua *ua, struct mbuf *buf, const char *uri)
 	err |= mbuf_write_str(buf, uri);
 
 	/* Append domain if missing and uri is not IP address */
-    uri_is_ip = inet_pton(AF_INET, uri, &(sa.sin_addr));	/* check if uri is valid IPv4 address */
+	/* check if uri is valid IPv4 address */
+	uri_is_ip = inet_pton(AF_INET, uri, &(sa.sin_addr));
 #if HAVE_INET6
-	uri_is_ip |= inet_pton(AF_INET6, uri, &(sa.sin_addr));  /* check if uri is valid IPv6 address */
+	/* check if uri is valid IPv6 address */
+	uri_is_ip |= inet_pton(AF_INET6, uri, &(sa.sin_addr));
 #endif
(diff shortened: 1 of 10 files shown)


alishir:
The better approach to test IP in uri is using regex. I tested [some IP validation](https://www.safaribooksonline.com/library/view/regular-expressions-cookbook/9780596802837/ch07s16.html) regex with `re_regex` but the result are always non zero.

src/ua.c
```diff
@@ -752,8 +755,15 @@ static int uri_complete(struct ua *ua, struct mbuf *buf, const char *uri)
...
-	if (0 != re_regex(uri, len, "[^@]+@[^]+", NULL, NULL)) {
+	/* Append domain if missing and uri is not IP address */
+	/* check if uri is valid IPv4 address */
+	uri_is_ip = inet_pton(AF_INET, uri, &(sa.sin_addr));
+#if HAVE_INET6
+	/* check if uri is valid IPv6 address */
+	uri_is_ip |= inet_pton(AF_INET6, uri, &(sa.sin_addr));
+#endif
+	if (0 != re_regex(uri, len, "[^@]+@[^]+", NULL, NULL) &&
+		1 != uri_is_ip) {
```
alfredh: it is better to use the `sa` functions from libre, they are portable and handle
both IPv4 and IPv6.

for example:

```
	struct sa tmp;

	uri_is_ip = (0 == sa_set_str(&tmp, uri, 0));
```

src/ua.c
```diff
@@ -739,6 +740,8 @@ static int uri_complete(struct ua *ua, struct mbuf *buf, const char *uri)
 {
 	size_t len;
 	int err = 0;
+	int uri_is_ip = 0;
```
alfredh: it is probably better to use a `bool` type here since the value
is either `true` or `false`.

alfredh:
hi @alishir 

thanks for your patch. I made a couple of comments to it :)


Alfred

alishir:
hi @alfredh 

Thanks for you comments, I updated the PR according to your comments.

PR #470, opened 2018-06-25
Sending and receiving custom headers

Encamy:
The PR https://github.com/alfredh/baresip/pull/425 was reworked

It was desirable to use the "X-" prefix with headers which are not defined by IETF/IANA, but not necessarily. It just removes possible collisions. But in this [RFC](https://tools.ietf.org/html/rfc6648) the "X-" prefix was deprecated. So I think that users should decide to use or not to use this prefix by themselves.

Now this feature uses list structure.

Interaction with custom headers changed:
Example of creating and sending custom headers:
```c++
int some_id = 7;
struct list *custom_hdrs;

err = custom_hdrs_alloc(&custom_hdrs);
err = custom_hdrs_add_int(custom_hdrs, "X-CALL_ID", some_id);
err = custom_hdrs_add(custom_hdrs, "X-HEADER_NAME", "VALUE");
err = ua_connect(user_agent, 0, NULL, uri, NULL, VIDMODE_OFF, custom_hdrs);

mem_deref(custom_hdrs);
```
Then the INVITE request will look like this:
image

Example of getting incoming headers:
```c++
//after creating an user agent:
ua_add_xhdr_filter(user_agent, "X-CALL_ID");

//when new call is incoming:
const struct list *hdrs;
hdrs = call_get_custom_hdrs(call); 

//Now we can interact with received headers
static int custom_header_extractor(const struct pl *name, const struct pl *val, void *arg)
{
    //do smth with "name" and "val" variables
    return 0;
}
custom_hdrs_apply(hdrs, custom_header_extractor, NULL);
```

Also a new test in baresip selftest was added

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -160,7 +160,7 @@ typedef int (custom_hdrs_h)(const struct pl *name, const struct pl *val,
 
 int custom_hdrs_add(struct list *hdrs, const char *name,
 	const char *fmt, ...);
-int custom_hdrs_apply(const struct list hdrs,
+int custom_hdrs_apply(const struct list *hdrs,
 	custom_hdrs_h *h, void *arg);
 
 
@@ -688,7 +688,7 @@ void ua_set_catchall(struct ua *ua, bool enabled);
 void ua_event(struct ua *ua, enum ua_event ev, struct call *call,
 	      const char *fmt, ...);
 int ua_add_xhdr_filter(struct ua *ua, const char *hdr_name);
-void ua_set_custom_hdrs(struct ua *ua, struct list custom_hdrs);
+void ua_set_custom_hdrs(struct ua *ua, struct list *custom_hdrs);
 
 
 /* One instance */
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -158,10 +158,9 @@ const struct list *call_get_custom_hdrs(const struct call *call);
 typedef int (custom_hdrs_h)(const struct pl *name, const struct pl *val,
 	void *arg);     /* returns error code if any */
 
-int custom_hdrs_alloc(struct list **hdrs);
 int custom_hdrs_add(struct list *hdrs, const char *name,
 	const char *fmt, ...);
-int custom_hdrs_apply(const struct list *hdrs,
+int custom_hdrs_apply(const struct list hdrs,
 	custom_hdrs_h *h, void *arg);
 
 
@@ -689,7 +688,7 @@ void ua_set_catchall(struct ua *ua, bool enabled);
 void ua_event(struct ua *ua, enum ua_event ev, struct call *call,
 	      const char *fmt, ...);
 int ua_add_xhdr_filter(struct ua *ua, const char *hdr_name);
-void ua_set_custom_hdrs(struct ua *ua, struct list *custom_hdrs);
+void ua_set_custom_hdrs(struct ua *ua, struct list custom_hdrs);
 
 
 /* One instance */
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -156,13 +156,11 @@ void call_set_current(struct list *calls, struct call *call);
 */
 const struct list *call_get_custom_hdrs(const struct call *call);
 typedef int (custom_hdrs_h)(const struct pl *name, const struct pl *val,
-	void *arg);     //returns error code if any
+	void *arg);     /* returns error code if any */
+
 int custom_hdrs_alloc(struct list **hdrs);
 int custom_hdrs_add(struct list *hdrs, const char *name,
-	const char *val);
-int custom_hdrs_add_pl(struct list *hdrs, const struct pl *name,
-	const struct pl *val);
-int custom_hdrs_add_int(struct list *hdrs, const char *name, int val);
+	const char *fmt, ...);
 int custom_hdrs_apply(const struct list *hdrs,
 	custom_hdrs_h *h, void *arg);
 
@@ -659,8 +657,7 @@ typedef void (ua_exit_h)(void *arg);
 int  ua_alloc(struct ua **uap, const char *aor);
 int  ua_connect(struct ua *ua, struct call **callp,
 		const char *from_uri, const char *uri,
-		const char *params, enum vidmode vmode,
-		struct list *custom_hdrs);
+		const char *params, enum vidmode vmode);
 void ua_hangup(struct ua *ua, struct call *call,
 	       uint16_t scode, const char *reason);
 int  ua_answer(struct ua *ua, struct call *call);
@@ -692,6 +689,7 @@ void ua_set_catchall(struct ua *ua, bool enabled);
 void ua_event(struct ua *ua, enum ua_event ev, struct call *call,
 	      const char *fmt, ...);
 int ua_add_xhdr_filter(struct ua *ua, const char *hdr_name);
+void ua_set_custom_hdrs(struct list *custom_hdrs);
 
 
 /* One instance */
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -149,12 +149,12 @@ void          call_enable_rtp_timeout(struct call *call, uint32_t timeout_ms);
 uint32_t      call_linenum(const struct call *call);
 struct call  *call_find_linenum(const struct list *calls, uint32_t linenum);
 void call_set_current(struct list *calls, struct call *call);
+const struct list *call_get_custom_hdrs(const struct call *call);
 
 
 /*
 * Custom headers
 */
-const struct list *call_get_custom_hdrs(const struct call *call);
 typedef int (custom_hdrs_h)(const struct pl *name, const struct pl *val,
 	void *arg);     /* returns error code if any */
 
@@ -689,7 +689,7 @@ void ua_set_catchall(struct ua *ua, bool enabled);
 void ua_event(struct ua *ua, enum ua_event ev, struct call *call,
 	      const char *fmt, ...);
 int ua_add_xhdr_filter(struct ua *ua, const char *hdr_name);
-void ua_set_custom_hdrs(struct list *custom_hdrs);
+void ua_set_custom_hdrs(struct ua *ua, struct list *custom_hdrs);
 
 
 /* One instance */
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -152,6 +152,22 @@ void call_set_current(struct list *calls, struct call *call);
 
 
 /*
+* Custom headers
+*/
+const struct list *call_get_custom_hdrs(const struct call *call);
+typedef int (custom_hdrs_h)(const struct pl *name, const struct pl *val,
+	void *arg);     //returns error code if any
+int custom_hdrs_alloc(struct list **hdrs);
+int custom_hdrs_add(struct list *hdrs, const char *name,
+	const char *val);
+int custom_hdrs_add_pl(struct list *hdrs, const struct pl *name,
+	const struct pl *val);
+int custom_hdrs_add_int(struct list *hdrs, const char *name, int val);
+int custom_hdrs_apply(const struct list *hdrs,
+	custom_hdrs_h *h, void *arg);
+
+
+/*
  * Conf (utils)
  */
 
@@ -643,7 +659,8 @@ typedef void (ua_exit_h)(void *arg);
 int  ua_alloc(struct ua **uap, const char *aor);
 int  ua_connect(struct ua *ua, struct call **callp,
 		const char *from_uri, const char *uri,
-		const char *params, enum vidmode vmode);
+		const char *params, enum vidmode vmode,
+		struct list *custom_hdrs);
 void ua_hangup(struct ua *ua, struct call *call,
 	       uint16_t scode, const char *reason);
 int  ua_answer(struct ua *ua, struct call *call);
@@ -674,6 +691,7 @@ void ua_set_media_af(struct ua *ua, int af_media);
 void ua_set_catchall(struct ua *ua, bool enabled);
 void ua_event(struct ua *ua, enum ua_event ev, struct call *call,
 	      const char *fmt, ...);
+int ua_add_xhdr_filter(struct ua *ua, const char *hdr_name);
 
 
 /* One instance */
--- a/src/call.c
+++ b/src/call.c
@@ -82,7 +82,7 @@ struct call {
 
 	uint32_t rtp_timeout_ms;  /**< RTP Timeout in [ms]                  */
 	uint32_t linenum;         /**< Line number from 1 to N              */
-	struct list *custom_hdrs; /**< List of custom headers if any */
+	struct list custom_hdrs;  /**< List of custom headers if any */
 };
 
 
@@ -398,7 +398,8 @@ static void call_destructor(void *arg)
 	mem_deref(call->sub);
 	mem_deref(call->not);
 	mem_deref(call->acc);
-	mem_deref(call->custom_hdrs);
+
+	list_flush(&call->custom_hdrs);
 }
 
 
@@ -690,14 +691,21 @@ int call_alloc(struct call **callp, const struct config *cfg, struct list *lst,
 }
 
 
-void call_set_custom_hdrs(struct call *call, struct list *hdrs)
+void call_set_custom_hdrs(struct call *call, struct list hdrs)
 {
 	if (!call)
 		return;
 
-	mem_deref(call->custom_hdrs);
-	call->custom_hdrs = hdrs;
-	mem_ref(hdrs);
+	list_flush(&call->custom_hdrs);
+
+	struct le *le;
+	LIST_FOREACH(&hdrs, le) {
+		struct sip_hdr *hdr = le->data;
+		char *buf = NULL;
+		re_sdprintf(&buf, "%r", &hdr->name);
+		custom_hdrs_add(&call->custom_hdrs, buf, "%r", &hdr->val);
+		mem_deref(buf);
+	}
 }
 
 
@@ -706,7 +714,7 @@ const struct list *call_get_custom_hdrs(const struct call *call)
 	if (!call)
 		return NULL;
 
-	return call->custom_hdrs;
+	return &call->custom_hdrs;
 }
 
 
--- a/src/call.c
+++ b/src/call.c
@@ -691,7 +691,7 @@ int call_alloc(struct call **callp, const struct config *cfg, struct list *lst,
 }
 
 
-void call_set_custom_hdrs(struct call *call, struct list hdrs)
+void call_set_custom_hdrs(struct call *call, const struct list *hdrs)
 {
 	if (!call)
 		return;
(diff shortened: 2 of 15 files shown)


alfredh:
many thanks for this work.

I will review it after baresip 0.5.10 is released.

include/baresip.h
```diff
@@ -151,6 +151,22 @@ struct call  *call_find_linenum(const struct list *calls, uint32_t linenum);
 void call_set_current(struct list *calls, struct call *call);
 
 
+/*
+* Custom headers
+*/
+const struct list *call_get_custom_hdrs(const struct call *call);
+typedef int (custom_hdrs_h)(const struct pl *name, const struct pl *val,
+	void *arg);     //returns error code if any
+int custom_hdrs_alloc(struct list **hdrs);
```
alfredh: please insert some empty lines here, between the typedef and
the function prototypes.

also please use C89 style comments (`/* xxx */`)

include/baresip.h
```diff
@@ -643,7 +659,8 @@ typedef void (ua_exit_h)(void *arg);
 int  ua_alloc(struct ua **uap, const char *aor);
 int  ua_connect(struct ua *ua, struct call **callp,
 		const char *from_uri, const char *uri,
-		const char *params, enum vidmode vmode);
+		const char *params, enum vidmode vmode,
+		struct list *custom_hdrs);
```
alfredh: instead of breaking the `ua_connect` api, is it better to add a new function to set
the custom headers ? e.g. `ua_set_custom_hdrs()`

src/call.c
```diff
@@ -82,6 +82,7 @@ struct call {
 
 	uint32_t rtp_timeout_ms;  /**< RTP Timeout in [ms]                  */
 	uint32_t linenum;         /**< Line number from 1 to N              */
+	struct list *custom_hdrs; /**< List of custom headers if any */
```
alfredh: is this for the outgoing call or the incoming call ?

src/call.c
```diff
@@ -688,6 +690,20 @@ int call_alloc(struct call **callp, const struct config *cfg, struct list *lst,
 }
 
 
+void call_set_custom_hdrs(struct call *call, struct list *hdrs)
+{
```
alfredh: please check `call` parameter, return if NULL.

src/call.c
```diff
@@ -688,6 +690,20 @@ int call_alloc(struct call **callp, const struct config *cfg, struct list *lst,
...
+	mem_deref(call->custom_hdrs);
+	call->custom_hdrs = hdrs;
+	mem_ref(hdrs);
+}
+
+
+const struct list *call_get_custom_hdrs(const struct call *call)
+{
```
alfredh: please check `call` parameter, return if NULL.

src/custom_hdrs.c
```diff
@@ -0,0 +1,114 @@
+#include <re.h>
```
alfredh: please add a copyright header here, similar to all the other .c files

src/custom_hdrs.c
```diff
@@ -0,0 +1,114 @@
...
+
+
+static void hdr_destructor(void *arg)
+{
+	struct sip_hdr *hdr = arg;
+
+	/*
+	list_unlink(&hdr->le);
+	hash_unlink(&hdr->he);
+	*/
```
alfredh: commented out code should be removed.

src/custom_hdrs.c
```diff
@@ -0,0 +1,114 @@
...
+					const char *val)
+{
+	struct pl name_pl;
+	struct pl val_pl;
+
+	pl_set_str(&name_pl, name);
+	pl_set_str(&val_pl, val);
+
+	return custom_hdrs_add_pl(hdrs, &name_pl, &val_pl);
+}
```
alfredh: I am wondering if we should add formatted arguments here, similar to:

```
int sip_treplyf(struct sip_strans **stp, struct mbuf **mbp, struct sip *sip,
		const struct sip_msg *msg, bool rec_route, uint16_t scode,
		const char *reason, const char *fmt, ...)
{
	va_list ap;
	int err;

	va_start(ap, fmt);
	err = vreplyf(stp, mbp, true, sip, msg, rec_route, scode, reason,
		      fmt, ap);
	va_end(ap);

	return err;
}
```

in this case we only need one add function, which becomes very flexible and powerful.

instead of custom_hdrs_add_int you can then write:

`custom_hdrs_add(hdrs, "X-Header", "%d", 42);`

what do you think ?

src/custom_hdrs.c
```diff
@@ -0,0 +1,114 @@
...
+
+int custom_hdrs_apply(const struct list *hdrs,
+					  custom_hdrs_h *h,
+					  void *arg)
+{
+	int err;
+	struct le *le;
+	for (le = list_head(hdrs); le; le = le->next) {
+		struct sip_hdr * hdr = (struct sip_hdr *)(le->data);
```
alfredh: the casting is not needed, can be removed.

src/srcs.mk
```diff
@@ -38,6 +38,7 @@ SRCS	+= timer.c
 SRCS	+= timestamp.c
 SRCS	+= ua.c
 SRCS	+= ui.c
+SRCS	+= custom_hdrs.c
```
alfredh: files in alphabetical order is nice :)

src/ua.c
```diff
@@ -31,6 +31,7 @@ struct ua {
 	int af_media;                /**< Preferred Address Family for media */
 	enum presence_status my_status; /**< Presence Status                 */
 	bool catchall;               /**< Catch all inbound requests         */
+	struct list *hdr_filter;     /**< List of custom headers             */
```
alfredh: is this for outgoing or incoming ?

src/ua.c
```diff
@@ -1311,6 +1323,35 @@ static void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
 		goto error;
 	}
 
+	if (ua->hdr_filter)
+	{
```
alfredh: please move curly braces to previous line, at the end:

```
if (x) {
}
```

src/ua.c
```diff
@@ -1311,6 +1323,35 @@ static void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
...
+		while (le) {
+			const struct sip_hdr *hdr_local;
+			const struct ua_xhdr_filter *filter = le->data;
+
+			le = le->next;
+		    hdr_local = sip_msg_xhdr_apply(msg, true, filter->hdr_name,
+				NULL, NULL);
+			if (hdr_local) {
+				custom_hdrs_add_pl(hdrs, &hdr_local->name,
+					&hdr_local->val);
```
alfredh: return value must be checked here

src/ua.c
```diff
@@ -1323,6 +1364,50 @@ static void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
 }
 
 
+static void ua_xhdr_filter_destructor(void *arg)
+ {
+struct ua_xhdr_filter *filter = arg;
+mem_deref(filter->hdr_name);
+}
```
alfredh: this function needs indentation (1 TAB per line)

src/ua.c
```diff
@@ -1323,6 +1364,50 @@ static void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
...
+		list_init(ua->hdr_filter);
+
+	}
+
+	char *buf = mem_alloc(sizeof(char) * (strlen(hdr_name) + 1), NULL);
+	if (!buf) {
+		return ENOMEM;
+
+	}
+	strcpy(buf, hdr_name);
```
alfredh: instead of these 6 lines you can use the `str_dup` function

alfredh:
some comments inline :)

another question is, should the custom headers also be added to the 200 OK
for incoming calls ?

Encamy:
Thanks for comments. I will improve code as soon as possible.
It is possible to add custom headers to response 200 OK, but I don't think that it is necessary. We just provide additional information to the client and do not expect that the client will respond to this information.

alfredh:
you can just add more commits to this branch.

when the review is complete I will squash and merge it master (as 1 commit).

src/call.c
Encamy: This is for outgoing calls. It is using in `ua_connect`.

src/ua.c
Encamy: This is for incoming calls

Encamy:
Hello @alfredh 
Using formatted arguments is awesome idea. It is much more flexible now.
Example of usage:
```c++
custom_hdrs_add(custom_hdrs, "X-HEADER_NAME", "%s", "VALUE");
custom_hdrs_add(custom_hdrs, "X-NAME", "%d", 12)
```
Also I have removed additional parameter from `ua_connect` (api is unchanged now)
Custom headers can be applied by using `ua_set_custom_hdrs(some_headers)` before `ua_connect`.
Is it ok, that custom headers is stored in `ua.c`? Or it's better to move `static struct list *custom_hdrs` to `custom_headers.c`?

src/ua.c
```diff
@@ -81,6 +87,7 @@ static struct {
 #endif
 };
 
+static struct list *custom_hdrs;
```
alfredh: it is better to move this to `struct ua` so that it becomes re-entrant.

the long term goal of baresip is to make it 100% re-entrant with no static/global data.

test/call.c
```diff
@@ -969,3 +980,75 @@ int test_call_mediaenc(void)
 
 	return err;
 }
+
+
+int test_call_custom_headers(void) {
```
alfredh: the curly brace should be moved to the next line.

functions are written like this:

```
void function(void)
{
}
```

test/call.c
```diff
@@ -969,3 +980,75 @@ int test_call_mediaenc(void)
...
+
+	ua_add_xhdr_filter(f->b.ua, "X-CALL_ID");
+	ua_add_xhdr_filter(f->b.ua, "X-HEADER_NAME");
+
+	f->behaviour = BEHAVIOUR_GET_HDRS;
+
+	/* Make a call from A to B
+	 * with some custom headers in INVITE message */
+	int some_id = 7;
+	struct list *custom_hdrs;
```
alfredh: declaration of variables should be moved to the beginning of the function.

test/call.c
```diff
@@ -969,3 +980,75 @@ int test_call_mediaenc(void)
...
+	ua_set_custom_hdrs(custom_hdrs);
+	err = ua_connect(f->a.ua, 0, NULL, f->buri, NULL, VIDMODE_OFF);
+
+	mem_deref(custom_hdrs);
+	TEST_ERR(err);
+
+	/* run main-loop with timeout, wait for events */
+	err = re_main_timeout(5000);
+
+	bool headers_matched = true;
```
alfredh: declaration of variables should be moved to the beginning of the function.

src/custom_hdrs.c
```diff
@@ -0,0 +1,95 @@
...
+int custom_hdrs_alloc(struct list **hdrs)
+{
+	struct list *h;
+	h = mem_zalloc(sizeof(*h), custom_hdrs_destructor);
+	if (!h)
+		return ENOMEM;
+	list_init(h);
+	*hdrs = h;
+	return 0;
+}
```
alfredh: I am wondering if it is better to use a compound list instead of allocated list.

Most of the code in baresip and libre using a list is declared like this:

```
struct foo {
    struct list my_list;
};
```

this way there is no need for an allocation. The list is empty by default.

To initialize the list:

`list_init();`

To destroy the list members:

`list_flush();`

To check if the list is empty or have any members:

`list_isempty()` or `list.head != NULL`

if you write the code like this you will find that it becomes simpler.
but it is not possible to use `mem_ref` on it, so you have to copy each element.

src/custom_hdrs.c
```diff
@@ -0,0 +1,95 @@
...
+	struct pl temp_pl = { NULL, 0 };
+	char *value = NULL;
+	int err = 0;
+
+	va_list ap;
+	va_start(ap, fmt);
+	err = re_vsdprintf(&value, fmt, ap);
+	va_end(ap);
+
+	struct sip_hdr *hdr;
```
alfredh: please move variable declaration to the beginning of the function

src/custom_hdrs.c
```diff
@@ -0,0 +1,95 @@
...
+
+	list_append(hdrs, &hdr->le, hdr);
+
+	return 0;
+error:
+	mem_deref(hdr);
+	return(err);
+}
+
+int custom_hdrs_apply(const struct list *hdrs,
```
alfredh: the distance between functions should be 2 empty lines.

src/custom_hdrs.c
```diff
@@ -0,0 +1,95 @@
...
+	struct le *le;
+	for (le = list_head(hdrs); le; le = le->next) {
+		struct sip_hdr * hdr = le->data;
+		err = h(&hdr->name, &hdr->val, arg);
+		if (err) {
+			return err;
+		}
+	}
+	return 0;
+}
```
alfredh: can you add some whitespace here, at least empty line before and after
the for loop. the code becomes a bit more readable.

include/baresip.h
```diff
@@ -151,6 +151,20 @@ struct call  *call_find_linenum(const struct list *calls, uint32_t linenum);
 void call_set_current(struct list *calls, struct call *call);
 
 
+/*
+* Custom headers
+*/
+const struct list *call_get_custom_hdrs(const struct call *call);
```
alfredh: perhaps move this prototype to the `struct call` block since it is accessing `call` ?

alfredh:
thanks for the update. it looks better now.

please have a second look at the formatting. all functions should have 2 empty
lines between them. The indentation is TAB. also insert some extra whitespace
to make the code more readable :)

Alfred

Encamy:
I have fixed code style and moved list of custom headers to ua struct.
It is possible to change lists from allocated list to compound list. But I have some troubles with this task. I will change them if it is necessary. Should I do that?

src/custom_hdrs.c
```diff
@@ -0,0 +1,105 @@
...
+	va_start(ap, fmt);
+	err = re_vsdprintf(&value, fmt, ap);
+	va_end(ap);
+
+	hdr = mem_zalloc(sizeof(*hdr), hdr_destructor);
+	if (!hdr || !value)
+		goto error;
+
+	pl_set_str(&temp_pl, name);
+	pl_dup(&hdr->name, &temp_pl);
```
alfredh: return value of `pl_dup` must be checked.

src/custom_hdrs.c
```diff
@@ -0,0 +1,105 @@
...
+{
+	int err;
+	struct le *le;
+
+	for (le = list_head(hdrs); le; le = le->next) {
+		struct sip_hdr * hdr = le->data;
+		err = h(&hdr->name, &hdr->val, arg);
+		if (err) {
+			return err;
+		}
```
alfredh: no need for curly braces here, since it has only 1 line.

src/ua.c
```diff
@@ -549,6 +556,8 @@ static void ua_destructor(void *arg)
 	if (list_isempty(&uag.ual)) {
 		sip_close(uag.sip, false);
 	}
+
+	mem_deref(ua->hdr_filter);
```
alfredh: also `custom_hdrs` must be derefed in the destructor

src/ua.c
```diff
@@ -833,13 +842,18 @@ int ua_connect(struct ua *ua, struct call **callp,
...
+
 	err = call_connect(call, &pl);
 
 	if (err)
 		mem_deref(call);
 	else if (callp)
 		*callp = call;
 
+	ua_set_custom_hdrs(ua, NULL);
+
```
alfredh: I think we should clear the list of custom headers here, this should be done by the application.

src/ua.c
```diff
@@ -1311,6 +1325,35 @@ static void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
...
+			goto error;
+
+		le = list_head(ua->hdr_filter);
+		while (le) {
+			const struct sip_hdr *hdr_local;
+			const struct ua_xhdr_filter *filter = le->data;
+
+			le = le->next;
+		    hdr_local = sip_msg_xhdr_apply(msg, true, filter->hdr_name,
+				NULL, NULL);
```
alfredh: you can also use `sip_msg_xhdr` here, it is perhaps a bit simpler.

src/ua.c
```diff
@@ -1323,6 +1366,47 @@ static void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
...
+
+		if (!ua->hdr_filter) {
+			return ENOMEM;
+		}
+		list_init(ua->hdr_filter);
+	}
+
+	char *buf;
+	if (str_dup(&buf, hdr_name))
+		return ENOMEM;
```
alfredh: instead of using a temp buf, you can instead dup the string
directly into `filter->hdr_name`

in case of errors, the destructor will deref the string.

src/ua.c
```diff
@@ -1965,3 +2049,9 @@ int uag_set_extra_params(const char *eprm)
 
 	return 0;
 }
+
+
+void ua_set_custom_hdrs(struct ua *ua, struct list *custom_headers)
+{
+	ua->custom_hdrs = custom_headers;
+}
```
alfredh: you should use mem_deref and mem_ref here.

the data in struct ua should not point to objects from the application.
the application object might go out of scope, which will make the
`ua->custom_hdrs` pointer point to invalid data (and perhaps a crash).

src/ua.c
```diff
@@ -31,6 +31,8 @@ struct ua {
 	int af_media;                /**< Preferred Address Family for media */
 	enum presence_status my_status; /**< Presence Status                 */
 	bool catchall;               /**< Catch all inbound requests         */
+	struct list *hdr_filter;     /**< Filter for incoming headers        */
```
alfredh: I think the filter can be converted to a compound list, that needs no allocation.

I have made an example of it here:

https://gist.github.com/alfredh/2577a5dadbde35a9ba47d96696ddb31d

src/ua.c
Encamy: Sorry, I can't understand how exactly to use `mem_deref` and `mem_ref`.
Should this look something like the following?
```c++
mem_ref(custom_headers);
ua->custom_hdrs = custom_headers;
mem_deref(custom_headers);
```

Encamy:
Thanks for a review. I have made some changes to the code.
Also I have tried to replace every allocated lists with compound lists. Code became more complicated. The most suitable method for copying all data is to add pairs `name:value` from one list to another. We can't just copy the memory of one header and paste to other because `mem_deref(hdr)` will break the application.
For example:
```c++
struct le *le;
LIST_FOREACH(&hdrs, le) {
	struct sip_hdr *hdr = le->data;
	custom_hdrs_add(&call->custom_hdrs, hdr->name.p, "%r", &hdr->val);
}
```
I don't think that it is good decision. May be I'm missing something? Should I replace other allocated lists with compound or we can leave it as it is?

Encamy:
I have found out the problem. This version is using only compound lists.

alfredh:
this looks very good. I have a small comment, see inline.

src/ua.c
```diff
@@ -1965,3 +2035,18 @@ int uag_set_extra_params(const char *eprm)
 
 	return 0;
 }
+
+
+void ua_set_custom_hdrs(struct ua *ua, struct list custom_headers)
```
alfredh: the custom_headers parameter should be of type pointer to a const struct list:

`const struct list *custom_headers`

it is not recommended to pass structs like this as function parameters.

src/call.c
```diff
@@ -688,6 +691,33 @@ int call_alloc(struct call **callp, const struct config *cfg, struct list *lst,
...
+	if (!call)
+		return;
+
+	list_flush(&call->custom_hdrs);
+
+	struct le *le;
+	LIST_FOREACH(&hdrs, le) {
+		struct sip_hdr *hdr = le->data;
+		char *buf = NULL;
+		re_sdprintf(&buf, "%r", &hdr->name);
```
alfredh: check return value, return if errors.

src/call.c
```diff
@@ -688,6 +691,33 @@ int call_alloc(struct call **callp, const struct config *cfg, struct list *lst,
...
+		custom_hdrs_add(&call->custom_hdrs, buf, "%r", &hdr->val);
```
alfredh: check return value, return if errors.

Encamy:
Thank you. I have fixed these shortcomings.

Encamy:
Thank you for all your reviews and for help

PR #472, opened 2018-06-25
Sdl2 threads

nicotyze:
This is a new proposal to address https://github.com/alfredh/baresip/pull/451 code reviews and last comment analyses related to performances issues.

All SDL2 functions are called from the main thread:
-  video frames are pushed through the new push_frame registered function
- a pushed frame is stored in an intermediary vidframe structure before being displayed
- once the video frame is copied, a timer is set to render the frame thanks to new display_handler function.

Tested with vidloops and calls on Win10 and Ubuntu 16.04


nicotyze:
A more generic approach should be implemented directly inside vidloop module => PR closed

PR #473, opened 2018-06-25
Vidloop threads

nicotyze:
This is a proposal to address  https://github.com/alfredh/baresip/pull/451 last comment analyses.
As suggested, video frames are rendered/displayed from main thread and the logic is moved to vidloop module.

Tested with SDL2 display, on win10 and ubuntu 16.04

### DIFF
--- a/modules/sdl2/sdl.c
+++ b/modules/sdl2/sdl.c
@@ -86,7 +86,9 @@ static void event_handler(void *arg)
 	struct vidisp_st *st = arg;
 	SDL_Event event;
 
+#ifndef WIN32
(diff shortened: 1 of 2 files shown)


nicotyze:
Not thread safe => closed.
Need more work on mutex positioning if we want to avoid frame copy before putting the rendering timer...

alfredh:
perhaps we can add a small vidqueue that holds vidframes. the vidqueue
can be thread-safe and have put and get operations.

the display timer can run with a 10ms interval and simply get video frames
from the queue and display them.

PR #476, opened 2018-06-26
dshow: remove qedit.h

Encamy:
Hello
According to this issue: https://github.com/alfredh/baresip/issues/464 I moved declarations of `ISampleGrabberCB` and `ISampleGrabber` into dshow.cpp so it does not require qedit.h file.
Here is [MSVS build log](https://pastebin.com/tCGb5BKn).
@nicotyze would you like to compile with mingw please? I have some troubles with compilation both in current state and this pr. Thank you.

Alexey

### DIFF
--- a/modules/dshow/dshow.cpp
+++ b/modules/dshow/dshow.cpp
@@ -27,17 +27,72 @@
  */
 
 
-/* a piece from Google WebM's qedit.h:
- *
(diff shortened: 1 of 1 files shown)


nicotyze:
Hello,
Successfully compiled (mingw) and tested (Win10) from my side as well.

Nicolas

alfredh:
thank you all !

also compile-time tested with mingw-w64 v5.0.4


Alfred

PR #481, opened 2018-06-28
Vidloop fix: video frame rendering is moved to main thread

nicotyze:
This is a proposal to address #451 last comment analyses.
As suggested, video frames are rendered/displayed from main thread and the logic is moved to vidloop module.
A mutex is used to protect vl->frame and vl->disp accesses
A frame copy is needed in order to provide the display module with final pixel values.

Tested with SDL2 display, on win10 and ubuntu 16.04

### DIFF
--- a/modules/vidloop/vidloop.c
+++ b/modules/vidloop/vidloop.c
@@ -5,6 +5,7 @@
  */
 #define _DEFAULT_SOURCE 1
 #define _BSD_SOURCE 1
+#include <pthread.h>
 #include <string.h>
 #include <time.h>
 #include <re.h>
@@ -64,9 +65,12 @@ struct video_loop {
 	struct list filtdecl;
 	struct vstat stat;
 	struct tmr tmr_bw;
+	struct tmr tmr_display;
 	struct vidsz src_size;
 	struct vidsz disp_size;
 	enum vidfmt src_fmt;
+	struct vidframe *frame;
+	pthread_mutex_t frame_mutex;
 	uint64_t ts_start;      /* usec */
 	uint64_t ts_last;       /* usec */
 	uint16_t seq;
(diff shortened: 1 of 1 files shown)


alfredh:
thanks,

I will review it after baresip 0.5.10 is released.

alfredh:
hi @nicotyze 

this looks very good. here are some comments:

* the current code will display the same frame many times. it is better to add a flag
  indicating if the frame contains valid content or not. The flag is set by the rendering code
  and cleared by the timer polling loop. we can also use this flag to detect overruns.

* the tmr_start should be called once from the alloc function and then from the display
  handler itself, so that the display handler is running at a 100Hz interval.
  we should not use tmr_start to signal a new frame, as it introduces a lot of jitter.
  (try with 120 fps and you will see for yourself)

* the code should handle changes in frame size.

modules/vidloop/vidloop.c
```diff
@@ -123,6 +127,26 @@ static double timestamp_state_duration(const struct timestamp_state *ts,
 }
 
 
+static void display_handler(void *arg)
+{
+	struct video_loop *vl = arg;
+	int err;
```
alfredh: you can add `tmr_start` here so that the loop runs all the time.

modules/vidloop/vidloop.c
```diff
@@ -123,6 +127,26 @@ static double timestamp_state_duration(const struct timestamp_state *ts,
...
+
+	pthread_mutex_lock(&vl->frame_mutex);
+
```
alfredh: check for a valid frame here. if no frame, just return.

modules/vidloop/vidloop.c
```diff
@@ -164,16 +188,18 @@ static int display(struct video_loop *vl, struct vidframe *frame)
...
 
-	/* display frame */
-	err = vidisp_display(vl->vidisp, "Video Loop", frame);
-	if (err == ENODEV) {
-		info("vidloop: video-display was closed\n");
-		vl->vidisp = mem_deref(vl->vidisp);
-		vl->err = err;
-		goto out;
+	pthread_mutex_lock(&vl->frame_mutex);
+
```
alfredh: check for changes in frame size here. if the size changed, you can just `mem_deref` the
current frame.

modules/vidloop/vidloop.c
```diff
@@ -164,16 +188,18 @@ static int display(struct video_loop *vl, struct vidframe *frame)
...
+	if (!vl->frame) {
+		vidframe_alloc(&vl->frame, frame->fmt, &frame->size);
 	}
+	vidframe_copy(vl->frame, frame);
+
+	pthread_mutex_unlock(&vl->frame_mutex);
+	/* NOTE: usually (e.g. SDL2),
+	 		 video frame must be rendered from main thread */
+	tmr_start(&vl->tmr_display, 10, display_handler, vl);
```
alfredh: we should not call `tmr_start` from a worker thread.

alfredh:
hi @nicotyze 

many thanks for your work. I merged your patch to master.

this work will most likely also improve things for other display modules
such as sdl and opengles on iOS/OSX.

Alfred

PR #488, opened 2018-07-23
audio: remove audio_srate and audio_channels

alfredh:
remove these config options:

  audio_srate            8000-48000
  audio_channels         1-2

these options has no practical usage, except to limit the
enabled audio codecs to the range of samplerate/channels
as specified.

if a module has multiple audio codecs, and you only want
to use one of them, this can now be done with the
"audio_codecs" parameter to the "accounts" file.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -16,8 +16,6 @@ sip_trans_bsize		128
 audio_player		alsa,default
 audio_source		alsa,default
 audio_alert		alsa,default
-audio_srate		8000-48000
-audio_channels		1-2
 #ausrc_srate		48000
 #auplay_srate		48000
 #ausrc_channels		0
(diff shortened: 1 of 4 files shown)


alfredh:
extra note: these config options were added a very long time ago, and is not really needed.
so I propose to remove them.

if someone is still using them, or disagrees with this PR, please state your opinion here :)

Alfred

juha-h:
Alfred E. Heggestad writes:

> remove these config options:
> 
>   audio_srate            8000-48000
>   audio_channels         1-2

How about ausrc_srate, auplay_srate, ausrc_channels, auplay_channels?
Are they too deprecated?

-- Juha

alfredh:
no, these config options are not deprecated. they are used to force the soundcard
to a specific sample rate/channels.

by default, the soundcard will be started with the same samplerate/channels
as the audio codec used, to get the best possible audio quality without latency.

if one of these options are specified, and it differs from the selected audio codec,
the audio resampler will be used.

juha-h:
Alfred E. Heggestad writes:

> by default, the soundcard will be started with the same samplerate/channels
> as the audio codec used, to get the best possible audio quality
> without latency. 
> 
> if one of these options are specified, and it differs from the
> selected audio codec, the audio resampler will be used.

Thanks for the explanation, Juha

PR #490, opened 2018-07-25
Vidloop: fix Windows compilation

Encamy:
Recently commit https://github.com/alfredh/baresip/commit/1e9152d8d9c004d2d55c543018135fcbfee04918 was added. As you know, there is no pthreads on Windows :) So this PR make compilation available.
Implementation was received from [this gist](https://gist.github.com/informationsea/f63c2b3854b20bd540ab)

### DIFF
--- a/modules/vidloop/vidloop.c
+++ b/modules/vidloop/vidloop.c
@@ -5,7 +5,9 @@
  */
 #define _DEFAULT_SOURCE 1
 #define _BSD_SOURCE 1
+#ifdef HAVE_PTHREAD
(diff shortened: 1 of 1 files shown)


alfredh:
many thanks for the PR.

in commit  117e4e1bec0c53f21ff6e5cab5d9c8b327aa74b2 I have changed the vidloop
from using pthread mutex to a portable "struct lock" from libre.

I believe this is a better solution and will make it compile on all platforms.


Alfred

Encamy:
Thank you! This solution is much better

PR #491, opened 2018-07-27
Vidsrc devices list

nicotyze:
This is a proposal for a media devices data structure to be listed from media sources (audio, video).
Here is a first usage/implementation with v4l2 module.
Device availability is checked from switch_video_source (+ existing devices printing).
This might be very useful for video and audio source modules where device naming is not known/standard (e.g dshow).

Tested with Ubuntu 16.04

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -390,6 +390,7 @@ struct mediadev {
 
 int mediadev_add(struct list *dev_list, const char *name);
 struct mediadev *mediadev_find(const struct list *dev_list, const char *name);
+struct mediadev *mediadev_get_default(const struct list *dev_list);
 int mediadev_print(struct re_printf *pf, const struct list *dev_list);
 
 
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -945,6 +945,7 @@ struct aucodec {
 	uint32_t srate;             /* Audio samplerate */
 	uint32_t crate;             /* RTP Clock rate   */
 	uint8_t ch;
+	uint8_t pch;                /* RTP packet channels */
 	const char *fmtp;
 	auenc_update_h *encupdh;
 	auenc_encode_h *ench;
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -377,6 +377,23 @@ struct media_ctx {
 
 
 /*
+ * Media Device
+ */
+
+struct mediadev;
+
+/** Defines a media device */
+struct mediadev {
+	struct le   le;
+	char  *name;
+};
+
+int mediadev_add(struct list *dev_list, const char *name);
+struct mediadev *mediadev_find(const struct list *dev_list, const char *name);
+int mediadev_print(struct re_printf *pf, const struct list *dev_list);
+
+
+/*
  * Message
  */
 
@@ -807,7 +824,6 @@ struct cmds *cmds_find(const struct commands *commands,
 
 struct vidsrc;
 struct vidsrc_st;
-struct mediadev;
 
 /** Video Source parameters */
 struct vidsrc_prm {
@@ -837,11 +853,6 @@ typedef void (vidsrc_update_h)(struct vidsrc_st *st, struct vidsrc_prm *prm,
 			       const char *dev);
 
 /** Defines a video source */
-struct mediadev {
-	struct le   le;
-	const char  *name;
-};
-
 struct vidsrc {
 	struct le         le;
 	const char       *name;
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -807,6 +807,7 @@ struct cmds *cmds_find(const struct commands *commands,
 
 struct vidsrc;
 struct vidsrc_st;
+struct mediadev;
 
 /** Video Source parameters */
 struct vidsrc_prm {
@@ -836,9 +837,15 @@ typedef void (vidsrc_update_h)(struct vidsrc_st *st, struct vidsrc_prm *prm,
 			       const char *dev);
 
 /** Defines a video source */
+struct mediadev {
+	struct le   le;
+	const char  *name;
+};
+
 struct vidsrc {
 	struct le         le;
 	const char       *name;
+	struct list      dev_list;
 	vidsrc_alloc_h   *alloch;
 	vidsrc_update_h  *updateh;
 };
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -765,16 +765,17 @@ static int switch_video_source(struct re_printf *pf, void *arg)
 			}
 
 			if (!le) {
-				re_hprintf(pf, 
-					   "no such device for %s video-source: %s\n",
-					   driver, device);
+				re_hprintf(pf,
+				  "no such device for %s video-source: %s\n",
+				  driver, device);
 				re_hprintf(pf, "\rAvailable devices are:\n");
 
 				for (le = list_head(&vs->dev_list); le;
 				     le = le->next) {
 
 						dev = le->data;
-						re_hprintf(pf, "%s\n", dev->name);
+						re_hprintf(pf, "%s\n",
+							   dev->name);
 				}
 				return 0;
 			}
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -718,7 +718,9 @@ static int switch_video_source(struct re_printf *pf, void *arg)
 	struct config_video *vidcfg;
 	struct config *cfg;
 	struct video *v;
+	const struct vidsrc *vs;
 	struct le *le;
+	struct mediadev *dev;
 	char driver[16], device[128] = "";
 	int err = 0;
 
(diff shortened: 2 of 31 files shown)


alfredh:
hi @nicotyze 

many thanks for this PR.

when checking the device name, there are 3 cases that needs to be handled:

1. module has one or more devices
2. module ignores the device parameter (such as cairo.so)
3. module does not implement this API (yet).

I suggest moving the mediadev related functions to mediadev.c
I wrote some sample code here:

https://gist.github.com/alfredh/e2faf29e34873958d5f554bb006a0115


more comments inline

include/baresip.h
```diff
@@ -836,9 +837,15 @@ typedef void (vidsrc_update_h)(struct vidsrc_st *st, struct vidsrc_prm *prm,
 			       const char *dev);
 
 /** Defines a video source */
+struct mediadev {
+	struct le   le;
+	const char  *name;
```
alfredh: remove const here

modules/menu/menu.c
```diff
@@ -745,10 +747,39 @@ static int switch_video_source(struct re_printf *pf, void *arg)
...
+
+			for (le = list_head(&vs->dev_list); le;
+			     le = le->next) {
+
+				dev = le->data;
+
+				if (!str_cmp(dev->name, device)) {
+					break;
+				}
+			}
```
alfredh: instead of this block, you can use `mediadev_find` instead.

modules/menu/menu.c
```diff
@@ -745,10 +747,39 @@ static int switch_video_source(struct re_printf *pf, void *arg)
...
+				  driver, device);
+				re_hprintf(pf, "\rAvailable devices are:\n");
+
+				for (le = list_head(&vs->dev_list); le;
+				     le = le->next) {
+
+						dev = le->data;
+						re_hprintf(pf, "%s\n",
+							   dev->name);
+				}
```
alfredh: here you can use `mediadev_print` instead

modules/v4l2/v4l2.c
```diff
@@ -411,6 +411,29 @@ static int read_frame(struct vidsrc_st *st)
...
+	char test_name[16];
+	struct mediadev *dev;
+
+	for (i=0;i < 16;i++) {
+
+		sprintf(test_name, "/dev/video%i", i);
+
+		if ((fd = open(test_name, O_RDONLY)) == -1) {
+			continue;
+		}
```
alfredh: the file descriptor `fd` must be closed, otherwise we will leak resources

modules/v4l2/v4l2.c
```diff
@@ -411,6 +411,29 @@ static int read_frame(struct vidsrc_st *st)
...
+		else {
+			str_dup(&name, test_name);
+			dev = mem_zalloc(sizeof(struct mediadev), NULL);
+			list_append(&vs->dev_list, &dev->le, dev);
+			dev->name = name;
```
alfredh: here you can use `mediadev_add` instead

modules/v4l2/v4l2.c
```diff
@@ -524,14 +547,30 @@ static int alloc(struct vidsrc_st **stp, const struct vidsrc *vs,
...
 static int v4l_close(void)
 {
+	struct le *le;
+	struct mediadev *dev;
+
+	for (le = list_head(&vidsrc->dev_list); le; le = le->next) {
+		dev = le->data;
+		mem_deref((char*)dev->name);
+	}
+	list_flush(&vidsrc->dev_list);
```
alfredh: perhaps we should move the `list_flush` to the vidsrc destructor ? then there is no need
to add any extra code in the close functions.

vidsrc.c:

```
static void destructor(void *arg)
{
	struct vidsrc *vs = arg;

	list_unlink(&vs->le);
}
```

nicotyze:
Hi @alfredh,

Many thanks for your review and the sample code. I've updated the PR following your comments.

/vidsrc command tested with v4l2, avformat and cairo video sources (ubuntu16.04).

modules/v4l2/v4l2.c
```diff
@@ -476,7 +495,7 @@ static int alloc(struct vidsrc_st **stp, const struct vidsrc *vs,
 		return EINVAL;
 
 	if (!str_isset(dev))
-		dev = "/dev/video0";
+		dev = ((struct mediadev*)list_head(&vs->dev_list)->data)->name;
```
alfredh: perhaps add a helper function in mediadev.c that will return the first device
or the name of the first device ?

modules/v4l2/v4l2.c
```diff
@@ -524,8 +543,15 @@ static int alloc(struct vidsrc_st **stp, const struct vidsrc *vs,
 
 static int v4l_init(void)
 {
-	return vidsrc_register(&vidsrc, baresip_vidsrcl(),
+	int err;
+
+	err = vidsrc_register(&vidsrc, baresip_vidsrcl(),
 			       "v4l2", alloc, NULL);
```
alfredh: check `err` here. if err then return err.

modules/v4l2/v4l2.c
```diff
@@ -411,6 +411,25 @@ static int read_frame(struct vidsrc_st *st)
...
 
 
+static void set_available_devices(struct vidsrc* vs)
+{
+	int i, fd;
+	char name[16];
+
+	for (i=0;i < 16;i++) {
+
+		sprintf(name, "/dev/video%i", i);
```
alfredh: better to use `snprintf` or `re_snprintf` instead so the buffer size can be specified.

modules/menu/menu.c
```diff
@@ -745,10 +746,23 @@ static int switch_video_source(struct re_printf *pf, void *arg)
 		pl_strcpy(&pl_driver, driver, sizeof(driver));
 		pl_strcpy(&pl_device, device, sizeof(device));
 
-		if (!vidsrc_find(baresip_vidsrcl(), driver)) {
+		vs = vidsrc_find(baresip_vidsrcl(), driver);
+		if (!vs) {
 			re_hprintf(pf, "no such video-source: %s\n", driver);
 			return 0;
 		}
+		else if (list_count(&vs->dev_list)) {
```
alfredh: it is more effient to use `!list_isempty` instead.

modules/v4l2/v4l2.c
```diff
@@ -411,6 +411,25 @@ static int read_frame(struct vidsrc_st *st)
...
+
+		if ((fd = open(name, O_RDONLY)) == -1) {
+			continue;
+		}
+		else {
+			close(fd);
+			mediadev_add(&vs->dev_list, name);
```
alfredh: return value must be checked. return err if error.

alfredh:
thanks for the update. I also added more review comments :)

there is also one more place where this code is useful, in config.c
where is the configuration is populated and parsed.

for example in `src/config.c`:


```
static const char *default_video_device(void)
{
#ifdef DARWIN

#ifdef QTCAPTURE_RUNLOOP
	return "qtcapture,nil";
#else
	return "avcapture,nil";
#endif

#elif defined (WIN32)
	return "dshow,nil";
#else
	return "v4l2,/dev/video0";
#endif
}
```

this value goes into the default config template. so instead of hardcoding
`v4l2,/dev...` we could prove the relevant driver for the default device name
(most likely the first one).

nicotyze:
Hi @alfredh ,

Thanks again for the reviews !

My last commit is a trial to to set the relevant default video source device in default config template.
In order to test it:
1/ launch baresip
2/ suppress the config file
3/ load v4l2 module
4/ reload config file

Not really intuitive... Perhaps /conf_reload should be tuned in order to allow config file re-generation.

src/mediadev.c
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	return NULL;
+}
+
+
+struct mediadev *mediadev_get_default(const struct list *dev_list)
+{
+	return (struct mediadev*)list_head(dev_list)->data;
+}
```
alfredh: if the list is empty, this function will crash.

the function should return NULL if the list is NULL or empty.

also, the casting is not necessary.

src/config.c
```diff
@@ -511,7 +511,14 @@ static const char *default_video_device(void)
...
 #else
-	return "v4l2,/dev/video0";
+	re_hprintf(pf, "%s", "v4l2,");
+	const struct vidsrc *vs = vidsrc_find(baresip_vidsrcl(), "v4l2");
+	if(vs) {
+		return re_hprintf(pf, "%s", mediadev_get_default(&vs->dev_list)->name);
+	}
+	else {
+		return re_hprintf(pf, "%s", "/dev/video0");
+	}
```
alfredh: this will break the cases for the other `#ifdef`s since the signature of the
function is changed.

I think this part is a nice improvement, but if you want we can also leave it to another patch.

the printing of the default device should be shared for all platforms, something like:

1. find the default vidsrc module for the current platform
2. find the default device name for that module.

nicotyze:
Hi @alfredh ,
In my last commit, handling of empty devices should be cleaner....
I've tested it with 0, 1, 2 webcams plugged (with and without video source device specified in config file).

Concerning, the implementation in default config template, I agree to separate/postpone this patch.

alfredh:
Thanks !

PR #494, opened 2018-07-29
speex: remove module

alfredh:
from www.speex.org:

—The Speex codec has been obsoleted by Opus. It will continue to be available, but since Opus is better than Speex in all aspects, users are encouraged to switch—

NOTE: this is a suggestion to remove the speex module. If you have strong opinions
about this, please let me know.


jungle-boogie:
Thanks for being proactive and removing old code that's better handled by newer things.

PR #496, opened 2018-08-01
Send DTMF code command

nicotyze:
This is the only way have found to send a conference call PIN from an http interface.

### DIFF
--- a/src/baresip.c
+++ b/src/baresip.c
@@ -73,9 +73,9 @@ static int rmmod_handler(struct re_printf *pf, void *arg)
 
 
 static const struct cmd corecmdv[] = {
-	{"quit", 'q', 0, "Quit",                     cmd_quit            },
-	{"insmod",  0, CMD_PRM, "Load module",       insmod_handler      },
-	{"rmmod",   0, CMD_PRM, "Unload module",     rmmod_handler       },
+	{"quit", 'q', 0, "Quit",                     cmd_quit             },
+	{"insmod", 0, CMD_PRM, "Load module",        insmod_handler       },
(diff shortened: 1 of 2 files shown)


alfredh:
can you move this code to the menu module ?

the baresip.c commands is reserved for core commands that is not possible
to implement in a module.

also could you implement it like the digit_handler:

```
static int digit_handler(struct re_printf *pf, void *arg)
{
	struct call *call;
	int err = 0;

	(void)pf;

	call = ua_call(uag_cur());
	if (call)
		err = call_send_digit(call, carg->key);

	return err;
}
```

i.e.g same loop as suggest, but call `call_send_digit` instead.


Alfred

src/baresip.c
```diff
@@ -73,9 +73,9 @@ static int rmmod_handler(struct re_printf *pf, void *arg)
 
 
 static const struct cmd corecmdv[] = {
-	{"quit", 'q', 0, "Quit",                     cmd_quit             },
-	{"insmod", 0, CMD_PRM, "Load module",        insmod_handler       },
-	{"rmmod",  0, CMD_PRM, "Unload module",      rmmod_handler        },
+	{"quit", 'q', 0, "Quit",                     cmd_quit            },
+	{"insmod",  0, CMD_PRM, "Load module",       insmod_handler      },
+	{"rmmod",   0, CMD_PRM, "Unload module",     rmmod_handler       },
```
alfredh: this is an unrelated change, please revert it

alfredh:
some minor comments from me, apart from that it looks good!

alfredh:
Thank You !

PR #501, opened 2018-08-08
consider empty contacts file as existing contacts file

juha-h:
Accept also empty contacts file as existing contacts file.  When baresip is started with empty contacts file,
it now does not create a new default one, but reports:

Populated 0 contacts


alfredh:
Thanks!

PR #503, opened 2018-08-08
Module: RTCP Summary

geoffstewart:
Background:
I am using baresip as a tool that is controlled from a java application.  I control it via input/output streams and collect the stats from the output of the app.  This PR helps make that task easier.  The PR has 3 elements:

- new module: rtcpsummary
  - At the end of a call, output the RTCP stats in a format that's easily parsed
  - expose some read-only info from stream.c
- do not buffer output/input to any of stdin, stdout, stderr
  - without this change, the new UI infrastructure buffers all output until the application is killed; makes it impossible to control programatically
- Allow the network interface to be specified as a command-line argument

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1108,7 +1108,13 @@ double video_timestamp_to_seconds(uint64_t timestamp);
 
 const struct rtcp_stats *stream_rtcp_stats(const struct stream *strm);
 struct call *stream_call(const struct stream *strm);
-
+const struct sdp_media *stream_sdp(const struct stream *strm);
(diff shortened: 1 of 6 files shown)


alfredh:
hi @geoffstewart 

many thanks for this patch.

we prefer that each logical change is in its own PR/commit.

I committed the network interface option here:

ea516ccd21753d994116b9c3d6f85d587a1b4ab8

could you please rebase or merge from master and then split your PR into 2 parts:

1. RTCP summary module
2. stdio buffering


Alfred

geoffstewart:
Split into two new PRs, as requested: #504 and #505.  Closing this one.

PR #504, opened 2018-08-09
Prevent buffering standard I/O

geoffstewart:
- do not buffer output/input to any of stdin, stdout, stderr
  - without this change, the new UI infrastructure buffers all output
until the application is killed; makes it impossible to control
programatically

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -79,6 +79,11 @@ int main(int argc, char *argv[])
 	size_t i;
 	int err;
 
+	/*
+	 * turn off buffering on stdout
+	 */
+	setbuf(stdout, NULL);
+
 	(void)re_fprintf(stdout, "baresip v%s"
 			 " Copyright (C) 2010 - 2018"
 			 " Alfred E. Heggestad et al.\n",


alfredh:
I would like to hear the opinion from other people about this change.

@juha-h what do you think ?

and where is the best place to put the code:

1. main.c
2. somewhere in libbaresip.a
3. stdio module

alfredh:
hi @geoffstewart 

I have some more comments about your patch. I would like to understand
what problem it is trying to solve first.

1. is it necessary to call setbuf on stderr ? From [1]:
"The standard error stream stderr is always unbuffered by default."

2. the stdout stream should be line buffered by default, meaning that for every newline
  the buffer is flushed.

3. is it an option to call `fflush` when baresip exits normally ? does baresip crash
  for you?

4. making stdout unbuffered will make the performance worse. is it worth it ?


I have tested the patch on Debian Linux and it works fine. So it is fine to merge
it but I just want to understand why it is really needed :)


[1] http://man7.org/linux/man-pages/man3/setbuf.3.html


Alfred

geoffstewart:
Hmmm, I will have to test the stderr.  I think stdin can likely be removed.
The issue I was seeing was when controlling baresip via java program.  In previous versions, there was no issue.  I assume the UI rework caused the calling java program to see nothing until baresip was exited or killed.  For a while, I just thought it was buffering on the java side.  But, it was in baresip.
The java apps needs to see the lines as they are output to be able to react and properly control baresip.
I'm completely open to better ways of achieving this result.  Baresip did not crash, no.

geoffstewart:
Replacing the 3 lines with: 
{code}
	setvbuf(stdout, NULL, _IOLBF, 0);
{code}
It behaves well, as well.  This is setting the stdout to line buffering.  Is that better for performance?

alfredh:
hi @geoffstewart many thanks for your answers :)

I made a simple test program and tested it with strace in a Linux terminal:

```c
#include <re.h>


int main(void)
{
        re_printf("Hello.");

        re_printf("World\n");
}
```

The default program gives 1 write operation:

```
write(1, "Hello.World\n", 12Hello.World
)           = 12
```

(i.e. line buffered)

Then I added this to the start:

```
        setvbuf(stdout, NULL, _IOLBF, 0);
```

and it gives the same result, i.e. still line buffered.

Then I added this:

```
        setbuf(stdout, NULL);
```

and it gives 2 write operations:

```
write(1, "Hello.", 6Hello.)                   = 6
write(1, "World\n", 6World
)                  = 6
```

Is it possible that a program inherits stdio settings from the shell/terminal ?

Is it possible that your Java program sets the stdio to "Block" buffered by default ?

Since most of the log-lines in baresip have a trailing newline, adding `setbuf` will not
degrade performance. So I think it is okay to add it, but might be worth to check your
Java terminal first :)


Alfred

geoffstewart:
With your hello world program, I don't need to make any changes to buffering when it's called from Java.  It simply outputs as expected. 
In that case, the piping of the output and error to the java program has no effect.
And, like I've mentioned above, in previous versions of baresip, the same was true.  I suspect that the UI rework somehow changed the behaviour when the output is piped... although, I'm not positive at all.
I'm not sure if I'm trying to avoid block buffering.  It was simply buffering *all* output, but perhaps it hadn't reached it's buffer limit yet. Either way, it was not workable for me.

Here's an example the reproduces the issue without needing my Java application:
```
mkfifo mypipe
baresip < mypipe
# in another shell
echo "/quit" > mypipe
```
There is no output on the first shell until you echo the "/quit" command via the pipe; then all the buffered text is output.

alfredh:
in the old days baresip printed everything to `stderr` (unbuffered).

then we changed the logging to print to `stdout`. Details are here:

https://github.com/alfredh/baresip/issues/320

I think we can reduce your patch to this:

```
	setbuf(stdout, NULL);
```

and then it should be okay to merge it to master :)

geoffstewart:
I'm not sure what failed on the build... it doesn't look related to the PR to me, but do let me know if I need to change something to make the build succeed.

alfredh:
thank you!

PR #505, opened 2018-08-09
RTCP summary module

geoffstewart:
- new module: rtcpsummary
   - At the end of a call, output the RTCP stats in a format that's easily parsed
   - expose some read-only info from stream.c

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1127,7 +1127,13 @@ double video_timestamp_to_seconds(uint64_t timestamp);
 
 const struct rtcp_stats *stream_rtcp_stats(const struct stream *strm);
 struct call *stream_call(const struct stream *strm);
-
+const struct sdp_media *stream_sdp(const struct stream *strm);
+uint32_t metric_get_tx_n_packets(const struct stream *strm);
+uint32_t metric_get_tx_n_bytes(const struct stream *strm);
+uint32_t metric_get_tx_n_err(const struct stream *strm);
+uint32_t metric_get_rx_n_packets(const struct stream *strm);
+uint32_t metric_get_rx_n_bytes(const struct stream *strm);
+uint32_t metric_get_rx_n_err(const struct stream *strm);
 
 /*
  * Media NAT
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1128,12 +1128,12 @@ double video_timestamp_to_seconds(uint64_t timestamp);
 const struct rtcp_stats *stream_rtcp_stats(const struct stream *strm);
 struct call *stream_call(const struct stream *strm);
 const struct sdp_media *stream_sdp(const struct stream *strm);
-uint32_t metric_get_tx_n_packets(const struct stream *strm);
-uint32_t metric_get_tx_n_bytes(const struct stream *strm);
-uint32_t metric_get_tx_n_err(const struct stream *strm);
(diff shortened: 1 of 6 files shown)


include/baresip.h
```diff
@@ -1127,7 +1127,13 @@ double video_timestamp_to_seconds(uint64_t timestamp);
...
 const struct rtcp_stats *stream_rtcp_stats(const struct stream *strm);
 struct call *stream_call(const struct stream *strm);
-
+const struct sdp_media *stream_sdp(const struct stream *strm);
+uint32_t metric_get_tx_n_packets(const struct stream *strm);
+uint32_t metric_get_tx_n_bytes(const struct stream *strm);
+uint32_t metric_get_tx_n_err(const struct stream *strm);
+uint32_t metric_get_rx_n_packets(const struct stream *strm);
+uint32_t metric_get_rx_n_bytes(const struct stream *strm);
+uint32_t metric_get_rx_n_err(const struct stream *strm);
```
alfredh: these metric functions should be prefixed with `stream_` which is the same as 
the other functions accessing `struct stream *strm` ..

mk/modules.mk
```diff
@@ -53,6 +53,7 @@
 #   USE_V4L2          Video4Linux2 module
 #   USE_WINWAVE       Windows audio driver
 #   USE_X11           X11 video output
+#   USE_RTCPSUMMARY   RTCP summary output after calls
```
alfredh: please try to use alphabetical order, if possible

modules/rtcpsummary/rtcpsummary.c
```diff
@@ -0,0 +1,122 @@
+/**
+ * @file rtcpsummary.c RTCP summary module
+ * Output RTCP stats at the end of a call if there are any
+ */
```
alfredh: a copyright statement is needed here.

you can for example add Creytiv.com and your own name, if you want

modules/rtcpsummary/rtcpsummary.c
```diff
@@ -0,0 +1,122 @@
...
+	(void)call;
+	struct le *le;
+	const struct rtcp_stats *rtcp;
+	const struct stream *s;
+
+	switch (ev) {
+
+		case UA_EVENT_CALL_CLOSED:
+			debug("rtcpsummary: CALL_CLOSED");
+			for (	le = call_streaml(call)->head;
```
alfredh: I suggest that you move this whole block into its own function that is called from here.

if will make the code more readable, with less indentation.

src/stream.c
```diff
@@ -756,3 +756,34 @@ struct call *stream_call(const struct stream *strm)
...
+ * @return SDP media object
+ */
+const struct sdp_media *stream_sdp(const struct stream *strm)
+{
+	return strm ? strm->sdp : NULL;
+}
+
+uint32_t metric_get_tx_n_packets(const struct stream *strm) {
+	return strm ? strm->metric_tx.n_packets : 0;
+}
```
alfredh: please move the starting curly bracket to the beginning of the next line, so that it looks like this:

```
void foo(void)
{
}
```

the same applies for all the other added functions below

include/baresip.h
geoffstewart: Makes sense.  Once question: Replace metric_ with stream_?  Or simply add stream_ to what's there now?

alfredh:
many thanks for your contributions :)

PR #511, opened 2018-08-20
Modified so that x11 displays on MobaXTerm.

rmundkowsky:


alfredh:
many thanks!

with this patch I am getting an error on Debian Linux:

```
X Error of failed request:  BadShmSeg (invalid shared segment parameter)
  Major opcode of failed request:  132 (MIT-SHM)
  Minor opcode of failed request:  3 (X_ShmPutImage)
  Segment id in failed request:  0xa00002
  Serial number of failed request:  24
  Current serial number in output stream:  25
```

rmundkowsky:
Are you running X11 server/client on same server? Are you running Docker?

And does it work with shared memory turned off?

export QT_X11_NO_MITSHM=1
export _X11_NO_MITSHM=1
export _MITSHM=0

Robert Mundkowsky

alfredh:
it is a regular Debian installation, no docker. The X server is running
on OSX via ssh tunnel:

`DISPLAY=localhost:10.0`

turning off shared memory gives the same result.

modules/x11/x11.c
```diff
@@ -221,18 +222,23 @@ static int x11_reset(struct vidisp_st *st, const struct vidsz *sz)
...
 	x11.errorh = XSetErrorHandler(error_handler);
 
-	if (!XShmAttach(st->disp, &st->shm)) {
-		warning("x11: failed to attach X to shared memory\n");
-		return ENOMEM;
+	if (st->xshmat){
+		if (!XShmAttach(st->disp, &st->shm)) {
+			warning("x11: failed to attach X to shared memory\n");
+			return ENOMEM;
+		}
```
alfredh: the flag xshmat is not set at this point, so it will always be false.

it is set a bit below, around line 240

rmundkowsky:
I created a new pull request with a fix for this.

Robert Mundkowsky

alfredh:
thanks for the update.

 I tested this patch on 2 different systems, OSX with X11 and shared memory enabled,
and Linux with shared memory disabled.

Result:

OSX X11: OK
Linux: Error

On linux (which has shared memory disabled) there is an error from xorg:


```
vidloop: NOTE: pixel-format conversion needed: rgb32  -->  yuv420p
x11: shared memory enabled
X Error of failed request:  BadAccess (attempt to access private resource denied)
  Major opcode of failed request:  132 (MIT-SHM)
  Minor opcode of failed request:  1 (X_ShmAttach)
  Serial number of failed request:  20
  Current serial number in output stream:  21
```

I think something goes wrong here:

```
        if (x11.shm_error || !XShmQueryExtension(st->disp)){
                info("x11: shared memory disabled\n");
                st->xshmat = false;
        }
        else{
                info("x11: shared memory enabled\n");
                st->xshmat = true;
        }
```

since that X11 has share memory disabled, but it still sets the xshmat flag to true.

alfredh:
btw, the x11 module is using the global error handler and XSync to figure out
if the call to XShmAttach worked or not.

perhaps this logic is not working for your terminal..

could you perhaps test this small patch and see if that helps ?


```diff
diff --git a/modules/x11/x11.c b/modules/x11/x11.c
index eafb0d6..5b1ec42 100644
--- a/modules/x11/x11.c
+++ b/modules/x11/x11.c
@@ -223,7 +223,7 @@ static int x11_reset(struct vidisp_st *st, const struct vidsz *sz)
 
        if (!XShmAttach(st->disp, &st->shm)) {
                warning("x11: failed to attach X to shared memory\n");
-               return ENOMEM;
+               x11.shm_error = 1;
        }
 
        XSync(st->disp, False);
```

rmundkowsky:
The if statement is the same as before, but with an added test that would make “st->xshmat = false;” , should work like before if shared memory is disabled. So I wonder if you have shared memory totally disabled.  Only little documentation I found states you have to set:

QT_X11_NO_MITSHM=1
_X11_NO_MITSHM=1
_MITSHM=0

I also read something that stated SSH tunnel fakes Xwindows into thinking that shared memory is turned on, but I am not sure if this is true.

Robert Mundkowsky

alfredh:
the function `XShmQueryExtension` returns 1 on both my systems, so I guess we cannot
use it to decide share mem or not. Perhaps it returns 0 on your system ?

could you try to run `xdpyinfo` in your terminal and check for the MIT-SHM flag ?

here is from my system:


```
number of extensions:    22
    Apple-DRI
    Apple-WM
    BIG-REQUESTS
    DAMAGE
    DOUBLE-BUFFER
    GLX
    Generic Event Extension
    MIT-SCREEN-SAVER
    MIT-SHM
    Present
    RANDR
    RENDER
    SGI-GLX
    SHAPE
    SYNC
    X-Resource
    XC-MISC
    XFIXES
    XINERAMA
    XInputExtension
    XKEYBOARD
    XVideo
```

alfredh:
hi,

I made a small patch here:

https://gist.github.com/alfredh/ffc74e2546f16c18eb1b38462c04539c

it basically checks XShmQueryExtension first to determine if we
should call XShmAttach or not.

Could you please test it on your system ?

if this works, then I guess it should work on most systems ..

rmundkowsky:
This patch works on my system.

Robert Mundkowsky

alfredh:
fixed in https://github.com/alfredh/baresip/commit/178f3edd4de528c6ee25cf1c89a76ac1e256c613

rmundkowsky:
Thanks!

Robert Mundkowsky

PR #512, opened 2018-08-21
call: Support also SIP URI with missing display name.

cspiel1:
In function call_connect SIP URIs like "<sip:user@host>" are not supported. The
log tells us:

 call: sipsess_connect: Function not implemented

This is because the angle brackets are present and the display name is missing.
The patch achieves that the empty display name is ignored and the URI is
accepted.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -742,7 +742,7 @@ int call_connect(struct call *call, const struct pl *paddr)
 	/* if the peer-address is a full SIP address then we need
 	 * to parse it and extract the SIP uri part.
 	 */
-	if (0 == sip_addr_decode(&addr, paddr) && addr.dname.p) {
+	if (0 == sip_addr_decode(&addr, paddr)) {
 		err = pl_strdup(&call->peer_uri, &addr.auri);
 	}
 	else {


juha-h:
Christian Spielberger writes:

> In function call_connect SIP URIs like "<sip:user@host>" are not
> supported.

Is the URI in call_connect a Request-URI?  If so, the above is not a
valid SIP-URI.

-- Juha

cspiel1:
Hi Juha,
it might be, but not coercibly. The call_connect paddr may contain the display name or not. But if so, the call to sip_addr_decode removes the display name and angle brackets in order to get a correct call->peer_uri.

regards,
chris.

juha-h:
Christian Spielberger writes:

> it might be, but not coercibly. The call_connect paddr may contain the
> display name or not. But if so, the call to sip_addr_decode removes
> the display name and angle brackets in order to get a correct
> call->peer_uri.

OK, in name-addr, display-name may be empty:

name-addr =  [ display-name ] LAQUOT addr-spec RAQUOT

-- Juha

juha-h:
Christian Spielberger writes:

> it might be, but not coercibly. The call_connect paddr may contain the
> display name or not. But if so, the call to sip_addr_decode removes
> the display name and angle brackets in order to get a correct
> call->peer_uri.

call_connect() is not API function.  does this mean that also some API
function would not accept name-addr without display-name?

alfredh:
thanks, well spotted @cspiel1 

I wanted to fix this for some time now. I will make some new testcases
and merge your patch.

alfredh:
Thanks @cspiel1 

I merged your patch and added a testcase for SIP uri with angle brackets here:

https://github.com/alfredh/baresip/commit/ee2afbd4a0b9268893ed0ebc08b726850d81e14e

PR #513, opened 2018-08-22
make struct ausrc and auplay public

nicotyze:
This is a proposal in order to be consistent with vidsrc and vidisp struct.

This patch is necessary for mediadev API support in audio modules.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -475,7 +475,6 @@ typedef int  (auplay_alloc_h)(struct auplay_st **stp, const struct auplay *ap,
 struct auplay {
 	struct le        le;
 	const char      *name;
-	struct list      dev_list;
 	auplay_alloc_h  *alloch;
 };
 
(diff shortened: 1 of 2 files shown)


include/baresip.h
```diff
@@ -464,6 +471,14 @@ typedef int  (auplay_alloc_h)(struct auplay_st **stp, const struct auplay *ap,
 			      struct auplay_prm *prm, const char *device,
 			      auplay_write_h *wh, void *arg);
 
+/** Defines an Audio Player */
+struct auplay {
+	struct le        le;
+	const char      *name;
+	struct list      dev_list;
```
alfredh: please remove `dev_list` from this patch, so that this patch has no functional
changes; it just moves the struct.

alfredh:
this looks fine to me.

dev_list should not be added though; it is better to do this in 2 steps,
first move the struct and then later add the dev_list type and logic.

PR #516, opened 2018-08-27
Add dev_list to auplay and ausrc struct

nicotyze:
- set_available_devices implementation in pulse module
  - check audio player/source devices availability in switch_audio_player/source

Tested on ubuntu 16.04.

### DIFF
--- a/modules/pulse/recorder.c
+++ b/modules/pulse/recorder.c
@@ -209,7 +209,7 @@ static void dev_list_cb(pa_context *c, const pa_source_info *l,
 
 	/* In pulseaudio every sink automatically has a monitor source
 	This "output" device must be filtered out */
-	if(!strstr(l->name,"output")) {
+	if (!strstr(l->name,"output")) {
 		err = mediadev_add(dev_list, l->name);
 		if (err) {
 			warning("pulse recorder: media device (%s) "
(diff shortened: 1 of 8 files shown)


modules/pulse/recorder.c
```diff
@@ -191,3 +195,75 @@ int pulse_recorder_alloc(struct ausrc_st **stp, const struct ausrc *as,
...
+	pa_mlapi = pa_mainloop_get_api(pa_ml);
+	pa_ctx = pa_context_new(pa_mlapi, "Baresip");
+
+	pa_context_connect(pa_ctx, NULL, 0, NULL);
+
+	while (pa_context_get_state(pa_ctx) != PA_CONTEXT_READY) {
+		err = pa_mainloop_iterate(pa_ml, 1, NULL);
+		if (err < 0) {
+			return err;
+		}
```
alfredh: when returning here, it will create a memory leak, the pa_op and pa_ctx will leak ..

alfredh:
hi @nicotyze 

thank you very much, really great work !

please keep the PRs coming :)

alfredh:
btw, I noticed that pulse.so now takes longer time to load.

the first time it was loaded, it took around 1 second.
the second and third time it was a bit faster.

any idea if this can be optimized ?

PR #526, opened 2018-09-02
h265: change rate control to ABR (Average Bitrate)

alfredh:


alishir:
Hi @alfredh,

I applied this pull request, the bandwidth usage is about `video_bitrate` configured value, but artifacts exist is received video. Also there exists lots of missing reference in hvec.

image

alfredh:
hi @alishir 

do you think it is better to use ABR ?

baresip module avcodec (H.264) is already using ABR, and also ffmpeg libavcodec
is using ABR.

do you think we should merge this one now to master, for the 0.5.11 release ?

alfredh:
if you tested H.265 two weeks ago and it was working, perhaps you can
test again with the previous release of baresip (v0.5.10) and see if that makes
a difference.

alishir:
In order to test my ffmpeg and x265 library, I set up a streamer/receiver with gstreamer that uses H.264 encoder/decoder, unfortunately the frame rate is too low, about 5 frame per second, and CPU usage is too high, about 178% on a dual core CPU. I think the current setting of x264/ffmpeg have some problem.

PR #529, opened 2018-09-05
Passing rtcpsummary info to user-defined callback

Encamy:
Hello @alfredh and @geoffstewart 
First of all, thank you for this useful module.
I ran into the need to use statistics in another function, and not just output it to the log.
So I think that calling a custom handler will solve this problem.
Is this PR will be appropriate?

Aleksei

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1151,6 +1151,24 @@ uint32_t stream_metric_get_rx_n_packets(const struct stream *strm);
 uint32_t stream_metric_get_rx_n_bytes(const struct stream *strm);
 uint32_t stream_metric_get_rx_n_err(const struct stream *strm);
 
+
+/*
+ * RTCP summary module
+ */
+
+typedef void (rtcpsummary_callback_h)(
+	const struct ua *ua,
+	size_t call_setup, size_t call_duration,
+	size_t packets_rx, size_t packets_tx,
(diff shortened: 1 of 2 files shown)


include/baresip.h
```diff
@@ -1151,6 +1151,24 @@ uint32_t stream_metric_get_rx_n_packets(const struct stream *strm);
...
+	const struct ua *ua,
+	size_t call_setup, size_t call_duration,
+	size_t packets_rx, size_t packets_tx,
+	size_t packets_loss_rx, size_t packets_loss_tx,
+	size_t packets_discarded_rx, size_t packets_discarded_tx,
+	double jitter_rx, double jitter_tx,
+	double rtt,
+	const struct sa *laddr, const struct sa *raddr);
+
+void rtcpsummary_set_callback(rtcpsummary_callback_h *callback);
```
alfredh: modules should not export global symbols. they only have 1 global entry point
which the module loader uses to initialize the modules.

alfredh:
if you also need this information, it is perhaps better to add a dedicated handler
for it in the baresip core (libbaresip).

The ua event UA_EVENT_CALL_RTCP is already implemented, and can be used
to fetch this information.


Alfred

Encamy:
Thank you for pointing at ua event UA_EVENT_CALL_RTCP. It is much better solution.

PR #530, opened 2018-09-10
- added support for per account mwi using ;mwi=on|off addr-param

juha-h:
- added account_mwi() and account_set_mwi() API functions
- added ;mwi account test


alfredh:
thanks, will start to review this after 0.5.11 is released.

modules/mwi/mwi.c
```diff
@@ -64,6 +64,8 @@ static void notify_handler(struct sip *sip, const struct sip_msg *msg,
 		ui_output(uis, "----- MWI for %s -----\n", ua_aor(mwi->ua));
 		ui_output(uis, "%b\n", mbuf_buf(msg->mb),
 			  mbuf_get_left(msg->mb));
+		ua_event(mwi->ua, UA_EVENT_MWI_NOTIFY, NULL, "%b",
+			  mbuf_buf(msg->mb), mbuf_get_left(msg->mb));
```
alfredh: adding this event is a good thing. then the application or other modules can use it.

perhaps you could also move the `ui_output` block above to the `menu` module
and let it also handle the UA_EVENT_MWI_NOTIFY event ?

modules/mwi/mwi.c
```diff
@@ -154,7 +156,8 @@ static void ua_event_handler(struct ua *ua,
 
 	if (ev == UA_EVENT_REGISTER_OK) {
 
-		if (!mwi_find(ua))
+		if (!mwi_find(ua) &&
+		    (strcmp(account_mwi(ua_account(ua)), "yes") == 0))
```
alfredh: please be aware that `strcmp` does not handle NULL input, in which case it will crash.

it is better to use `str_casecmp` or let the `account_mwi` function return `bool`.

src/event.c
```diff
@@ -167,6 +170,7 @@ const char *uag_event_str(enum ua_event ev)
 	case UA_EVENT_REGISTER_OK:          return "REGISTER_OK";
 	case UA_EVENT_REGISTER_FAIL:        return "REGISTER_FAIL";
 	case UA_EVENT_UNREGISTERING:        return "UNREGISTERING";
+	case UA_EVENT_MWI_NOTIFY:           return "MWI_NOFIFY";
```
alfredh: small typo here in the returned string :)

alfredh:
this looks very nice. some minor comments inline.

just a question, if the mwi parameter is missing is the mwi enabled
or disabled by default ?

juha-h:
Alfred E. Heggestad writes:
> 
> > @@ -64,6 +64,8 @@ static void notify_handler(struct sip *sip, const struct sip_msg *msg,
>  		ui_output(uis, "----- MWI for %s -----\n", ua_aor(mwi->ua));
>  		ui_output(uis, "%b\n", mbuf_buf(msg->mb),
>  			  mbuf_get_left(msg->mb));
> +		ua_event(mwi->ua, UA_EVENT_MWI_NOTIFY, NULL, "%b",
> +			  mbuf_buf(msg->mb), mbuf_get_left(msg->mb));
> 
> adding this event is a good thing. then the application or other modules can use it.
> 
> perhaps you could also move the `ui_output` block above to the `menu` module
> and let it also handle the UA_EVENT_MWI_NOTIFY event ?

Yes, I thought that too.  I can try to do that tomorrow.

juha-h:
Alfred E. Heggestad writes:

> just a question, if the mwi parameter is missing is the mwi enabled
> or disabled by default ?

Enabled by default so the current behavior is not changed,

juha-h:
Alfred E. Heggestad writes:
> alfredh commented on this pull request.
> 
> 
> 
> > @@ -154,7 +156,8 @@ static void ua_event_handler(struct ua *ua,
>  
>  	if (ev == UA_EVENT_REGISTER_OK) {
>  
> -		if (!mwi_find(ua))
> +		if (!mwi_find(ua) &&
> +		    (strcmp(account_mwi(ua_account(ua)), "yes") == 0))
> 
> please be aware that `strcmp` does not handle NULL input, in which
> case it will crash.
> 
> it is better to use `str_casecmp` or let the `account_mwi` function
> return `bool`.

account_mwi always returns "yes" or "no" and I don't think that in the
above ua_account(ua) can return NULL.  I can change comparison function
to str_casecmp.

PR #535, opened 2018-09-16
per account support for call transfer (REFER method)

juha-h:


src/account.c
```diff
@@ -945,6 +978,19 @@ const char *account_mwi(const struct account *acc)
...
 
+/**
+ * Get call transfer capability of an account
+ *
+ * @param acc User-Agent account
+ *
+ * @return "yes" or "no"
+ */
+const char *account_call_transfer(const struct account *acc)
+{
```
alfredh: the input argument (acc) must be checked for NULL.

if NULL you can return either NULL or the string "no".

alfredh:
thanks !

I merged it to master.

Issue #155, opened 2016-10-03
build error (master branch on Ubuntu 14 32bit latest updates)


cypa:
src/video.c: In function ‘video_stream_decode’:
src/video.c:597:3: warning: implicit declaration of function ‘vidframe_copy’ [-Wimplicit-function-declaration]
   vidframe_copy(frame_filt, frame);
   ^
src/video.c:597:3: warning: nested extern declaration of ‘vidframe_copy’ [-Wnested-externs]
  CC      build-i686/src/vidcodec.o
  CC      build-i686/src/vidfilt.o
  CC      build-i686/src/vidisp.o
  CC      build-i686/src/vidsrc.o
  CC      build-i686/src/main.o
  LD      baresip
build-i686/src/video.o: In function `video_stream_decode':
/home/cypa/tmp/baresip/src/video.c:597: undefined reference to `vidframe_copy'
collect2: error: ld returned 1 exit status
make: *** [baresip] Ошибка 1

cypa:
how do I disable video? 
USE_VIDEO := 0 in Makefile doesn't help

cypa:
commented next line in 2 files to build successfully

//                      vidframe_copy(frame_filt, frame);

czarkoff:
@cypa: you should have updated [librem](http://creytiv.com/rem.html) to version 0.4.7 or later.

P.S.: to disable video you need to call make with `USE_VIDEO=` flag.

cypa:
OK

PR #543, opened 2018-09-27
Use extra CA file for TLS transport

wnetbal:
Added a new option to baresip configuration called sip_ca that is used the set an extra CA file for TLS transport.

In combination with pull request [tls: set verify methode for openSSL 1.1.0 #155](https://github.com/creytiv/re/pull/155) from creytiv/re
a TLS connection to a server can use the CA file checking if the pem file is created with
this CA and is still valid.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -11,6 +11,7 @@ poll_method		epoll		# poll, select, epoll ..
 sip_trans_bsize		128
 #sip_listen		0.0.0.0:5060
 #sip_certificate	cert.pem
+#sip_ca			ca.crt
 
 # Audio
 audio_player		alsa,default
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -220,6 +220,7 @@ struct config_sip {
 	char uuid[64];          /**< Universally Unique Identifier  */
 	char local[64];         /**< Local SIP Address              */
 	char cert[256];         /**< SIP Certificate                */
+	char ca[256];           /**< SIP CA                         */
 };
 
 /** Call config */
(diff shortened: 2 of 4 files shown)


wnetbal:
Hi Alfred,

my merge failed because of a line is longer than 79 signs.

warning("ua: tls_add_ca() failed: %m\n", err);
what makes sens is to move the parameters to the next line 
warning("ua: tls_add_ca() failed: %m\n",
                 err);

But this doesn't fix the warning because the line is then 80 signs long.

Do you have a best practice for this ?

Kind regards
Wolfgang

alfredh:
one solution is to break the string:

```
warning("foo"
           " bar\n");
```

alfredh:
do you have a SIP server up and running that I can test against ?

do you have some sample PEM and CRT files that I can use ?

alfredh:
please also have a look at the function `tls_peer_verify`:

https://github.com/creytiv/re/blob/master/src/tls/openssl/tls.c#L683

I have tested locally with baresip against a service that supports TLS (antisip.com)
and the verify function fails with this error code:

`X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY`

I guess it means they are using self-signed certs or there is local
problem getting the root CAs ?

alfredh:
perhaps the name of the config item should be `sip_cafile` ?

```
sip_cafile      server.crt
```

alfredh:
thanks @wnetbal 

I added your patch here:

https://github.com/alfredh/baresip/commit/1aba4e95bae18d07924da0fe325f5834048f08c1

with a small change, using `sip_cafile` as the config name.

PR #550, opened 2018-09-30
dshow: mediadev support

nicotyze:
Hi,
This PR is an  implementation of set_available_devices in dshow module.

Testing environment:
- Binaries tested on Win10

### DIFF
--- a/modules/dshow/dshow.cpp
+++ b/modules/dshow/dshow.cpp
@@ -179,7 +179,8 @@ private:
 static struct vidsrc *vsrc;
 
 
-static int get_device(struct vidsrc_st *st, const char *name)
+static int enum_devices(struct vidsrc_st *st, const char *name,
+			struct list *dev_list
 {
 	ICreateDevEnum *dev_enum;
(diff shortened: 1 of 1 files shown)


alfredh:
Thank you :)

alfredh:
@nicotyze got a build error:

modules/dshow/dshow.cpp:184:1: error: expected ‘)’ before ‘{’ token

could you please have a look ?

nicotyze:
Sorry for this hasty commit...
I've just made a new PR to correct it.

alfredh:
I just tested this again with linux+wine and 1 webcam attached.

the device is added twice:

```
dshow: device name='HD Pro Webcam C920' id=0
dshow: device name='HD Pro Webcam C920' id=1
```

perhaps it is a bug in wine.

in any case, before calling mediadev_add should we call mediadev_find to
check if it already exists and avoid duplicates ?

nicotyze:
I've done a test as well with linux+wine, without being able to reproduce the problem.
But your solution is OK for me.

Issue #539, opened 2018-09-16
module aufile closes connection after playing wave file


hallerbachj:
Hi Alfred,

when I make a call using the /d parameter, wait until the connection is established, then change the audio source with /ausrc aufile, test.wav the wave file is beeing played, but after that the connection is closed. Is it possible, to not close the connection to give the user the chance to send for instance dtmf codes and close the connection if the right code is entered? Is it possible to play more than one wave file one after one by repeating the /ausrc command? I think, these functions maybe useful for other users too.

Are there some examples, how to get dtmf codes and make use of them?

Thanks

alfredh:
Hi @hallerbachj 

this code in aufile module is calling error handler when the file is done:

```c
	/* check if audio buffer is empty */
	if (aubuf_cur_size(st->aubuf) < (2 * st->sampc)) {

		info("aufile: end of file\n");

		/* error handler must be called from re_main thread */
		if (st->errh)
			st->errh(0, "end of file", st->arg);
	}
```

furthermore, in the audio.c this handler is called:

```c
static void ausrc_error_handler(int err, const char *str, void *arg)
{
	struct audio *a = arg;
	MAGIC_CHECK(a);

	if (a->errh)
		a->errh(err, str, a->arg);
}
```

and finally in call.c will disconnect the call:

```
static void audio_error_handler(int err, const char *str, void *arg)
{
	struct call *call = arg;
	MAGIC_CHECK(call);

	if (err) {
		warning("call: audio device error: %m (%s)\n", err, str);
	}

	call_stream_stop(call);
	call_event_handler(call, CALL_EVENT_CLOSED, str);
}
```


the intended scenario was that the active call is playing a wav file
back to the remote peer, and when the file is done the call is disconnected.

if you prefer a different logic, we can discuss that

one work around for you is to use a WAV file with your audio and then
append a lot of silence to it :)

hallerbachj:
Hi Alfred,

thanks for your reply. After commenting out 

> 	call_stream_stop(call);
> 	call_event_handler(call, CALL_EVENT_CLOSED, str);
the behaviour is exactly like I need it. I have tested it in the following way:

Make a call with the „d“ command
If the call ist accepted, switch the audio source with /ausrc aufile,“test.wav“ -> after playing is finished, the call is not terminated -> listening to dtmf-codes -> play the wave file again with /ausrc, „test.wav“ -> end the call with the „b“ command.

[...]

PR #555, opened 2018-10-05
add UA_EVENT_AUDIO_ERROR (ref #539)

alfredh:
add an event for audio errors, so the application can handle it.

currently the active call is disconnected in the event of an audio error.
some use cases wants the call to continue, if audio error happens.


juha-h:
Alfred E. Heggestad writes:

> add an event for audio errors, so the application can handle it.
> 
> currently the active call is disconnected in the event of an audio error.
> some use cases wants the call to continue, if audio error happens.

Just a question: what are typical causes of audio errors?

juha-h:
Ok, then in normal call it makes always sense to disconnect the call at this error event (like menu module now does).

Issue #554, opened 2018-10-05
Rpi Zero W / PiCam v2 / v4l2_codec does not work


henriz:
**Use Case**
Use case: a SIP capable video intercom doorbel, based on a Rpi Zero W, ReSpeaker 2 Mic pi HAT and the pi cam.

**Problem**

I have got this working with video, using v4l2. However, it's terribly slow and doesn't use hardware h264 encoding then, which is a shame, as the Rpi has that. It also has h264 hardware decoding btw.

If I set the config to

video_source         v4l2,/dev/video0
module                  v4l2.so
module                  avcodec.so

this works, abeit very slow. It can't even do 160x120 @ 20fps.

Now if I change it to

video_source         v4l2_codec,/dev/video0
module                  v4l2_codec.so
module                  avcodec.so

It does claim to be working, but the receiving end gets no video. I have tested this with a Grandstream GXV3175 and with linphone.

Note: this goes both for the epoll as poll polling setting. Epoll works fine with v4l2.

**Output using v4l2**

Set video encoder: H264 packetization-mode=0 (500000 bit/s, 25.00 fps)
avcodec: video encoder H264: 25.00 fps, 500000 bit/s, pktsize=1024
video: no video display
v4l2: /dev/video0: found valid V4L2 device (160 x 120) pixfmt=YU12
v4l2: Current input: "Camera 0"
v4l2: current framerate is 30.00 fps
call: stream start (active=1)
audio: start
Set video decoder: H264 packetization-mode=0
avcodec: video decoder H264 ()
777@192.168.178.50: Call established: sip:333@192.168.178.50

**Output using v4l2_codec**

Set video encoder: H264 packetization-mode=0 (500000 bit/s, 25.00 fps)
v4l2_codec: video encoder H264: 25.00 fps, 500000 bit/s, pktsize=1024
video: no video display
v4l2_codec: Driver Caps:
  Driver:        "bm2835 mmal"
  Card:          "mmal service 16.1"
  Bus:           "platform:bcm2835-v4l2"
  Version:       4.0
  Capabilities:  0x85200005
  Formats:
     YU12:    'Planar YUV 4:2:0'
     YUYV:    'YUYV 4:2:2'
     RGB3:    '24-bit RGB 8-8-8'
     JPEG: C  'JFIF JPEG'
  -  H264: C  'H.264'
     MJPG: C  'Motion-JPEG'
     YVYU:    'YVYU 4:2:2'
[...]

PR #558, opened 2018-10-13
v4l2_codec: use thread instead of fd event

alfredh:
from a user:

"Also you need to be using poll. Epoll doesn't work."

reference #554


alfredh:
thanks to @henriz for testing :)

PR #559, opened 2018-10-14
winwave: add support for FLOAT sample format

alfredh:
This PR adds support for FLOAT sample format.

See here for a list of audio drivers and sample formats:

https://github.com/alfredh/baresip/wiki/Audio-sample-formats

### DIFF
--- a/modules/winwave/play.c
+++ b/modules/winwave/play.c
@@ -7,6 +7,7 @@
 #include <rem.h>
 #include <windows.h>
 #include <mmsystem.h>
+#include <mmreg.h>
 #include <baresip.h>
 #include "winwave.h"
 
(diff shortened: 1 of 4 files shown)


alfredh:
hi @nicotyze do you want to review/test this patch ?


Alfred

nicotyze:
Hi @alfredh,

The patch looks good.

I've tested from win10 with following config:
>ausrc_format		float		# s16, float, ..
auplay_format		float		# s16, float, ..
auenc_format		s16		# s16, float, ..
audec_format		s16		# s16, float, ..

/auloop => OK, float format seems to be well supported
My baresip version does not support opus or mpa so I can not test if it is ok with these codecs.
But when I do a call test with g711,  FLOAT  -->  S16LE and S16LE  -->  FLOAT conversions seem to work.

Nicolas

alfredh:
thanks for helping @nicotyze  !

PR #566, opened 2018-10-26
opengl: properly handle linesize

alfredh:
the current code assumes that linesize == width.

if the linesize is different from width, the rendered picture
will be broken.

adding a call to glPixelStorei(GL_UNPACK_ROW_LENGTH, linesize)
seems to fix it.


alfredh:
help needed to test this patch, for example on OSX.

PR #567, opened 2018-10-30
pass CALL_EVENT_TRANSFER_FAILED event to handler and handle it in men…

juha-h:
…u module


alfredh:
Thank you, this looks really nice !

PR #569, opened 2018-11-01
ctrl_tcp: Fix netstring frame handling

jmillan:


jmillan:
Before this fix, when multiple netstring strings are received in the tcp receiver handler, only the first one is processed.

alfredh:
Thank you, I verified your patch with this command:

`21:{"command":"sysinfo"},21:{"command":"memstat"},`

which said:

```
ctrl_tcp: handle_command:  cmd='sysinfo', params:'', token=''
ctrl_tcp: handle_command:  cmd='memstat', params:'', token=''
```

Alfred

PR #578, opened 2018-11-21
/play will always stop current file playing. 

Ola-Palm:
The current file play via /play is always stopped before new file is played. 
Empty filename will stop current file and then return.


modules/debug_cmd/debug_cmd.c
```diff
@@ -84,19 +84,28 @@ static int cmd_ua_debug(struct re_printf *pf, void *unused)
...
-		return err;
+	int err = 0;
 
-	err = play_file(NULL, baresip_player(), filename, 0);
-	if (err) {
-		warning("debug_cmd: play_file(%s) failed (%m)\n",
-			filename, err);
-		return err;
+	/* Stop the current tone, if any */
+	g_play = mem_deref(g_play);
```
Encamy: You are derefing a variable that was not allocated. Is this is a correct behaviour?

modules/debug_cmd/debug_cmd.c
```diff
@@ -84,19 +84,28 @@ static int cmd_ua_debug(struct re_printf *pf, void *unused)
...
+
+	if (str_isset(filename))
```
Encamy: If `filename` will not be set, this block would not change value of `err`. So you are going to return 0. But this means that the function which purpose was to play the file did not perform its action. What about returning some error code?
(It is just an idea)

alfredh:
many thanks @Ola-Palm 

I will merge this to master after the Baresip version 0.5.12 is released.

alfredh:
thanks @Ola-Palm !

I decided to merge it into the 0.6.0 release

PR #581, opened 2018-11-22
cmd_quit: ua_stop_all call moved to main thread

nicotyze:
This is a patch to fix a segmentation fault which occurs when quit command is pushed from httpd UI.

I reproduced the bug (seg fault) from the two environments:

- Mac OS (10.13.6)
- Ubuntu (16.04)

### DIFF
--- a/src/baresip.c
+++ b/src/baresip.c
@@ -18,6 +18,7 @@ static struct baresip {
 	struct commands *commands;
 	struct player *player;
 	struct message *message;
+	struct mqueue *mq;
 	struct list mnatl;
 	struct list mencl;
 	struct list aucodecl;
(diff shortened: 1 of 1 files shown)


src/baresip.c
```diff
@@ -32,15 +32,27 @@ static struct baresip {
...
 	int err;
 
 	(void)unused;
 
+	tmr_init(&tmr);
+
 	err = re_hprintf(pf, "Quit\n");
 
-	ua_stop_all(false);
+	tmr_start(&tmr, 100, stop_handler, (void*)&forced);
```
alfredh: you also need to call `tmr_cancel` in a close function.

if the timer is started but it does not fire, the timer object will "leak".

alfredh:
do you intend to make this thread-safe ?

calling tmr_start must be done from the main thread, so this patch will not
make it thread safe, it will just make the execution asyncronous.

if you want to make it thread-safe I suggest you use the `mqueue` instead
of timer, because the mqueue input can come from any thread.

nicotyze:
@alfredh , many thanks for the review.
 As suggested, I've tried to implement it with mqueue.

alfredh:
hi,

I have made an alternative patch, could you please try this:

https://github.com/alfredh/baresip/tree/module_app_unload


I think we should unload all modules from main, after all the
shutdown sequence is completely done. this will avoid the segfault
if 'q' is called from httpd.

nicotyze:
Hi,
Thanks for this alternative patch. 
I have tested from Ubuntu, OSX and it solves the segfault issue

alfredh:
thanks for the PR and testing.

This commit fixes the issue:

https://github.com/alfredh/baresip/commit/78b66009b536a73a38097a833e86218b9d5c3c58

closing this PR.

PR #582, opened 2018-11-23
opus

jmillan:


modules/opus/encode.c
```diff
@@ -100,8 +100,7 @@ int opus_encode_update(struct auenc_state **aesp, const struct aucodec *ac,
 		aes->ch = ac->ch;
 
 		aes->enc = opus_encoder_create(ac->srate, ac->ch,
-					       /* this has big impact on cpu */
-					       OPUS_APPLICATION_AUDIO,
+					       OPUS_APPLICATION_VOIP,
```
sreimers: The value should then also be configurable imho. I prefer better quality over cpu usage in my project.

jmillan: I don't mind creating such a config param.

jmillan: > The value should then also be configurable imho. I prefer better quality over cpu usage in my project.

I've added such parameter which defaults to `OPUS_APPLICATION_AUDIO`

alfredh:
thank you @jmillan 

should we include this in baresip 0.6.0 ?

jmillan:
> should we include this in baresip 0.6.0 ?

Sure, 

Thanks!

PR #583, opened 2018-11-23
make 'audio' and 'stream' independent of 'call'

jmillan:
With these few changes baresip `audio` and `stream` do not depend on `call`

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -365,7 +365,7 @@ int stream_alloc(struct stream **sp, const struct stream_param *prm,
 	struct stream *s;
 	int err;
 
-	if (!sp || !prm || !cfg || !call || !rtph)
+	if (!sp || !prm || !cfg || !rtph)
 		return EINVAL;
 
 	s = mem_zalloc(sizeof(*s), stream_destructor);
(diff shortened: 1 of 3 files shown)


src/stream.c
```diff
@@ -383,7 +383,7 @@ int stream_alloc(struct stream **sp, const struct stream_param *prm,
 	s->rtcp  = s->cfg.rtcp_enable;
 
 	if (prm->use_rtp) {
-		err = stream_sock_alloc(s, call_af(call));
+		err = stream_sock_alloc(s, call ? call_af(call) : AF_INET);
```
jmillan: I know this is limited to use `AF_INET` family when `call` is not used, yet I find it a good trade-off.

alfredh:
thanks @jmillan this is a good improvement.

the best would be to make `stream` and `audio` and `video` compile-time
independent of `call`.

jmillan:
> I guess that this also applies to video_alloc().

Indeed.

jmillan:
> the best would be to make stream and audio and video compile-time
independent of call.

One may want to use both kind of streams :-)

alfredh:
hi @jmillan 

I added address family and cname to the struct stream_param so that they
can be transported directly from call to stream:

```c
/** Common parameters for media stream */
struct stream_param {
	bool use_rtp;
	int af;
	const char *cname;
};
```

can you please rebase your PR on top of master ?

some of the changes in your PR are no longer necessary. perhaps the only
change you need to keep is the removal of `if (!call)`.


Alfred

alfredh:
hi @jmillan 

I made an alternative solution. Could you please test
commit 6d752df5bd6630a19ef82ae22cb6668123c21579
and see if it works for you ?

jmillan:
Hi Alfred,

I'm on vacation right now, almost out of Internet.

I'll check it once I'm back.

Thanks

jmillan:
It's working perfectly @alfredh.

Nice solution!

PR #587, opened 2018-11-26
sdl2: make window resizable

nicotyze:
Not a lot, but It would be nice to have resizable window with SDL2.

Tested with win 10, Ubuntu 16.04 and OSX 10.13.6


alfredh:
Thanks!

PR #589, opened 2018-12-03
Module app unload

alfredh:


alfredh:
This PR fixes two things:

1. move call to module_app_unload to main, it is not related to ua.c

2. fix a crash in case of 'q' from httpd module.


in general, unloading modules should only be done when the application
is completely finished with the shutdown sequence, and the re_main loop
has stopped.

closes https://github.com/alfredh/baresip/pull/581

juha-h:
Alfred E. Heggestad writes:

> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/alfredh/baresip/pull/589
> 
> -- Commit Summary --
> 
>   * unload application modules after exit

This affects also code in

https://github.com/alfredh/baresip/wiki/Using-baresip-as-a-library

Can I go ahead and add module_app_unload(); there:
```
out:
	if (err)
		ua_stop_all(true);

	ua_close();
    module_app_unload();
	conf_close();

	baresip_close();
```

alfredh:
Yeah, thank you very much :)

PR #598, opened 2018-12-08
added audio_codec api function

juha-h:


src/audio.c
```diff
@@ -2360,3 +2360,18 @@ void audio_set_hold(struct audio *au, bool hold)
...
 }
+
+
+/**
+ * Get audio codec of audio stream
+ *
+ * @param au    Audio object
+ * @param tx    True to get transmit codec, false to get receive codec
+ */
+const struct aucodec *audio_codec(const struct audio *au, bool tx)
```
alfredh: can you also add a line with " @return " that describes what is returned ?

alfredh:
thanks, looks fine except 1 small comment.

PR #614, opened 2018-12-23
avcodec: force intra (H.263, H.264) frame request if no key frame received

nicotyze:
This patch adds a check in decode_h263 and decode_h264 in order to return an error while the decoding process has not been initiated with a key frame (~intra frame).
This error is returned in video_stream_decode (video.c), triggering a RTCP FIR.

This aims at avoiding infinite loop with no key frame received/processed (=> no decoded video)


alfredh:
thank you.

I was having the same problem here and trying to fix it.
Your fix has also solved it for me :)

PR #617, opened 2018-12-25
webrtc aec

alfredh:


juha-h:
Alfred E. Heggestad writes:

> -- Commit Summary --
> 
>   * webrtc aec

Thanks for the code.  Could you provide a pointer to the source of
webrtc_full lib?  I could try to add it to baresip-android.  Perhaps
someone has already done the NDK build?

alfredh:
I used a pre-compiled library on Linux:

https://sourcey.com/precompiled-webrtc-libraries/

juha-h:
Alfred E. Heggestad writes:

> I used a pre-compiled library on Linux:
> 
> https://sourcey.com/precompiled-webrtc-libraries/

OK, looks like it is not solution for Android and perhaps not even for
Linux if size of the application grows by 50mb:

  The WebRTC repository is a monster weighing in at over 10gb with a
  myriad of dependencies that no developer without a fat paycheck from
  Google should ever have to mess with. However, once the excess fat has
  been trimmed it’s possible to get the archived build size down to
  about 50mb including all static libraries and headers. Jenny Craig
  would be proud!

-- Juha

alfredh:
baresip will link statically to libwebrtc.a and the linker will remove unused symbols.

* 3.4M	baresip [default]
* 9.8M	baresip [with webrtc_aec]

so the difference is around 6.4Mbytes

(compiled with:)

```
$ make STATIC=yes EXTRA_MODULES=webrtc_aec
```

juha-h:
I found github project that can be used to generate libwebrtc for Android that only includes aec stuff:
```
https://github.com/thepacific/webrtc-android-jni
```
This lib is used by pjsip project and is based on a rather old version of webrtc sources.

I built the lib and got webrtc_aec module loaded by baresip. I haven't tested it yet (waiting for solution to this Linux related issue https://github.com/alfredh/baresip/issues/712 first).

PR #627, opened 2019-01-06
menu: create_ua doesn't use dialbuf

tverlaan:
I think I actually copied this one as template. Shouldn't use `dialbuf` then as well.


alfredh:
well spotted! thank you!

PR #628, opened 2019-01-06
menu: sndcode should signal release of key

tverlaan:
You currently can't issue DTMF via httpd or ctrl_tcp. With this patch you can.

Without this patch:
```
/sndcode 123
audio: send DTMF digit: '1'
audio: send DTMF digit: '2'
audio: send DTMF digit: '3'
audio: send DTMF digit end: '3'
```

With this patch:
```
/sndcode 123
audio: send DTMF digit: '1'
audio: send DTMF digit end: '1'
audio: send DTMF digit: '2'
audio: send DTMF digit end: '2'
audio: send DTMF digit: '3'
audio: send DTMF digit end: '3'
```

The "short" command that you can also issue over httpd will still not work. But changing that will also touch code that currently works in the console. Since "sndcode" is a good solution for httpd anyway I don't think that's a problem.

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -898,6 +898,9 @@ static int send_code(struct re_printf *pf, void *arg)
 	if (call) {
 		for (i = 0; i < str_len(carg->prm) && !err; i++) {
 			err = call_send_digit(call, carg->prm[i]);
+			if (!err) {
+				err = call_send_digit(call, KEYCODE_REL);
+			}
 		}
 	}
 


alfredh:
sending a DTMF tone should end the previous tone.

for example sending '2' should end '1'.

tverlaan:
My main use case is actually to use this command via ctrl_tcp and it didn't work for me like you describe. I actually issued the command multiple times, so not exactly as described above.

However, what if I want to send a single DTMF? How to end it via ctrl_tcp? Also, should the ctrl_tcp module expose specific DTMF characteristics?

Perhaps a different approach is better, but I don't know which approach atm.

alfredh:
the best thing is to use both KEY_DOWN and KEY_UP events.
send dtmf start on KEY_DOWN, and then dtmf end on KEY_UP.

but if KEY_UP is not available, you can use a timeout of e.g. 100ms

1. when sending a digit, start the timer (100ms).
2. when sending a new digit, re-start the timer.
3. if the timer fires, send dtmf end.

tverlaan:
I'm controlling baresip via the ctrl_tcp API so I don't have KEY_DOWN or KEY_UP events. I'll give an example of how I expect to use `sndcode`:
`echo -n '34:{"command":"sndcode","params":"2"},' | nc 127.0.0.1 4444`

But in baresip I only see:
`audio: send DTMF digit: '2'`
There is no entry with *end*.

When I do it twice in a row I get:
```
audio: send DTMF digit: '2'
audio: send DTMF digit: '2'
```
No *end* here as well

I can definitely implement a timer, but I'm not sure if and why that would be beneficial as opposed to sending `end` straight away. When you issue `sndcode` as per my example twice with the same digit, I would expect that two digits will be send. Not that the second one possibly cancels sending `end` of the first one. Also a timeout on the `end` that can not be canceled also doesn't seem to have any benefit for the `sndcode` command.

I also see that currently each *input* module (except remote inputs like httpd and ctrl_tcp) is using a timer, so each *input* module is repeating the exact same functionality if you ask me. If we want to add the timer to ctrl_tcp as well, should we try to make a generic timer that can be called by each *input* module as to avoid duplicating the same timer everywhere?

In short two questions;
- Are we really sure we want `sndcode` to work with timer?
- If we want a timer, should we try to make 1 timer solution which can be used by other modules?

PR #637, opened 2019-01-17
Adding g729

edantie:
I'm trying to add g729 codec to baresip.
It's using bcg729 from http://www.linphone.org/technical-corner/bcg729/overview
Testing it and decode is working, but I don know why, encoding doesn't work.


Encamy:
bcg729 are developed under GPL v2.0 licence, baresip is using BSD licence
So if project uses some GPL-based code, it ought to be distributed under GPL too
IMHO, this implementation of g729 is unacceptable

alfredh:
hi @smorlat

would it be possible for us to use libbcg729 with a BSD license ?

Alfred

edantie:
I've tested it against an asterisk pbx.
I've seen pjsip (GPL) and freeswitch also (MPL 1.1) have integrated G.729 and thought is was convenient for  baresip to have. I hadn't seen the licensing problem. But you are right. Let's hope Linphone permit using their software. 
PS, I've based this version from mod_g729 of freeswitch.

alfredh:
@edantie did you get the encoder working ?

I have asked Belledonne Communications for an exception of the GPL,
so we can distribute it with BSD.

edantie:
Great news from Belledonne Communications, but I'm still stucked with encoding: it's doing metallic sounds. But I don't know why. Decoding is working nicely. I've also tested it also calling a grandstream phone, but I've obtained the same effect. Maybe someone can find out why it isn't working...

tomek-o:
> I've seen pjsip (GPL) and freeswitch also (MPL 1.1) have integrated G.729 and thought is was convenient for baresip to have. I hadn't seen the licensing problem.

PJSIP is supporting G.729 with IPP (Intel Performance Primitives) library. It seems to be licensed under https://software.intel.com/en-us/license/intel-simplified-software-license , I haven't tried and I'm not sure if it is distributed as source or compiled libraries.
FreeSWITCH has mod_com_g729 (commercial) and also unofficial, third party module based on ITU code. This "third party" solution might work here - end user would need to download G.729 code on its own and build it, possibly as shared library.

edantie:
> PJSIP is supporting G.729 with IPP (Intel Performance Primitives) library. It seems to be licensed under https://software.intel.com/en-us/license/intel-simplified-software-license , I haven't tried and I'm not sure if it is distributed as source or compiled libraries.

The source included with PJSIP is using bcg729 as well:
http://svn.pjsip.org/repos/pjproject/trunk/pjmedia/src/pjmedia-codec/bcg729.c

> FreeSWITCH has mod_com_g729 (commercial) and also unofficial, third party module based on ITU code.
You're right about it.

> This "third party" solution might work here - end user would need to download G.729 code on its own and build it, possibly as shared library.

If you don't have the bcg729 library installed this code is skipped.

edantie:
I've been doing more tests with no success.
I don't understand why the encoding is always called with 160 bytes encoded as 10 bytes, when decoding is called with 20 bytes decoded in 320 bytes.

Encamy:
Hello @edantie 

> If you don't have the bcg729 library installed this code is skipped.

What about if external project (non-GPL) is using baresip as library? Then this project won't be ever able to use this module.
As for me, I found this addition very useful and it would be nice if it won't have GPL license

alfredh:
Hi, I got the following answer from Belladonne Communications:

```
Hi Alfred,

Thank you for your interest in our products and services.

bcg729 is only available under GPLv2 license or proprietary license. We never distribute it under the BSD license, which is quite different from the GPL model.
Alternatively, you could have a specific version of your product integrating bcg729 distributed under GPLv2.

Best regards,

Elisa Nectoux
Sales & Marketing
```

Therefore we are not able to add this module, as it uses GPL code.

For now I would suggest that you can maintain the code/module in your own tree.

tomek-o:
@alfredh 
> Therefore we are not able to add this module, as it uses GPL code.

Technically this wouldn't be the first module depending on copyleft-licensed code. For one, libzrtp is using either AGPLv3 or commercial/private license with exception for FreeSWITCH. I don't see this as a big problem, I'm just omitting this module if I want to keep BSD license.

@edantie
> I don't understand why the encoding is always called with 160 bytes encoded as 10 bytes, when decoding is called with 20 bytes decoded in 320 bytes.

G.729 is internally working with 10ms frames. Most devices use 20ms framing (aka ptime) as default, so they are sending two G.729 frames in one RTP packet. Other direction is the same - you need to encode two frames at a time.
Alternatively (at least for testing) you might set ptime in baresip account configuration to 10ms, although from my experience not all devices/PABXs handle 10ms G.729 correctly.

alfredh:
The best thing would be to find a G.729 library with a liberal license compatible with BSD,
that is also a technically good implementation.


in any case, this patch can be maintained outside the baresip master tree for now.

PR #642, opened 2019-01-23
Improve mqtt module

oej:
The current mqtt module use the same client ID for all logins, making it impossible to have multiple baresip instances connected to the same broker. It also makes it impossible to separate instances when sending commands or reading events.

This patch adds configuration options for client ID, username, password as well as a base topic making it possible to control a cluster of Baresip instances on the same broker.

When configured, the base topic is used for both events and commands.

```
mqtt: connecting to broker at sollentuna.example.com:1883 as baresip01 topic baresip/olle01
mqtt: Publishing on /baresip/olle01/event, subscribing to /baresip/olle01/command/+
```

Tested with mosquitto. Original defaults are the same.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -162,3 +162,11 @@ ice_mode		full	# {full,lite}
 
 # sndfile #
 snd_path 		/tmp/
+
+# MQTT
+#mqtt_broker_host        sollentuna.example.com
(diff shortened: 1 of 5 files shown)


oej:
Fixed format to pass all checks.

alfredh:
thank you @oej !

PR #644, opened 2019-01-27
audio: calculate maximum samples from srate/channels

alfredh:
this PR calculates the maximum samples from samplerate, channels and max ptime.

@tonylangley could you please test it ?

it is a solution for:

https://github.com/alfredh/baresip/issues/625

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -53,11 +53,7 @@
  */
 
 enum {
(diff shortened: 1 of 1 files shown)


alfredh:
after some thinking, the patch has some problems that must be fixed first.
mainly related to the resampler.

PR #649, opened 2019-02-02
Added opus_multistream module

tonylangley:
Added opus_multistream module which has been tested to support 4 channels using ALSA, PortAudio and JACK. This should theoretically support up to 254 channels.

### DIFF
--- a/modules/opus_multistream/opus_multistream.c
+++ b/modules/opus_multistream/opus_multistream.c
@@ -1,0 +1,209 @@
+/**
+ * @file opus_multistream.c Opus Multistream Audio Codec
+ *
+ * Copyright (C) 2019 Creytiv.com
+ */
+
+#include <re.h>
+#include <baresip.h>
+#include <opus/opus_multistream.h>
+#include "opus_multistream.h"
+
+
+/**
+ * @defgroup opus opus
+ *
+ * The OPUS multistream audio codec
+ *
+ * Supported version: libopus 1.0.0 or later
+ *
+ * Configuration options:
+ *
+ \verbatim
+  opus_stereo        yes     # Request peer to send stereo
+  opus_sprop_stereo  yes     # Sending stereo
+  opus_bitrate    128000     # Average bitrate in [bps]
+  opus_cbr        {yes,no}   # Constant Bitrate (inverse of VBR)
+  opus_inbandfec  {yes,no}   # Enable inband Forward Error Correction (FEC)
+  opus_dtx        {yes,no}   # Enable Discontinuous Transmission (DTX)
+  opus_complexity {0-10}     # Encoder's computational complexity (10 max)
+  opus_application {audio, voip} # Encoder's intended application
+ \endverbatim
+ *
+ * References:
+ *
+ *    RFC 6716  Definition of the Opus Audio Codec
+ *    RFC 7587  RTP Payload Format for the Opus Speech and Audio Codec
+ *
+ *    http://opus-codec.org/downloads/
+ */
+
+
+static bool opus_mirror;
+static char fmtp[256] = "";
+static char fmtp_mirror[256];
+
+uint32_t opus_complexity = 10;
+opus_int32 opus_application = OPUS_APPLICATION_AUDIO;
+
+uint32_t opus_ms_channels = 2;
+uint32_t opus_ms_streams = 2;
+uint32_t opus_ms_c_streams = 2;
+
+static int opus_multistream_fmtp_enc(struct mbuf *mb, const struct sdp_format *fmt,
+			 bool offer, void *arg)
+{
+	bool mirror;
+
+	(void)arg;
+	(void)offer;
+
+	if (!mb || !fmt)
+		return 0;
+
+	mirror = !offer && str_isset(fmtp_mirror);
+
+	return mbuf_printf(mb, "a=fmtp:%s %s\r\n",
+			   fmt->id, mirror ? fmtp_mirror : fmtp);
+}
+
+
+static struct aucodec opus_multistream = {
+	.name      = "opus_multistream",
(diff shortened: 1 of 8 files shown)


tonylangley:
@alfredh this is my first pull request so I hop I've done it right!!

Do you need me to resolve the errors in the Travis CI build? They mostly seem to be related to line character lengths. Is this actually still an issue?

alfredh:
@tonylangley many thanks for this

I will start the review and testing

@Ola-Palm and @juha-h this is perhaps also interesting for you :)

alfredh:
do you have an example of how this codec can be used ?
has it been tested with other SIP clients?

I did a quick test with 2 channels, seems to work.

modules/opus_multistream/opus_multistream.c
```diff
@@ -0,0 +1,209 @@
...
+
+	mirror = !offer && str_isset(fmtp_mirror);
+
+	return mbuf_printf(mb, "a=fmtp:%s %s\r\n",
+			   fmt->id, mirror ? fmtp_mirror : fmtp);
+}
+
+
+static struct aucodec opus_multistream = {
+	.name      = "opus_multistream",
```
alfredh: is this the correct codec name to use ? this string will be used in the SDP.

modules/opus_multistream/opus_multistream.c
```diff
@@ -0,0 +1,209 @@
...
+	.srate     = 48000,
+	.crate     = 48000,
+	.ch        = 2,
```
alfredh: this field says that the codec has 2 channels. I am wondering how this can work with 4 channels ..

modules/opus_multistream/opus_multistream.c
tonylangley: If your question is in terms of forming the `a=rtpmap ...` string, I am unable find any reference of how to use the multistream codec over RTP.

4 channels of opus_multistream encoded audio just works for me between 2 hosts running the same custom baresip code without doing anything special. Is this an "accidental" hack rather than the SDP working "properly"?

modules/opus_multistream/opus_multistream.c
tonylangley: The `.ch` parameter of the` opus_multistream` struct gets updated from the default value ("2") in line 183 to the the `opus_ms_channels` value in the config file (e.g. "4").

tonylangley:
> do you have an example of how this codec can be used ?

Using a multi-channel sound card with 4+ inputs and outputs connected via JACK, ALSA or PortAudio
```
# Opus codec parameters
opus_bitrate		128000 # 6000-510000
opus_stereo		yes
opus_sprop_stereo	yes
#opus_cbr		no
#opus_inband_fec        yes
#opus_dtx		no
#opus_mirror		no
opus_complexity		10
opus_application	audio	# {voip,audio}

# Opus Multistream codec parameters
opus_ms_channels    4
opus_ms_streams     2
opus_ms_c_streams   2
```
Calling another baresip host with the same setup will result in 4 channels of bi-directional audio

> has it been tested with other SIP clients?

I haven't

modules/opus_multistream/opus_multistream.c
alfredh: I think the opus payload format RFC (RFC 7587) covers both standard opus
and multistream opus.

so the codec name should probably be just "opus".

the SDP just has one line for opus, even if different sampling rates or channels is used:

`opus/48000/2`

I also think it should be possible to send multistream to a regular opus decoder, which
will resample and mix channels to match the decoders capabilities.
Perhaps we should check this with the Opus folks.

modules/opus_multistream/opus_multistream.c
alfredh: aha, I got it :)

alfredh:
@tonylangley this is really great work :)

I need to think a bit more about how we should handle the channels.

Couple of more comments for now:

1. the ccheck warnings must be fixed. you can use this tool:

2. some symbols needs to be renamed to avoid clash with
  existing opus module. try a static build where all modules are statically
  linked into the baresip binary:

```
$ make STATIC=yes

  LD      baresip
duplicate symbol _opus_complexity in:
    build-x86_64/modules/opus/opus.o
    build-x86_64/modules/opus_multistream/opus_multistream.o
duplicate symbol _opus_application in:
    build-x86_64/modules/opus/opus.o
    build-x86_64/modules/opus_multistream/opus_multistream.o
ld: 2 duplicate symbols for architecture x86_64
make: *** [baresip] Error 1

```

alfredh:
@tonylangley I added some extra checks in travis in master

you can get this in your branch by merging from master:

```
$ git fetch upstream
$ git merge upstream/master
$ git push
```

alfredh:
@tonylangley do you want to fix the ccheck warnings ?

tonylangley:
> @tonylangley do you want to fix the ccheck warnings ?

@alfredh I should get a chance in the next couple of weeks to look at this. I'm keen to progress this PR, just need some free time to give it attention.

alfredh:
I had a look at ffmpeg how they use the libopus api.

it looks like they only use the opus multistream api:

https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/libopusenc.c#L348


for baresip it is fine to use both apis I guess.

alfredh:
@tonylangley 

I took your patch and massaged the code a bit. The result was merged to master
in commit 07028f9976b0f263059fa058abb6337261091e17

the code is still experimental but I think it is okay to keep it in master now.

we can still do small/big changes on it, just send me more Pull Requests if you
want to change something :)

Thanks for your contributions !

Alfred

mk6032:
@alfredh  How do you specify payload type? For example in ffmpeg -c:a libopus -payload_type 105 ?

PR #650, opened 2019-02-03
Added missing mem_ref on menu.message

trampster:
message_listen constructs a message_lsnr using mem_zalloc (ref count 1)

Then it is added to the message->lsnrl list. 

When this list is cleaned up using list_flush (in re) then mem_deref is called on le->data (ref count 0)

Then in menu.c module_close mem_deref is called on menu.message (now we are accessing memory after free)

menu.c needs to mem-ref it's menu.message because the reference added by mem_zelloc is consumed by the list it is added to.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -409,6 +409,7 @@ struct message_lsnr;
 int  message_init(struct message **messagep);
 int  message_listen(struct message_lsnr **lsnrp, struct message *message,
 		    message_recv_h *h, void *arg);
+void message_stop_listening(struct message_lsnr *lsnrp);
 int  message_send(struct ua *ua, const char *peer, const char *msg,
 		  sip_resp_h *resph, void *arg);
 
--- a/modules/gtk/gtk_mod.c
+++ b/modules/gtk/gtk_mod.c
@@ -1050,6 +1050,7 @@ static int module_close(void)
 		pthread_join(mod_obj.thread, NULL);
 	mod_obj.mq = mem_deref(mod_obj.mq);
 	aufilt_unregister(&vumeter);
+	message_stop_listening(mod_obj.message);
 	mod_obj.message = mem_deref(mod_obj.message);
 
 #ifdef USE_LIBNOTIFY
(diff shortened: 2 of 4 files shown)


alfredh:
@trampster thanks for the PR

I am not sure if the patch is needed. The `struct message` container is supposed
to be destroyed last, i.e. after all listeners are removed.

```c
baresip_init:

	err = message_init(&baresip.message);
	if (err) {
		warning("baresip: message init failed: %m\n", err);
		return err;
	}

menu.so module_init:

	err = message_listen(&menu.message, baresip_message(),
			     message_handler, NULL);
	if (err)
		return err;

menu.so module_close:

	menu.message = mem_deref(menu.message);


baresip_close:

	baresip.message = mem_deref(baresip.message);

```

are you using the `module_app` directive to load `menu.so` ?

if you want to allow the two objects to be destroyed in random order,
another alternative could be to add a `message_unlisten` function,
keep the listener object hidden inside message.c and use the callback handler
function as the input argument.

trampster:
The order doesn't matter. There are two mem_derefs on something that has only one ref.

If we change the order of the mem_deref then we don't avoid the problem we just have the some problem in occurring is a different order, the second mem_deref still causes a use after free. This still corrupts the heap. If you have two mem_derefs then you need two refs, there is only one.

It's not that we are actually using using it after free it's that the second mem_deref itself is a uses after free. This was found using valgrind because we had a crash on close in malloc because the heap was corrupt.

trampster:
OK I think I know why the order matters, The listener_destructor in message.c removes it from the list. That's a side effect I didn't expect.

We are using the baresip config to load the module, I'll need to have a look at the our code to determine why we are not unloading the modules first when I'm back a  work tomorrow. It's a public holiday here today.

alfredh:
if you want to change this, the new solution should be more robust so that
any closing order is possible.

this api is perhaps more robust, you can look at it for inspiration:

```
int  uag_event_register(ua_event_h *eh, void *arg);
void uag_event_unregister(ua_event_h *eh);
```

alfredh:
@trampster are you planning to update your PR ?

trampster:
Yes sorry, am working on it now.

alfredh:
@trampster please have a look at commit bf5ab14d19851c9e30e5d18defbc2836ac790e60

I have solved the problem in a different way ..

PR #655, opened 2019-02-07
Update MSVS project

Encamy:
Hello @alfredh 
In this commit 18c79180628c90889bd1610742a2e21bd366d818 file `realtime.c` was removed. This patch will remove it from MS Visual Studio solution


alfredh:
Thanks again :)

PR #662, opened 2019-02-20
vidloop: update vidsrc

nicotyze:
This patch makes possible to switch between video sources without having to restart the loop.

Tested on Ubuntu16.04 and Win10.

### DIFF
--- a/modules/vidloop/vidloop.c
+++ b/modules/vidloop/vidloop.c
@@ -636,11 +636,13 @@ static void update_vidsrc(void *arg)
 	struct config *cfg = conf_config();
 	int err;
 
-	tmr_start(&vl->tmr_update_src, 10, update_vidsrc, vl);
+	tmr_start(&vl->tmr_update_src, 100, update_vidsrc, vl);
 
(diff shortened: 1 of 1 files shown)


modules/vidloop/vidloop.c
```diff
@@ -627,6 +629,29 @@ static int vsrc_reopen(struct video_loop *vl, const struct vidsz *sz)
...
 
 
+static void update_vidsrc(void *arg)
+{
+	struct video_loop *vl = arg;
+	struct vidsz size;
+	struct config *cfg = conf_config();
+	int err;
+
+	tmr_start(&vl->tmr_update_src, 10, update_vidsrc, vl);
```
alfredh: I think the interval should be a bit higher, for example 50 or 100 ms

alfredh:
I tested this on OSX but it is not working.

Tried switching from avcapture to cairo using the /vidsrc command.
The change is detected, but the video source is still stuck at avcapture.

nicotyze:
@alfredh , thanks for the review.
Indeed, a check on the video module was missing... 
This is solved in my last commit.

alfredh:
@nicotyze thanks!

PR #664, opened 2019-02-28
Command not found returns error

weili-jiang:
If a command does not exist, the response should be a failure, not "ok", for clients only looking at success/failure and not the content.


alfredh:
@weili-jiang thanks !

PR #674, opened 2019-03-21
Added debug cmd to print UUID

AndreasHansson73:
Added a command that prints the UUID

@alfredh What do you think about this?


alfredh:
great, thanks !

PR #680, opened 2019-03-31
menu: add uadel to delete a uac by aor

tverlaan:
This addition allows to delete ua's from within baresip. I removed `sip_close` as it would otherwise stop Baresip on deleting the last ua. I believe `sip_close` is not necessary here as it is called from `ua_stop_all` already which is called when you quit Baresip.

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -407,6 +407,31 @@ static int cmd_ua_find(struct re_printf *pf, void *arg)
 }
 
 
+static int cmd_ua_delete(struct re_printf *pf, void *arg)
+{
+	const struct cmd_arg *carg = arg;
+	struct ua *ua = NULL;
+
+	if (str_isset(carg->prm)) {
+		ua = uag_find_aor(carg->prm);
+	}
+
+	if (!ua) {
+		return ENOENT;
+	}
+
+	if (ua == uag_cur()) {
+		(void)cmd_ua_next(pf, NULL);
+	}
+
+	(void)re_hprintf(pf, "deleting ua: %s\n", carg->prm);
+	mem_deref(ua);
+	(void)ua_print_reg_status(pf, NULL);
+
+	return 0;
+}
+
+
 static int print_commands(struct re_printf *pf, void *unused)
 {
 	(void)unused;
@@ -460,6 +485,7 @@ static const struct cmd cmdv[] = {
 {NULL,        KEYCODE_ESC,0, "Hangup call",             cmd_hangup           },
 {"uanext",    'T',        0, "Toggle UAs",              cmd_ua_next          },
 {"uanew",     0,    CMD_PRM, "Create User-Agent",       create_ua            },
+{"uadel",     0,    CMD_PRM, "Delete User-Agent",       cmd_ua_delete        },
 {"uafind",    0,    CMD_PRM, "Find User-Agent <aor>",   cmd_ua_find          },
 {"ausrc",     0,    CMD_PRM, "Switch audio source",     switch_audio_source  },
 {"auplay",    0,    CMD_PRM, "Switch audio player",     switch_audio_player  },


tverlaan:
I see the tests now fail. Not sure how to solve deleting the last User Agent without quitting Baresip.. Any ideas?

modules/menu/menu.c
```diff
@@ -407,6 +407,31 @@ static int cmd_ua_find(struct re_printf *pf, void *arg)
...
+	}
+
+	(void)re_hprintf(pf, "deleting ua: %s\n", carg->prm);
+	mem_deref(ua);
+	(void)ua_print_reg_status(pf, NULL);
+
+	return 0;
+}
+
+
```
alfredh: this block looks fine. keep in mind that if the last UA is deleted, the current ua pointer
will point to an invalid object.

src/ua.c
```diff
@@ -533,10 +533,6 @@ static void ua_destructor(void *arg)
 	mem_deref(ua->pub_gruu);
 	mem_deref(ua->acc);
 
-	if (list_isempty(&uag.ual)) {
-		sip_close(uag.sip, false);
-	}
-
```
alfredh: the logic here is that if the last UA is deleted, it should also start the closing
of the SIP stack (call to `sip_close`)...

I guess in your usecase this is probably not wanted ? I guess you want to add and delete
UAs many times ?

alfredh:
Thanks for your PR! see some comments inline ..

alfredh:
the shutdown code of UA/SIP-stack is perhaps a bit hard to understand
and a bit too complex.

the call to sip_close from ua_destructor is perhaps not necessary,
we need to test it properly that it does not have any side-effects.

this patch should make your selftest pass:

```diff
diff --git a/test/ua.c b/test/ua.c
index a1c50da..ae134b4 100644
--- a/test/ua.c
+++ b/test/ua.c
@@ -79,6 +79,8 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
                        t->srvv[i]->terminate = true;
 
                t->ua = mem_deref(t->ua);
+
+               ua_stop_all(false);
        }
        else if (ev == UA_EVENT_REGISTER_FAIL) {
 
```

modules/menu/menu.c
tverlaan: True. It would be better to set it to `NULL` in that case, however when the list is empty then `uag_current()` will return `NULL` so this shouldn't be a problem. It would require more logic to implement setting it to `NULL`.

src/ua.c
tverlaan: Yes, my idea is to have my application load UAs on startup and remove them on shutdown. Without removing baresip would just append only :-)

tverlaan:
Thanks! I applied the patch.

I was just trying to find out how and why sip_close would be called there, but couldn't find another reason then "just to be safe". On shutdown/quitting sip_close is already called. Removing the last UA shouldn't necessitate shutting down. You can also start Baresip without any UAs in a working state.

modules/menu/menu.c
tverlaan: What would you think about this:
```
diff --git a/modules/menu/menu.c b/modules/menu/menu.c
index 752ef62..dee80f8 100644
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -367,8 +367,10 @@ static int cmd_ua_next(struct re_printf *pf, void *unused)
 
        if (!menu.le_cur)
                menu.le_cur = list_head(uag_list());
-       if (!menu.le_cur)
+       if (!menu.le_cur) {
+               uag_current_set(NULL);
                return 0;
+       }
 
        menu.le_cur = menu.le_cur->next ?
                menu.le_cur->next : list_head(uag_list());
```

juha-h:
Alfred E. Heggestad writes:

> the call to sip_close from ua_destructor is perhaps not necessary,
> we need to test it properly that it does not have any side-effects.

It should be possible to call ua_destroy(ua) also for the last ua and
then call ua_alloc(uri) for a new ua without a need to do anything else.
Is it like this now?

alfredh:
this is the current code from `ua_destructor`:

```c
	if (list_isempty(&uag.ual)) {
		sip_close(uag.sip, false);
	}
```

i.e. if the last UA is destroyed, it will close the SIP stack.
if this is not the wanted behaviour, we can change that.

please help to test that the change does not have any side-effects.
we need to test various SIP modules like mwi, presence and
SIP register with regint=0 and regint=N

juha-h:
Alfred E. Heggestad writes:

> please help to test that the change does not have any side-effects.
> we need to test various SIP modules like mwi, presence and
> SIP register with regint=0 and regint=N

Could you commit the change to master or create a new branch for it.

alfredh:
one of the reasons for sip_close in ua_destructor, is the logic in the mwi module.
when the shutdown sequence starts, mwi receives a `UA_EVENT_SHUTDOWN` event
which will send un-subscribe and asyncronously deref the ua object. the un-REGISTER
will happen _after_ this.


a safer solution might be to add a `bool terminated` flag to ua.c

```c
void ua_stop_all(bool forced)
{
    uag.terminated = true;

   /* ... */
}
```

and then check this flag in the destructor:

```c
	if (uag.terminated && list_isempty(&uag.ual)) {
		sip_close(uag.sip, false);
	}
```

alfredh:
here is also one candidate for testing:

```diff
git diff
diff --git a/src/ua.c b/src/ua.c
index aecf9b8..266295e 100644
--- a/src/ua.c
+++ b/src/ua.c
@@ -533,10 +533,6 @@ static void ua_destructor(void *arg)
        mem_deref(ua->pub_gruu);
        mem_deref(ua->acc);
 
-       if (list_isempty(&uag.ual)) {
-               sip_close(uag.sip, false);
-       }
-
        list_flush(&ua->custom_hdrs);
        list_flush(&ua->hdr_filter);
 }
@@ -1587,7 +1583,6 @@ void ua_stop_all(bool forced)
 
        if (ext_ref) {
                info("ua: in use (%u) by app module\n", ext_ref);
-               return;
        }
 
        if (forced)
```

alfredh:
@tverlaan I fixed a bug in the testcode, you can rebase your patch if you want ..

juha-h:
Before testing the proposed alternatives, I tried what happens when I
try to delete a user agent in my baresip Android app by calling
mem_deref(ua).  The result was that nothing happened. No un-REGISTER or
mwi un-SUBSCIBE message was sent out.

What might be the reason for that? What is the correct way to delete a
user agent, i.e., do the reverse of ua_alloc()?

alfredh:
@juha-h this depends on the reference count for the ua object.

only if the reference count goes from 1 -> 0 will the destructor be called.
if there are more references, nothing will happen.

you can check this by pressing 'u' in the console:
(in my case where mwi.so is loaded)

```
 nrefs:     2
```

or printing out the number before calling `mem_deref`

```c
re_printf("ua nrefs = %u\n", mem_nrefs(ua));
mem_deref(ua);
```

alfredh:
@tverlaan I have added a delayed_close flag to ua.c to fix your issue

it should now be possible to add your uadel command without making
any other changes.

adding/deleting uas should also be possible without shutting down the sip stack.

juha-h:
Alfred E. Heggestad writes:

> only if the reference count goes from 1 -> 0 will the destructor be called.
> if there are more references, nothing will happen.
> 
> you can check this by pressing 'u' in the console:
> (in my case where mwi.so is loaded)
> 
> ```
>  nrefs:     2
> ```
> 
> or printing out the number before calling `mem_deref`
> 
> ```c
> re_printf("ua nrefs = %u\n", mem_nrefs(ua));
> mem_deref(ua);

I too got "ua nrefs = 2".  So how to first un-SUBSCRIBE mwi?

alfredh:
currently the only way to un-SUBSCRIBE mwi is to unload the mwi.so module ..

juha-h:
Alfred E. Heggestad writes:

> currently the only way to un-SUBSCRIBE mwi is to unload the mwi.so
> module ..

It should be possible to un-SUBSCRIBE only one UA, when it is destroyed
or when it un-REGISTERs and thus cannot anymore receive NOTIFYs.

tverlaan:
> @tverlaan I have added a delayed_close flag to ua.c to fix your issue

I updated the PR. Thanks a lot!

test/ua.c
```diff
@@ -79,6 +79,8 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 			t->srvv[i]->terminate = true;
 
 		t->ua = mem_deref(t->ua);
+
+		ua_stop_all(false);
```
alfredh: please also remove this chunk

tverlaan: done!

juha-h:
juha-h writes:

> > currently the only way to un-SUBSCRIBE mwi is to unload the mwi.so
> > module ..
> 
> It should be possible to un-SUBSCRIBE only one UA, when it is destroyed
> or when it un-REGISTERs and thus cannot anymore receive NOTIFYs.

The key here is un-REGISTER, which can happen with or without
destruction of the UA.  If the UA upon un-REGISTER has any subscriptions
or publishes, those should be undone before un-REGISTER takes place.

PR #692, opened 2019-04-19
added prefer_ipv6 config variable

juha-h:


alfredh:
I am planning to rewrite this part and make it more flexible.

the command line arguments should be similar to other apps:

```
    -4      Forces ssh to use IPv4 addresses only.

     -6      Forces ssh to use IPv6 addresses only.
```


and the config should be also more flexible, something like this:

```c
struct config_net {
    bool enable_ipv4;
    bool enable_ipv6;
};
```

juha-h:
FIne, but if the rewrite takes long time, could prefer_ipv6 be used in the meantime?

alfredh:
okay I will make the release first, and then merge it after that

juha-h:
If both ipv4 and ipv6 are enabled, can dns_server variable contain both ipv4 and ipv6 addresses of name servers?

alfredh:
currently dualstack is enabled by default. if the client has both IPv4 and IPv6 addresses
then both will be used, but without any specific priority.

the dns_server variable can contain both IPv4 and IPv6 addresses, but the
prefer_ipv6 flag will not be used here.

the prefer_ipv6 flag is currently mainly for the SIP stack and RTP.

in summary, the networking code is a bit messy and would need some re-design.

PR #702, opened 2019-04-26
Count ua references after UA_EVENT_SHUTDOWN in ua_destroy

### LINKED ISSUE #701: Quit command hangs if a module has references to the UA and mem_derefs them synchronously on UA_EVENT_SHUTDOWN
The recent change to add `ua_destroy` appears to have introduced a regression whereby the global command to close the application (e.g. `quit`) will hang if there is a module that is holding a reference and expecting the `UA_EVENT_SHUTDOWN` to trigger clean-up.

The issue is that counting of the number of external references is now done too early, prior to `UA_EVENT_SHUTDOWN` which other modules may be relying on to perform clean-up. Even if these modules get this event and clean up, it is too late as the counting as already happened.

weili-jiang:
Resolves #701

Count ua references after UA_EVENT_SHUTDOWN in case that triggers clean-up in other modules.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -717,7 +717,7 @@ int  ua_update_account(struct ua *ua);
 int  ua_register(struct ua *ua);
 void ua_unregister(struct ua *ua);
 bool ua_isregistered(const struct ua *ua);
-void ua_destroy(struct ua *ua);
+unsigned ua_destroy(struct ua *ua);
 void ua_pub_gruu_set(struct ua *ua, const struct pl *pval);
 const char     *ua_aor(const struct ua *ua);
 const char     *ua_cuser(const struct ua *ua);
--- a/src/ua.c
+++ b/src/ua.c
@@ -273,11 +273,15 @@ bool ua_isregistered(const struct ua *ua)
  * send the SHUTDOWN event.
  *
  * @param ua User-Agent object
+ *
+ * @return Number of remaining references
  */
-void ua_destroy(struct ua *ua)
+unsigned ua_destroy(struct ua *ua)
 {
+	unsigned nrefs;
+
 	if (!ua)
-		return;
+		return 0;
 
 	list_unlink(&ua->le);
 
@@ -287,7 +291,12 @@ void ua_destroy(struct ua *ua)
 	/* terminate all calls now */
 	list_flush(&ua->calls);
 
+	/* number of remaining references (excluding this one) */
+	nrefs = mem_nrefs(ua) - 1;
+
 	mem_deref(ua);
+
+	return nrefs;
 }
 
 
@@ -1591,11 +1600,9 @@ void ua_stop_all(bool forced)
 		struct ua *ua = le->data;
 		le = le->next;
 
-		if (mem_nrefs(ua) > 1) {
+		if (ua_destroy(ua) != 0) {
 			++ext_ref;
 		}
-
-		ua_destroy(ua);
 	}
 
 	if (ext_ref) {


alfredh:
there should not be a need for this change.

the change that was done in
commit 1406caf404a0d402210e86529feb0e8b6e0ac207
does not change the behaviour.

can you please include the full log of a session where your problem appears?

weili-jiang:
The problem occurs with my proprietary application module which holds a reference to the UA instance, so a relevant log is not easily available.

When the `quit` command is invoked, `baresip` simply does a `ua_stop_all`. My module removes all UA references upon receiving a  `UA_EVENT_SHUTDOWN` event (generated by `ua_stop_all`), thus allowing shutdown to proceed.

---

Referring to the diff in 1406caf:

Prior to the change, `UA_EVENT_SHUTDOWN` is first generated in `ua_stop_all`, allowing modules to remove their references before `mem_nrefs(ua)` is called to determine whether clean-up can proceed.

Afterwards, `ua_stop_all` has been changed such that `mem_nrefs(ua)` is done first, prior to `ua_destroy`. For application modules that hold references to the UA, this means that they will always contribute to the initial reference count, even if they remove all their references upon processing `UA_EVENT_SHUTDOWN`. This all means that another `quit` command is required to actually shutdown the application.

---

I notice that the `mwi` module also holds a reference and would hold up shutdown in the same manner as my module. That one seems even more problematic as it does not always immediately `mem_deref` its UA reference even upon receiving a `UA_EVENT_SHUTDOWN` as it may have further asynchronous actions to perform on shutdown.

With that module enabled and active, I cannot see how `baresip` can be shutdown without invoking `quit` repeatedly - it would be counted as holding a reference at the point of `quit` being invoked, causing `baresip` shutdown to be deferred. Upon `mwi` clean-up completing, it calls `mem_deref` on its UA reference, but there is nothing to tell `baresip` to continue application shutdown from there.

Have I misunderstood how such application modules are supposed to tie into shutdown mechanisms, particularly regarding the `quit` command? Is the idea that you just have to invoke the `quit` command repeatedly until everything has cleaned up to allow it to finally proceed? Please advise.

juha-h:
weili-jiang writes:

> I notice that the `mwi` module also holds a reference and would hold
> up shutdown in the same manner as my module. That one seems even more
> problematic as it does not always immediately `mem_deref` its UA
> reference even upon receiving a `UA_EVENT_SHUTDOWN` as it may have
> further asynchronous actions to perform on shutdown.

I have not noticed any shutdown problems with mwi module in my app.  At
quit, the app calls ua_stop_all(false) and sets five second timer.  If
it fires before baresip has died, the app calls ua_stop_all(true).  I
haven't seen the latter happening.

Below is an example of Quit:

```
4-28 19:55:04.036  7679  7679 D Baresip : Quiting
04-28 19:55:04.086  7679  7679 D Baresip Service: Received onStartCommand action Stop
04-28 19:55:04.090  7679  7679 D Baresip Lib: ua_stop_all upon baresipStop
04-28 19:55:04.090  7679  7847 D Baresip Lib: calling ua_stop_all with force 0
04-28 19:55:04.090  7679  7847 D Baresip Lib: ua event (SHUTDOWN) 
04-28 19:55:04.090  7679  7848 D Baresip Lib: ua: stop all (forced=0)
04-28 19:55:04.090  7679  7848 D Baresip Lib: mwi: shutdown of sip:foo@test.tutpro.com
04-28 19:55:04.091  7679  7847 D Baresip Lib: ua event (SHUTDOWN) 
04-28 19:55:04.092  7679  7848 D Baresip Lib: mwi: shutdown of sip:bar@test.tutpro.com
04-28 19:55:04.092  7679  7679 D Baresip : Seconds remaining: 4
04-28 19:55:04.092  7679  7848 D Baresip Lib: ua: in use (2) by app module
04-28 19:55:04.591  7679  7847 D Baresip Lib: ua event (UNREGISTERING) 
04-28 19:55:04.591  7679  7847 D Baresip Lib: sending ua/call 506863377936/0 event unregistering
04-28 19:55:04.592  7679  7847 D Baresip Service: uaEvent got event unregistering/506863377936/0
04-28 19:55:04.595  7679  7847 D Baresip Lib: ua event (UNREGISTERING) 
04-28 19:55:04.595  7679  7847 D Baresip Lib: sending ua/call 506863379536/0 event unregistering
04-28 19:55:04.595  7679  7847 D Baresip Service: uaEvent got event unregistering/506863379536/0
04-28 19:55:04.852  7679  7847 D Baresip Lib: ua event (EXIT) 
04-28 19:55:04.852  7679  7847 D Baresip Lib: ua exited -- stopping main runloop
04-28 19:55:04.852  7679  7847 D Baresip Lib: main loop exit
04-28 19:55:04.852  7679  7847 D Baresip Lib: closing ...
04-28 19:55:04.852  7679  7848 D Baresip Lib: ua: sip-stack exit
04-28 19:55:04.871  7679  7848 D Baresip Lib: module: unloading app mwi
04-28 19:55:04.875  7679  7848 D Baresip Lib: module: unloading app debug_cmd
04-28 19:55:04.875  7679  7848 D Baresip Lib: module: unloading app contact
04-28 19:55:04.875  7679  7847 D Baresip Lib: unloading modules ...
04-28 19:55:04.876  7679  7848 D Baresip Lib:  [   zrtp main]: DESTROYING LIBZRTP...
04-28 19:55:04.877  7679  7848 D Baresip Lib:  [   zrtp main]: DESTROYING LIBZRTP - DONE
04-28 19:55:04.877  7679  7847 D Baresip Lib: tell main that baresip has stopped
04-28 19:55:04.878  7679  7847 D Baresip Service: 'stopped' from baresip
04-28 19:55:04.884  7679  7679 D Baresip : Handling service event 'stopped'
```

weili-jiang:
Thanks @juha-h for your example. I think my problem is caused because my modified `main` function is no longer calling the right shutdown routines as per the default `baresip` `main`. I will update it to follow the `baresip` one.

weili-jiang:
On further analysis, this change is still required. See my update to 701.

alfredh:
@weili-jiang thanks for the patch and the detailed report.

it turns out that you are correct.

a module that deref ua syncronously on shutdown event, will need a fix like this.

PR #705, opened 2019-05-01
menu: default config params

f-squirrel:


alfredh:
thanks,

could you please rebase the commit on top of master HEAD ?

the diff contains also the previous PR ..

f-squirrel:
@alfredh , done

modules/menu/menu.c
```diff
@@ -1253,6 +1253,12 @@ static int module_init(void)
 	struct pl val;
 	int err;
 
+	menu.bell = true;
+	menu.redial_attempts = 3;
+	menu.redial_delay = 5;
+	menu.ringback_disabled = true;
```
alfredh: the default value for redial should be 0 and 0, i.e. disabled

ringback should be enabled by default (i.e. disabled=false).

f-squirrel: But it is different from defaults printed in src/config.c:
```
	(void)re_fprintf(f,
			"\n# Menu\n"
			"#menu_bell\t\tyes\n"
			"#redial_attempts\t3 # Num or <inf>\n"
			"#redial_delay\t\t5 # Delay in seconds\n"
			"#ringback_disabled\tyes\n"
			"#statmode_default\toff\n");
```
Are you sure you want to change it as well?

alfredh: the config in the example is commented out, meaning that it will not be read.

the default config should have sane values. for example the redial feature
should be switched off by default.

the config template shows how a user can enable redial feature, by
uncommenting the line.

f-squirrel: Updated together with the config template. I think config template has to reflect the default values for the optional parameters.

PR #707, opened 2019-05-06
call: reset streams on call hold

cspiel1:
The audio data in the jitter buffer has to be flushed when putting a call into
on hold state. Otherwise it is played on the playback device after the call is
resumed.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -1175,6 +1175,7 @@ static int sipsess_offer_handler(struct mbuf **descp,
 {
 	const bool got_offer = (0 != mbuf_get_left(msg->mb));
 	struct call *call = arg;
+	struct le *le;
 	int err;
 
 	MAGIC_CHECK(call);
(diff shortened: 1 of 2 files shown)


alfredh:
@cspiel1 thank you very much for this patch!

in addition to this, please also consider this:

1. if the RTP marker bit is set for incoming audio packets, the jitter buffer should
  be reset.

2. there is also a audio buffer (aubuf) for decoded audio frames, perhaps this buffer
  should be reset also:

https://github.com/alfredh/baresip/blob/master/src/audio.c#L1810

cspiel1:
Dear Alfred,
your welcome!
For 1. and 2. I will have a look later.
r, chris

cspiel1:
@alfredh: Ad 2. Why not call call_stream_stop (which calls audio_stop and video_stop) ?

PR #710, opened 2019-05-09
audio: increase buffer size for audio device string

nicotyze:
Hi,
This is a small patch to fix problems with long audio device names.
Tested with ubuntu 16.04


alfredh:
thank you !

PR #716, opened 2019-05-14
Bugfix/flush buffers on call hold

cspiel1:
This is a rework on pull request https://github.com/alfredh/baresip/pull/707 .

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -882,6 +882,11 @@ static void stream_recv_handler(const struct rtp_header *hdr,
 			return;
 	}
 
+	if (hdr->m) {
+		stream_reset(a->strm);
+		debug("audio: rtp marker bit set\n");
+	}
+
 	/* RFC 5285 -- A General Mechanism for RTP Header Extensions */
 	for (i=0; i<extc; i++) {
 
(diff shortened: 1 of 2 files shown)


src/stream.c
```diff
@@ -258,6 +258,11 @@ static void rtp_handler(const struct sa *src, const struct rtp_header *hdr,
 			     sdp_media_name(s->sdp), s->ssrc_rx, hdr->ssrc,
 			     mbuf_get_left(mb), src);
 		}
+		else if (hdr->m) {
+			flush = true;
+			info("stream: %s: marker bit set\n",
+				sdp_media_name(s->sdp));
+		}
```
alfredh: the marker bit has different semantics for audio/video stream.

audio: signal start of a new talk spurt
video: signal a complete video frame

I think checking RTP header marker bit should be done in audio.c instead.

cspiel1: Right! Now it looks better.

alfredh:
@cspiel1 thanks for helping out with these improvements :)

cspiel1:
you're welcome!
r,chris

PR #724, opened 2019-05-28
audio: handle marker bit in stream.c

alfredh:
when audio stream is receiving RTP with marker bit set, the following
should happen:

1. flush jitter buffer (re-ordering buffer)
2. flush aubuf for playback.

@cspiel1 could you please test this patch ?

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -882,11 +882,6 @@ static void stream_recv_handler(const struct rtp_header *hdr,
 			return;
 	}
 
-	if (hdr->m) {
-		stream_reset(a->strm);
-		debug("audio: rtp marker bit set\n");
-	}
-
 	/* RFC 5285 -- A General Mechanism for RTP Header Extensions */
 	for (i=0; i<extc; i++) {
 
--- a/src/stream.c
+++ b/src/stream.c
@@ -252,6 +252,10 @@ static void rtp_handler(const struct sa *src, const struct rtp_header *hdr,
 	if (!(sdp_media_ldir(s->sdp) & SDP_RECVONLY))
 		return;
 
+	/* The marker bit indicates the beginning of a talkspurt. */
+	if (hdr->m && s->type == MEDIA_AUDIO)
+		flush = true;
+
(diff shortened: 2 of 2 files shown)


alfredh:
@weili-jiang FYI.

weili-jiang:
Looks good.

The marker bit may be lost, however, on lossy wireless IP networks. A more complete solution would also look at computed time delta and/or sequence numbers, as implied by the following: https://www.cs.columbia.edu/~hgs/rtp/faq.html#marker

> The marker bit is a hint; the beginning of a talkspurt can also be computed by comparing the difference in timestamps and sequence numbers between two packets, assuming the timestamp clock rate is known. 

I think that if an incoming RTP packet has a timestamp or sequence number that puts it beyond the the acceptable reordering duration, it should also be treated as the beginning of a new talk burst and flush.

alfredh:
@weili-jiang you are right, we should also detect large jumps in sequence number
and flush in this case.


perhaps also the `aubuf` should be flushed in audio.c ?

```c
if (hdr->m) {
	aubuf_flush(rx->aubuf);
}
```

this buffer will contain decoded audio samples from the "previous" talkspurt.

weili-jiang:
> perhaps also the `aubuf` should be flushed in audio.c ?

@alfredh I do not think that flush is necessary - the remaining samples in the audio buffer are all valid, and should be drained out by the audio driver naturally as part of the end of the talk spurt anyway. When new RTP packets arrive as part of the new talk spurt, they can be decoded and put at the end of the audio buffer - they will  be played out naturally after the existing audio is drained.

Actually, the last of the RTP packets in the jitter buffer would also be valid, if they had been decoded into the audio buffer at the end of the talk spurt rather than being left in the jitter buffer, but that is a separate matter...

alfredh:
thanks @weili-jiang 

I will address the sequence number jump in a separate patch soon ..

weili-jiang:
> 
> 
> thanks @weili-jiang
> 
> I will address the sequence number jump in a separate patch soon ..

Great @alfredh. Don't forget about the Timestamp field also.

PR #749, opened 2019-07-03
menu: sndcode should signal release of key

tverlaan:
Coming back on PR #628. The DTMFs arrive, but the last sent digit really needs the `KEYCODE_REL` in order for the far end to receive the DTMF end marker. I changed this PR to the suggestion you made in that PR :-)


alfredh:
many thanks for the fix :)

PR #754, opened 2019-07-13
menc: add wait_secure flag

alfredh:
this patch adds a flag `bool wait_secure` to the mediaenc struct

the purpose is to control when the media-stream should start.
if the flag is false, the media stream will start when the SIP session
is established. if the flag is true, the media stream will start when
the menc is secure.

for example DTLS-SRTP has an async handshake after the SIP session
is established, so the flag should be set to true here.

here is a list of Mediaenc modules and the suggested flag value:

| Module | wait_secure |
| --- | --- |
| srtp | no |
| zrtp | no |
| dtls_srtp | yes |

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -604,6 +604,7 @@ struct menc {
 	struct le le;
 	const char *id;
 	const char *sdp_proto;
+	bool wait_secure;
 	menc_sess_h *sessh;
 	menc_media_h *mediah;
 };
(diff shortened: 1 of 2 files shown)


juha-h:
Alfred E. Heggestad writes:

> here is a list of Mediaenc modules and the suggested flag value:
> 
> | Module | wait_secure |
> | --- | --- |
> | srtp | no |
> | zrtp | no |
> | dtls_srtp | yes |

They look good to me.

PR #785, opened 2019-08-27
Use default PulseAudio device 

trampster:
Passing NULL to pa_simple_new as the device instructs PulseAudio to use the default. This is better then the current baresip behaviour which just uses the first device if no device was specified. In my case the first device was not the default.


alfredh:
thanks, your patch also fixes a potential NULL dereference
in case `mediadev_get_default` returns NULL.

@nicotyze can you please review and/or test ?

nicotyze:
@alfredh , yes I will have a look and test it  (beginning of next week...)

alfredh:
I merged the patch to master. 

@trampster many thanks !

@nicotyze please feel free to test it and report and issues.

PR #790, opened 2019-09-02
Add support for aes_256_gcm

trampster:
libre already had support, was just missing from baresip.

### DIFF
--- a/modules/srtp/srtp.c
+++ b/modules/srtp/srtp.c
@@ -34,8 +34,9 @@ struct menc_sess {
 struct menc_st {
 	/* one SRTP session per media line */
 	const struct menc_sess *sess;
-	uint8_t key_tx[32];
-	uint8_t key_rx[32];
+	uint8_t key_tx[32+12];
+	/* base64_decoding worst case encoded 32+12 key */
+	uint8_t key_rx[46];
 	struct srtp *srtp_tx, *srtp_rx;
 	bool use_srtp;
 	bool got_sdp;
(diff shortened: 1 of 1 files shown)


alfredh:
many thanks !

can you please tell me how this was tested ?

modules/srtp/srtp.c
```diff
@@ -34,8 +34,9 @@ struct menc_sess {
 struct menc_st {
 	/* one SRTP session per media line */
 	const struct menc_sess *sess;
-	uint8_t key_tx[32];
-	uint8_t key_rx[32];
+	uint8_t key_tx[32+12];
+	/* base64_decoding worst case encoded 32+12 key */
+	uint8_t key_rx[46];
```
alfredh: is the comment correct? 32+12=44 and not 46 ?

trampster: 32+12 is the size without base64 encoding, This is a result of the padding. we started with 44 and found out the hard way we needed 46.

trampster:
Testing was performed as follows:

- We have a number of baresip based clients which have the preferred_suite (line 48 opf srtp.c) set to aes_256_gcm
- These talk to a Kamailio SIP server
- Media all goes to a central Conference Controller which handles the fanout for the group) this uses libre (not baresip) directly to do the aes_256_gcm
- We also have a client which bridges to a LMR (Land Mobile Radio) network, this also uses libre
- We have been running this for about a week here with about 5 units without issue.

What we haven't done:
- Tested against a non baresip/libre client
- Run a field trial

alfredh:
Great work :)

PR #796, opened 2019-09-14
Account specific audio source and playback

alfredh:
Add support for specifying audio source and player per account.
The global config will be used by default.

Syntax:

```
;audio_source="module,device"
;audio_player="module,device"
```

Example with multiple accounts/devices:

```
<sip:user1@example.com>;audio_source=alsa,plughw:0,1
<sip:user2@example.com>;audio_source=alsa,plughw:0,2
<sip:user3@example.com>;audio_source=alsa,plughw:0,3
```

Please review and test.


premultiply:
Nice!
But will there also be account specific playback device added?

alfredh:
thanks for your feedback.

I added also audio_player per account, same syntax:

```
;audio_player="alsa,default"
```

premultiply:
This works fine now! Thank you!

PR #797, opened 2019-09-15
mpa: switch encoder to use lame

alfredh:
libmp3lame uses MPEG2 Layer III.

### DIFF
--- a/mk/modules.mk
+++ b/mk/modules.mk
@@ -152,9 +152,9 @@ HAVE_SPEEXDSP := \
 endif
 ifneq ($(USE_MPG123),)
 ifneq ($(HAVE_SPEEXDSP),)
-USE_MPA  := $(shell [ -f $(SYSROOT)/include/twolame.h ] || \
-	[ -f $(SYSROOT)/local/include/twolame.h ] || \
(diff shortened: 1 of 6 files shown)


hoene:
The problem with lame is that it only support layer 3 not layer 2 anymore.
Thus, actually both libraries are need to support MP2 and MP3.

premultiply:
Encoding MPA Layer III itself works fine now. Checked it with Orban Opticodec 7600 as counterpart.
But MPA Layer II encoding support has gone. mpa_layer setting in config does only set another fmtp layer value but does not switch encoding to the selected layer at all.
As @hoene has written both librarys are required to support Layer II and Layer III.

And there are major issues with the SIP signalisation.
The MPA fmtp line does not always strictly reflect the internal encoder settings as it should be.
When calling from baresip without all mpa_* settings set in config the fmtp string is incomplete even if internal default settings are used by the encoder.
When baresip is called with specific MPA settings in SIP INVITE in fmtp string it should adapt its own encoder settings to this properties and should reflect this with the same fmtp string in SIP OK in response to the SIP INVITE. And it should not read these properties from the config file when answering a call.

PR #804, opened 2019-09-18
ua: add support for SIP trace

alfredh:
this requires a special version of libre:

  https://github.com/alfredh/re

To run baresip with SIP trace enabled:

```
$ baresip -i
```

(the letter `i` is open to better ideas).

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -745,6 +745,7 @@ int  ua_init(const char *software, bool udp, bool tcp, bool tls);
 void ua_close(void);
 void ua_stop_all(bool forced);
 void uag_set_exit_handler(ua_exit_h *exith, void *arg);
+void uag_enable_sip_trace(bool enable);
 int  uag_reset_transp(bool reg, bool reinvite);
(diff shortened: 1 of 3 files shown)


alfredh:
sip-trace

juha-h:
Thanks, good feature for testing.  Why is special version of libre needed, i.e., why can't trace support be included in the standard libre?

alfredh:
do you have any opinion of the `i` letter ? it was reused from `curl`:

```
       -i, --include
              Include the HTTP response headers in the output. The HTTP response headers can
              include things like server name, cookies, date of the document,  HTTP  version
              and more...

```

we can use a different letter if you have better suggestions.

the feature is experimental for now, hence special version of libre.

juha-h:
How about -s (for sip trace)?

As long as this new feature does not affect SIP protocol operation of re, it could still be included there even if it experimental.  It does not really matter if printing of SIP messages to console has some defects.

alfredh:
thanks for your input, I changed it to -s and merged it to master.

we might include it in re master later ..

PR #822, opened 2019-10-15
Search amr include dir also from directory pointed by AMR_PATH environment variable

juha-h:


juha-h:
Only nb include dir is searched, since searching wb dir is pointless, because it does not include the required file enc_if.h.

PR #839, opened 2019-11-01
MPA layer 3 encoding fixes

premultiply:
Please note: The MPA module currently does layer 3 encoding only (using `libmp3lame`).
It tries to advertise this by SDP and apply constraints even to mirrored fmtp encoding.

There are not any constraints decoding other audio layers.

Encoding layer 2 (MP2) would require dynamic switching to `libtwolame` encoding which is not implemented (yet).


alfredh:
great, thanks a lot!

I tested this locally and it works fine.

PR #844, opened 2019-11-04
MPA L2 and L3 encoding

premultiply:
Fixes https://github.com/alfredh/baresip/issues/840

Maybe needs some love.
Working for me but please feel free to cleanup or improve.

### DIFF
--- a/mk/modules.mk
+++ b/mk/modules.mk
@@ -30,7 +30,7 @@
 #   USE_ISAC          iSAC audio codec
 #   USE_JACK          JACK Audio Connection Kit audio driver
 #   USE_L16           L16 audio codec
-#   USE_MPA           MPA audo codec
+#   USE_MPA           MPA audio codec
 #   USE_MPG123        Use mpg123
 #   USE_OMX_RPI       RaspberryPi VideoCore display driver
 #   USE_OMX_BELLAGIO  libomxil-bellagio xvideosink driver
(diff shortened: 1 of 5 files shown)


premultiply:
And please have a intense look at the encoder resource cleanup.
I am really not shure about if it will work fine when updating fmt.

alfredh:
many thanks for this!

building on OSX gives me an error:

```
  CC [M]  build-x86_64/modules/mpa/mpa.o
  CC [M]  build-x86_64/modules/mpa/decode.o
  CC [M]  build-x86_64/modules/mpa/sdp.o
modules/mpa/sdp.c:10:10: fatal error: 'twolame.h' file not found
#include <twolame.h>
         ^
1 error generated.
make: *** [build-x86_64/modules/mpa/sdp.o] Error 1
```

lame is installed, but not twolame.

also there are some `ccheck` warnings that must be addressed ..

mk/modules.mk
```diff
@@ -160,7 +160,10 @@ HAVE_SPEEXDSP := \
 endif
 ifneq ($(USE_MPG123),)
 ifneq ($(HAVE_SPEEXDSP),)
-USE_MPA  := $(shell [ -f $(SYSROOT)/include/lame/lame.h ] || \
+USE_MPA  := $(shell [ -f $(SYSROOT)/include/twolame.h ] || \
+	[ -f $(SYSROOT)/local/include/twolame.h ] || \
+	[ -f $(SYSROOT_ALT)/include/twolame.h ] || \
+	[ -f $(SYSROOT)/include/lame/lame.h ] || \
```
alfredh: I think this logic should be AND instead of OR.

perhaps add another test e.g. `HAVE_TWOLAME`

this modules depends on many libraries:

- libspeexdsp
- libmpg123
- libmp3lame
- libtwolame

premultiply: Very true.
Hopefully this is better...

premultiply:
Hopefully done.

PR #849, opened 2019-11-10
audio: add ptime to struct aucodec

alfredh:
- add optional ptime (in ms) packet time to each codec,
  which can be used to override the default

- update l16.so with some ptime values to keep the
  packet size below a common MTU (1500):

  48000/2 -> 5ms
  44100/2 -> 5ms
  32000/2 -> 10ms
  16000/2 -> 20ms
  ... etc ...

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -986,6 +986,7 @@ struct aucodec {
 	uint32_t crate;             /* RTP Clock rate   */
 	uint8_t ch;
 	uint8_t pch;                /* RTP packet channels */
+	uint32_t ptime;             /* Packet time in [ms] (optional) */
 	const char *fmtp;
 	auenc_update_h *encupdh;
 	auenc_encode_h *ench;
--- a/modules/l16/l16.c
+++ b/modules/l16/l16.c
@@ -71,16 +71,16 @@ static int decode(struct audec_state *st, int fmt, void *sampv, size_t *sampc,
 
 /* See RFC 3551 */
 static struct aucodec l16v[NR_CODECS] = {
-{LE_INIT,    0, "L16", 48000, 48000, 2, 2, 0, 0, encode, 0, decode, 0, 0, 0},
-{LE_INIT, "10", "L16", 44100, 44100, 2, 2, 0, 0, encode, 0, decode, 0, 0, 0},
-{LE_INIT,    0, "L16", 32000, 32000, 2, 2, 0, 0, encode, 0, decode, 0, 0, 0},
-{LE_INIT,    0, "L16", 16000, 16000, 2, 2, 0, 0, encode, 0, decode, 0, 0, 0},
-{LE_INIT,    0, "L16",  8000,  8000, 2, 2, 0, 0, encode, 0, decode, 0, 0, 0},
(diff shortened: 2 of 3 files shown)


alfredh:
anyone who can help to test the L16 codec against some legacy endpoints ?

premultiply:
Yes. I can do some tests tomorrow.

djhenley:
This is just great! Absolutely, will have the possibility to test this tomorrow as well.

djhenley:
We have now done tests against the following IP-audio codecs using **l16**.

*Prodys Quantum
*Prodys Pronotnet
*ME UMAC M2
*Luci Live

Result, works very well, ptime sets to 5ms (48000/2) instead of 20ms as it was before (default ptime) - perfect!!!

Notice: in the SDP (INVITE) Baresip is announcing (a)ptime=20 as a general ptime value (if not set in the account file of course)

All the above IP-audio codecs do ignore this (incorrect) value and set their ptime values correctly calculated internally.

As we know the only "correct" way in an SDP INVITE containing several codec formats would be to set corresponding plength value per format as an ebuacip paramter (eg a=ebuacip:plength). However, we are not sure how many IP audio codecs actually supports this in practice.

Very well done Alfred, we think this is a major leap forward and finally we can add PCM along with OPUS with Baresip in our production environment! // Jim & Ola

alfredh:
many thanks for testing.

we have this code in `audio.c`:

```c
	/* This is probably only meaningful for audio data, but
	   may be used with other media types if it makes sense. */
	attr = sdp_media_rattr(stream_sdpmedia(a->strm), "ptime");
	if (attr) {
		struct autx *tx = &a->tx;
		uint32_t ptime_tx = atoi(attr);

		if (ptime_tx && ptime_tx != a->tx.ptime
		    && ptime_tx <= MAX_PTIME) {

			info("audio: peer changed ptime_tx %ums -> %ums\n",
			     a->tx.ptime, ptime_tx);

			tx->ptime = ptime_tx;

			if (tx->ac) {
				size_t sz;

				sz = aufmt_sample_size(tx->src_fmt);

				tx->psize = sz * calc_nsamp(tx->ac->srate,
							    tx->ac->ch,
							    ptime_tx);
			}

			sdp_media_set_lattr(stream_sdpmedia(a->strm), true,
					    "ptime", "%u", ptime_tx);
		}
	}
```

the code will adjust the sending ptime, to the ptime value of the incoming SDP.
this was useful a long time ago, but it is not so useful anymore.
perhaps we should remove the code, and just ignore the ptime SDP attribute.

Ola-Palm:
Hi Alfred. 
I think that we still benefits from this code. 
If the peer suggests a ptime other than we had planed for, we are better off If we follow. As long as we keep it below a value that would mean fragmented packages.
So from our point of view, we are in favor for keeping this section.

premultiply:
I had little time the last two days but I did quick tests with

Orban Opticodec 7600
AVT Magic ACip3

16/48000/2 RX Audio was fine on both.

Will try the Luci iOS App hopefully tomorrow.

premultiply:
Luci on iOS ("ARD mupro App") is also playing correctly.
But baresip repeatly claims about
`stream: unknown ext type ignored (0x4c53)`

PR #854, opened 2019-11-20
aptX audio codec module

premultiply:
libopenaptx
https://github.com/pali/libopenaptx/

Supports aptX and aptX HD variants

### DIFF
--- a/modules/aptx/aptx.c
+++ b/modules/aptx/aptx.c
@@ -29,29 +29,31 @@
  */
 
 static struct aucodec aptx = {
-	.name = "aptx",
-	.srate = APTX_SRATE,
-	.crate = APTX_SRATE,
-	.ch = APTX_CHANNELS,
-	.pch = APTX_CHANNELS,
-	.ptime = 4,
-	.encupdh = aptx_encode_update,
-	.ench = aptx_encode_frm,
-	.decupdh = aptx_decode_update,
-	.dech = aptx_decode_frm,
-	.fmtp_ench = aptx_fmtp_enc,
-	.fmtp_cmph = aptx_fmtp_cmp,
+    .name = "aptx",
+    .srate = APTX_SRATE,
+    .crate = APTX_SRATE,
+    .ch = APTX_CHANNELS,
+    .pch = APTX_CHANNELS,
+    .ptime = 4,
+    .encupdh = aptx_encode_update,
+    .ench = aptx_encode_frm,
+    .decupdh = aptx_decode_update,
+    .dech = aptx_decode_frm,
+    .fmtp_ench = aptx_fmtp_enc,
+    .fmtp_cmph = aptx_fmtp_cmp,
 };
 
 
-static int module_init(void) {
+static int module_init(void)
+{
 	aucodec_register(baresip_aucodecl(), &aptx);
 
 	return 0;
 }
 
 
-static int module_close(void) {
+static int module_close(void)
+{
 	aucodec_unregister(&aptx);
 
 	return 0;
@@ -59,8 +61,8 @@ static int module_close(void) {
 
 
 EXPORT_SYM const struct mod_export DECL_EXPORTS(aptx) = {
-	"aptx",
-	"audio codec",
-	module_init,
-	module_close,
+    "aptx",
+    "audio codec",
+    module_init,
+    module_close,
 };
--- a/modules/aptx/aptx.c
+++ b/modules/aptx/aptx.c
@@ -4,9 +4,9 @@
  * Copyright (C) 2019 Hessischer Rundfunk
  */
 
-#include "aptx.h"
-#include <baresip.h>
 #include <re.h>
+#include <baresip.h>
+#include "aptx.h"
 
 /**
  * @defgroup aptx aptx
--- a/modules/aptx/aptx.c
+++ b/modules/aptx/aptx.c
@@ -29,18 +29,18 @@
  */
 
 static struct aucodec aptx = {
-    .name = "aptx",
-    .srate = APTX_SRATE,
-    .crate = APTX_SRATE,
-    .ch = APTX_CHANNELS,
-    .pch = APTX_CHANNELS,
-    .ptime = 4,
-    .encupdh = aptx_encode_update,
-    .ench = aptx_encode_frm,
-    .decupdh = aptx_decode_update,
-    .dech = aptx_decode_frm,
-    .fmtp_ench = aptx_fmtp_enc,
-    .fmtp_cmph = aptx_fmtp_cmp,
+	.name = "aptx",
+	.srate = APTX_SRATE,
+	.crate = APTX_SRATE,
+	.ch = APTX_CHANNELS,
+	.pch = APTX_CHANNELS,
+	.ptime = 4,
+	.encupdh = aptx_encode_update,
+	.ench = aptx_encode_frm,
+	.decupdh = aptx_decode_update,
+	.dech = aptx_decode_frm,
+	.fmtp_ench = aptx_fmtp_enc,
+	.fmtp_cmph = aptx_fmtp_cmp,
 };
 
 
@@ -61,8 +61,8 @@ static int module_close(void)
 
 
 EXPORT_SYM const struct mod_export DECL_EXPORTS(aptx) = {
-    "aptx",
-    "audio codec",
-    module_init,
-    module_close,
+	"aptx",
+	"audio codec",
+	module_init,
+	module_close,
 };
--- a/modules/aptx/aptx.c
+++ b/modules/aptx/aptx.c
@@ -1,0 +1,66 @@
+/**
+ * @file aptx.c aptX Audio Codec
+ *
+ * Copyright (C) 2019 Hessischer Rundfunk
+ */
+
+#include <re.h>
+#include <baresip.h>
+#include "aptx.h"
+
+/**
+ * @defgroup aptx aptx
+ *
+ * aptX audio codec (Standard and HD variant)
+ *
+ * Supported version:
+ *      libopenaptx 0.1.0 or later
+ *
+ * References:
+ *
+ *    RFC 7310  RTP Payload Format for Standard apt-X and Enhanced apt-X Codecs
+ *
+ * 
+ * TODOs:
+ *    - Code check, cleanup and error handling
+ *    - Add SDP fmt negotiation & config preconfiguration
+ *    - Check and implement other sampling rates and channel modes
+ *    - Add real 24 bit I/O support
+ *    
+ */
+
+static struct aucodec aptx = {
+	.name = "aptx",
+	.srate = APTX_SRATE,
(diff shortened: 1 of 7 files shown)


modules/aptx/aptx.c
```diff
@@ -0,0 +1,66 @@
...
+	.fmtp_cmph = aptx_fmtp_cmp,
+};
+
+static int module_init(void)
+{
+	aucodec_register(baresip_aucodecl(), &aptx);
+
+	return 0;
+}
+
```
alfredh: please use 2 empty lines between functions.

modules/aptx/aptx.h
```diff
@@ -0,0 +1,31 @@
+/**
+ * @file aptx.h aptX Audio Codec -- internal interface
+ *
+ * Copyright (C) 2019 Hessischer Rundfunk
+ */
+
+#define APTX_VARIANT 0
+
+enum
+{
```
alfredh: the first curly brace should be on the previous line, example:

```
enum {
};
```

modules/aptx/decode.c
```diff
@@ -0,0 +1,93 @@
...
+
+#include <re.h>
+#include <rem.h>
+#include <baresip.h>
+#include <openaptx.h>
+#include <stdlib.h>
+#include "aptx.h"
+
+struct audec_state
+{
```
alfredh: move curly brace to previous line

modules/aptx/decode.c
```diff
@@ -0,0 +1,93 @@
...
+		return 0;
+
+	ads = mem_zalloc(sizeof(*ads), destructor);
+	if (!ads)
+		return ENOMEM;
+
+	ads->dec = aptx_init(APTX_VARIANT);
+	if (!ads->dec)
+	{
+		warning("Cannot initialize aptX decoder.\n");
```
alfredh: perhaps add module name as prefix, i.e. `aptx: cannot ...`

modules/aptx/decode.c
```diff
@@ -0,0 +1,93 @@
...
+		warning("aptX decoding stopped in the middle of the sample, dropped %u bytes\n", (unsigned int)(len - processed));
+
+	if (*sampc > 0)
+	{
+		/* remap S24 to S16 in same buffer */
+		for (size_t s = 0; s < *sampc; s++)
+		{
+			s16[s * 2]     = s16[(s * APTX_WORDSIZE) + 1];
+			s16[s * 2 + 1] = s16[(s * APTX_WORDSIZE) + 2];
+		}
```
alfredh: is the 24-bit format host-endian or little-endian ?

will this code work on big-endian machines ?

we already supports S24_LE and FLOAT, perhaps we should also add those formats here?
(maybe later).

converting from S24 -> S16 means loss of information.

alfredh:
hi @premultiply 

this is a really nice work, many thanks. I am happy to merge it to master
after some minor things are fixed.

Is it also okay to add your name as a maintainer of this module:

https://github.com/alfredh/baresip/wiki/Modules

modules/aptx/aptx.c
premultiply: OK, thanks for your input!
This was Visual Studio Code autoformat.
Is there any standard named codestyle preset or plugin I should set the autoformater to?

Anyway I will reformat it as required of course.

modules/aptx/decode.c
premultiply: libopenaptx documentation says 24-bit little-endian only.
I am not even shure if libopenaptx does work on big-endian machines at all...?

It is only the buffer interface which is always 24-bit wide but most of the time the data is only 16-bit. AptX Standard is 16-bitso the LSB is all zero and this is memory remapping only.
Only aptX HD has support for 24-bit where currently 16-bit/8-bit input will be padded or cropped.

Anyway if it should support native 24_LE I would need some assistance how to implement it.

Maybe L24 module may be also added then ;-)

premultiply:
> this is a really nice work, many thanks. I am happy to merge it to master
> after some minor things are fixed.

Thank you very much for your feedback and this nice project!
I will do the requested modifications during the next days.


> Is it also okay to add your name as a maintainer of this module:
> 
> https://github.com/alfredh/baresip/wiki/Modules

I am fine with that.

alfredh:
I tested this locally on OSX with libopenaptx and it is working.

```
audio           Transmit:     Receive:
packets:           1465         1454
avg. bitrate:     384.1        383.8  (kbit/s)
errors:               0            0
pkt.report:        1258         1247
lost:                 0            0
jitter:             3.0          2.6  (ms)
```

alfredh:
this looks much better.

I will have a look after 0.6.5 is released.

modules/aptx/encode.c
```diff
@@ -0,0 +1,108 @@
...
+	size_t intermediate_len;
+
+	if (!aes || !buf || !len || !sampv)
+		return EINVAL;
+
+	if (fmt != AUFMT_S16LE)
+		return ENOTSUP;
+
+	intermediate_len = sampc * APTX_WORDSIZE;
+	intermediate_buf = (uint8_t *)malloc(intermediate_len);
```
alfredh: please use `mem_alloc` instead of malloc.

also the casting is not necessary.

modules/aptx/encode.c
```diff
@@ -0,0 +1,108 @@
...
+	for (size_t s = 0; s < sampc; s++) {
+		intermediate_buf[s * APTX_WORDSIZE] = 0;
+		intermediate_buf[s * APTX_WORDSIZE + 1] = s16[s * 2];
+		intermediate_buf[s * APTX_WORDSIZE + 2] = s16[s * 2 + 1];
+	}
+
+	processed = aptx_encode(aes->enc, intermediate_buf, intermediate_len,
+	                        buf, *len, &written);
+
+	free(intermediate_buf);
```
alfredh: use `mem_deref` here

alfredh:
here is an example of a decode function that supports
both S16 and S24_LE:

```c
int aptx_decode_frm(struct audec_state *ads, int fmt, void *sampv,
                    size_t *sampc, const uint8_t *buf, size_t len)
{
        size_t processed = 0;
        size_t written = 0;
        uint8_t *s16 = sampv;

        if (!ads || !sampv || !sampc || !buf)
                return EINVAL;

        processed = aptx_decode(ads->dec, buf, len, s16, *sampc, &written);

        *sampc = written / APTX_WORDSIZE;

        if (written == 0)
                return 0;

        if (processed != len)
                warning("aptx: Decoding stopped in the middle of the sample, "
                        "dropped %u bytes\n",
                        (unsigned int)(len - processed));

        switch (fmt) {

        case AUFMT_S16LE:
                /* remap S24 to S16 in same buffer */
                for (size_t s = 0; s < *sampc; s++) {
                        s16[s * 2] = s16[(s * APTX_WORDSIZE) + 1];
                        s16[s * 2 + 1] = s16[(s * APTX_WORDSIZE) + 2];
                }
                break;

        case AUFMT_S24_3LE:
                break;

        default:
                return ENOTSUP;
        }

        return 0;
}
```

the advantage of using S24 is the large bit-depth, and also that
there is no need to process the samples one extra time.

pali:
Hello! I'm afraid but code in this pull request does not work.

My libopenaptx library provides support only for **Bluetooth aptX** and **Bluetooth aptX HD** codecs. Library does not support **Standard aptX** nor **Enhanced aptX** codecs for which is defined RTP payload in RFC 7310.

Yes, these codecs are different. **Standard aptX** seems that use different quantization tables as are used in **Bluetooth aptX**. I have **Standard aptX** samples and when they are decoded by libopenaptx as **Bluetooth aptX** variant I hear base sound, but with lot of noise.

So if you are targetting RFC 7310 support, I would suggest you to disable this aptX code as it produce different format, not what is expected by RFC 7310 receiver.

premultiply:
This works fine - at least from baresip to baresip.

Are you able to extend your code that it supports what you call "Standard aptX" or even "Enhanced aptX"?

There is another feature that some aptX HW devices use: "autosync" on/off.
This also disturbes the audio when it is set wrong.

pali:
> This works fine - at least from baresip to baresip.

Of course. Both sides uses **Bluetooth aptX**. But it is not compatible with other implementations of RTP/SDP which expects **Standard aptX** as defined in **RFC 7310**.

I would really suggest to not use RFC 7310 definitions of **Standard aptX** for transmission of **Bluetooth aptX**. Just use different RTP/SDP parameters to not confise **Standard aptX** receiver.

> Are you able to extend your code that it supports what you call "Standard aptX" or even "Enhanced aptX"?

I have not figured out which coefficients and quantization values these codecs use.

> There is another feature that some aptX HW devices use: "autosync" on/off.
> This also disturbes the audio when it is set wrong.

No aptX _autosync_ feature is there to detect if some bytes in aptX stream were dropped or modified. It is very primitive _checksum_ which allows aptX receiver/decoder to find out where is the next (encoded) aptX word. It is _not_ there to disturbes the audio when it is set wrong.

Basically all aptX family codecs has ability to _include_ these autosync checksums into aptX stream. Parameters (as defined in RFC 7310) then specify if this autosync feature is used in stream or not.

**Bluetooth aptX** and **Bluetooth aptX HD** codecs always put these _autosync_ marks into aptX stream. New version of libopenaptx 0.2.0 has improved decoder support to use them for dropping incomplete words and prevents decoding noise.

If you really want to use **Bluetooth aptX** or **Bluetooth aptX HD** codecs for SIP/RTP streaming then switch to new decoding function introduced in libopenaptx 0.2.0 which takes care about those _autosync_ marks.

pali:
> Are you able to extend your code that it supports what you call "Standard aptX" or even "Enhanced aptX"?

If you want to hack with it, I can give you "Standard aptX" and "Enhanced aptX" encoded audio samples and you can try to figure out how to decode it. I also have their correct decoded form so can compare...

premultiply:
Oh we have these hardware boxes here which talks "Enhanced aptX" only. So I never had a chance to compare libopenaptx aptX with "Standard aptX" but I could provide you even with any live signal you want. But my own knowlege of aptX bitstreams is near zero.

pali:
"Standard aptX" is supported by Trial/demo version of [dcsTools](http://www.dcstools.com/Products/Prod_dcsTools_AudioImporter.html) software which supports wav to aptX and aptX to wav file conversion. Internally this software uses oficial aptX implementation. So via this tool one can verify that custom aptX encoder/decoder is compatible with world. This I already used to verify that "Standard aptX" is a different then "Bluetooth aptX".

I have just some "Enhanced aptX" stream (plus their decoded forms) which I was not able to decode. And I have not found any software which can generate "Enhanced aptX" audio. And I do not know if that stream was correctly captured (so I do not know if I thave correct input for decoding).

If you have SW or HW tools which can convert arbitrary audio wav file to "Enhanced aptX" then it would be great for codec inspection and I could try to look at "Enhanced aptX" again.

PR #863, opened 2019-12-02
aptx: 24-bit audio i/o support (S24_3LE)

premultiply:


alfredh:
there are some ccheck warnings that needs to be fixed:

./modules/aptx/encode.c:95: line is too wide (83 - max 79)
1580

premultiply:
Yes I saw it but the funny thing is that it seems that only the tab/space-calculation of the checker is wrong.

alfredh:
I tested the patch and it works well, tested both S16 and S24_3LE sample format.

alfredh:
I am ready to merge the patch, if you can fix the ccheck warnings.

The ccheck.py tool assumes that a TAB is 8 spaces.

premultiply:
Isn't it a bit too much?
Normal TAB-spacing is 4 spaces.

PR #871, opened 2019-12-11
mk: allow sndio also on non openbsd systems

Johnnynator:


alfredh:
Thank you!

I tested that it compiles on Debian 9.

PR #873, opened 2019-12-14
add config for audio receive buffer size

alfredh:
The audio receive/decode path has an audio buffer with de-jitter.

This patch makes the size run-time configurable:

`audio_buffer            20-100          # ms`

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -1541,19 +1541,17 @@ static int start_player(struct aurx *rx, struct audio *a)
 		prm.fmt        = rx->play_fmt;
 
 		if (!rx->aubuf) {
-			const size_t ssz = aufmt_sample_size(rx->play_fmt);
-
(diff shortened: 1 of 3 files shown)


src/audio.c
```diff
@@ -1541,17 +1541,33 @@ static int start_player(struct aurx *rx, struct audio *a)
...
+			size_t min_sz;
+			size_t max_sz;
 
-			err = aubuf_alloc(&rx->aubuf, psize * 1,
-					  rx->aubuf_maxsz);
-			if (err)
+			if (!ptime_min || !ptime_max)
+				return EINVAL;
+
+			min_sz = ssz*calc_nsamp(prm.srate, prm.ch, ptime_min);
```
sreimers: Maybe `min_ssz` and `max_ssz` would be more consistent to `ssz` (sample_size). Or `sz` as before.

PR #875, opened 2019-12-16
change message log level of unknown RTP extension

premultiply:
Stop spamming to console and logs when unknown RTP extension type is used.

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -194,7 +194,7 @@ static void handle_rtp(struct stream *s, const struct rtp_header *hdr,
 		int err;
 
 		if (hdr->x.type != RTPEXT_TYPE_MAGIC) {
-			info("stream: unknown ext type ignored (0x%04x)\n",
+			debug("stream: unknown ext type ignored (0x%04x)\n",
 			     hdr->x.type);
 			goto handler;
 		}


alfredh:
I would be very interested to see what kind of RTP extension you are receiving.
Do you have a log or network trace ?

in any case, I suggest that we remove the logline.

premultiply:
Oh it is just the software from https://www.technicadelarte.com/ (Luci Live, Luci Studio, ARD muPro App, ...) which is just a kind of buggy.
They just set a wrong extension type.

`stream: unknown ext type ignored (0x4c53)`

src/stream.c
```diff
@@ -194,7 +194,7 @@ static void handle_rtp(struct stream *s, const struct rtp_header *hdr,
 		int err;
 
 		if (hdr->x.type != RTPEXT_TYPE_MAGIC) {
-			info("stream: unknown ext type ignored (0x%04x)\n",
+			debug("stream: unknown ext type ignored (0x%04x)\n",
```
alfredh: I suggest that we remove the log-line instead.

premultiply: That would be also fine for me.

PR #877, opened 2019-12-17
Add support for DTMF audio generation

hselasky:


alfredh:
many thanks for this patch :)

I have some comments:

1. what is your usecase for using in-band DTMF generation?
 is it possible to use out-of-band Telephone event instead?
(Out-of-band DTMF is more reliable)

2. the DTMF sending mode should be configurable, and/or
  negotiated per call. For example Telephone event is offered by default.

3. We should consider to make this code into a module
  (for example dtmfgen.so). The audio filter API has access to the audio samples.
  We could add a callback for outgoing DTMF events.

4. is it possible to use the DTMF generator in librem instead ? (autone_dtmf)

5. How has this code been tested ?


Alfred

hselasky:
> many thanks for this patch :)
> 
> I have some comments:
> 
>     1. what is your usecase for using in-band DTMF generation?
>        is it possible to use out-of-band Telephone event instead?
>        (Out-of-band DTMF is more reliable)

I use baresip with a GSM gateway and the DTMF inband by the GSM module is broken. I've frequently had to use linphone for telephone calls, because baresip cannot generate DTMF audio events.
> 
>     2. the DTMF sending mode should be configurable, and/or
>        negotiated per call. For example Telephone event is offered by default.

Can you guide me how I can do this?

> 
>     3. We should consider to make this code into a module
>        (for example dtmfgen.so). The audio filter API has access to the audio samples.
>        We could add a callback for outgoing DTMF events.

I have no idea how to do this. Can you help?
> 
>     4. is it possible to use the DTMF generator in librem instead ? (autone_dtmf)

I don't know. It is not just about DTMF generation, but also about keeping a queue of DTMF events. I've never used autone_dtmf() before, so I don't know.

> 
>     5. How has this code been tested ?

Built baresip, made a call and pressed some keys and verified Asterisk was able to detect those DTMF digits.

alfredh:
we should add a config option so that people can choose which DTMF method
to use.

example:

```c
enum dtmf_method {
    DTMF_METHOD_NONE,
    DTMF_METHOD_TELEV,
    DTMF_METHOD_INBAND
};
```

this method could be added to global config or per account.

for each call/session we should probably also negotiate this with the peer,
during the SDP offer/answer exchange.
for example if remote SDP contains `telephone-event/8000`
the local dtmf method is set to DTMF_METHOD_TELEV.

I am happy to help with this part.

hselasky:
Feel free to update my patch.

alfredh:
I tested the DTMF generator against the DTMF detector in librem (rem_dtmf.h)
The digits 1-6 are able to detect properly, but digits 7-9 are detected twice.

I recorded the outgoing audio using sndfile.so and it looks like
the DTMF tone is sent twice, one with normal amplitude and one
with small amplitude.

Here is a screenshot from Audacity:

Picture: 
dtmf

alfredh:
could you please also fix these ccheck errors:

```
./src/dtmf.c:146: else: ending if bracket should be on previous line
1609./src/dtmf.c:151: else: ending if bracket should be on previous line
1610./src/audio.c:75: line is too wide (85 - max 79)
1611./src/audio.c:555: line is too wide (80 - max 79)
1612./src/audio.c:1956: line is too wide (83 - max 79)
1613./src/audio.c:1958: line is too wide (86 - max 79)
```

the if/else should be written like this:

```
	if (pg->input_pos == pg->output_pos) {
		return (EINVAL);
	}
	else if (pg->duration == 0) {
```

hselasky:
> I tested the DTMF generator against the DTMF detector in librem (rem_dtmf.h)
> The digits 1-6 are able to detect properly, but digits 7-9 are detected twice.
> 
> I recorded the outgoing audio using sndfile.so and it looks like
> the DTMF tone is sent twice, one with normal amplitude and one
> with small amplitude.
> 
> Here is a screenshot from Audacity:
> 
> Picture:
> dtmf

That must be a bug. Maybe it is echo from the other end. Did you verify that?

hselasky:
> /src/dtmf.c:146: else: ending if bracket should be on previous line
> 1609./src/dtmf.c:151: else: ending if bracket should be on previous line

Should be fixed now.

modules/gst/gst.c
```diff
@@ -4,7 +4,9 @@
  * Copyright (C) 2010 - 2015 Creytiv.com
  */
 #define _DEFAULT_SOURCE 1
+#ifndef __FreeBSD__
 #define _POSIX_C_SOURCE 199309L
+#endif
```
alfredh: can you please make a separate PR for this section ?

src/audio.c
```diff
@@ -24,6 +24,7 @@
 #define MAGIC 0x000a0d10
 #include "magic.h"
 
+#include "dtmf.h"
```
alfredh: should not be needed, the API should be defined in core.h

src/audio.c
```diff
@@ -544,6 +546,17 @@ static void poll_aubuf_tx(struct audio *a)
 		sampc = sampc_rs;
 	}
 
+	/* check if we should generate DTMF audio */
+	if (tx->enc_fmt == AUFMT_S16LE &&
+	    dtmf_is_empty(&tx->dtmfgen) == false) {
```
alfredh: comparing with false is not needed, better to write it like this:

`!dtmf_is_empty`

src/audio.c
```diff
@@ -1940,6 +1953,11 @@ int audio_send_digit(struct audio *a, char key)
 			return EINVAL;
 		}
 
+		err = dtmf_queue_digit(&a->tx.dtmfgen, key, a->tx.ac->crate, 0, 0);
```
alfredh: ac->srate should be used here, instead of ac->crate.

does it support both 1 and 2 channels ?

src/dtmf.c
```diff
@@ -0,0 +1,160 @@
...
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
```
alfredh: please use a similar header to all the other .c files

src/dtmf.c
```diff
@@ -0,0 +1,160 @@
...
+	{852, 1336, '8'},
+	{852, 1477, '9'},
+	{697, 1633, 'A'},
+	{770, 1633, 'B'},
+	{852, 1633, 'C'},
+	{941, 1633, 'D'},
+	{0, 0, 0},
+};
+
+static void
```
alfredh: please use 2 empty lines between functions

src/dtmf.c
```diff
@@ -0,0 +1,160 @@
...
+	float p1 = two_pi * (float)f1 / (float)sample_rate;
+
+	ps->kx[0] = cosf(p0);
+	ps->ky[0] = sinf(p0);
+	ps->kx[1] = cosf(p1);
+	ps->ky[1] = sinf(p1);
+
+	ps->duration = duration;
+}
+
```
alfredh: please use 2 empty lines between functions

src/dtmf.c
```diff
@@ -0,0 +1,160 @@
...
+dtmf_queue_digit(struct dtmf_generator *pg,
+		 uint8_t digit, uint32_t sample_rate,
+		 uint16_t tone_duration, uint16_t gap_duration)
+{
+	uint32_t next_pos0 = (pg->output_pos + 1) % DTMF_MAX_DIGITS;
+	uint32_t next_pos1 = (pg->output_pos + 2) % DTMF_MAX_DIGITS;
+	uint32_t n;
+
+	if (next_pos0 == pg->input_pos || next_pos1 == pg->input_pos)
+		return (ENOMEM);
```
alfredh: please remove paranthesis, should be written like this:

`return ENOMEM;`

src/dtmf.c
```diff
@@ -0,0 +1,160 @@
...
+	dtmf_set_state(pg->state + pg->output_pos,
+	    sample_rate, tone_duration,
+	    dtmf_to_freq[n].f0, dtmf_to_freq[n].f1);
+
+	dtmf_set_state(pg->state + next_pos0,
+	    sample_rate, gap_duration, 0, 0);
+
+	pg->output_pos = next_pos1;
+
+	return (0);
```
alfredh: please remove paranthesis.

src/dtmf.c
```diff
@@ -0,0 +1,160 @@
...
+
+	pg->x[0] = nx[0];
+	pg->y[0] = ny[0];
+
+	pg->x[1] = nx[1];
+	pg->y[1] = ny[1];
+
+	pg->duration++;
+
+	return (ret * 8192.0f);
```
alfredh: please remove paranthesis

src/dtmf.c
```diff
@@ -0,0 +1,160 @@
...
+}
+
+int
+dtmf_get_sample(struct dtmf_generator *pg, int16_t *psamp)
```
alfredh: is it possible to make a function that returns a bunch of samples instead?
this make the API a bit more user friendly...

src/dtmf.h
```diff
@@ -0,0 +1,66 @@
...
+{
+	return (pg->input_pos == pg->output_pos);
+}
+
+int	dtmf_queue_digit(struct dtmf_generator *, uint8_t digit,
+			 uint32_t sample_rate, uint16_t tone_duration,
+			 uint16_t gap_duration);
+int	dtmf_get_sample(struct dtmf_generator *pg, int16_t *psamp);
+
+#endif			/* __DTMF_GEN_H__ */
```
alfredh: please move the API to `core.h`

alfredh:
I am wondering if we can make the DTMF method automatic
for each call.

1. if the remote SDP contains "telephone-event", use out-of-band RFC4733
2. if not, use inband DTMF tone generation (your code).

do you think this would work with your GSM gateway ?

the logic can be placed in `audio_send_digit`:

```c
int audio_send_digit(struct audio *a, char key)
{
    if (sdp_media_rformat(stream_sdpmedia(audio_strm(a)), telev_rtpfmt))
       telev_send();
    else
        dtmf_queue_digit();
}
```

hselasky:
> do you think this would work with your GSM gateway ?

I don't know yet. I'm using Asterisk 1.3.

Last time I looked into this area I found nothing but a big mess. Some applications like Asterisk will even detect inband DTMF and then try to mute those tones away from the audio stream and convert to SIP DTMF events instead.

As time permits I'll try to implement your suggestions.

alfredh:
I am happy to help you with the integration of the patch.
please have a look at my review comments first.

please re-open when you are ready to work on this.

hselasky:
Sure, I just need some time for this :-)

PR #896, opened 2020-01-12
Jack connect config

AaronHorn:
Second attempt... Fixed formatting.

Add option to allow baresip not to make jack connections once the call is up and running.  My use case is radio broadcasting where the routing will be switched by the playout system.


alfredh:
Thanks!

PR #904, opened 2020-01-17
Set ebuacip parameters according to audio buffer settings. 

Ola-Palm:
Hi.
This PR will set announced ebuacip parameters according to the new audio buffer values. 
The ebuacip handler will set parameters according to incoming sdp, "ebuacip:jbdef and its values."

<hr><h3>Note:</h3><br> I'm not sure how the stream jb acctually works. 
If incoming packets are forwarded as soon as the next correct incoming packet has arrived, I think it's safe to always set as small stream buffer as possible. 
The min value is therefore always one.
The max value is set from incoming sdp max jb value / tx ptime. If this calculated value is less or equal to 1 ==> set max to 2 frames. (could perhaps be a value taken from config?) min_stream_jb=17
something like:

`if(frames <= min_stream_jb ) {`
`frames = min_stream_jb;`
`}`

If this isn't how the stream jb works, this pr might not be accurate. 
/ Cheers Ola


alfredh:
thanks for the patch, I merged it to master.

the only comment I have is that it is not necessary to set the stream jb:

```
	stream_jbuf_reset(au->strm, 1, frames);
```

the stream jbuf is share for audio/video and is mainly for re-ordering.
it should be kept as small as possible, e.g. 1-2 packets.

PR #905, opened 2020-01-20
add numeric version number

paresy:
Submitting a PR as requested in https://github.com/alfredh/baresip/issues/893

Thanks!

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -15,6 +15,9 @@ extern "C" {
 /** Defines the Baresip version string */
 #define BARESIP_VERSION "0.6.5"
 
+/** Defines the Baresip version number */
+#define BARESIP_VERSION_NUMBER 0x000605
+
 
 #ifndef NET_MAX_NS
 #define NET_MAX_NS (4)


alfredh:
the master version number is kept in the Makefile:

https://github.com/alfredh/baresip/blob/master/Makefile#L13

for every version bump, we run make version:

```
$ make version
updating version number to 0.6.5
```

which updates the version number in baresip.h

would it perhaps be easier to define:

```
#define BARESIP_VERSION_MAJOR 0
#define BARESIP_VERSION_MINOR 6
#define BARESIP_VERSION_PATCH 5
```

paresy:
Sure that should also work well. 

If i understood correctly at the moment you bump the BARESIP_VERSION_STRING define and run "make version" to update references.

I can update the Makefile magic to properly update your proposed splitted defines or would you prefer to update the splitted defines by hand and then run the Makefile magic to automatically update the BARESIP_VERSION_STRING and other references?

The second variant should be easier to implement/maintain but would imply a small workflow change. Is this documented somewhere? If yes, i can update the docs as well.

alfredh:
I was thinking something like this:


Makefile:

old:

```
VERSION   := 0.6.5
```


new:

```
# Master version number
VER_MAJOR := 0
VER_MINOR := 6
VER_PATCH := 5

VERSION   := $(VER_MAJOR).$(VER_MINOR).$(VER_PATCH)
```

and then update the "make version" target with
the 3 sub-version numbers.

alfredh:
please reopen when you have a new patch

PR #911, opened 2020-02-02
make reg->af available for use by best_effort_af

juha-h:
Do not update ua->af_media by reg.c  register_handler().  Instead make reg->af available for use by ua.c best_effort_af().


alfredh:
Thanks, this look great !

PR #915, opened 2020-02-08
add support for STUN uri used in MNAT api

alfredh:
https://tools.ietf.org/html/rfc7064
https://tools.ietf.org/html/rfc7065

ice_turn yes|no config has been replaced by per account
stunserver:

    ;stunserver="turn:example.org:8000"

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,8 @@ zrtp          ZRTP media encryption module
 * RFC 6464  A RTP Header Extension for Client-to-Mixer Audio Level Indication
 * RFC 6716  Definition of the Opus Audio Codec
 * RFC 6886  NAT Port Mapping Protocol (NAT-PMP)
+* RFC 7064  URI Scheme for STUN Protocol
+* RFC 7065  TURN Uniform Resource Identifiers
 * RFC 7310  RTP Payload Format for Standard apt-X and Enhanced apt-X Codecs
 * RFC 7587  RTP Payload Format for the Opus Speech and Audio Codec
 * RFC 7741  RTP Payload Format for VP8 Video
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -186,7 +186,6 @@ video_selfview		window # {window,pip}
 #selfview_size		64x64
 
 # ICE
-ice_turn		no
 ice_debug		no
 ice_nomination		regular	# {regular,aggressive}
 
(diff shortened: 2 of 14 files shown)


juha-h:
Name of the parameter is ;stunserver, but the URI can also point to a turn server, which is a bit confusing.

Could it be that an account needs both a stun server and a turn server?

GGGO:
What about iceserver ?
Like webrtc api: https://www.w3.org/TR/webrtc/#rtciceserver-dictionary

;iceserver="stun:example.org"
;iceserver="turn:example.org"

alfredh:
Please also see here:

https://github.com/alfredh/baresip/wiki/Media-NAT-Traversal

natpmp.so and pcp.so does not use the STUN-server configuration.

The STUN-server configuration is used by:

* stun.so
* turn.so
* ice.so

stun.so is ONLY stun.
turn.so is ONLY turn.
ice.so is stun OR turn.

Please note; a TURN server is also a STUN server, TURN is using STUN
as the base protocol.

ICE is *using* STUN/TURN but there is no such thing as an "ICE" Server.
In WebRTC terminology, an ICE-server is a TURN/STUN server.

Confusing ?

After this commit I am planning to add support for multiple STUN/TURN servers.

some candidates for parameter name:

```
stunserver=xxx
turnserver=xxx
iceserver=xxx
mnatserver=xxx
```

juha-h:
Yes, confusing.  Wki page has example account:
```
<sip:user@example.com>;medianat=turn;stunserver="stun:my.turn.server";stunuser=USER;stunpass=SECRET
```
If turn is used, can the server's URI scheme still be `stun:`?

My preference would be to match parameter name with server name, i.e., have both ;stunserver and ;turnserver parameters.

alfredh:
do you mean that "turnserver" is an alias for "stunserver" ?

in this case it would be possible to introduce config errors:

`turnserver=stun:example.com`

currently we only have storage for 1 stun/turn server, the plan is to change
this to a list of stun/turn servers.

the type of server is identified by the scheme ( `stun:` or `turn:` )

please read the patch in modules/turn to understand how the scheme is used:

```
if (srv->scheme != STUN_SCHEME_TURN)
 		return ENOTSUP;
```

alfredh:
I merged the patch to master.

The next step is to change from 1 to N servers.

Issue #758, opened 2019-07-19
Callcontrol trough SIP (answer/hold/unhold)


jnaeff:
We are playing with the Baresip client on raspberry. The experience is directly really great! This makes it possible to put the audio through the raspberry and the interface on the Remote Desktop Client. This way you get great audio and interface experience. The only thing we are missing is some commands to control the call completely through SIP. The commands are Answer, Hold and Unhold. If we have these commands we can control the phone completely. 

The specific documentation you can find [here](http://lists.digium.com/pipermail/asterisk-dev/attachments/20080102/25aaa8a1/attachment-0001.pdf). You need to look into chapter 5 (answer) and 6 (hold).

Is it possible for you guys to implement this? We are also able to pay a little bit of money for this functionality.

alfredh:
the referenced document is a Broadsoft SIP server.

it is not really clear to me what do you are doing or what you are trying to achieve?
is there a bug somewhere that should be fixed ?

are you asking us to implement a new feature ?

can you please describe properly here in the issue what you want to achieve ?

jnaeff:
_it is not really clear to me what do you are doing or what you are trying to achieve?_
We want to run Baresip as a daemon on the Raspberry PI. We want to build a desk phone based on that without a display.  To control Baresip on the Raspberry PI we want to use sip commands todo call control.  We only miss the commands (answer/hold/unhold) to make this possible. We do already this trick with desk phones (Yealink and Snom ). We can control the complete phone from our platform through SIP without using, for example, using the Embedded web-server with HTTP interface or the Command-line console over UDP/TCP. We think it is nice that you reuse the   SIP session that you have already with the platform. 

_is there a bug somewhere that should be fixed ?_
No, it is a new feature. 

_are you asking us to implement a new feature ?_
See the answer above.

_can you please describe properly here in the issue what you want to achieve ?_
We want to have solutions for our client that don't want to have a desk phone. And with the solution with Baresip running native on the Raspberry PI we think we have a great solution. The audio goes directly through the Rasperry with Baresip and you can control your Phone (Rasperry) trough sip for example on a remote desktop without audio issues.

alfredh:
I think this feature is a bit product spesific, and probably outside the scope
for a generic SIP client.

it is probably better if you use the mailing list to ask this question
if anyone can help you. Baresip is a open source project with contributions
from people, if you feel that this feature is very important to the project
you can implement the feature yourself and submit a Pull Request.

here is the mailing list:

[...]

PR #928, opened 2020-02-25
Implement Broadsoft call-control (answer/hold/unhold)

gaaf:
Implement call control according to Broadsoft as requested in #758.

This depends on changes in creytiv/re (creytiv/re#235). I'm not completely sure how creytiv/re is supposed to work with in-dialog requests, so i requested clarification on the re-devel mailinglist and as well in a pull-request (creytiv/re#235). Both got no response at all.


alfredh:
many thanks for the PR, this is really nice work!

we should first try to establish how the incoming NOTIFY message looks like
and if we need to make changes to libre or not.

1. do you have a reference to the Broadsoft SIP specification ?
2. do you have an example SIP NOTIFY message ?

I understand that your code uses the in-dialog NOTIFY message
from the sipsess layer (requires re patching). This is also called an
unsolicited NOTIFY. Normally a NOTIFY is only sent after the client
has sent SUBSCRIBE to subscribe to a resource.

if we can use a normal SUBSCRIBE/NOTIFY event package,
it might be easier to use the sipevent api:

https://github.com/creytiv/re/blob/master/include/re_sipevent.h

we can also use `sip_listen` to listen for all incoming SIP messages:

```c
static bool request_handler(const struct sip_msg *msg, void *arg)
{
	if (!pl_strcmp(&msg->met, "NOTIFY")) {
		/* lookup call from call-id, to/from tag */
	}

	return true;
}

	err = sip_listen(&uag.lsnr, uag.sip, true, request_handler, NULL);

```

.gitignore
```diff
@@ -23,3 +23,10 @@ selftest
 src/static.c
 build*
 *.pc
+
+# Backup files
+*~
+*.bak
+*.old
+*.orig
+*.rej
```
alfredh: nit: this can be a separate PR

gaaf:
>     1. do you have a reference to the Broadsoft SIP specification ?
It's in the referenced issue (#928).

>     2. do you have an example SIP NOTIFY message ?
```
NOTIFY sip:test-0x556394ecd6e0@172.16.1.1:7007 SIP/2.0.
Via: SIP/2.0/UDP 172.16.1.1:7001;branch=z9hG4bKb1-3.
From: "banonymous" <sip:+12345678@172.16.1.1:7001>;tag=1.
To: sut <sip:test@invalid>;tag=b01011672f1aa7be.
Call-ID: 1-522843@172.16.1.1.
CSeq: 2 NOTIFY.
Max-Forwards: 70.
Subscription-State: terminated.
Event: talk.
```


> I understand that your code uses the in-dialog NOTIFY message
> from the sipsess layer (requires re patching). This is also called an
> unsolicited NOTIFY.

I'm more used to that term if the NOTIFY is sent completely out-of-dialog. This NOTIFY is in-dialog.


> Normally a NOTIFY is only sent after the client
> has sent SUBSCRIBE to subscribe to a resource.

It is a bit like the NOTIFY's related to REFER. I only found out later that re/baresip handle those outside of the INVITE-dialog (even though they're part of it). This is what confused me and I tried to get clarified by my email to re-devel.

I got the impression re doesn't really know the concept of in-dialog requests like other SIP stacks handle them. It seems requests are dispatched based on the method and then, by each method handler separately, checked against current dialogs. I'm more used to the other way around.

All fine to me. But for me, being new to re and baresip, it is a bit difficult to determine where in-dialog requests should be added. As the NOTIFY really belongs to the INVITE dialog, I added it there. Just like the in-dialog INFO requests are handled.


> if we can use a normal SUBSCRIBE/NOTIFY event package,
> it might be easier to use the sipevent api

I looked at that, but concluded that was for SUBSCRIBE dialogs. Decided not to use it because my NOTIFY really belongs to the INVITE dialog.


If the conclusion is sip_event should be used, I'll rewrite the patchset.

.gitignore
gaaf: That got in by accident, I'll remove it.

alfredh:
thanks, I understand the concept now.

from Broadsoft spec section 5.6:

```
Subscriptions are implied by sessions established through INVITE requests and last throughout the duration of the session. When a UAC sends an INVITE to a UAS, it adds an Allow-Events header to the request, indicating all of the event packages it supports. This implicitly creates the subscription between the UAC and the UAS. When a UAS responds to the INVITE with an 18x provisional response or a 200 OK response, it adds an Allow-Events header indicating all of the event packages it supports. This implicitly creates the subscription between the UAS and the UAC. These subscriptions last until the session has been terminated with a BYE or CANCEL transaction.
```

this means that we cannot use the `re_sipevent` API.

I think we have 2 options:

1. use sip_listen to listen for incoming NOTIFY messages

2. update sipsess module to handle incoming NOTIFY messages.
  this option is less attractive, as the Broadsoft is not a standard SIP behaviour.

alfredh:
after some thinking, there might be another solution...

is it acceptable for the server to receive a SUBSCRIBE ?

```c
int sipevent_dsubscribe(struct sipsub **subp, struct sipevent_sock *sock,
			struct sip_dialog *dlg, const char *event,
			const char *id, uint32_t expires, const char *cuser,
			sip_auth_h *authh, void *aarg, bool aref,
			sipsub_notify_h *notifyh, sipsub_close_h *closeh,
			void *arg, const char *fmt, ...);
```

this function can be used to create a Subscription for an existing dialog.
The dialog in your case would be the Call dialog.

if the server can ignore/receive the SUBSCRIBE, you can call this
function from call.c and receive incoming NOTIFY via notifyh.

gaaf:
>     2\. update sipsess module to handle incoming NOTIFY messages.

This is what my PR's implement.


>         this option is less attractive, as the Broadsoft is not a standard SIP behaviour.

Well, SIP implementations are free to send any request method (even those not in RFC's, someone may send the "KEEPALIVE" method).

Even though it is not "standard", the way re is setup now is a bit restrictive wrt in-dialog requests, especially for a generic SIP stack. Preferably, re would call a callback for all in-dialog requests it does not handle itself (or for _all_ requests).

gaaf:
> is it acceptable for the server to receive a SUBSCRIBE ?
> if the server can ignore/receive the SUBSCRIBE, you can call this
> function from call.c and receive incoming NOTIFY via notifyh.

I'm not sure. I don't know which, not even how many, servers implement this protocol. The client I'm implementing this for will use Asterisk as server (for now).

alfredh:
I would try to find a solution that does not need any changes to libre.

Please look at this example code where the broadsoft call control
is implemented as a module:

```c
/**
 * @file broadsoft.c  Broadsoft call-control
 *
 * Copyright (C) 2010 Creytiv.com
 */

#include <re.h>
#include <baresip.h>


static struct sip_lsnr *lsnr;


static struct call *lookup_call(const struct sip_msg *msg)
{
	struct ua *ua;
	struct le *le;

	ua = uag_find(&msg->uri.user);
	if (!ua)
		return NULL;

	for (le = list_head(ua_calls(ua)); le; le = le->next) {
		struct call *call = le->data;

		if (0 == pl_strcmp(&msg->callid, call_id(call)))
			return call;
	}

	return NULL;
}


static bool request_handler(const struct sip_msg *msg, void *arg)
{
 	struct sipevent_event event;
 	const struct sip_hdr *hdr;
	struct call *call;

	if (pl_strcmp(&msg->met, "NOTIFY"))
		return false;

 	hdr = sip_msg_hdr(msg, SIP_HDR_EVENT);
 	if (!hdr || sipevent_event_decode(&event, &hdr->val))
 		return false;

 	if (pl_strcmp(&event.event, "talk") == 0) {
		call = lookup_call(msg);
		if (call) {
			if (call_state(call) == CALL_STATE_INCOMING)
				call_answer(call, 200);
			else
				call_hold(call, false);
			sip_reply(uag_sip(), msg, 200, "OK");
		}
		else {
			sip_reply(uag_sip(), msg, 404, "Not Found");
		}
 	}
	else if (pl_strcmp(&event.event, "hold") == 0) {
		call = lookup_call(msg);
		if (call) {
			call_hold(call, true);
			(void)sip_reply(uag_sip(), msg, 200, "OK");
		}
		else {
			sip_reply(uag_sip(), msg, 404, "Not Found");
		}
 	}
	else
		return false;

 	return true;
}


static int module_init(void)
{
	int err;

	err = sip_listen(&lsnr, uag_sip(), true, request_handler, NULL);
	if (err) {
		warning("broadsoft: sip listen error (%m)\n", err);
		return err;
	}

	return 0;
}


static int module_close(void)
{
	lsnr = mem_deref(lsnr);
	return 0;
}


const struct mod_export DECL_EXPORTS(broadsoft) = {
	"broadsoft",
	"application",
	module_init,
	module_close
};
```

when the incoming NOTIFY arrives, it looks up the correct call using Call-ID.
for matching calls, it sets the call state according to the event type (talk/hold).

The advantage of this approach is;

1. The implementation details of Broadsoft stays in the broadsoft.so module
and does not affect any code in libbaresip core

2. Any future changes to Broadsoft will only affect this module

3. The module is only loaded if you want to enable this feature, hence the need
for "remote_control" account parameter is obsolete.

gaaf:
Thanks. I'll test it.

Some remarks/questions:

1) Standards say NOTIFY must be answered first, before handling the event. I'll change that.

2) Is it possible to enable/disable modules per account? If not, I think the option must stay.

3) CSeq should be checked. Requests arriving out-of-order can break a call. This should be stuff handled by a SIP stack library, but apparently it isn't. This means replicating even more sipsession code in the module. This was the reason I hooked it into sipsession in the first place.

gaaf:
Well, I tried.

After getting it to compile it doesn't seem to do anything. In-dialog requests are not delivered to the request_handler.

Also, even though the module approach looks nice, the interfaces are not sufficient to efficiently build the needed functionality. As partly mentioned in my previous comment under 3), the whole dialog matching needs to be reimplemented from scratch (not only callid matching, but also tags and CSeq). If there was a possibility for the module to just receive _verified_ in-dialog messages, this would be a lot more feasible.

And how to get the "Allowed-Events" part into the module?

I my proposed libre changes would be accepted, maybe the baresip part could be moved into a module.

alfredh:
there is a problem regarding incoming NOTIFY messages.

ua.c:

```c
	err = sipevent_listen(&uag.evsock, uag.sip, bsize, bsize,
			      sub_handler, NULL);
	if (err)
		goto out;
```

re/src/sipevent/listen.c:

```
	else if (!pl_strcmp(&msg->met, "NOTIFY")) {

		notify_handler(sock, msg);
		return true;
	}
```

an incoming SIP request is dispatched to all listeners
calling `sip_listen`, and used by the first one that returns true.

other users of NOTIFY is for example Call Transfer (REFER/NOTIFY)
and Presence (SUBSCRIBE/NOTIFY).

can you please check if this is still working with your patch applied ?

you can run this testcase as an example:

```
$ ./selftest test_call_transfer
running baresip selftest version 0.6.5 with 1 tests
[ RUN      ] test_call_transfer
.... sipevent: recv NOTIFY
.... sipevent: recv NOTIFY
[       OK ]
OK. 1 tests passed successfully
```

src/account.c
```diff
@@ -1150,6 +1183,21 @@ const char *account_call_transfer(const struct account *acc)
...
+ * @param acc User-Agent account
+ *
+ * @return "yes" or "no"
+ */
+const char *account_remote_control(const struct account *acc)
+{
+	if (!acc)
+		return "no";
+
+	return acc->remote_control ? "yes" : "no";
```
alfredh: I think it is better to use "broadsoft" as a value here, then it is possible to add other
profiles in the future

src/ua.c
```diff
@@ -2029,6 +2029,29 @@ int ua_print_allowed(struct re_printf *pf, const struct ua *ua)
...
 
+/**
+ * Print the supported events header
+ *
+ * @param pf  Print function
+ * @param ua  User-Agent object
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int ua_print_allowed_events(struct re_printf *pf, const struct ua *ua)
```
alfredh: perhaps move this function to account.c ?

alfredh:
hi, did you see my review comments ?

Alfred

gaaf:
Sorry, had no time yet to look into it. I hope next week.

src/ua.c
gaaf: Initially I had that function in another file, but is felt so lonely there between all those dissimilar looking functions that I moved it to its current place between its ua_print_* friends. But I happily move it if deemed necessary.

alfredh:
this PR will not be accepted as is.

The patch creates a potential large security hole in the core,
where the remote caller can answer the call on behalf of the user,
just by sending a specially crafted NOTIFY message.
The current code is protected by a special setting in "account",
but it is very easy later to make mistakes so that the check is removed
from the code and the security hole is exposed.

If you want this feature into baresip, you need to:

1. Create the feature in a module (e.g. broadsoft.so) so that the module has to
be explicitly loaded by the user in order to use the feature.

2. use the existing SUBSCRIBE/NOTIFY framework or lookup call-id.

3. dont make any changes to creytiv/re

PR #930, opened 2020-02-27
prompt password only if auth_user is set and auth_pass is not

juha-h:
There are use cases where account does not need to authenticate itself when sending SIP requests, for example, when it is only used to call some public service URIs.

So far I have not figured out how to create an account without auth_user and auth_pass so that password is not prompted when account module is loaded.  

This pull requests solves the issue by causing password prompting only if auth_user is set and auth_pass is not set.  Better solutions are, of course, welcome .

### DIFF
--- a/modules/account/account.c
+++ b/modules/account/account.c
@@ -144,8 +144,9 @@ static int line_handler(const struct pl *addr, void *arg)
 		}
 	}
 
-	/* optional password prompt */
-	if (!str_isset(account_auth_pass(acc))) {
+	/* prompt password if auth_user is set, but auth_pass is not  */
+	if (str_isset(account_auth_user(acc)) &&
+	    !str_isset(account_auth_pass(acc))) {
 		char *pass = NULL;
 
 		(void)re_printf("Please enter password for %s: ",


alfredh:
thanks, this is a good solution for now ..

anarcat:
i believe this has caused some sort of regression on my end. before I had this in `accounts`:

```
<sip:REDACTED@voip.ms>;sipnat=outbound;stunserver=stun:stun.ekiga.net;mediaenc=srtp-mand;outbound="sip:toronto6.voip.ms:5060:transport=tls"
```

and it would prompt me for the password of the `REDACTED` account. since I upgraded to the latest baresip (in the Debian bullseye upgrade), I need this instead:

```
<sip:REDACTED@voip.ms>;sipnat=outbound;stunserver=stun:stun.ekiga.net;mediaenc=srtp-mand;outbound="sip:toronto6.voip.ms:5060:transport=tls";auth_user=REDACTED
```

which is kind of silly because it has two copies of that username...

juha-h:
SIP URI userpart does not need to be same as authentication
username. Adding auth_user should not be a too big trouble.

anarcat:
Oh I know. I did it, and it works. It's just that I couldn't figure out
why it was failing to authenticate at first and I went through *many*
other routes to try to figure out what was going on (including
rebuilding the entire configuration, reseting the passwords on the VoIP
provider and so on).

Having a clearer idea of what is going on would have helped here. I
don't understand why auth_username absolutely needs to be set for the
prompt to happen, it seems unrelated. It would be better to have a clear
option to prompt for the password, say "auth_password=PROMPT" or
something.

a.
-- 
Your injured body has become the burden of your digital soul.
                        - Yin Aiwen, 2013, The Massage is the Medium

juha-h:
anarcat writes:

> Having a clearer idea of what is going on would have helped here. I
> don't understand why auth_username absolutely needs to be set for the
> prompt to happen, it seems unrelated. It would be better to have a clear
> option to prompt for the password, say "auth_password=PROMPT" or
> something.

I agree with that.  Also, implementation of password prompting should be
based on events.

anarcat:
alternatively, maybe we can just update the examples file to document this feature...

PR #938, opened 2020-03-13
coreaudio: Fallback to CFStringGetCString when CFStringGetCStringPtr fails

njeisecke:
This became apparent when using a bluetooth headset. As soon as its
devicename was listed loading the module started to fail.

According to the CFStringGetCStringPtr documentation it is normal for
the function to fail in some cases. Using CFStringGetCString as a backup
is recommended.

Might also be the reason for https://github.com/alfredh/baresip/issues/914.


alfredh:
Many thanks!

njeisecke:
Alfred, thanks a lot for developing baresip. It's now part of our home office enabling efforts.

PR #948, opened 2020-03-24
Win32 fixes

njeisecke:
Hi Alfred,

these are my fixes to build baresip with VS 2017.

Not sure about the HAVE_GETOPT though - it requires to install the LGPL getopt via vcpkg:

vcpkg.exe install getopt

There's an alternative getopt implementation that works too, but it's using "AT&T public" license which I think is not compatible with baresip's BSD.


alfredh:
thank you!

njeisecke:
Thank you Alfred, there are similar PRs for libre and librem.

PR #954, opened 2020-04-03
Aubuf current jb value to mqtt

Ola-Palm:
When reporting VU RX values the current audio buffer size in ms is also included.

### DIFF
--- a/modules/mqtt/publish.c
+++ b/modules/mqtt/publish.c
@@ -34,6 +34,15 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 	if (err)
 		goto out;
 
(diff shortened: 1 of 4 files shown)


modules/mqtt/publish.c
```diff
@@ -34,6 +34,15 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 	if (err)
 		goto out;
 
+	//send audio jitter buffer values together with 
+	// VU rx values.
+	if ((int)ev == UA_EVENT_VU_RX) {
```
alfredh, line 39: is the casting to (int) necessary ?

PR #956, opened 2020-04-04
auframe: add timestamp

alfredh:
Add timestamp to audio frame. The clock-rate of the timestamp is in microseconds:

`#define AUDIO_TIMEBASE 1000000U`

Updated audio source modules:

* [x] alsa
* [x] aubridge
* [x] audiounit
* [x] aufile
* [x] avformat
* [x] coreaudio
* [ ] gst
* [ ] i2s
* [x] jack
* [ ] opensles
* [x] oss
* [x] portaudio
* [x] pulse
* [x] rst
* [ ] sndio
* [x] winwave

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -22,6 +22,11 @@ extern "C" {
 
 
 /*
+ * Clock-rate for audio timestamp
+ */
+#define AUDIO_TIMEBASE 1000000U
+
+/*
  * Clock-rate for video timestamp
  */
 #define VIDEO_TIMEBASE 1000000U
@@ -106,6 +111,7 @@ struct auframe {
 	int fmt;       /**< Sample format (enum aufmt)        */
 	void *sampv;   /**< Audio samples (must be mem_ref'd) */
 	size_t sampc;  /**< Total number of audio samples     */
+	uint64_t timestamp;  /**< Timestamp in AUDIO_TIMEBASE units */
 };
 
 void   auframe_init(struct auframe *af, int fmt, void *sampv, size_t sampc);
--- a/modules/alsa/alsa_src.c
+++ b/modules/alsa/alsa_src.c
@@ -53,6 +53,7 @@ static void ausrc_destructor(void *arg)
 static void *read_thread(void *arg)
 {
 	struct ausrc_st *st = arg;
+	uint64_t frames = 0;
 	int num_frames;
 	int err;
 
@@ -82,6 +83,9 @@ static void *read_thread(void *arg)
 		af.fmt   = st->prm.fmt;
 		af.sampv = st->sampv;
 		af.sampc = n * st->prm.ch;
+		af.timestamp = frames * AUDIO_TIMEBASE / st->prm.srate;
+
+		frames += n;
 
 		st->rh(&af, st->arg);
 	}
--- a/modules/aubridge/device.c
+++ b/modules/aubridge/device.c
@@ -104,7 +104,8 @@ static void *device_thread(void *arg)
 			struct auframe af = {
 				.fmt   = dev->ausrc->prm.fmt,
 				.sampv = sampv_in,
-				.sampc = sampc_in
+				.sampc = sampc_in,
+				.timestamp = ts * 1000
 			};
 			dev->ausrc->rh(&af, dev->ausrc->arg);
 		}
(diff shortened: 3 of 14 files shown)


alfredh:
The following modules needs also needs to be updated:

* gst
* i2s
* opensles
* sndio

Does anyone want to have a look at that ?

You need to set the audio timestamp field of this struct:

```c
/**
 * Defines a frame of audio samples
 */
struct auframe {
	int fmt;       /**< Sample format (enum aufmt)        */
	void *sampv;   /**< Audio samples (must be mem_ref'd) */
	size_t sampc;  /**< Total number of audio samples     */
	uint64_t timestamp;  /**< Timestamp in AUDIO_TIMEBASE units */
};
```

FYI: @cspiel1 @juha-h

juha-h:
Alfred E. Heggestad writes:

> The following modules needs also needs to be updated:
> 
> * gst
> * i2s
> * opensles
> * sndio

I looked at opensles, but unfortunately don't have a clue to what value
timestamp should be set in bqRecorderCallback and opensles_player_alloc
functions.

alfredh:
if we cannot get the hardware timestamp, I think there are 2 options:

1. count the total number of samples, and convert to microseconds.
  See alsa_src.c for example

2. use `tmr_jiffies_usec`

juha-h:
Alfred E. Heggestad writes:

> if we cannot get the hardware timestamp, I think there are 2 options:

This gives time in usecs:
```
uint64_t getTimeUSec() {
    struct timestamp now;
    clock_gettime(CLOCK_MONOTONIC, &now);
    return now.tv_sec*1000000U + now.tv_usec;
}
```
Would it qualify has hw timestamp?

juha-h:
Here is another version, but it returns int64_t:
```
static int64_t getTimeUSec() {
    struct timespec now;
    clock_gettime(CLOCK_MONOTONIC, &now);
    return (int64_t)now.tv_sec*1000000U + now.tv_nsec / 1000;
}
```

alfredh:
you can also use this function, it does the same thing:

```c
uint64_t tmr_jiffies_usec(void);
```

juha-h:
So this would be the diff?
```
diff --git a/modules/opensles/recorder.c b/modules/opensles/recorder.c
index 824b45f..c47aa15 100644
--- a/modules/opensles/recorder.c
+++ b/modules/opensles/recorder.c
@@ -58,7 +58,8 @@ static void bqRecorderCallback(SLAndroidSimpleBufferQueueItf bq, void *context)
    struct auframe af = {
        .fmt   = AUFMT_S16LE,
        .sampv = st->sampv[st->bufferId],
-       .sampc = st->sampc
+       .sampc = st->sampc,
+       .timestamp = tmr_jiffies_usec()
    };
    (void)bq;
 
```

alfredh:
yes, great. same as test ausrc mock:

https://github.com/alfredh/baresip/blob/master/test/mock/mock_ausrc.c#L40

juha-h:
Unfortunately it turned out that after the timestamps were added,
baresip is crashing on all Androind phones that have 32 bit
armeabi-v7a cpu architecture:

********** Crash dump: **********
Build fingerprint: 'motorola/cedric/cedric:8.1.0/OPPS28.85-13-6/04442:user/release-keys'
#00 0x0004a1dc /system/lib/libc.so (tgkill+12)
#01 0x003929bb /data/app/com.tutpro.baresip-2806o0ACdhSoNhoZQSJEnw==/lib/arm/libbaresip.so
__aeabi_ldiv0
/buildbot/src/android/gcc/toolchain/build/../gcc/gcc-4.9/libgcc/config/arm/lib1funcs.S:1337:0
#02 0x000a4378 /data/app/com.tutpro.baresip-2806o0ACdhSoNhoZQSJEnw==/lib/arm/libbaresip.so
audio_codec
/usr/src/libbaresip-android/baresip/src/audio.c:2304:6
Crash dump is completed

juha-h:
juha-h writes:

> Unfortunately it turned out that after the timestamps were added,
> baresip is crashing on all Androind phones that have 32 bit
> armeabi-v7a cpu architecture:

Actually, the crashes occur already at commit:

https://github.com/alfredh/baresip/commit/3f2a18227adef0b4b9db2edc660c69862532bf2d

No crashed at previous commit:

https://github.com/alfredh/baresip/commit/915171ab791df8e87c40fff1780434455909057b

PR #966, opened 2020-04-25
Added vidmode param to call_answer(), ua_answer() and ua_hold_answer() functions

juha-h:
Tried to preserve backwards compatibility.


alfredh:
Many thanks @juha-h Great work :)

Issue #959, opened 2020-04-07
MQTT / TCP on connection client state request


Roog:
Hi Alfred, 
I'm working with @djhenley and @Ola-Palm at Swedish Radio on our implementation of Baresip. As I hope you are aware, we are using Baresip on a daily basis and it's at the core of radio production at [Sveriges Radio](https://sverigesradio.github.io). 

So I just want to extend an enormous **THANK YOU** for the continuous improvements of Baresip. This magical application helps our journalist do their normal day to day work at ease with amazing sound quality and in these Covid-19 times also providing remote production capabilities!

So to the actual point.. :)

We are using the MQTT or TCP connection with a Node.js app (hoping to distribute this soon). While connected to the "bus" the messages arrives in a nice JSON formatted way. But when connecting at a later stage, like during a call, we need to get the current state. Therefor our app is sending the "/callstat" and "/uastat" command to get the current state. While this comes back as a "sort of" formatted JSON blob, there could be some improvements. We are now parsing the text blob to get some key things like,
- are we registered and with what credentials?
- are we in a call and with who?
- are we in a call with what settings?
In the best of worlds, we would like to get a JSON object with the current state.

So my question to you, 
- What is needed and where should we put a receiver for commands coming from the MQTT/TCP connection? (So that we could respond with a full JSON response).

- I'm suggesting a JSON response for the general state request, what is your opinion on the contents? and as mentioned above, the basic things are most important, like registration status and call status. 

- Is there anything that need to be considered for implementations not using SIP?

Best regards
Roger

```js
{
  accountId: "cuser-id|rogersan-0x7fc96e722770",
  accountName: "Display Name Account",
  settings {
    answermode: "auto|manual|..",
    callTransfer: true|false,
  },
  sip: {
    state: "registered|unregistered",
    registered: true|false,
    uri: "me@sip-example.com",
    displayName: "Me Nice Name",
    domain: "sip-example.com",
    authUser: "me",
    responseCode: "200 OK",
    SRV: sip-example.com.
    AF: "v4",
    regInterval: 3779,
    pubInterval: "??",
    nat: true|false,
    stun: {
      user: "",
      server: ""
    },
  },
  call {
    state: "disconnected|connecting|connected|disconnecting",
    peerUri: "you@sip-example.com",
    peerDisplayName: "You Nice Name",
    direction: "incoming|outgoing",
    started: "unix-timestamp",
    localAttributes: {
      minptime: "20"
      ptime: "20",
[...]

PR #973, opened 2020-05-04
JSON initial codec state command and response

Roog:
Hi!
Thanks for answering the 'issue' #959 . "MQTT / TCP on connection client state request". 

As mentioned before in the "issue" the idea here is to have a command with a JSON-formatted response. The command could be used when you are starting an api connection to Baresip, via TCP or MQTT. Similair like how '/uastat' and '/callstat' is used today, so an external application could get up and running without a lot of parsing of those commands. The new 'response' could also be the default response on registration and call events...

I should start with an apology, I don't have a deep knowledge of C-programming. So the pull-request is a very early one, maybe @alfredh you can put me in the right direction. Please let me know if I should look up something before you give a proper response, I don't want to bother you with too trivial things..

* How memory is working and how the code i wrote is effecting that, I couldn't tell you..
* I'm probably not using 'pointers' and references in the correct way..
* I implemented a command in the debug_cmd, and I hope the placement of the other functions is ok/correct, maybe there is a better place or even in some custom module?
  * menu '/apistate' command added in debug_cmd.c
  * cmd_api_uastate added in devug_cmd.c
  * ua_state_json_api added in ua.c
  * account_json_api added in account.c
  * reg_json_api added in reg.c
* I will add some comments on specific rows, where I could not get the code to work in the way I'm thinking. There is some things that is not working in the way I think.
* This "object" only contains the user-agent, settings and account information. Call information will come later and maybe even in another function / "object". Depending how this goes.

```json
{
    "cuser":"rogersan-0x7f81ded10c30",
    "selected_ua":true,
    "aor":"sip:rogersan@domain.com",
    "display_name":"Roger San",
    "settings":{
        "sip_nat":"outbound",
        "sip_nat_outbound":[
            "sip:primary.example.com",
            "sip:secondary.example.com"
        ],
        "stun_host":"",
        "stun_port":0,
        "stun_user":"rogersan",
        "stun_pass":"***",
        "answer_mode":"manual",
        "call_transfer":true,
        "packet_time":20
    },
    "registration":{
        "interval":3697,
        "q_value":0.000000,
        "id":1,
        "scode":999,
        "af":"v4"
    }
}
```
Best regards
Roger

### DIFF
--- a/modules/debug_cmd/debug_cmd.c
+++ b/modules/debug_cmd/debug_cmd.c
@@ -82,6 +82,36 @@ static int cmd_ua_debug(struct re_printf *pf, void *unused)
 }
 
 
+/**
+ * Returns the current User-Agent and general codec state.
+ * Formatted as JSON, for use with TCP / MQTT API interface.
+ *
+ * @return Current User-Agent, NULL if none
+ */
+static int cmd_api_uastate(struct re_printf *pf, void *unused)
+{
+	(void)unused;
+	int err;
+	struct odict *od = NULL;
+
+	err = odict_alloc(&od, 8);
+	if (err)
+		return err;
+
+	struct le *le;
+	for (le = list_head(uag_list()); le && !err; le = le->next) {
+		const struct ua *ua = le->data;
+
+		err |= ua_state_json_api(od, pf, ua);
+		err |= json_encode_odict(pf, od);
+
+		if (err)
+			warning("debug: failed to encode json (%m)\n", err);
+	}
+	return re_hprintf(pf, "\n");
+}
+
+
 static int cmd_play_file(struct re_printf *pf, void *arg)
 {
 	static struct play *g_play;
@@ -187,6 +217,7 @@ static const struct cmd debugcmdv[] = {
 {"timers",      0,       0, "Timer debug",            tmr_status          },
 {"uastat",     'u',      0, "UA debug",               cmd_ua_debug        },
 {"uuid",        0,       0, "Print UUID",             print_uuid          },
+{"apistate",    0,       0, "User Agent state",       cmd_api_uastate     },
 };
 
 
--- a/modules/debug_cmd/debug_cmd.c
+++ b/modules/debug_cmd/debug_cmd.c
@@ -83,10 +83,10 @@ static int cmd_ua_debug(struct re_printf *pf, void *unused)
 
 
 /**
- * Returns the current User-Agent and general codec state.
+ * Returns all the User-Agents and their general codec state.
  * Formatted as JSON, for use with TCP / MQTT API interface.
  *
- * @return Current User-Agent, NULL if none
+ * @return All User-Agents available, NULL if none
  */
 static int cmd_api_uastate(struct re_printf *pf, void *unused)
 {
@@ -102,7 +102,8 @@ static int cmd_api_uastate(struct re_printf *pf, void *unused)
 	for (le = list_head(uag_list()); le && !err; le = le->next) {
 		const struct ua *ua = le->data;
 
-		err |= ua_state_json_api(od, pf, ua);
+		err |= ua_state_json_api(od, ua);
+		/* todo: add to list */
 		err |= json_encode_odict(pf, od);
 
 		if (err)
--- a/modules/debug_cmd/debug_cmd.c
+++ b/modules/debug_cmd/debug_cmd.c
@@ -93,12 +93,12 @@ static int cmd_api_uastate(struct re_printf *pf, void *unused)
 	(void)unused;
 	int err;
 	struct odict *od = NULL;
+	struct le *le;
 
 	err = odict_alloc(&od, 8);
 	if (err)
 		return err;
 
-	struct le *le;
 	for (le = list_head(uag_list()); le && !err; le = le->next) {
 		const struct ua *ua = le->data;
 
@@ -108,6 +108,8 @@ static int cmd_api_uastate(struct re_printf *pf, void *unused)
 		if (err)
 			warning("debug: failed to encode json (%m)\n", err);
 	}
+
+	mem_deref(od);
 	return re_hprintf(pf, "\n");
 }
 
--- a/modules/debug_cmd/debug_cmd.c
+++ b/modules/debug_cmd/debug_cmd.c
@@ -85,15 +85,16 @@ static int cmd_ua_debug(struct re_printf *pf, void *unused)
 /**
  * Returns all the User-Agents and their general codec state.
  * Formatted as JSON, for use with TCP / MQTT API interface.
+ * JSON object with 'cuser' as the key.
  *
  * @return All User-Agents available, NULL if none
  */
 static int cmd_api_uastate(struct re_printf *pf, void *unused)
 {
-	(void)unused;
-	int err;
 	struct odict *od = NULL;
 	struct le *le;
+	int err;
+	(void)unused;
 
 	err = odict_alloc(&od, 8);
 	if (err)
@@ -101,16 +102,21 @@ static int cmd_api_uastate(struct re_printf *pf, void *unused)
 
 	for (le = list_head(uag_list()); le && !err; le = le->next) {
 		const struct ua *ua = le->data;
+		struct odict *odua;
 
-		err |= ua_state_json_api(od, ua);
-		/* todo: add to list */
-		err |= json_encode_odict(pf, od);
+		err = odict_alloc(&odua, 8);
 
-		if (err)
-			warning("debug: failed to encode json (%m)\n", err);
+		err |= ua_state_json_api(odua, ua);
+		err |= odict_entry_add(od, ua_aor(ua), ODICT_OBJECT, odua);
+		mem_deref(odua);
 	}
 
+	err |= json_encode_odict(pf, od);
+	if (err)
+		warning("debug: failed to encode json (%m)\n", err);
+
 	mem_deref(od);
+
 	return re_hprintf(pf, "\n");
 }
 
--- a/src/account.c
+++ b/src/account.c
@@ -1259,40 +1259,54 @@ int account_debug(struct re_printf *pf, const struct account *acc)
  *
  * @return 0 if success, otherwise errorcode
  */
-int account_json_api(struct odict *od, struct odict *odcfg, const struct account *acc)
+int account_json_api(struct odict *od, struct odict *odcfg,
+		const struct account *acc)
 {
 	int err = 0;
+	struct odict *obn = NULL;
+	const char *stunhost = "";
 
 	if (!acc)
 		return 0;
 
 	/* account */
 	err |= odict_entry_add(od, "aor", ODICT_STRING, acc->aor);
-	err |= odict_entry_add(od, "display_name", ODICT_STRING, acc->dispname);
+	if (acc->dispname) {
+		err |= odict_entry_add(od, "display_name", ODICT_STRING,
+				acc->dispname);
+	}
 
 	/* config */
-	err |= odict_entry_add(odcfg, "sip_nat", ODICT_STRING, account_sipnat(acc));
+	if (acc->sipnat) {
+		err |= odict_entry_add(odcfg, "sip_nat", ODICT_STRING,
+				acc->sipnat);
+	}
 
-	struct odict *obn = NULL;
 	err |= odict_alloc(&obn, 8);
 	for (size_t i=0; i<ARRAY_SIZE(acc->outboundv); i++) {
 		if (acc->outboundv[i]) {
-			err |= odict_entry_add(obn, "outbound", ODICT_STRING, acc->outboundv[i]);
+			err |= odict_entry_add(obn, "outbound", ODICT_STRING,
+					acc->outboundv[i]);
 		}
 	}
 	err |= odict_entry_add(odcfg, "sip_nat_outbound", ODICT_ARRAY, obn);
 
-	const char *stunhost = account_stun_host(acc) ? account_stun_host(acc) : "";
+	stunhost = account_stun_host(acc) ? account_stun_host(acc) : "";
 	err |= odict_entry_add(odcfg, "stun_host", ODICT_STRING, stunhost);
-	err |= odict_entry_add(odcfg, "stun_port", ODICT_INT, account_stun_port(acc));
-	err |= odict_entry_add(odcfg, "stun_user", ODICT_STRING, account_stun_user(acc));
-	const char *stunpwd = account_stun_pass(acc) != NULL ? "***" : "";
-	err |= odict_entry_add(odcfg, "stun_pass", ODICT_STRING, stunpwd);
+	err |= odict_entry_add(odcfg, "stun_port", ODICT_INT,
+			account_stun_port(acc));
+	if (acc->stun_user) {
+		err |= odict_entry_add(odcfg, "stun_user", ODICT_STRING,
+				acc->stun_user);
+	}
 
-	err |= odict_entry_add(odcfg, "answer_mode", ODICT_STRING, answermode_str(acc->answermode));
+	err |= odict_entry_add(odcfg, "answer_mode", ODICT_STRING,
+			answermode_str(acc->answermode));
 	err |= odict_entry_add(odcfg, "call_transfer", ODICT_BOOL, acc->refer);
 
-	err |= odict_entry_add(odcfg, "packet_time", ODICT_INT, account_ptime(acc));
+	err |= odict_entry_add(odcfg, "packet_time", ODICT_INT,
+			account_ptime(acc));
 
+	mem_deref(obn);
 	return err;
 }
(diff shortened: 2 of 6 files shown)


src/ua.c
```diff
@@ -1263,6 +1263,8 @@ int ua_state_json_api(struct odict *od, struct re_printf *pf, const struct ua *u
 
 	/* user-agent info */
 	err |= odict_entry_add(od, "cuser", ODICT_STRING, ua->cuser);
+	/* TODO: 'selected_ua' not working, it's always true? */
+	err |= odict_entry_add(od, "selected_ua", ODICT_BOOL, ua == uag_current());
```
Roog, line 1267: The comparing of ua with uag_current() seems to always be true, maybe I'm referencing the wrong thing here.

src/ua.c
```diff
@@ -1239,6 +1239,60 @@ int ua_debug(struct re_printf *pf, const struct ua *ua)
...
+
+	for (le = ua->regl.head; le; le = le->next) {
+		/* TODO: how to get only current ua register state? */
+		err |= reg_json_api(reg, le->data);
+	}
+	if (err)
+		warning("ua: failed to encode json registration (%m)\n", err);
+
+	/* package */
+	/* TODO: this is complaining, how to build these objects? */
```
Roog, line 1286: The adding of:
* odict_entry_add(od, "settings", ODICT_OBJECT, cfg)
* odict_entry_add(od, "registration", ODICT_OBJECT, reg)
is causing Invalid argument, but seems to do the job, I'm missing something

src/ua.c
```diff
@@ -1239,6 +1239,60 @@ int ua_debug(struct re_printf *pf, const struct ua *ua)
...
+
+	/* registration info */
+	err |= odict_entry_add(reg, "interval", ODICT_INT, ua->acc->regint);
+	err |= odict_entry_add(reg, "q_value", ODICT_DOUBLE, ua->acc->regq);
+	/* TODO: current not working */
+	err |= odict_entry_add(reg, "current", ODICT_BOOL, ua == uag_current());
+
+	for (le = ua->regl.head; le; le = le->next) {
+		/* TODO: how to get only current ua register state? */
+		err |= reg_json_api(reg, le->data);
```
Roog, line 1280: This one is going through all the user agents and their registration statuses. I would like to just have the one that is valid for this user-agent.

modules/debug_cmd/debug_cmd.c
```diff
@@ -82,6 +82,36 @@ static int cmd_ua_debug(struct re_printf *pf, void *unused)
...
+{
+	(void)unused;
+	int err;
+	struct odict *od = NULL;
+
+	err = odict_alloc(&od, 8);
+	if (err)
+		return err;
+
+	struct le *le;
```
alfredh, line 101: please move the declaration to the top of the function

modules/debug_cmd/debug_cmd.c
```diff
@@ -82,6 +82,36 @@ static int cmd_ua_debug(struct re_printf *pf, void *unused)
...
+	for (le = list_head(uag_list()); le && !err; le = le->next) {
+		const struct ua *ua = le->data;
+
+		err |= ua_state_json_api(od, pf, ua);
+		err |= json_encode_odict(pf, od);
+
+		if (err)
+			warning("debug: failed to encode json (%m)\n", err);
+	}
```
alfredh, line 110: please destroy the object allocated in line 97:

```c
mem_deref(od);
```

src/account.c
```diff
@@ -1241,10 +1241,58 @@ int account_debug(struct re_printf *pf, const struct account *acc)
...
+int account_json_api(struct odict *od, struct odict *odcfg, const struct account *acc)
+{
+	int err = 0;
+
+	if (!acc)
+		return 0;
+
+	/* account */
+	err |= odict_entry_add(od, "aor", ODICT_STRING, acc->aor);
+	err |= odict_entry_add(od, "display_name", ODICT_STRING, acc->dispname);
```
alfredh, line 1271: acc->dispname can be NULL, so please add a conditional for this line:

```c
if (acc->dispname)
```

src/account.c
```diff
@@ -1241,10 +1241,58 @@ int account_debug(struct re_printf *pf, const struct account *acc)
...
+
+	/* config */
+	err |= odict_entry_add(odcfg, "sip_nat", ODICT_STRING, account_sipnat(acc));
+
+	struct odict *obn = NULL;
```
alfredh, line 1276: move declaration to top.

src/account.c
```diff
@@ -1241,10 +1241,58 @@ int account_debug(struct re_printf *pf, const struct account *acc)
...
+	err |= odict_alloc(&obn, 8);
+	for (size_t i=0; i<ARRAY_SIZE(acc->outboundv); i++) {
+		if (acc->outboundv[i]) {
+			err |= odict_entry_add(obn, "outbound", ODICT_STRING, acc->outboundv[i]);
+		}
+	}
+	err |= odict_entry_add(odcfg, "sip_nat_outbound", ODICT_ARRAY, obn);
+
+	const char *stunhost = account_stun_host(acc) ? account_stun_host(acc) : "";
```
alfredh, line 1285: move declaration to the top

src/account.c
```diff
@@ -1241,10 +1241,58 @@ int account_debug(struct re_printf *pf, const struct account *acc)
...
+	err |= odict_entry_add(odcfg, "stun_host", ODICT_STRING, stunhost);
+	err |= odict_entry_add(odcfg, "stun_port", ODICT_INT, account_stun_port(acc));
+	err |= odict_entry_add(odcfg, "stun_user", ODICT_STRING, account_stun_user(acc));
+	const char *stunpwd = account_stun_pass(acc) != NULL ? "***" : "";
```
alfredh, line 1289: perhaps we should consider dropping printing of stun password ?

src/reg.c
```diff
@@ -260,6 +268,30 @@ int reg_debug(struct re_printf *pf, const struct reg *reg)
...
+ */
+int reg_json_api(struct odict *od, const struct reg *reg)
+{
+	int err = 0;
+
+	if (!reg)
+		return 0;
+
+	err |= odict_entry_add(od, "id", ODICT_INT, reg->id);
+	err |= odict_entry_add(od, "scode", ODICT_INT, reg->scode);
```
alfredh, line 287: odict_entry_add is using va_list, so for type `ODICT_INT` you need to cast
the value to `int64_t`

(both id and scode)

src/ua.c
```diff
@@ -1239,6 +1239,60 @@ int ua_debug(struct re_printf *pf, const struct ua *ua)
...
+	/* TODO: 'selected_ua' not working, it's always true? */
+	err |= odict_entry_add(od, "selected_ua", ODICT_BOOL, ua == uag_current());
+
+	/* account info */
+	err |= account_json_api(od, cfg, ua->acc);
+	if (err)
+		warning("ua: failed to encode json account (%m)\n", err);
+
+	/* registration info */
+	err |= odict_entry_add(reg, "interval", ODICT_INT, ua->acc->regint);
```
alfredh, line 1275: please cast value to `int64_t`

alfredh:
this looks really nice, well done for writing this code!

I have done a code review and added some comments inline.

it is mainly 2 things:

- add check for optional strings
- free dictionaries after encoding (mem_deref)

if I run the patch locally, here is how it looks:

```
$ ./baresip -e/apistate -eq
```

```
baresip is ready.
/apistate
ua: failed to encode json account (Invalid argument)
ua: failed to encode json registration (Invalid argument)
ua: failed to encode json package (Invalid argument)
{"cuser":"aeh-0x7fbfe8f31e20","selected_ua":true,"aor":"sip:aeh@iptel.org","settings":{"sip_nat_outbound":[],"stun_host":"","stun_port":0,"stun_user":"aeh","stun_pass":"***","answer_mode":"manual","call_transfer":true,"packet_time":20},"registration":{"interval":3761,"q_value":-0.000000,"id":0,"scode":0,"af":"v?"}}debug: failed to encode json (Invalid argument)

Quit
ua: stop all (forced=0)
mem: Memory leaks (55):
  0x7fbfe8f37140: nrefs=1  size=24      [90 75 f3 e8 bf 7f 00 00 50 89 f3 e8 bf 7f 00 00 ] [.u......P.......]
  0x7fbfe8f371a0: nrefs=1  size=16      [f0 71 f3 e8 bf 7f 00 00 08 00 00 00 00 00 00 00 ] [.q..............]
  0x7fbfe8f371f0: nrefs=1  size=128     [00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ] [................]
  0x7fbfe8f372b0: nrefs=2  size=24      [60 83 f3 e8 bf 7f 00 00 20 87 f3 e8 bf 7f 00 00 ] [`....... .......]
...
```

alfredh:
there are also some formating warnings that should be fixed:

```
ccheck.py 
./src/account.c:1262: line is too wide (86 - max 79)
./src/account.c:1271: line is too wide (80 - max 79)
./src/account.c:1274: line is too wide (84 - max 79)
./src/account.c:1280: line is too wide (97 - max 79)
./src/account.c:1285: line is too wide (84 - max 79)
./src/account.c:1287: line is too wide (86 - max 79)
./src/account.c:1288: line is too wide (89 - max 79)
./src/account.c:1292: line is too wide (100 - max 79)
./src/account.c:1295: line is too wide (84 - max 79)
./src/ua.c:1251: line is too wide (82 - max 79)
./src/ua.c:1267: line is too wide (83 - max 79)
./include/baresip.h:62: line is too wide (90 - max 79)
./include/baresip.h:741: line is too wide (84 - max 79)
Statistics:
~~~~~~~~~~~
Number of files processed:     c: 243  cpp: 9  h: 49  mk: 96  m4: 0  py: 0  m: 4  s: 0  java: 0 
Number of lines with errors:   13
```

modules/debug_cmd/debug_cmd.c
Roog: Ah, seems important, as mentioned not so familiar with how to deal with the "cleanup". Adding it after the for-loop.

src/account.c
Roog: Cool! Added some other null check ones as well.

src/account.c
Roog: Yes, I agree, did not think it's a good idea to send it out in the api, so obscuring it to show it's there, is maybe not needed.

Roog:
Thank you @alfredh ! 

Your comments, really helped :)
- I can not see any more "memory leeks", and I think i placed the mem_deref in the right places and freed up dictionaries.
- I think most formatting is solved, hopefully according to the rest of the code.
- Most values should be checked for NULL and casted in the correct format.

Suggested next step:
* Format and add User-Agents and information to a JSON list, for multiple user-agents.
* The 'selected_ua' property is not working, should only be true for the user-agent that is selected in Baresip. Comparing (ua == uag_current()) is not working, not sure what i'm doing wrong. It is always true on each iteration.
* Only filter out registration-status for one user-agent at a time, while iterating through list. Seems to exhaust the list on 'round two', guess the list needs to be pointed in some way to the beginning.. Not really sure how to yet. 
* Is there a unique id for each 'account' added?
* What is the 'cuser' exactly? Is it the computer user?

Thank you for taking your time to reply and comment!

Best regards
Roger

src/reg.c
```diff
@@ -242,6 +242,14 @@ static const char *print_scode(uint16_t scode)
 }
 
 
+/**
+ * Print the registration debug information
+ *
+ * @param pf  Print function
+ * @param ua  Registration object
```
alfredh, line 249: the param "ua" is wrong, it should be "reg".

src/ua.c
```diff
@@ -1239,6 +1239,62 @@ int ua_debug(struct re_printf *pf, const struct ua *ua)
...
+		return 0;
+
+	err |= odict_alloc(&reg, 8);
+	err |= odict_alloc(&cfg, 8);
+
+	/* user-agent info */
+	err |= odict_entry_add(od, "cuser", ODICT_STRING, ua->cuser);
+	/* TODO: 'selected_ua' not working, it's always true? */
+	err |= odict_entry_add(od, "selected_ua", ODICT_BOOL,
+			ua == uag_current());
```
alfredh, line 1267: this logic is working for me, when using the command line interface and 2 accounts.

```
"selected_ua":true,"aor":"sip:aeh@iptel.org"
"selected_ua":false,"aor":"sip:alfredh@sip.antisip.com"
```

modules/debug_cmd/debug_cmd.c
```diff
@@ -82,6 +82,39 @@ static int cmd_ua_debug(struct re_printf *pf, void *unused)
...
+	err = odict_alloc(&od, 8);
+	if (err)
+		return err;
+
+	for (le = list_head(uag_list()); le && !err; le = le->next) {
+		const struct ua *ua = le->data;
+
+		err |= ua_state_json_api(od, ua);
+		/* todo: add to list */
+		err |= json_encode_odict(pf, od);
```
alfredh, line 107: this logic will not work. it will append the ua object to the dict and then print it.
if you have 3 uas the result will be:

1
1,2
1,2,3

instead you should build an array, and then encode it as the last step

alfredh:
hi @Roog I added some more review comments, do they make sense to you ?

Roog:
Hi @alfredh , they do make sense! I'm just spending a little bit time reading up on some C related things. I wanted to understand arrays and pointers in C better. Hopefully I'll have some code soon :)

alfredh:
great,

regarding the function that encodes the JSON, here is an example:

```c
static int cmd_api_uastate(struct re_printf *pf, void *unused)
{
	struct odict *od = NULL;
	struct le *le;
	int err;
	(void)unused;

	err = odict_alloc(&od, 8);
	if (err)
		return err;

	for (le = list_head(uag_list()); le && !err; le = le->next) {
		const struct ua *ua = le->data;
		struct odict *odua;

		err = odict_alloc(&odua, 8);

		err |= ua_state_json_api(odua, ua);

		err |= odict_entry_add(od, ua_aor(ua), ODICT_OBJECT, odua);

		mem_deref(odua);
	}

	err |= json_encode_odict(pf, od);
	if (err)
		warning("debug: failed to encode json (%m)\n", err);

	mem_deref(od);

	return re_hprintf(pf, "\n");
}
```

it adds multiple objects to the root object, using the SIP aor as the key.
this is just an idea, you can solve it differently if you want.

the output is something like this:

```
{
   "sip:aeh@iptel.org":{
      "cuser":"aeh-0x7fe892e49cd0",
      "selected_ua":true,
      "aor":"sip:aeh@iptel.org",
      "settings":{
         "sip_nat_outbound":[

         ],
         "stun_host":"",
         "stun_port":0,
         "stun_user":"aeh",
         "answer_mode":"manual",
         "call_transfer":true,
         "packet_time":20
      },
      "registration":{
         "interval":3759,
         "q_value":0.000000,
         "id":0,
         "scode":0,
         "ip_version":"v?"
      }
   },
   "sip:alfredh@sip.antisip.com":{
      "cuser":"alfredh-0x7fe892e4a700",
      "selected_ua":false,
      "aor":"sip:alfredh@sip.antisip.com",
      "display_name":"Anti Alf",
      "settings":{
         "sip_nat_outbound":[

         ],
         "stun_host":"",
         "stun_port":0,
         "stun_user":"alfredh",
         "answer_mode":"manual",
         "call_transfer":true,
         "packet_time":20
      },
      "registration":{
         "interval":600,
         "q_value":0.000000,
         "id":0,
         "scode":0,
         "ip_version":"v?"
      }
   }
}
```

I have used an online JSON validator to check the JSON output.

src/ua.c
Roog: That is true! I must have got the wrong result since I did not append it to the array correctly. or just read it wrong.

src/ua.c
Roog: I realised that this one was listing two registrations for one user-agent / account. Not sure how i managed to get that. But that is what caused some confusion. Adding a check for this, unless it's a feature.

Roog:
Thank you for the suggestion!

- I think this makes sense, an alternative would be an array, instead of an object. But with the assumption that the SIP id, is unique, your suggestion is good I think. At least something we could start with. 

- I managed in some way to get one "user-agent" in my accounts file to be registered twice. So for one "registration" object i got two, so there was an "id:0 and id:1" with the same aor (not listed as multiple lines in the account file) except two registrations. I've added a check for this, assuming that it is not something that should happen? So this might have confused me a bit.

- I see that your "cuser" is unique per account, how do you make it so?

If you don't see anything that you'd like to change.
Maybe the next step is to question if the "debug_cmd.c" is the right location for this kind of data, thinking it can be available over TCP / HTTP and MQTT modules?

Best regards
Roger

alfredh:
thanks, I have merged the code to master.

- cuser (contact user) is unique per baresip instance. it is used to dispatch
incoming SIP requests to the correct UA.

- it is possible to have multiple UAs with the same aor, so the sip aor is not
really unique. we can change the JSON to an array if you want, or use
a numeric index.

you are welcome to make more changes and submit a PR for review.

Many thanks for your contributions :)

PR #988, opened 2020-05-27
call/update_video() enhancements

juha-h:
1) Before update, check that video stream has not been disabled, for example, by ua_event UA_EVENT_CALL_REMOTE_SDP.  Needs pull request https://github.com/creytiv/re/pull/248 to re.

2) If video stream has been disabled, stop also video display.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -360,10 +360,14 @@ static int update_audio(struct call *call)
 
 static int update_video(struct call *call)
 {
-	const struct sdp_format *sc;
+	const struct sdp_format *sc = NULL;
 	int err = 0;
 
-	sc = sdp_media_rformat(stream_sdpmedia(video_strm(call->video)), NULL);
+	struct sdp_media *m = stream_sdpmedia(video_strm(call->video));
+
+	if (!sdp_media_disabled(m))
+		sc = sdp_media_rformat(m, NULL);
+
 	if (sc) {
 		err = video_encoder_set(call->video, sc->data,
 					sc->pt, sc->params);
@@ -385,6 +389,7 @@ static int update_video(struct call *call)
 	else if (call->video) {
 		info("video stream is disabled..\n");
 		video_stop(call->video);
+		video_stop_display(call->video);
 	}
 
 	return err;


alfredh:
is it possible to make a patch that does not require changes in libre ?

if you can make a PR that does not need changes in libre, the PR can be merged very fast.
otherwise it may take weeks and months before the change is approved in libre.

juha-h:
libre has both set and get functions for most of the sdp_media attributes, but not for disabled attribute.  What I proposed in the libre pull request is not to change anything, but add a very simple get function for disabled attribute similar to the other already existing getters.

I don't know of any trivial alternative to this pull request, i.e., how to allow processing of  UA_EVENT_CALL_REMOTE_SDP event to affect the outcome of update_video() call.  I need that functionality so that the user can decide if he/she wants to accept addition of video media to a call via re-INVITE.

Last commit to libre was on April 23.  If libre is not actively maintained anymore, would it be better to fork libre for baresip use?  If I have understood correctly, changes to libre would also be needed for adding proper websocket transport for baresip.

alfredh:
hi @juha-h 

we can easily make a fork of libre, and use it for baresip.

but the question is, are you willing to maintain it ?
we need to have commitment from multiple people first, that they are willing to
maintain it for the unforseable future.

if you are okay with this, then please go a head and make a fork.
suggested location:

`github.com/baresip/re`

sreimers:
`but the question is, are you willing to maintain it ?`

I can also help with maintaining, at the moment I already have some patches to maintain. What about rem and retest? Are they actively maintained at the moment? Or should we also consider to fork?

sreimers:
I just started the `libre` fork: https://github.com/baresip/re and made some first commits (setup travis etc.) to align with baresip.

alfredh:
@sreimers many thanks!

@juha-h can you please commit your patch to that new fork ?

juha-h:
Alfred E. Heggestad writes:

> @juha-h can you please commit your patch to that new fork ?

Will do in a day or two.  Just returned from a five day hiking trip to
the high country.

juha-h:
Will close this pull request and create a new one due to baresip git repo move.

PR #996, opened 2020-05-31
net: add fallback dns nameserver

sreimers:
Sometimes the system name server recognition fails (is empty) or a defect name server is used.
Now it is possible to define a separate fallback DNS server (without overwriting the system-side ones).


alfredh:
looks fine, as long as it works :)

please feel free to push/merge to master.

PR #1010, opened 2020-06-10
fix dtmf_tone in gtk and add tones as feedback

mbattista:
Call window in gtk is not sending the correct key to end the DTMF tone which leads to unexpected behavior when using it in a call.

I also added wav files I have created using the library wavfile "https://www3.nd.edu/~dthain/courses/cse20211/fall2013/wavfile/" to give more feedback to the gtk user.


modules/gtk/call_window.c
```diff
@@ -217,22 +218,78 @@ static gboolean call_on_window_close(GtkWidget *widget, GdkEventAny *event,
...
+	case '0':
+		(void)play_file(&win->play_dtmf_tone, baresip_player(),
+			"sound0.wav", -1, cfg->audio.alert_mod,
+			cfg->audio.alert_dev);
+		break;
+	case '#':
+		(void)play_file(&win->play_dtmf_tone, baresip_player(),
+			"soundroute.wav", -1, cfg->audio.alert_mod,
+			cfg->audio.alert_dev);
+		break;
```
alfredh, line 286: this block can be simplified by printing the dtmf key to a string, and then calling play_file once.

mbattista: True. That makes it way cleaner.

PR #1012, opened 2020-06-12
GTK notification buttons do not work on Systems

mbattista:
On my KDE System I see the notification, but the accept / reject buttons will not work.

While debugging I found, that the call is not found after the pointer to int conversation.
With this PR i could make the buttons work and should not break any other Systems.

### DIFF
--- a/modules/gtk/gtk_mod.c
+++ b/modules/gtk/gtk_mod.c
@@ -71,25 +71,17 @@ static void reject_activated(GSimpleAction *, GVariant *, gpointer);
 static void denotify_incoming_call(struct gtk_mod *, struct call *);
 
 static GActionEntry app_entries[] = {
-	{"answer", answer_activated, "x", NULL, NULL, {0} },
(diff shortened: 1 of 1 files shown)


robert-scheck:
Behaviour for me without this patch:

 * MATE on RHEL 7 (x86_64): Accept/Reject buttons in GNotification work as expected
 * MATE or GNOME 3 on Fedora 32 (x86_64): Accept/Reject buttons in GNotification do not work at all

Behaviour for me with this patch:

 * MATE on RHEL 7 (x86_64): Accept/Reject button in GNotification work as expected (like before)
 * MATE or GNOME 3 on Fedora 32 (x86_64): Accept/Reject buttons in GNotification work as expected (yay!)

Thanks also to Sebastian Keller aka 'borschty' on Freenode IRC in #GNOME for his time to debug the DBus-related GNotification issue with me and also for the valuable input for the patch improvement.

mbattista:
Changed the functions again, so that the call is now stored and searched with the call_id.

PR #1013, opened 2020-06-14
Added call_state() API function that returns enum state of the call.

juha-h:


include/baresip.h
```diff
@@ -149,6 +149,18 @@ enum call_event {
...
+enum state {
+	STATE_IDLE = 0,
+	STATE_INCOMING,
+	STATE_OUTGOING,
+	STATE_RINGING,
+	STATE_EARLY,
+	STATE_ESTABLISHED,
+	STATE_TERMINATED,
+	STATE_UNKNOWN
+};
```
alfredh, line 162: this type belongs to "call.c" so it should be prefixed with call_ and CALL_:

enum call_state;

CALL_STATE_IDLE,
etc.

PR #1014, opened 2020-06-15
1. Changed "jack:" to "jacksrc:" and "jackdst:" for info and warnings

eafunk:
1. Changed "jack:" to "jacksrc" and "jackdst:" for info and warnings
generated by either jack_src.c and jack_play.c, so that interested ui
watchers and determine if the messages are from a new jack source or
destination.
2. Always send an info message when a new jack source or destination is
port created, allowing a ui watcher to manage jack connections.  The
previous code only send port names on the second or later port creation
which prevented the first calls first port name from being captured by a ui
watcher for connection managment.

### DIFF
--- a/modules/jack/jack_play.c
+++ b/modules/jack/jack_play.c
@@ -67,7 +67,7 @@ static void auplay_destructor(void *arg)
 {
 	struct auplay_st *st = arg;
 
-	info("jack: destroy\n");
+	info("jacksrc: destroy\n");
 
 	if (st->client)
 		jack_client_close(st->client);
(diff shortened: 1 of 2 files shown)


eafunk:
NOTE: jacksrc (source) and jackdst (destination) are in reference to the jack audio data flow: audio flows into a destination port, and out of a source port.

alfredh:
I am sorry but this PR is rejected.

The debug text should be prefixed with the name of the module (jack).

eafunk:
Would it be acceptable to leave "jack:" as is, and to instead change
the single message: 
info("jack: unique source name `%s' assigned\n", client_name);
and for the other direction of audio flow:
info("jack: unique destination name `%s' assigned\n", client_name);
So that my application can parse this message from the UI and determine
which jack clients go with which calls, and which client names contain
the source and destination ports?
Is is also acceptable to pull  the above message code out of the if
statement, so that is is always passed to the UI when a jack client is
generated, regardless of the jack status unique name flag?
Thanks,-Ethan

PR #1019, opened 2020-06-19
Again, modifying info message on jack client creation

eafunk:
Simpler version of my previous request to modify info message generation when a new jack client is created:
1. Always message the new client name, even if this is the first jack client created.
2. Specify in the message the audio flow direction as source (audio flow out of client) or destination (audio flow into client).

Unlike my previous pull request, this info change is only for the new client creation message and more closely conforms to the project info message format.

### DIFF
--- a/modules/jack/jack_play.c
+++ b/modules/jack/jack_play.c
@@ -108,10 +108,8 @@ static int start_jack(struct auplay_st *st)
 	if (status & JackServerStarted) {
 		info("jack: JACK server started\n");
 	}
-	if (status & JackNameNotUnique) {
-		client_name = jack_get_client_name(st->client);
-		info("jack: unique name `%s' assigned\n", client_name);
-	}
+	client_name = jack_get_client_name(st->client);
+	info("jack: source unique name `%s' assigned\n", client_name);
 
 	jack_set_process_callback(st->client, process_handler, st);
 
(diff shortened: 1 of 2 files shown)


modules/jack/jack_play.c
```diff
@@ -108,10 +108,8 @@ static int start_jack(struct auplay_st *st)
 	if (status & JackServerStarted) {
 		info("jack: JACK server started\n");
 	}
-	if (status & JackNameNotUnique) {
-		client_name = jack_get_client_name(st->client);
-		info("jack: unique name `%s' assigned\n", client_name);
-	}
+	client_name = jack_get_client_name(st->client);
+	info("jack: source unique name `%s' assigned\n", client_name);
```
alfredh, line 112: changing the info text is fine.

but why did you change this part `if (status & JackNameNotUnique)`

eafunk: This is to allow the message to be posted every time baresip creates a new jack client.  Because the first jack client created by baresip likely has no name conflict with any other jack clients on the jack session at the time, it's JackNameNotUnique is NOT set, and the info message for the name of that client is not posted.  Subsequent jack clients created by baresip will have jack name conflict with the first client create by baresip, so jack appends a -1, -2, etc. to the name and sets the JackNameNotUnique flag so that baresip can note that the actual name assigned is slightly different than the one requested.  It is very useful (to me) to always get the jack client name assigned via an info message, when ever baresip creates a new jack client, even if it is the requested name without the added -1, -2, etc.

PR #1038, opened 2020-07-15
Add avfilter module/command

mkurkov:
Hi,

Here is my attempt to add FFmpeg filter graphs integration to baresip.

Idea is to allow dynamically set video filter graph for current video stream. And then remove it when needed.
It's implemented as baresip filter module and command `/avfilter`. For example to enable `vflip` simple filter graph just call `/avfilter vflip` at any time and see how video is flipped. To disable filter graph just send `/avfilter` command without any filter string.
As more complex example you can apply some png overlay on top of current video stream with
`/avfilter movie=/path/to/some/image.png[pic];[in][pic]overlay=(W-w)/2:H-h-20` command.
Full syntax and list of filters are available on FFmpeg Filtegraph's page: https://ffmpeg.org/ffmpeg-filters.html#Filtergraph-description

I run into several issues during development:

### 1. Bottom-up FFmpeg frames.

It seems that FFmpeg frame structure could keep both top-down and bottom-up plane lines orders. In the last case
linesizes are negative and data pointers point to the last rows in the buffer. Baresip uses unsigned linesizes, so as I understand
it doesn't support bottom-up order. As workaround I had to rearrange plane lines in top-down order for such frames and
update linesizes/data pointers. So my question is how Baresip works with bottom-up frames and is my approach valid.

### 2. Moving data between AVFrame/vidframe structures

I wonder how properly work with FFmpeg AVFrame structure. Currently as I get new AVFrame from filter graph pipeline, I just
update incoming filter frame, setting data pointers and linesizes to AVframe data. It seems to work well and without any additional copying. Main question is when I supposed to unref AVFrame structure. Normally it should be done in the end of encoding pipeline, but filter module sits in the middle of pipeline and doesn't control it. Now I unref previous AVFrame before processing new one, assuming that previous encoding cycle is finished and it's safe to clean memory. Am I right or there is a more proper way to do it?

I don't have much experience with baresip modules and video processing, so any suggestions are greatly appreciated.

Thanks,
Mikhail

### DIFF
--- a/modules/avfilter/avfilter.c
+++ b/modules/avfilter/avfilter.c
@@ -41,7 +41,7 @@
  *
  */
 
-static struct lock *lock = (void*)0;
+static struct lock *lock;
 static char filter_descr[MAX_DESCR] = "";
 static bool filter_updated = false;
 
@@ -142,7 +142,7 @@ static int module_init(void)
 	int err;
 	err = lock_alloc(&lock);
 	if (err)
-		return ENOMEM;
+		return err;
 
 	vidfilt_register(baresip_vidfiltl(), &avfilter);
 	return cmd_register(baresip_commands(), cmdv, ARRAY_SIZE(cmdv));
@@ -151,7 +151,7 @@ static int module_init(void)
 
 static int module_close(void)
 {
-	mem_deref(lock);
+	lock = mem_deref(lock);
 	vidfilt_unregister(&avfilter);
 	cmd_unregister(baresip_commands(), cmdv);
 	return 0;
--- a/modules/avfilter/avfilter.c
+++ b/modules/avfilter/avfilter.c
@@ -125,6 +125,7 @@ static int avfilter_command(struct re_printf *pf, void *arg)
 	return 0;
 }
 
+
 static struct vidfilt avfilter = {
 	.name    = "avfilter",
 	.ench    = encode,
--- a/modules/avfilter/avfilter.c
+++ b/modules/avfilter/avfilter.c
@@ -17,14 +17,16 @@
 /**
  * @defgroup avfilter avfilter
  *
- * Apply FFmpeg filter graph to outcoming stream
+ * Video filters using libavfilter
  *
+ * This module allows to dynamically apply complex video filter graphs
+ * to outcoming stream using libavfilter from FFmpeg project.
  *
  * Commands:
  *
  \verbatim
- avfilter [filter_string] Apply avfilter to outcoming stream
- avfilter [Empty filter_string] Stop avfilter
+ avfilter <FILTER> - Enable avfilter for outcoming stream
+ avfilter          - Disable avfilter
  \endverbatim
  *
  * Example:
@@ -32,6 +34,11 @@
  \verbatim
  avfilter movie=watermark.png[pic];[in][pic]overlay=10:10[out]
  \endverbatim
+ *
+ * References:
+ *
+ *     https://ffmpeg.org/ffmpeg-filters.html
+ *
  */
 
 static struct lock *lock = (void*)0;
@@ -49,8 +56,8 @@ static void st_destructor(void *arg)
 
 
 static int update(struct vidfilt_enc_st **stp, void **ctx,
-	const struct vidfilt *vf, struct vidfilt_prm *prm,
-	const struct video *vid)
+		  const struct vidfilt *vf, struct vidfilt_prm *prm,
+		  const struct video *vid)
 {
 	struct avfilter_st *st;
 	(void)vid;
@@ -73,7 +80,7 @@ static int update(struct vidfilt_enc_st **stp, void **ctx,
 
 
 static int encode(struct vidfilt_enc_st *enc_st, struct vidframe *frame,
-	uint64_t *timestamp)
+		  uint64_t *timestamp)
 {
 	struct avfilter_st *st = (struct avfilter_st *)enc_st;
 	int err=0;
@@ -106,8 +113,8 @@ static int avfilter_command(struct re_printf *pf, void *arg)
 	if (str_isset(carg->prm)) {
 		str_ncpy(filter_descr, carg->prm, sizeof(filter_descr));
 		info("avfilter: enabled for %s\n", filter_descr);
-  }
-  else {
+	}
+	else {
 		str_ncpy(filter_descr, "", sizeof(filter_descr));
 		info("avfilter: disabled\n");
 	}
@@ -119,8 +126,8 @@ static int avfilter_command(struct re_printf *pf, void *arg)
 }
 
 static struct vidfilt avfilter = {
-	.name = "avfilter",
-	.ench = encode,
+	.name    = "avfilter",
+	.ench    = encode,
 	.encupdh = update
 };
 
--- a/modules/avfilter/avfilter.c
+++ b/modules/avfilter/avfilter.c
@@ -1,0 +1,159 @@
+/**
+ * @file avfilter.c	 Video filter using libavfilter
+ *
+ * Copyright (C) 2020 Mikhail Kurkov
+ */
+
+#include <string.h>
+#include <libavformat/avformat.h>
+#include <libavfilter/buffersink.h>
+#include <libavfilter/buffersrc.h>
+#include <libavutil/opt.h>
+#include <re.h>
+#include <rem.h>
+#include <baresip.h>
+#include "avfilter.h"
+
+/**
+ * @defgroup avfilter avfilter
+ *
+ * Apply FFmpeg filter graph to outcoming stream
+ *
+ *
+ * Commands:
+ *
+ \verbatim
+ avfilter [filter_string] Apply avfilter to outcoming stream
+ avfilter [Empty filter_string] Stop avfilter
+ \endverbatim
+ *
+ * Example:
+ *
+ \verbatim
+ avfilter movie=watermark.png[pic];[in][pic]overlay=10:10[out]
+ \endverbatim
+ */
+
+static struct lock *lock = (void*)0;
+static char filter_descr[MAX_DESCR] = "";
+static bool filter_updated = false;
+
+
+static void st_destructor(void *arg)
+{
+	struct avfilter_st *st = arg;
+
+	list_unlink(&st->vf.le);
+	filter_reset(st);
+}
+
+
+static int update(struct vidfilt_enc_st **stp, void **ctx,
+	const struct vidfilt *vf, struct vidfilt_prm *prm,
+	const struct video *vid)
+{
+	struct avfilter_st *st;
+	(void)vid;
+
+	if (!stp || !ctx || !vf || !prm)
+		return EINVAL;
+
+	if (*stp)
+		return 0;
+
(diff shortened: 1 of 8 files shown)


modules/avfilter/avfilter.c
```diff
@@ -0,0 +1,166 @@
...
+ avfilter movie=watermark.png[pic];[in][pic]overlay=10:10[out]
+ \endverbatim
+ *
+ * References:
+ *
+ *     https://ffmpeg.org/ffmpeg-filters.html
+ *
+ */
+
+static struct lock *lock = (void*)0;
```
alfredh, line 44: no need to initialise lock, static/global vars are inited to null.

modules/avfilter/avfilter.c
```diff
@@ -0,0 +1,166 @@
...
+	{"avfilter", 0, CMD_PRM, "Start avfilter", avfilter_command}
+};
+
+
+static int module_init(void)
+{
+	int err;
+	err = lock_alloc(&lock);
+	if (err)
+		return ENOMEM;
```
alfredh, line 145: you should return `err` here, instead of ENOMEM

modules/avfilter/avfilter.c
```diff
@@ -0,0 +1,166 @@
...
+
+	vidfilt_register(baresip_vidfiltl(), &avfilter);
+	return cmd_register(baresip_commands(), cmdv, ARRAY_SIZE(cmdv));
+}
+
+
+static int module_close(void)
+{
+	mem_deref(lock);
```
alfredh, line 154: also setting lock to NULL makes it more robust:

`lock = mem_deref(lock)`

modules/avfilter/filter.c
```diff
@@ -0,0 +1,195 @@
...
+#include <libavformat/avformat.h>
+#include <libavfilter/buffersink.h>
+#include <libavfilter/buffersrc.h>
+#include <libavutil/opt.h>
+#include <re.h>
+#include <rem.h>
+#include <baresip.h>
+#include "avfilter.h"
+#include "util.h"
+
```
alfredh, line 17: 2 empty lines between functions

modules/avfilter/filter.c
```diff
@@ -0,0 +1,195 @@
...
+	    !st->vframe_in || !st->vframe_out) {
+		err = AVERROR(ENOMEM);
+		goto end;
+	}
+
+	/* buffer video source */
+	snprintf(args, sizeof(args),
+		 "video_size=%dx%d:pix_fmt=%d:"
+		 "time_base=%d/%d:pixel_aspect=1/1",
+		 frame->size.w, frame->size.h, src_format, 1, 25);
```
alfredh, line 50: time_base = 1/25 is hardcoded.

if you want to use the timestamps from the API you can use 1/1000000 as time_base.
(or `VIDEO_TIMEBASE` constant)

modules/avfilter/filter.c
```diff
@@ -0,0 +1,195 @@
...
+	info("avfilter: filter graph initialized for %s\n", filter_descr);
+
+ end:
+	avfilter_inout_free(&inputs);
+	avfilter_inout_free(&outputs);
+
+	return err;
+}
+
+int filter_reset(struct avfilter_st* st)
```
alfredh, line 114: 2 empty lines between functions

modules/avfilter/filter.c
```diff
@@ -0,0 +1,195 @@
...
+	if (st->filter_graph)
+		avfilter_graph_free(&st->filter_graph);
+	if (st->vframe_in)
+		av_frame_free(&st->vframe_in);
+	if (st->vframe_out)
+		av_frame_free(&st->vframe_out);
+	st->enabled = false;
+	info("avfilter: filter graph reset\n");
+	return 0;
+}
```
alfredh, line 127: since filter_reset always returns 0, you can change the return type to `void`

modules/avfilter/filter.c
```diff
@@ -0,0 +1,195 @@
...
+
+bool filter_valid(struct avfilter_st* st, struct vidframe *frame)
```
alfredh, line 129: the two input arguments can be made const

modules/avfilter/util.c
```diff
@@ -0,0 +1,92 @@
...
+
+#include <string.h>
+#include <re.h>
+#include <rem.h>
+#include <baresip.h>
+#include <libavutil/frame.h>
+#include "util.h"
+
+static int swap_lines(uint8_t *a, uint8_t *b, size_t size) {
+	uint8_t tmp[size];
```
alfredh, line 15: I think it is dangerous to declare a variable size array on the stack.

probably better to use a heap buffer, or a fixed size stack buffer

modules/avfilter/util.c
```diff
@@ -0,0 +1,92 @@
...
+	memcpy(tmp, a, size);
+	memcpy(a, b, size);
+	memcpy(b, tmp, size);
+	return 0;
+}
+
+static int reverse_lines(uint8_t *data, int linesize, int count) {
```
alfredh, line 22: curly brace at the end `{` should be moved to next line

modules/avfilter/util.c
```diff
@@ -0,0 +1,92 @@
...
+	for (int i = 0; i < count/2; i++)
+		swap_lines(data + linesize * i,
+			   data + linesize * (count - i - 1),
+			   abs(linesize));
+	return 0;
+}
+
+/*
+ * Sometimes AVFrame contains planes with lines in bottom-up order.
+ * Then linesize is negative and data points to the last row in buffer.
```
alfredh, line 32: this sounds strange, I have never seen a negative linesize. Which plane was it (0 to 3) ?

the linesize is the number of pixels in a horizontal line, for each plane.
example: [320, 160, 160, 0]

modules/avfilter/util.c
```diff
@@ -0,0 +1,92 @@
...
+
+				return EPROTO;
+			}
+		}
+	}
+
+	return 0;
+}
+
+enum AVPixelFormat vidfmt_to_avpixfmt(const enum vidfmt fmt)
```
alfredh, line 68: const is not needed

modules/avfilter/util.c
```diff
@@ -0,0 +1,92 @@
...
+
+	case VID_FMT_YUV420P: return AV_PIX_FMT_YUV420P;
+	case VID_FMT_YUV444P: return AV_PIX_FMT_YUV444P;
+	case VID_FMT_NV12:    return AV_PIX_FMT_NV12;
+	case VID_FMT_NV21:    return AV_PIX_FMT_NV21;
+	default:              return AV_PIX_FMT_NONE;
+	}
+}
+
+enum vidfmt avpixfmt_to_vidfmt(const enum AVPixelFormat pix_fmt)
```
alfredh, line 81: const is not needed

src/config.c
```diff
@@ -806,6 +806,7 @@ int config_write_template(const char *file, const struct config *cfg)
 	(void)re_fprintf(f, "#module\t\t\t" "snapshot" MOD_EXT "\n");
 	(void)re_fprintf(f, "#module\t\t\t" "swscale" MOD_EXT "\n");
 	(void)re_fprintf(f, "#module\t\t\t" "vidinfo" MOD_EXT "\n");
+  (void)re_fprintf(f, "#module\t\t\t" "avfilter" MOD_EXT "\n");
```
alfredh, line 809: use TAB for indentation

alfredh:
hi @mkurkov many thanks for this good and useful piece of work. we will merge it to master.

regarding bottom-up frames, I have never seen negative linesize in the wild.
baresip linesize is only unsigned.

regarding copying of buffers, it is fine to update data pointers in struct vidframe to point
to the real buffers. the buffer should be valid until the vidfilt encode function has returned,
because the application will use the data. The next time the vidfilt encode function is called,
they can be overwritten with new pointers.

I tested this with ffmpeg 4.1.4 and it works just fine:

```
$ ./baresip -m avfilter.so -e/vidloop -e"/avfilter boxblur"
```

I added some review comments inline, please have a look

alfredh:
please also add your name and the new module to this list:

modules/avfilter/util.c
mkurkov: For bottom-up frames all planes have negative linesizes, like this:  `linesize = {-1920, -960, -960, 0, 0, 0, 0, 0}` (this is from gdb inspecting AVFrame structure).

mkurkov:
@alfredh thanks for reviewing, will fix mentioned issues.

Regarding bottom-up frames, I run into them trying simple filters like `hflip` - it worked properly, but with `vflip` I constantly got segfault. Looking into AVFrame from `vflip` in gdb I saw this:
```C
{data = {0x7fffe7675cc0 '\353' <repeats 199 times>, <incomplete sequence \353>..., 
    0x7fffe76f85a0 '\200' <repeats 200 times>..., 0x7fffe7777dc0 '\200' <repeats 200 times>..., 0x0, 0x0, 0x0, 0x0, 0x0}, 
  linesize = {-1920, -960, -960, 0, 0, 0, 0, 0}, extended_data = 0x7fffec0e7600, width = 1920, height = 1080,
```
I guess vflip switches frame to bottom-up as optimization. Looking at vflip [implementation](https://ffmpeg.org/doxygen/trunk/vf__vflip_8c_source.html), it's visible that they also change data to point on last row in buffer:
```C
static int filter_frame(AVFilterLink *link, AVFrame *frame)
{
     FlipContext *flip = link->dst->priv;
     int i;
 
     for (i = 0; i < 4; i ++) {
         int vsub = i == 1 || i == 2 ? flip->vsub : 0;
         int height = FF_CEIL_RSHIFT(link->h, vsub);
 
         if (frame->data[i]) {
             frame->data[i] += (height - 1) * frame->linesize[i];
             frame->linesize[i] = -frame->linesize[i];
         }
     }
 
     return ff_filter_frame(link->dst->outputs[0], frame);
}
```
Not sure if this could be a problem in other parts of Baresip, but avfilter needs to support bottom-up frames too.

mkurkov:
@alfredh issues should be resolved in last commits.

alfredh:
thanks

I get an error when compiling now:

```
  CC [M]  build-x86_64/modules/avfilter/filter.o
modules/avfilter/filter.c:119:3: error: void function 'filter_reset' should not return a value
      [-Wreturn-type]
                return 0;
                ^      ~
1 error generated.
make: *** [build-x86_64/modules/avfilter/filter.o] Error 1
```

alfredh:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20170806190921.10897-1-cus@passwd.hu/

librem/baresip supports only positive linesize:

```c
/** Video frame */
struct vidframe {
	uint8_t *data[4];      /**< Video planes        */
	uint16_t linesize[4];  /**< Array of line-sizes */
	struct vidsz size;     /**< Frame resolution    */
	enum vidfmt fmt;       /**< Video pixel format  */
};
```

this has worked fine for many years.

adding support for negative linesize, would mean updating and testing
all modules and code that is involved with video processing.

I think for now it is best that you work around it in avfilter.so

alfredh:
thanks a lot for your contribution :)

PR #1039, opened 2020-07-16
Add cmake build system

michaelmalonenz:
Add enough of a cmake build to compile a useful library.  It is not
exhaustive and will require some modification for various systems.
This has been tested using Linux ARM, x86_64 and Android ndk r21.

It assumes that `re` and `rem` are provided / built by a parent `CMakeLists.txt` which might be a sticking point here.  i.e. they are believed to appear from nowhere without having actually "found" them.

It does include the static module magic which should be a solid start for iOS builds.


alfredh:
hi @michaelmalonenz many thanks for your PR.

in the past I have been sceptical about adding extra build-systems.
but if you can commit to maintain it in baresip, we can always add it.

some comments:

- please put all the new makefiles under `mk/`

testing with cmake 3.17.2 does not work:

```
-- The C compiler identification is AppleClang 9.0.0.9000039
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Found PkgConfig: /opt/local/bin/pkg-config (found version "0.29.2") 
-- Checking for module 'libpulse-simple'
--   Found libpulse-simple, version 13.0-rebootstrapped
-- Configuring done
CMake Error at CMakeLists.txt:70 (add_library):
  Cannot find source file:

    /Users/alfredh/tmp/baresip/src/floorctrl.c

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx


CMake Error at CMakeLists.txt:70 (add_library):
  No SOURCES given to target: baresip


CMake Generate step failed.  Build files cannot be regenerated correctly.
```

michaelmalonenz:
Oh dear, I appear to have tested against our branch with modifications.

If it's just me, then I can't commit to maintaining the cmake files. My contract with the company who's working with baresip will finish before the end of the year, and I probably won't look at it thereafter. But, as mentioned, the team is still using it (including said cmake files) so I'll ask on Monday if they're willing to take it on.

michaelmalonenz:
OK, I've chatted to the team and because the cmake (as it stands) is currently very specific to our system, we don't think it's likely to be useful to anyone else just yet.  i.e. the defines are hard-coded instead of read from autoconf.  So we'll close this PR, but have made it an internal goal that every time we touch the build we'll try to generalise it further with the intent to upstream it at some point in the future.

PR #1043, opened 2020-07-22
fix search path for avdevice.h

alberanid:
The first search path for avdevice.h should look in the "avdevice" directory, instead of "avformat".


alfredh:
Thank You :)

PR #1045, opened 2020-07-22
fixes #1044: correctly link against ffmpeg libraries

### LINKED ISSUE #1044: fail to load avcodec and avformat modules on ARM
I have compiled (in a Yocto environment for an ARM target, using gcc 5.4) Bareip 0.6.6 using the av* libraries from ffmpeg version 3.2.4 and it fails to load the modules:
```
# baresip
baresip v0.6.6 Copyright (C) 2010 - 2020 Alfred E. Heggestad et al.
Local network address:  IPv4=eth1|192.168.1.99 
aucodec: PCMU/8000/1
aucodec: PCMA/8000/1
aucodec: L16/48000/2
aucodec: L16/44100/2
aucodec: L16/32000/2
aucodec: L16/16000/2
aucodec: L16/8000/2
aucodec: L16/48000/1
aucodec: L16/44100/1
aucodec: L16/32000/1
aucodec: L16/16000/1
aucodec: L16/8000/1
ausrc: alsa
auplay: alsa
dl: mod: /usr/lib/baresip/modules/avcodec.so (/usr/lib/baresip/modules/avcodec.so: undefined symbol: avcodec_find_encoder)
module avcodec.so: No such file or directory
dl: mod: /usr/lib/baresip/modules/avformat.so (/usr/lib/baresip/modules/avformat.so: undefined symbol: av_frame_unref)
module avformat.so: No such file or directory
   [...]
```

The compiled modules seems to be unaware of the required libraries (that are available both at compile time and runtime); for example:
```
# objdump -x /usr/lib/baresip/modules/avcodec.so | grep NEED
  NEEDED               libre.so
  NEEDED               libc.so.6
  VERNEED              0x000010c0
  VERNEEDNUM           0x00000001
```

I've tried to compile Baresip setting STATIC=1; it compiled succesfully, but failed on the install:
```
| install -m 0644  /build/tmp-glibc/work/cortexa7hf-neon-divus-linux-gnueabi/baresip/0.6.6-r19/image/usr/lib/baresip/modules
| install: missing destination file operand after '/build/tmp-glibc/work/cortexa7hf-neon-divus-linux-gnueabi/baresip/0.6.6-r19/image/usr/lib/baresip/modules'
| Try 'install --help' for more information.
```

this may or may not be a different bug; however, I've decided to not pursue the static solution and got back looking at the dynamic compiling.

I've fixed it adding `-static -fPIC` to the CFLAGS of both modules.

Now I have:
```
# objdump -x /usr/lib/baresip/modules/avcodec.so | grep NEED
  NEEDED               libavcodec.so.57
  NEEDED               libavutil.so.55
  NEEDED               libre.so
  NEEDED               libc.so.6
  VERNEED              0x000010fc
  VERNEEDNUM           0x00000003

# objdump -x /usr/lib/baresip/modules/avformat.so | grep NEED
  NEEDED               libavformat.so.57
  NEEDED               libavcodec.so.57
  NEEDED               libswresample.so.2
  NEEDED               libavutil.so.55
  NEEDED               libavdevice.so.57
  NEEDED               libre.so
  NEEDED               libc.so.6
  VERNEED              0x00000d60
  VERNEEDNUM           0x00000006

```

and the modules are loaded correctly.

I'll provide a pull request to fix it, but since I've just found the fix and not thought too deeply about all the implications of these changes, you may want to think/test it a little, before the merge.

Some things to consider:
* probably the *-fPIC* is all it's needed, but I've not tried removing the *-static* option
* I'll include in the PR also the same fix for avfilter, even if I'm not using it right now
* I've not tested it on x86
* while I'm using some old gcc and ffmpeg, I don't think that newer versions change anything

alberanid:
Compile avformat, avcodec and avfilter modules so that they
are correctly linked to the ffmpeg libraries.

### DIFF
--- a/modules/avcodec/module.mk
+++ b/modules/avcodec/module.mk
@@ -11,7 +11,7 @@ $(MOD)_SRCS	+= encode.c
 $(MOD)_SRCS	+= h263.c
 $(MOD)_SRCS	+= h265.c
 $(MOD)_SRCS	+= sdp.c
-$(MOD)_CFLAGS	+= -isystem /usr/local/include
+$(MOD)_CFLAGS	+= -isystem /usr/local/include -static -fPIC
 $(MOD)_LFLAGS	+= `pkg-config --libs libavcodec libavutil`
 
 include mk/mod.mk
--- a/modules/avfilter/module.mk
+++ b/modules/avfilter/module.mk
@@ -8,6 +8,7 @@ MOD		:= avfilter
 $(MOD)_SRCS	+= avfilter.c
 $(MOD)_SRCS	+= filter.c
 $(MOD)_SRCS	+= util.c
+$(MOD)_CFLAGS	+= -static -fPIC
 $(MOD)_LFLAGS	+= -lavfilter -lavutil
 
 include mk/mod.mk
--- a/modules/avformat/module.mk
+++ b/modules/avformat/module.mk
@@ -8,6 +8,7 @@ MOD		:= avformat
 $(MOD)_SRCS	+= avformat.c
 $(MOD)_SRCS	+= audio.c
 $(MOD)_SRCS	+= video.c
+$(MOD)_CFLAGS	+= -static -fPIC
 $(MOD)_LFLAGS	+= \
 	`pkg-config --libs libavformat libavcodec libswresample \
 		libavutil libavdevice libavfilter libswscale libpostproc`


juha-h:
They compile fine w/o -static -fPIC on amd64 Linux.  So I don't think it is OK to add those flags for all architectures.

alberanid:
> They compile fine w/o -static -fPIC on amd64 Linux. So I don't think it is OK to add those flags for all architectures.

I think that FFmpeg is by default compiled without -fPIC on some platforms like x86 and with it in others like ARM (not sure if this is still true with recent versions of FFmpeg, though).

Maybe a better option would be to be able to pass it setting CFLAGS using an environment variable or something like that.

alfredh:
I dont think that this change is needed.

can you paste the output of your pkg-config output for your compiler ?

on my machine OSX 10.14 it is:

```
$ pkg-config --cflags libavcodec
-I/usr/local/include

$ pkg-config --libs libavcodec
-L/usr/local/lib -lavcodec
```

please check if those commands on your platform contains -static

you can also pass additional cflags via `EXTRA_CFLAGS`

`make EXTRA_CFLAGS=-static`

alberanid:
@alfredh @juha-h I guess that's correct and you can close this PR.

Probably the best solution is to just document the fact that if FFmpeg is compiled with -fPIC, the same options must be used to compile the baresip modules.

PR #1051, opened 2020-08-02
Add back Registration shortcut for API

lmangani:
It's currently impossible (AFAIK) to add a registration endpoint via `uanew` command and HTTP API. This patch re-implements a shortcut for it (I'm pretty sure it was there earlier on) as well as methods for Deleting and setting answer-mode with no known side effects.


alfredh:
thanks @lmangani 

can you please also update this page with overview of commands:

https://github.com/baresip/baresip/wiki/Commands-registry

alfredh:
long commands are also supported in the HTTP api:

`http://127.0.0.1:8000/?/sysinfo`

I dont think this patch is needed.

lmangani:
@alfredh ACK if that's the case but where is that documented? :) will test and confirm by reopening if i find any shortcomings.

lmangani:
@alfredh `http://127.0.0.1:8000/?/uanew` won't take any parameters for me. Could you confirm with an example?

alfredh:
try to escape the space by using %20:

```
http://127.0.0.1:8000/?/play%20ring.wav
```

PR #1053, opened 2020-08-06
fix compilation without deprecated APIs

neheb:
Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/modules/debug_cmd/debug_cmd.c
+++ b/modules/debug_cmd/debug_cmd.c
@@ -7,6 +7,10 @@
 #include <time.h>
 #ifdef USE_OPENSSL
 #include <openssl/crypto.h>
+#if OPENSSL_API_COMPAT < 0x10100000L && !defined(LIBRESSL_VERSION_NUMBER)
+#define OPENSSL_VERSION SSLEAY_VERSION
+#define OpenSSL_version SSLeay_version
+#endif
 #endif
 #include <re.h>
 #include <baresip.h>
@@ -56,7 +60,7 @@ static int print_system_info(struct re_printf *pf, void *arg)
 
 #ifdef USE_OPENSSL
 	err |= re_hprintf(pf, " OpenSSL:  %s\n",
-			  SSLeay_version(SSLEAY_VERSION));
+			  OpenSSL_version(OPENSSL_VERSION));
 #endif
 
 	return err;


alfredh:
can you please include which OS and openssl version the warning comes from ?

please include the relevant build output.

neheb:
Output is basically the opposite of this: https://travis-ci.org/github/baresip/baresip/jobs/715650996

Tested on OpenWrt with OpenSSL deprecated APIs disabled.

This can be simulated by passing -DOPENSSL_API_COMPAT=0x10100000L and -Werror=implicit-function-declaration .

alfredh:
can you make sure that the patch works with all supported versions of OpenSSL:

- OpenSSL version 1.0.2
- OpenSSL version 1.1.0
- OpenSSL version 1.1.1
- LibreSSL version 2.x

https://github.com/baresip/baresip/wiki/Supported-platforms#supported-versions-of-openssl

modules/debug_cmd/debug_cmd.c
```diff
@@ -56,7 +60,7 @@ static int print_system_info(struct re_printf *pf, void *arg)
 
 #ifdef USE_OPENSSL
 	err |= re_hprintf(pf, " OpenSSL:  %s\n",
-			  SSLeay_version(SSLEAY_VERSION));
+			  OpenSSL_version(OPENSSL_VERSION));
```
alfredh, line 63: is it not better to check for the version that the function was added ?

```c
#if VERSION >= X
OpenSSL_version(OPENSSL_VERSION);
#else
SSLeay_version(SSLEAY_VERSION);
#endif
```

alfredh:
this fails to build on OpenBSD 6.6. with LibreSSL 3.0.2:

```
modules/debug_cmd/debug_cmd.c:11:1: warning: "OPENSSL_VERSION" redefined
In file included from modules/debug_cmd/debug_cmd.c:9:
/usr/include/openssl/crypto.h:329:1: warning: this is the location of the previous definition
```

modules/debug_cmd/debug_cmd.c
neheb: That's done above.

neheb:
> 
> 
> this fails to build on OpenBSD 6.6. with LibreSSL 3.0.2:
> 
> ```
> modules/debug_cmd/debug_cmd.c:11:1: warning: "OPENSSL_VERSION" redefined
> In file included from modules/debug_cmd/debug_cmd.c:9:
> /usr/include/openssl/crypto.h:329:1: warning: this is the location of the previous definition
> ```

I'll fix when I can.

alfredh:
hi @neheb when are you planning to update the PR ?

neheb:
Maybe now?

Issue #1059, opened 2020-08-13
Segfault because of mem_deref(call) in module menu ua_event_handler


cspiel1:
In case of UA_EVENT_CALL_TRANSFER_FAILED, UA_EVENT_AUDIO_ERROR the pointer call given as function parameter is de-referenced.

Other ua_event_handler that use information from call will crash with a segfault if there are called after menus ua_event_handler.

It depends on the order of modules initialization. If a module with an ua_event_handler is loaded after menu, the segfault occurs.

Commits that cause the issue:
commit 7b6b8a05ebf44d566810b3a60b8799b6f8da5ddd
Author: Alfred E. Heggestad <alfred.heggestad@gmail.com>
Date:   Wed Oct 10 17:57:26 2018 +0200

    add UA_EVENT_AUDIO_ERROR (ref #539) (#555)

commit 4dd9bc9783ecb591eebdb421d6fdef548ecb2026
Author: juha-h <jh@tutpro.com>
Date:   Tue Oct 30 11:42:29 2018 +0200

    pass CALL_EVENT_TRANSFER_FAILED event to handler and handle it in menu module (#567)

alfredh:
fixed in commit f42021893713bba85f5730c95545485f436f51a6

PR #1061, opened 2020-08-14
menu: fix segfault in ua_event_handler (#1059)

cspiel1:
A segfault occurs in modules loaded after menu and have an ua_event_handler
that use call object (to collect information about the call). The call object
may not be de-referenced in any ua_event_handler.

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -1273,7 +1273,6 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 
 	case UA_EVENT_CALL_TRANSFER_FAILED:
 		info("menu: transfer failure: %s\n", prm);
-		mem_deref(call);
 		break;
 
 	case UA_EVENT_REGISTER_OK:
@@ -1290,7 +1289,6 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 
 	case UA_EVENT_AUDIO_ERROR:
 		info("menu: audio error (%s)\n", prm);
-		mem_deref(call);
 		break;
 
 	default:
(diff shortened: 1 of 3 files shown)


alfredh:
thanks for the patch.

the patch fixes the problem, but there is one thing that I think we should avoid.

call.c calls mem_deref(call), in general it is bad practice and dangerous to destroy a "self" object
inside the object's code. mem_deref should be done on the outside.

a better solution is to add a call_close() function that closes the call object,
but does not destroy it. sample code:

```c
void call_close(struct call *call)
{
	if (!call)
		return;

	list_unlink(&call->le);
	tmr_cancel(&call->tmr_dtmf);

	call->sess = mem_deref(call->sess);

	set_state(call, CALL_STATE_TERMINATED);

	etc ...
}
```

this function can then be called from ua.c

cspiel1:
Hi Alfred,

right! The call is allocated from ua. So it would be best to deref it in ua. What you suggest with call_close() is already done in the  call_destructor(). I will push again.
r,c

cspiel1:
Hi Alfred,

you are right! The call_destructor() does all these things you suggest for call_close(). Maybe the video_error_handler() shows the solution. That's how it was done before the problematic commits mentioned in #1059.

I will push again.

r,c

alfredh:
is this the final code that you want merged ?

cspiel1:
Yes, if it's ok for you.

alfredh:
yes, great. Thanks for fixing the bug!

cspiel1:
We have much more. ;-) Thanks for review and merging!

PR #1070, opened 2020-08-22
event: send un-register events

cspiel1:
Add UA events for the result of un-registering from a SIP proxy.

This PR is based on the libre PR https://github.com/baresip/re/pull/15.


alfredh:
I dont think this is needed.

The system has no knowledge of "Unregistration success or failure"...

it is simply... unregistered.

if you try to put too much logic into this, bad things may happen.
for example if you try to unregister from a dead proxy, you have to wait until the client transaction
times out. during this period, is the client registered or not registered ?

cspiel1:
That's a good argument. I have to discuss this with my colleagues. As I already stated we have a code for registering to multiple proxies not in parallel as it is done normally, but also in serial. That means a client has a main SIP proxy and one (or multiple) fallback proxies. If the registration to a proxy fails the next fallback proxy is used for a REGISTER. If the main proxy becomes available again, baresip un-registers from the fallback proxy. The Cisco Mode is an extension to the serial mode which adds keep-alive checks to fallback proxies.

We have to check if these modes are still possible without the UNREGISTER states. But I am with you. It should be not important if the un-REGISTER fails or not. I hope, that we will soon send a PR for serial/Cisco mode.

PR #1073, opened 2020-08-24
call, stream, menu: add cmd to set the direction of video stream

cHuberCoffee:
This commit adds a command for the user to specify the video stream direction.
Possible directions:

    none: disable the video at all
    send: the client does not accept a video stream but will send a video steam
    receive: the client will receive and display a video stream but will NOT send a video stream
    duplex: normal bidirectional video call

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -838,6 +838,34 @@ static int call_video_debug(struct re_printf *pf, void *unused)
 }
 
 
+static int set_video_dir(struct re_printf *pf, void *arg)
+{
+	const struct cmd_arg *carg = arg;
+	int err = 0;
+
+	if (0 == str_cmp(carg->prm, sdp_dir_name(SDP_INACTIVE))) {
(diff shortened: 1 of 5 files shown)


modules/menu/menu.c
```diff
@@ -969,6 +997,7 @@ static const struct cmd callcmdv[] = {
 {"statmode",    'S',       0, "Statusmode toggle",    toggle_statmode      },
 {"transfer",    't', CMD_PRM, "Transfer call",        call_xfer            },
 {"video_debug", 'V',       0, "Video stream",         call_video_debug     },
+{"video_dir",     0, CMD_PRM, "Set video direction",  set_video_dir        },
```
cspiel1, line 1000: Maybe we should inform the user about the possible values. E.g.
```
{"video_dir",     0, CMD_PRM, "Set video direction"
    " <inactive|recvonly|sendonly|senrecv>",  set_video_dir}
```

PR #1078, opened 2020-09-04
ua, menue: new command to print certificate issuer and subject

cHuberCoffee:
With these two commands a user can print the issuers and subjects of the certificate chain stored in SIP_certificate


alfredh:
lets do this one after 1.0.0 is released

cHuberCoffee:
I did a rebase because of a typo in the commit msg
also to rerun the travis checker with the merged [baresip/re#18](https://github.com/baresip/re/pull/18) PR

PR #1079, opened 2020-09-04
opengles: remove module (not working)

alfredh:
the module is not working, proposal to remove it unless anyone wants to maintain it ...


magnatesage:
I used for ios same baresip library from the master branch I get the following error

```
Makefile:134: modules/opengles/module.mk: No such file or directory
make[1]: *** No rule to make target `modules/opengles/module.mk'.  Stop.
make: *** [baresip] Error 2
```

could you please help me.
@alfredh

PR #1083, opened 2020-09-09
Serial registration

cspiel1:
Consider a setup with one or more main SIP proxies and one or more fallback proxies.

Serial registration mode
-  Accounts optionally have priorities (default prio=0). Accounts with prio  0 are registered at startup.
-  Prio 1 accounts are fallback accounts and are registered if all of the prio 0 accounts registrations fail.
-  Prio 2 accounts are fallback accounts second stage, and so on.
-  If a (re-)REGISTER fails, then serreg switches to the next priority UA.
 
Cisco mode
 -  Additionally to the serial mode sends Cisco REGISTER keep-alives to not-registered UAs in order to poll their availability. This is only a name for a REGISTER with expires zero, thus a periodic un-REGISTER.
-   If a (re-)REGISTER with positive expires value fails, serreg switches to the next available UA.
-   If a UA with prio lower than the current becomes available again, serreg switches to the UA with the lower prio.

This PR is based on https://github.com/baresip/re/pull/19.


alfredh:
this one can be merged after 1.0.0 is released

cspiel1:
I did a lot of tests with up to 5 instances of routr SIP server running in docker.
1. serial mode (fbregint was not set)
- Switch to next proxy if all of current prio proxies go offline.
- Switch from last prio proxy to prio=0 proxy.
- Multiple proxies with same prio --> parallel registration
- If a prio p is missing and there is prio p-1 and p+1, then if proxy with p-1 goes offline it switches to prio p+1.
- None of the proxies registration is successful. --> prevent fast endless registration loop
- Then switch to first available proxy.
2. Cisco mode (fbregint was set)
- Periodic REGISTER with expire=0 is done with configured fbregint.
- Switch to next proxy if all of current prio proxies go offline. But this time it knows if the fallback is available.
- If current prio is p and all prio p proxies go offline and all p+1 proxies are also offline but p+2 proxy is available, it switch to p+2 proxy. This switch is done immediately, without trying p+1 proxies.
- If current prio is p and p-1 proxy becomes available again (this is detected by the periodic REGISTER with expire 0), it switches back to p-1 proxy.
- Switching from maxprio to prio 0 works also.
- Prevent fast endless loop REGISTER if none of the proxies is available. Then switch to first available proxy.

alfredh:
should be possible to merge now ?

cspiel1:
> should be possible to merge now ?

Yes. I will make travis happy first. I recently find the handy tool "make ccheck". ;-)

alfredh:
many thanks for your contributions :)

cspiel1:
Many thanks for review and merge! :-)

PR #1085, opened 2020-09-11
Ctrl dbus

cspiel1:
A module that enables to control baresip over dbus.

Added also generated source files which has to be excluded from ccheck. That's why there is also a commit for mod.mk and Makefile.

Let me know if you don't want to have the generated files in the git repo! An alternative would be to generate them during the build.


alfredh:
thanks for your contribution.

do you have a simple example of how I can use/test this module ?

cspiel1:
Because of permission problems with the system bus, its easier to use the session bus for a test. So set:
ctrl_dbus_use  session

Then start baresip. Example is already in modules/ctrl_dbus/ctrl_dbus.c doxygen:
gdbus call -e -d com.creytiv.Baresip -o /baresip -m com.creytiv.Baresip.invoke reginfo

alfredh:
I get this error when loading module:

```
ctrl_dbus: name com.creytiv.Baresip acquired on the session-bus bus_owner=1
ctrl_dbus: dbus name lost com.creytiv.Baresip
ctrl_dbus: could not export dbus interface
```

cspiel1:
This is because you don't have a dbus session bus running. E.g. on an ubuntu system this is often started by an xsession. But if you have a system without X or maybe other constellations I don't know, it might be that there is no session bus started.

You can always start a session bus by means of dbus-launch. If you run dbus-launch it prints out two environment variables. You have to export these. See man page of dbus-launch. It shows how this can be done in a one-liner.

```
eval `dbus-launch --sh-syntax`
```

alfredh:
thanks, works fine in Debian 9 👍

PR #1092, opened 2020-09-23
ua: use configurable TCP idle timeout

Naegolus:
The idle timeout of the TCP connection can be changed
by the user if needed.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -55,6 +55,7 @@ rtp_stats		no
 #dns_server		1.1.1.1:53
 #dns_server		1.0.0.1:53
 #net_interface		wlan1
+#tcp_timeout		180
 
 #------------------------------------------------------------------------------
 # Modules
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -277,6 +277,7 @@ struct config_sip {
 	char local[64];         /**< Local SIP Address              */
 	char cert[256];         /**< SIP Certificate                */
 	char cafile[256];       /**< SIP CA-file                    */
+	uint16_t tcp_timeout;   /**< TCP idle timeout               */
 };
 
 /** Call config */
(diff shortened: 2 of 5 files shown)


alfredh:
closing this for now, please create a new PR when the re patch is ready.

IMO I think having a config option to control this is too detailed, it is probably better to find a good value
that works for most cases.

cspiel1:
> closing this for now, please create a new PR when the re patch is ready.
> 
> IMO I think having a config option to control this is too detailed, it is probably better to find a good value
> that works for most cases.

Right! Already prepaired a new PR https://github.com/baresip/baresip/pull/1098. But then I again was thinking about this. The TCP connection might not be the reason for the problem. Remember, our goal was to reduce the delay until the registration succeeds after the network connection comes back from offline.

From https://github.com/baresip/re/pull/23:
"We found in our test setup, that when baresip is offline (in this scenario for 4 minutes), it lasts about 11min until SIP registration is performed again. This patch makes it possible to reduce/set the TCP idle timeout which we found is the reason for the long registration delay."

Currenlty I am looking in sipregs retry mechanism for the reason of the long delay. I saw that failwait() increases the retry delay. But could not find a real failure. My colleagues an me will discuss this and compare this with other SIP stack solutions.

juha-h:
Just to add comment, my baresip Android app doesn't have any re-REGISTER delays due to network connectivity changes, since it gets notifications from Android OS whenever changes in connectivity happen and can thus issue re-REGISTER requests immediately when lost network connectivity is restored.

I don't know if such callbacks are available in Linux.

cspiel1:
Thanks juha-h for this hint! And sure there are messages in kernel log when plugging/unplugging:
```
[1144779.857437] r8169 0000:08:00.0 eth0: link down
[1144805.530065] r8169 0000:08:00.0 eth0: link up
```
I have to investigate further how this information goes into user space, into tcp.c and from their to reg.c.

PR #1097, opened 2020-10-06
Updating .vcxproj file for windows builds

johnjuuljensen:
Added missing files.
Removed HAVE_GETOPT define.
Added dirent.h, since it's not present on windows.
Updated .gitignore to exclude visual studio build files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,9 @@ tags
 
 # Vim swp files
 *.swp
+
+# Window build folder
+Win32/*
+
+# Visual studio config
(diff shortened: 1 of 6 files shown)


johnjuuljensen:
The test fails on the code conventions check of "dirent.h" that I've added.
The license on this version of "dirent.h" is liberal enough, but I don't know if you're interested in having it.
I have no idea how to exclude that file from the ccheck.

alfredh:
I merged the patch to master and removed dirent.h

we should not add copies of OS files in the repo, unless it is absolutely necessary.

PR #1098, opened 2020-10-06
config, ua: add config setting sip_timeout

cspiel1:
A timeout can be specified for detecting TCP connection loss. Default is 900
seconds. Lower boundary is 180 seconds to comply with RFC3261.

This commit is a rework of https://github.com/baresip/baresip/pull/1092.

This commit is related to https://github.com/baresip/re/pull/29.


cspiel1:
baresip/re#29 closed.

PR #1099, opened 2020-10-12
Add a HTTP request module with authorization

cspiel1:
Provides the HTTP request feature of https://github.com/baresip/re/pull/33 in form of a new baresip module "httpreq".
- Makes use of struct network which also was moved to libre.
- GET, POST requests
- TLS
- basic, digest and bearer authentication

PR https://github.com/baresip/re/pull/34 adds struct network to libre.

Travis will be happy if https://github.com/baresip/re/pull/33 and https://github.com/baresip/re/pull/34 are merged.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -152,6 +152,7 @@ module_app		menu.so
 module_app		vidloop.so
 # module_app		ctrl_dbus.so
 # ctrl_dbus_use		system       # system, session
+# module_app		httpreq.so
 
 
 #------------------------------------------------------------------------------
@@ -220,3 +221,4 @@ video_selfview		window # {window,pip}
 
 # EBU ACIP
 #ebuacip_jb_type     	fixed   # auto,fixed
+
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -222,3 +222,11 @@ video_selfview		window # {window,pip}
 # EBU ACIP
 #ebuacip_jb_type     	fixed   # auto,fixed
 
+# HTTP request module
+# httpreq_ca		trusted1.pem
+# httpreq_ca		trusted2.pem
+# httpreq_dns		1.1.1.1
+# httpreq_dns		8.8.8.8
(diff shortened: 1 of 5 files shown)


oej:
You only have a setting for client cert. I would like to see a setting for client key to have them in separate files.

cspiel1:
> You only have a setting for client cert. I would like to see a setting for client key to have them in separate files.

Okay, makes sense!

alfredh:
what is the purpose of this
new module?

cspiel1:
> what is the purpose of this
> new module?

Allows to interact with a remote REST interface.

include/baresip.h
```diff
@@ -333,17 +333,6 @@ struct config_avt {
...
-struct config_net {
-	int af;                 /**< AF_UNSPEC, AF_INET or AF_INET6 */
-	char ifname[64];        /**< Bind to interface (optional)   */
-	struct {
-		char addr[64];
-		bool fallback;
-	} nsv[NET_MAX_NS];      /**< Configured DNS nameservers     */
-	size_t nsc;             /**< Number of DNS nameservers      */
-};
-
```
alfredh, line 346: I dont agree with this change.

This code belongs to baresip. if you need similar code in a different project, it is better
to write the code from scratch or to copy parts of it.

cspiel1: Do you mean you don't want to have "struct network" in libre? I could revert this commit and leave network in baresip.
Or do you mean you don't find the module useful?
Or do you mean that you don't find request.c in libre useful at all?

alfredh:
I dont think that we should move `struct config_net` to re.

the module itself is useful.

the patch itself is quite large, it would be easier to review if the patch was smaller...

cspiel1:
I removed the network commit.
I will try to split up the large commit "httpreq: add HTTP request module" into handsome chunks.

cspiel1:
Thanks Alfred for review and merge!

PR #1102, opened 2020-10-15
Menu: corrections for ring tones and call status by means of a global call counter

cspiel1:
ua: count calls globally for max_calls check
menu: for an outgoing call only play ringback for the current call
menu: remove redundant stop of play tone in cmd_hangup
menu: on call termination resume adequate play tone
menu: stop play tone also for call progress
menu: remove redundant call to menu_set_incall()
menu: correct incall decision for update_callstatus()
menu: do not start ringback tone during established call


alfredh:
this one looks good

cspiel1:
Thank you Alfred!

PR #1107, opened 2020-10-30
qtcapture: remove deprecated module

alfredh:
qtcapture is using QTKit (QuickTime Kit) which was deprecated in OS X v10.9

suggest removing the module, use avcapture.so instead.


alfredh:
see https://github.com/baresip/baresip/wiki/Supported-platforms#supported-platforms-1

PR #1112, opened 2020-11-11
Adaptive jbuf

cspiel1:
audio, stream: put to and get from jitter buffer in two threads
    
    During a call there is the RTP receive thread (A) and the auplay thread (B).
    Previously the two threads had been connected by an aubuf. Thread A put the
    RTP packets in the jbuf and also did a get in the same thread. Thus jbuf did
    not really work as jitter buffer. Thread A did also the decoding and pushed the
    data into the aubuf. Thread B fetched the data from the aubuf and pushed it to
    the audio device.
    
aubuf-2threads

    Now in libre the jbuf was made thread safe and it works now as adaptive jitter
    buffer. What means that the latency is controlled in reaction on the measured
    network jitter.

Thread B has real time constraints and decoding could be to time consuming (e.g. on slow HW). We added a third thread X for decoding. Like shown in this image: 
jbuf-3threads
    
    
    Adapting of the latency in the jbuf is done during periods of silence. This
    commit also adds a silence detection. Which is disabled by default because it
    is the only O(n) complexity part, where n is the number of samples (sampc).

Related to: https://github.com/baresip/re/pull/41.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -143,7 +143,9 @@ struct aurx {
 	const struct aucodec *ac;     /**< Current audio decoder           */
 	struct audec_state *dec;      /**< Audio decoder state (optional)  */
 	struct aubuf *aubuf;          /**< Incoming audio buffer           */
+	size_t aubuf_minsz;           /**< Minimum aubuf size in [bytes]   */
 	size_t aubuf_maxsz;           /**< Maximum aubuf size in [bytes]   */
+	size_t num_bytes;             /**< Size of one frame in [bytes]    */
 	volatile bool aubuf_started;  /**< Aubuf was started flag          */
 	struct auresamp resamp;       /**< Optional resampler for DSP      */
 	struct list filtl;            /**< Audio filters in decoding order */
@@ -167,6 +169,9 @@ struct aurx {
 		uint64_t aubuf_underrun;
 		uint64_t n_discard;
 	} stats;
+
+	struct tmr tmr;       /**< Timer for audio decoding        */
+	volatile int32_t wcnt;
 };
 
 
@@ -289,7 +294,7 @@ static void stop_rx(struct aurx *rx)
 	/* audio player must be stopped first */
 	rx->auplay = mem_deref(rx->auplay);
 	rx->aubuf  = mem_deref(rx->aubuf);
-
+	tmr_cancel(&rx->tmr);
 	list_flush(&rx->filtl);
 }
 
@@ -652,6 +657,28 @@ static bool silence(const void *sampv, size_t sampc, int fmt)
 }
 
 
+static void audio_decode(void *arg)
+{
+	struct audio *a = arg;
+	struct aurx *rx = &a->rx;
+	int err = 0;
+
+	while (rx->wcnt > 0 || err == EAGAIN ||
+			(!err && aubuf_cur_size(rx->aubuf) < rx->num_bytes)) {
+
+		rx->wcnt--;
+		if (err == EAGAIN)
+			++rx->again;
+
+		err = stream_decode(a->strm);
+
+		if (err && err != EAGAIN)
+			break;
+
+	}
+}
+
+
 /*
  * Write samples to Audio Player.
  *
@@ -669,22 +696,12 @@ static void auplay_write_handler(void *sampv, size_t sampc, void *arg)
 	int err = 0;
 	struct audio *a = arg;
 	struct aurx *rx = &a->rx;
-	size_t num_bytes = sampc * aufmt_sample_size(rx->play_fmt);
+	rx->num_bytes = sampc * aufmt_sample_size(rx->play_fmt);
 
-	err = stream_decode(a->strm);
-
-	while (err == EAGAIN ||
-			(!err && aubuf_cur_size(rx->aubuf) < num_bytes)) {
-
-		if (err == EAGAIN)
-			++rx->again;
-
-		err = stream_decode(a->strm);
-	}
-
-	if (rx->aubuf_started && aubuf_cur_size(rx->aubuf) < num_bytes) {
+	if (rx->aubuf_started && aubuf_cur_size(rx->aubuf) < rx->num_bytes) {
 
 		++rx->stats.aubuf_underrun;
+		err = ENOENT;
 
 #if 0
 		debug("audio: rx aubuf underrun (total %llu)\n",
@@ -692,24 +709,27 @@ static void auplay_write_handler(void *sampv, size_t sampc, void *arg)
 #endif
 	}
 
-	aubuf_read(rx->aubuf, sampv, num_bytes);
+	aubuf_read(rx->aubuf, sampv, rx->num_bytes);
 
 	/* Reduce latency after EAGAIN? */
-	if (rx->again && silence(sampv, sampc, rx->play_fmt)) {
+	if (rx->again && (err || silence(sampv, sampc, rx->play_fmt))) {
 
 		rx->again--;
-		if (aubuf_cur_size(rx->aubuf) >= num_bytes) {
-			aubuf_read(rx->aubuf, sampv, num_bytes);
+		if (aubuf_cur_size(rx->aubuf) >= rx->aubuf_minsz) {
+			aubuf_read(rx->aubuf, sampv, rx->num_bytes);
 			debug("Dropped a frame to reduce latency\n");
 		}
 	}
 
 #ifdef USE_SILENCE_DETECTION
 	/* decide if we have silence */
-	if (!err)
-		stream_silence_on(a->strm, silence(sampv, sampc,
-					rx->play_fmt));
+	stream_silence_on(a->strm, err ? true :
+			silence(sampv, sampc, rx->play_fmt));
 #endif
+
+	/* as soon as possible decode aubuf_minsz bytes in polling thread */
+	rx->wcnt++;
+	tmr_start(&rx->tmr, 0, audio_decode, a);
 }
 
 
@@ -1515,6 +1535,7 @@ static int start_player(struct aurx *rx, struct audio *a,
 				return err;
 			}
 
+			rx->aubuf_minsz = min_sz;
 			rx->aubuf_maxsz = max_sz;
 		}
 
--- a/src/audio.c
+++ b/src/audio.c
@@ -170,6 +170,7 @@ struct aurx {
 		uint64_t n_discard;
 	} stats;
 
+	enum jbuf_type jbtype;       /**< Jitter buffer type               */
 	volatile int32_t wcnt;       /**< Write handler call count         */
 
 #ifdef HAVE_PTHREAD
@@ -657,6 +658,39 @@ static void check_telev(struct audio *a, struct autx *tx)
 }
 
 
+/*
+ * Write samples to Audio Player. This version of the write handler is used
+ * for the configuration jitter_buffer_type JBUF_FIXED.
+ *
+ * @note This function has REAL-TIME properties
+ *
+ * @note The application is responsible for filling in silence in
+ *       the case of underrun
+ *
+ * @note This function may be called from any thread
+ *
+ * @note The sample format is set in rx->play_fmt
+ */
+static void auplay_write_handler(void *sampv, size_t sampc, void *arg)
+{
+	struct audio *a = arg;
+	struct aurx *rx = &a->rx;
+	size_t num_bytes = sampc * aufmt_sample_size(rx->play_fmt);
+
+	if (rx->aubuf_started && aubuf_cur_size(rx->aubuf) < num_bytes) {
+
+		++rx->stats.aubuf_underrun;
+
+#if 0
+		debug("audio: rx aubuf underrun (total %llu)\n",
+			rx->stats.aubuf_underrun);
+#endif
+	}
+
+	aubuf_read(rx->aubuf, sampv, num_bytes);
+}
+
+
 static bool silence(const void *sampv, size_t sampc, int fmt)
 {
 	const int16_t *v;
@@ -738,18 +772,13 @@ static void *rx_thread(void *arg)
 
 
 /*
- * Write samples to Audio Player.
+ * Write samples to Audio Player. This version of the write handler is used
+ * for the configuration jitter_buffer_type JBUF_ADAPTIVE.
  *
- * @note This function has REAL-TIME properties
+ * @note See auplay_write_handler()!
  *
- * @note The application is responsible for filling in silence in
- *       the case of underrun
- *
- * @note This function may be called from any thread
- *
- * @note The sample format is set in rx->play_fmt
  */
-static void auplay_write_handler(void *sampv, size_t sampc, void *arg)
+static void auplay_write_handler2(void *sampv, size_t sampc, void *arg)
 {
 	int err = 0;
 	struct audio *a = arg;
@@ -1237,6 +1266,7 @@ int audio_alloc(struct audio **ap, struct list *streaml,
 
 	tx->enc_fmt = cfg->audio.enc_fmt;
 	rx->dec_fmt = cfg->audio.dec_fmt;
+	rx->jbtype  = cfg->avt.jbtype;
 
 	err = stream_alloc(&a->strm, streaml,
 			   stream_prm, &cfg->avt, sdp_sess,
@@ -1622,6 +1652,8 @@ static int start_player(struct aurx *rx, struct audio *a,
 		err = auplay_alloc(&rx->auplay, auplayl,
 				   rx->module,
 				   &prm, rx->device,
+				   rx->jbtype == JBUF_ADAPTIVE ?
+				   auplay_write_handler2 :
 				   auplay_write_handler, a);
 		if (err) {
 			warning("audio: start_player failed (%s.%s): %m\n",
@@ -2383,21 +2415,21 @@ int audio_set_source(struct audio *au, const char *mod, const char *device)
  * Set the audio player state to a new audio player module and device.
  * The current audio player will be stopped.
  *
- * @param au     Audio object
+ * @param a      Audio object
  * @param mod    Audio player module (NULL to stop)
  * @param device Audio player device name
  *
  * @return 0 if success, otherwise errorcode
  */
-int audio_set_player(struct audio *au, const char *mod, const char *device)
+int audio_set_player(struct audio *a, const char *mod, const char *device)
 {
 	struct aurx *rx;
 	int err;
 
-	if (!au)
+	if (!a)
 		return EINVAL;
 
-	rx = &au->rx;
+	rx = &a->rx;
 
 	/* stop the audio device first */
 	rx->auplay = mem_deref(rx->auplay);
@@ -2406,7 +2438,9 @@ int audio_set_player(struct audio *au, const char *mod, const char *device)
 
 		err = auplay_alloc(&rx->auplay, baresip_auplayl(),
 				   mod, &rx->auplay_prm, device,
-				   auplay_write_handler, au);
+				   rx->jbtype == JBUF_ADAPTIVE ?
+				   auplay_write_handler2 :
+				   auplay_write_handler, a);
 		if (err) {
 			warning("audio: set_player failed (%s.%s): %m\n",
 				mod, device, err);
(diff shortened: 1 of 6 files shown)


premultiply:
Just a note:
Received (encoded) RTP packets should be never dropped but decoded audio samples may be dropped or silence may be added if necessary.

The aubuffer is always read with the clock of the local audio output device. So it must underrun or overflow from time to time if the audio output clock does not exactly match the senders input audio clock.
The jitterbuffer does make shure that also delayed packets, packet bursts and reordered packets can be decoded in line.

sreimers:
> So the three thread version would not have any benefits.

In my use case the auplay thread uses samples from 64 to 960 @48khz, i need to test if opus decoding is fast enough to
decode packets with ptime 20ms with deadline 64 samples (real time window lower than 1,3ms@48kHz).

alfredh:
please note that the so called "jitter buffer" has a misleading name, and should
in reality be called a "reordering" buffer to reorder incoming UDP/RTP packets.

the auplay handler code must be very fast, and only copy audio samples from a buffer.

there is a chance that auplay handler may run in the context of a real-time thread,
see Audiounit or Portaudio documentation.

src/audio.c
```diff
@@ -1105,9 +1127,9 @@ int audio_alloc(struct audio **ap, struct list *streaml,
 
 	err = stream_alloc(&a->strm, streaml,
 			   stream_prm, &cfg->avt, sdp_sess,
-			   MEDIA_AUDIO, label,
+			   MEDIA_AUDIO, label, true, ptime,
```
alfredh, line 1130: I dont think that `ptime` should be sent into stream.
ptime is audio specific, and stream is generic.

src/core.h
```diff
@@ -289,6 +289,7 @@ enum {STREAM_PRESZ = 4+12}; /* same as RTP_HEADER_SIZE */
 typedef void (stream_rtp_h)(const struct rtp_header *hdr,
 			    struct rtpext *extv, size_t extc,
 			    struct mbuf *mb, unsigned lostc, void *arg);
+typedef int (stream_pt_h)(uint8_t pt, struct mbuf *mb, void *arg);
```
alfredh, line 292: is it possible to make a separate PR for this change, including the code that is
handling payload-type changes?

src/stream.c
```diff
@@ -502,10 +524,10 @@ int stream_alloc(struct stream **sp, struct list *streaml,
 		goto out;
 
 	/* Jitter buffer */
-	if (cfg->jbuf_del.min && cfg->jbuf_del.max) {
+	if (jbuf && cfg->jbuf_del.min && cfg->jbuf_del.max) {
 
 		err = jbuf_alloc(&s->jbuf, cfg->jbuf_del.min,
-				 cfg->jbuf_del.max);
+				 cfg->jbuf_del.max, cfg->jbuf_wish, ptime);
```
alfredh, line 530: is there a need to send `ptime` to jbuf ?

src/video.c
```diff
@@ -897,9 +893,10 @@ int video_alloc(struct video **vp, struct list *streaml,
 	tmr_init(&v->tmr);
 
 	err = stream_alloc(&v->strm, streaml, stream_prm,
-			   &cfg->avt, sdp_sess, MEDIA_VIDEO, label,
+			   &cfg->avt, sdp_sess, MEDIA_VIDEO, label, false, 0,
```
alfredh, line 896: the video stream also needs a jitter/reordering buffer ?

src/stream.c
cspiel1: Yes, the adaptive jitter buffer algorithm needs ptime. See libre PR! Earlier I had a version where ptime was somehow computed from the RTP timestamps. But it was not very reliable.

src/video.c
cspiel1: Maybe yes. So I should drop the bool jbuf from stream_alloc()?

src/audio.c
cspiel1: The ptime is passed to jbuf. It is the time difference between two RTP packets. For audio it is the same like the frame size. For Video it is the time how long a video frame is viewed.
In this sense, the packet time is stream specific. No matter if video or audio. Right?

cspiel1:
> > So the three thread version would not have any benefits.
> 
> In my use case the auplay thread uses samples from 64 to 960 @48kHz, i need to test if opus decoding is fast enough to
> decode packets with ptime 20ms with deadline 64 samples (real time window lower than 1,3ms@48kHz).

You mean that jbuf_put() is called much more often then jbuf_get()?
I think, you have seen already: The stream_decode() is called on demand and in a while loop until the num_bytes is reached in the auplay write handler.

Our assumption is: If the decoding (e.g. opus) is not fast enough it would also be to slow for the current jbuf and same for the three thread version. At least for single core CPUs. But we can discuss about a three threaded version.

src/core.h
cspiel1: I'll try so. Anyway, I forgot to call s->pth() during preparing the PR.

cspiel1: Currently try to divide the big commit into:
- pt-handler
- silence detection
- stream decode

src/audio.c
```diff
@@ -629,6 +631,27 @@ static void check_telev(struct audio *a, struct autx *tx)
...
+	}
+#else
+	(void) sampv;
+	(void) sampc;
+#endif
+
+	return true;
+}
+
+
```
alfredh, line 654: would it be possible to use this API for silence detection:

`double aulevel_calc_dbov(int fmt, const void *sampv, size_t sampc);`

it supports more audio sample formats, and it also calculates the
correct energy/power level of the signal.

alfredh:
after some thinking I am wondering if we should use the new jbuf only for the audio stream.
the video stream can still use the current jbuf.

also the new code can be placed in baresip/src/. and given a new name.

it might also be possible to add a config item to select the jbuf type used.

The ideal receive pipeline for incoming audio:

```
[ RTP ] -> [ Reordering buffer ] -> [ Decode ] -> [ Adaptive Playout buffer ] -> [ auplay ]
```

The new logic that you have implemented, should be done on the decoded audio samples.

The re-ordering buffer can be e.g. 2-4 slots, and all incoming RTP packets must be decoded.
It can also be skipped if you know that the network has perfect UDP transport.

As @premultiply said, all rtp packets must be decoded.

premultiply:
The "reordering buffer" is more or less a timeout waiting for a delayed, reordered or lost packet which should be as long as the maximum expected jitter to avoid skipping late packets but handle real lost packets on the receiver side by continuing with the next received rtp packet in the buffer in proper order.

cspiel1:
Now pushed again:
- Activated jbuf also for video stream. Tested already.
- ptime is computed from RTP timestamps and works reliable now. So stream_alloc() does'nt need the ptime parameter.
- Split the commit into parts: pt-handler, jbuf with two threads, wish size, silence detection

cspiel1:
> after some thinking I am wondering if we should use the new jbuf only for the audio stream.
> the video stream can still use the current jbuf.
> 
> also the new code can be placed in baresip/src/. and given a new name.
> 
> it might also be possible to add a config item to select the jbuf type used.
> 
> The ideal receive pipeline for incoming audio:
> 
> ```
> [ RTP ] -> [ Reordering buffer ] -> [ Decode ] -> [ Adaptive Playout buffer ] -> [ auplay ]
> ```
> 
> The new logic that you have implemented, should be done on the decoded audio samples.
> 
> The re-ordering buffer can be e.g. 2-4 slots, and all incoming RTP packets must be decoded.
> It can also be skipped if you know that the network has perfect UDP transport.
> 
> As @premultiply said, all rtp packets must be decoded.

- Another buffer means more latency. My solution has the benefit that reorder and jitter control is done in one buffer. The minimum value is to guarantee that reordering can be done. And the maximum value can be chosen relatively high and let the jbuf extend if a high jitter is determined. If the jbuf increases the number of stored packets when the network jitter increases, also the reordering can be done more effectively. 
- Dropping RTP packets is only done to reduce latency if the network condition improves. It is also and already done for late out-of-order packets, for old packets in case of an overflow and anyway it is also possible that packets are lost during the transmission. Anyway, the decoder works also with missing packets.
- The adaptive jitter buffer algorithm needs the RTP header (timestamp, seq-nr), thus the decoded frames would be to less information for computing the network jitter. Please see the libre PR!

A customer of our VOIP products reported audio artifacts and we found that a high network jitter was the problem. But increasing the jitterbuffer minimum value was not applicable because of the high latency. In coordination with our audio specialists I developed the first version of the adaptive jbuf in June 2019. I also compared with gstreamer. Gstrtpjitterbuffer also works with RTP packets as input and output. Also gstreamer has to drop too old packets. After a few improvements we have a stable version in productive systems since this year February.

cspiel1:
Added a commit that avoids packet drops in jbuf.

src/stream.c
```diff
@@ -505,7 +542,7 @@ int stream_alloc(struct stream **sp, struct list *streaml,
 	if (cfg->jbuf_del.min && cfg->jbuf_del.max) {
 
 		err = jbuf_alloc(&s->jbuf, cfg->jbuf_del.min,
-				 cfg->jbuf_del.max);
+				 cfg->jbuf_del.max, cfg->jbuf_wish);
```
alfredh, line 545: perhaps avoid breaking the api by adding a new function, e.g.:

```c
void jbuf_set_wish(struct jbuf *jb, uint32_t wish);
```

cspiel1: I'll add a commit to avoid the push force. Maybe later we should make a interactive rebase before merge.

alfredh:
I would like to suggest that you make a separate PR for the stream pt handler code.

this PR is quite big, and it is difficult to review with many unrelated changes.

as a first step, please make a PR only for the pt handler.
then we merge it to master, and rebase this patch.

src/audio.c
```diff
@@ -643,9 +668,22 @@ static void check_telev(struct audio *a, struct autx *tx)
...
+	while (err == EAGAIN ||
+			(!err && aubuf_cur_size(rx->aubuf) < num_bytes)) {
+
+		if (err == EAGAIN)
+			++rx->again;
+
+		err = stream_decode(a->strm);
+	}
+
```
alfredh, line 686: we should not do audio decode in the audio play handler.

this handler may be called from audio DSP realtime interrupt-handler.

I suggest instead to do decoding in:

1. a new thread
2. a `struct tmr` polling loop

alfredh:
the PT handler patch was merged to master.

please rebase this PR on top of master :)

src/audio.c
cspiel1: Ok, I also agree and will add a commit that does one of this suggestions.

cspiel1: I think both with #ifdef HAVE_PTHREAD, right?

Found that the union around the "pthread_t tid" is obsolete because the AUDIO_MODE_TMR was removed. But I could use the original union approach to select tmr vs thread. Ok?

cspiel1:
Tested with different network jitter values. Tested also the tmr version and the thread version.
jbuf-jitter-50ms

cspiel1:
Some of the commits should be squashed before merging.

src/audio.c
```diff
@@ -172,8 +172,20 @@ struct aurx {
...
+		struct {
+			pthread_t tid;
+			bool run;
+			pthread_cond_t cond;
+			pthread_mutex_t mutex;
+		} thr;                /**< Audio decode thread             */
+#else
+		struct tmr tmr;       /**< Timer for audio decoding        */
+#endif
+	} u;
```
sreimers, line 188: I think with ifdef/else there is only one union member at time, so the union looks a bit needless?

cspiel1: Yes. I'll remove this. Same for the struct autx? I ask already 3 days ago.

src/audio.c
```diff
@@ -629,6 +661,88 @@ static void check_telev(struct audio *a, struct autx *tx)
...
+	struct aurx *rx = &a->rx;
+	int err = 0;
+
+	while (rx->wcnt > 0) {
+		err = stream_decode(a->strm);
+		rx->wcnt--;
+	}
+
+	while (err == EAGAIN ||
+			(!err && aubuf_cur_size(rx->aubuf) < rx->num_bytes)) {
```
sreimers, line 698: I have segfault with call hangup:
```
#0  0x00007fc9c2d02a76 in pthread_rwlock_rdlock () at /usr/lib/libpthread.so.0
#1  0x000055bff209f1b0 in lock_read_get (l=0xb5b5b5b5b5b5b5b5) at src/lock/rwlock.c:66
#2  0x000055bff20add0a in aubuf_cur_size (ab=0x55bff2c78d80) at src/aubuf/aubuf.c:339
#3  0x000055bff1fe5c35 in audio_decode (arg=0x55bff2c695f0) at src/audio.c:698
#4  0x000055bff1fe5d70 in rx_thread (arg=0x55bff2c695f0) at src/audio.c:738
#5  0x00007fc9c2cfe3e9 in start_thread () at /usr/lib/libpthread.so.0
#6  0x00007fc9bf162293 in clone () at /usr/lib/libc.so.6
```

cspiel1: Tried hangup on both sides. For the moment I am not able to reproduce this. It seems that the decoding thread was not stopped cleanly. I should replace two lines in stop_rx(). Like this:
```
		rx->thr.run = false;
		pthread_cond_signal(&rx->thr.cond);
```

sreimers: Looks good now. Thanks.

cspiel1: Ok, also saw the segfault if it was an outgoing call.

sreimers:
with rx_thread I see a very high CPU usage (100% instead of 10% without patches), I haven't looked further (or test tmr), let me know if you need debug information.

cspiel1:
Thanks! Again this is only for outgoing call. I am able to reproduce this.

cspiel1:
Had to move the
```
	pthread_mutex_destroy(&a->rx.thr.mutex);
	pthread_cond_destroy(&a->rx.thr.cond);
```
from stop_rx() to the destructor. Which makes e.g. a difference for outgoing calls.

alfredh:
hi again Christian,

many thanks for your efforts, this is a very nice feature.

I apologise for my forth and back answers here, during the past couple of days I did some
more thinking about this. Please bear with me :)

my biggest worry is that this is a large change, and will break something that we did not think
about, or break something for some production users.

it is also a bit annoying that some code lives in baresip and some in re.

I therefore propose the following:

1. We move the jbuf module from re to baresip/src, perhaps with a new name (e.g. ajbuf)
2. We apply your patch on this new module
3. Add a config item to control which jbuf type is used:
```
enum jbuf_type {
    JBUF_OFF,
    JBUF_FIXED,
    JBUF_ADAPTIVE
};
```
this way users can choose to use the old jbuf method (fixed).

4. After merge, the development continues in baresip/src/ajbuf.c and we can delete re/src/jbuf

This way you can make atomic PR's in baresip that does not depend on other changes.

Please let me know what you think, it is just an idea from my side.

PS: Christian, do you also want to have commit-rights to the baresip project ?

cspiel1:
Hi Alfread, indeed it is a change which goes deep into the core of baresip. I also think there could be done more testing with different codecs, different min/max/wish settings for jbuf and aubuf. Also different network behavior.

Making an ajbuf beside of jbuf and choosing between them is a good idea. But I am not sure and I have to check if it is possible to choose between old and new behavior of stream/audio.
Old behavior: The RTP thread does jbuf_put and jbuf_get and decoding. So also pushes into aubuf.
New behavior: The three threads method as shown in above figure.

On the other side for stream/video I think we still have the old behavior. So maybe a few if-conditions will be enough for stream/audio. So I have to check/try this and add a commit.

But more concretely to the ajbuf:
- The adaptive jbuf is useful for incoming RTP streams in general. Are you sure that we should move it from re to baresip?
- To your point 4.: You suggest to delete re/src/jbuf. Do you mean that from then on the old jbuf and the new ajbuf are living in baresip side-by-side?
- Did you also think about making a baresip module ajbuf? We could simply decide that it has to go into the baresip core. But we could also invent a new filter type and add a "register_jbuf" or "select_jbuf" or "set_jbuf" method for a jitter buffer. This would make it possible for module developer to add different jitter buffer implementations and then choose one of them during runtime.


Thanks that you want to grant me commit rights! The current workflow with PRs and reviews surely is advantageous for the quality. We have to re-think about the usefulness of our patches, improve our changes and try to deliver simple and understandable PRs. You and the other maintainers are very fast and responsible. Your suggestions are helpful and useful. What would be the benefit if I have commit rights and when should I make use of it?

regards, Christian.

src/audio.c
```diff
@@ -112,12 +114,10 @@ struct autx {
...
-	union {
-		struct {
-			pthread_t tid;/**< Audio transmit thread           */
-			bool run;     /**< Audio transmit thread running   */
-		} thr;
-	} u;
+	struct {
+		pthread_t tid;/**< Audio transmit thread           */
+		bool run;     /**< Audio transmit thread running   */
+	} thr;
```
alfredh, line 120: this change is unrelated, I fixed it in master:

4b8a6d4180252ba7662f66da37feecb6fb1bbc31

cspiel1:
Hi Alfred, so the most important question we have to solve before I can proceed is if we should move the jitter buffer from re to baresip. Most annoying for me is that travis fails if the API of re changes. In this case the function jbuf_set_wish() was added. But @sreimers has a good plan to solve this in general for PRs. He wrote:

> It looks like there is already a nice github action for our needs: https://github.com/marketplace/actions/github-action-build-chain-cross-repo-builds (i will take a closer look at this)

So should we really move jbuf to baresip? Ok, it is above re_rtp.h. But usable for any incoming RTP stream. Not only for baresip. Maybe also @sreimers has an opinion about this? Who else?

cspiel1:
Second question: Should jbuf be replace able by different implementations (like ausrc is)? I started to implement an adapter:
```
/** The jitter buffer adapter makes jitter buffer implementations exchange-
 * able.
 * */
struct jbuf_ad {
	struct jbuf *st;
	
	jbuf_set_wish_h * set_wish;
	jbuf_put_h *put;
	jbuf_get_h *get;
	jbuf_set_silence_h *set_silence;
	jbuf_flush_h *flush;
	jbuf_stats_h *stats;
	jbuf_debug_h *debug;
};
...
/* use like this: */
err = s->jbuf->put(s->jbuf->st, hdr, mb);
```
But then I stopped. It seems that to be too much flexibility, which we don't need. And with the old and the new jbuf living in parallel (in re or baresip that doesn't really matter for this question) it would be a lot of code duplication. 

Instead I suggest to add a new method to the new jbuf which select between the old and the new behavior.
```
jbuf_set_method(struct jbuf *jbuf, enum jbuf_type jbtype);
```

@alfredh, please let me know which approach you prefer!

Additionally as mentioned above stream/audio will get the configuration for the jbuf type.

alfredh:
it should be as simple and straightforward as possible.

adding a jbuf abstraction I think is not needed, it is a bit overengineering.

a good solution is to add a jbuf type, that can be used to select the wanted behaviour:

```c
enum jbuf_type {
    JBUF_OFF,
    JBUF_FIXED,
    JBUF_ADAPTIVE
};
```

the default behaviour should be "FIXED".

cspiel1:
Ok, I pushed in re/baresip the selectable jbuf type. Tested all three variants (with incoming/outgoing calls this time).

Let me know if I should move jbuf to baresip! The failing travis should not be the reason for that, because @sreimers is about to solve this.

sreimers:
@cspiel1 I just merged my pull request dependency ci action code. Can you try to rebase from master? It should detect the baresip/re pull request (same github pr title) automatically.

https://github.com/baresip/baresip/pull/1132

cspiel1:
@sreimers Yes! It works. Great solution!

cspiel1:
@alfredh All tests are green now with the new github action of @sreimers . Selecting type JBUF_FIXED vs JBUF_ADAPTIVE work. Also in retest there are now two separate tests for fixed/adaptive. Do you still prefer that I move jbuf to baresip?

I let the PR to draft, because if you are ready with review I want to make a cleaner git history.

alfredh:
thanks. I will take a look at this during the next couple of days ..

cspiel1:
Is it ok, if I start the interactive rebase to squash some of the commits and then push force?

alfredh:
yes, okay :)

alfredh:
merged to master.

Many thanks to Christian and his team for this feature 👍

cspiel1:
You're welcome! Thanks for your review!

PR #1117, opened 2020-11-12
Valgrind ci

sreimers:


.github/workflows/valgrind.yml
```diff
@@ -0,0 +1,42 @@
...
+                            libgtk2.0-dev \
+                            libjack-jackd2-dev \
+                            libmpg123-dev \
+                            libopus-dev \
+                            libpulse-dev \
+                            libsndfile1-dev \
+                            libspandsp-dev \
+                            libssl-dev \
+                            libglib2.0-dev \
+                            valgrind
```
alfredh, line 25: selftest only depends on libbaresip and no modules.
perhaps some -dev packages can be skipped ?

.github/workflows/valgrind.yml
```diff
@@ -0,0 +1,42 @@
...
+    
+    - name: install aac
+      run: |
+        sudo apt-get install -y libfdk-aac-dev
```
alfredh, line 29: aac is probably not needed ..

alfredh:
thanks Sebastian, this is really awesome :)

PR #1119, opened 2020-11-12
modify event type and check if peeruri null 

nanguantong:
modify event type and check if peeruri null


src/ua.c
```diff
@@ -460,6 +462,7 @@ static void call_event_handler(struct call *call, enum call_event ev,
 			break;
 
 		case ANSWERMODE_MANUAL:
+		default:
```
alfredh, line 465: why is this needed ? answer module has 3 values that are all handled in this switch:

```c
enum answermode {
	ANSWERMODE_MANUAL = 0,
	ANSWERMODE_EARLY,
	ANSWERMODE_AUTO
};
```

PR #1125, opened 2020-11-20
Cmd acceptdir

cHuberCoffee:
Implementation of a new command to accept a call. This acceptdir command is designed such that the user controls the media stream directions via arguments of the command.

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -73,6 +73,35 @@ static const char *translate_errorcode(uint16_t scode)
 }
 
 
+/**
+ * Decode a SDP direction
+ *
+ * @param pl  SDP direction as string
+ *
+ * @return sdp_dir SDP direction
+ */
+static enum sdp_dir decode_sdp_enum(struct pl *pl)
+{
+	if (!pl)
+		return SDP_INACTIVE;
+
+	if (!pl_strcmp(pl, "inactive")) {
+		return SDP_INACTIVE;
+	}
+	else if (!pl_strcmp(pl, "sendonly")) {
+		return  SDP_SENDONLY;
+	}
+	else if (!pl_strcmp(pl, "recvonly")) {
+		return SDP_RECVONLY;
+	}
+	else if (!pl_strcmp(pl, "sendrecv")) {
+		return SDP_SENDRECV;
+	}
+
+	return SDP_INACTIVE;
+}
+
+
 static void menu_play(const char *fname, int repeat)
 {
 	struct config *cfg = conf_config();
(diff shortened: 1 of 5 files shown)


juha-h:
Good pull request.  I have had to play lots of tricks in baresip+ app to control call video.

Would it be possible to also show to user when call comes in if the offer includes video and to which direction?  It would help the user to decide how to answer.

juha-h:
Question about this: `save the user decision of the media directions`.

Let's say call comes in and don't want to accept it with video.  Then later during the call, I want to turn on video by re-INVITE or I get re-INVITE with video from the other party.  Is it possible to get video turned on using re-INVITE if I originally accepted or made the call w/o video?

Regarding incoming re-INVITE with video, baresip+ currently replies to it immediately without video.  Then it asks the user, would he/she like to turn video on.  If yes baresip+ sends re-INVITE with video back.  I have not found a better solution, since reply to re-INVITE cannot take long time

cHuberCoffee:
That commit is about saving the users choice, in case of a call on-hold scenario. Since the streams will get new defined media directions. At call resume, the users choice should be restored.

Short answer of your question: Yes a user can enable and set the direction of the video stream via the command "/video_dir <SDP-Direction>". The underlying structure will send a re-INVITE, informing the opponent about the change of this media stream.

An incoming re-INVITE should not influence the users choice about the video stream at the local instance. A re-INVITE is to inform the opponent that some media streams have changed, which does not necessarily mean that the local user wants to have a video stream. So the rdir of the video stream will change, but the ldir will be the same. Depending on ldir in the first place, will decide the behavior of the media stream. Like A initiate the call with video=recvonly and B accept it with video=recvonly. Nothing will happen since nobody is able to send a stream. But with a re-INIVTE of A to B with video=sendonly/sendrecv, B will receive the stream immediately. Again the behavior of B can be changed with the video_dir command.

modules/menu/menu.c
```diff
@@ -73,6 +73,35 @@ static const char *translate_errorcode(uint16_t scode)
...
 
 
+/**
+ * Decode a SDP direction
+ *
+ * @param pl  SDP direction as string
+ *
+ * @return sdp_dir SDP direction
+ */
+static enum sdp_dir decode_sdp_enum(struct pl *pl)
```
alfredh, line 83: input argument can be const

src/call.c
```diff
@@ -1092,7 +1092,6 @@ int call_answer(struct call *call, uint16_t scode, enum vidmode vmode)
 			call->linenum, call->peer_uri, scode);
 
 	if (call->got_offer) {
-
```
alfredh, line 1095: this change is unrelated

src/call.c
```diff
@@ -2350,3 +2349,32 @@ void call_set_current(struct list *calls, struct call *call)
...
 	list_append(calls, &call->le, call);
 }
+
+
+/**
+ * Set stream sdp media line direction attribute
+ *
+ * @param call	Call object
+ * @param a	Audio SDP direction
+ * @param v	Video SDP direction if video available
```
alfredh, line 2359: please align the parameter descriptions

alfredh:
thanks for the PR! see my comments inline..

one unrelated note; the file menu.c is now becoming quite large and it might be good
to split it up into smaller files. for example a new file can implement all call-related commands.

cHuberCoffee:
> Good pull request. I have had to play lots of tricks in baresip+ app to control call video.
> 
> Would it be possible to also show to user in menu when call comes in if the offer includes video and to which direction? It would help the user to decide how to answer.

We added more infos in menu.c at an incoming call to show both the audio and video remote stream directions.

PR #1127, opened 2020-11-21
Added g729 audio module

juha-h:
Added g729 based on version 1.1.0 of https://github.com/BelledonneCommunications/bcg729 library. Debian packages of the
library can be created using debian directory from https://github.com/juha-h/bcg729-deb.

Decoding of audio works fine. Encoding may have some cracks, but they may be due to the built-in mic of my laptop.  Please review, since I don't know much about audio coding.

### DIFF
--- a/mk/modules.mk
+++ b/mk/modules.mk
@@ -24,6 +24,7 @@
 #   USE_G722          G.722 audio codec
 #   USE_G722_1        G.722.1 audio codec
 #   USE_G726          G.726 audio codec
+#   USE_G729          G.729 audio codec
 #   USE_GSM           GSM audio codec
 #   USE_GST           Gstreamer audio module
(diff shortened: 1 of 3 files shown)


juha-h:
Did more tests and looks like encoding does produce cracks.  Have not been able to figure out why.

alfredh:
what about the license ?

bcg729 is GPL license. See here for previous discussions:

https://github.com/baresip/baresip/pull/637

juha-h:
OK, I'll close this pull request and only include the module in my baresip Android app.

alfredh:
IMO we should kindly ask Belledonna communications to re-license it under BSD to the Baresip project.

juha-h:
Such request was rejected already once: https://github.com/baresip/baresip/pull/637#issuecomment-457267742.

PR #1128, opened 2020-11-23
audio, video, stream: check payload type before put to jbuf

cspiel1:


src/core.h
```diff
@@ -361,6 +364,7 @@ void stream_enable_rtp_timeout(struct stream *strm, uint32_t timeout_ms);
 int  stream_jbuf_reset(struct stream *strm,
 		       uint32_t frames_min, uint32_t frames_max);
 bool stream_is_ready(const struct stream *strm);
+int stream_decode(struct stream *strm);
```
sreimers, line 367: stream_decode looks like a leftover from the jbuf pull request

src/stream.c
```diff
@@ -295,8 +295,12 @@ static void rtp_handler(const struct sa *src, const struct rtp_header *hdr,
 		flush = true;
 	}
 
-	if (s->jbuf) {
+	/* payload-type changed? */
+	err = s->pth(hdr->pt, mb, s->arg);
+	if (err)
+		return;
```
alfredh, line 301: I think that the pth handler should only be called, if the pt has changed.
something similar to this logic:

```
	/* Audio payload-type changed? */
	if (hdr->pt != rx->pt) {

		err = update_payload_type(a, rx->pt, hdr->pt);
		if (err)
			return;
	}
```

cspiel1: ok.

src/core.h
cspiel1: Oops. You're right!

src/stream.c
```diff
@@ -295,8 +295,15 @@ static void rtp_handler(const struct sa *src, const struct rtp_header *hdr,
 		flush = true;
 	}
 
-	if (s->jbuf) {
+	/* payload-type changed? */
+	if (s->pt_dec != hdr->pt) {
```
alfredh, line 299: pt = 0 is also a valid payload type. If the first incoming packet is pt=0, this logic will not work.

s->pt_dec should be set to -1, similar to audio.c

PR #1129, opened 2020-11-24
menu: module refactoring

cHuberCoffee:
In the PR #1125 @alfredh mentioned a refactoring of the menu module would make sense.
My proposal is to split it in the following way:
split menu into static and dynamic part
put both parts into separate files
command arrays are put into c files
command registration via register utility functions
ua_event_handler and related funtions stay in menu.c

### DIFF
--- a/modules/menu/dynamic_menu.c
+++ b/modules/menu/dynamic_menu.c
@@ -1,0 +1,289 @@
+/**
+ * @file dynamic_menu.c  dynamic menu related functions
+ *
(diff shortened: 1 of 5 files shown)


alfredh:
really nice, thanks for doing this :)

alfredh:
after merge to master, the build failed:

```
  404  Not Found
```

sreimers:
Should be fixed now. (missing apt-get update)

PR #1138, opened 2020-12-03
mqtt: fix printing port and add tls support

agranig:
The port was printed in a debug message before the default value
got overwritten by our config value, which confuses users if a
port other than the default port is used.

When passing the CA certificate of the broker with the new
mqtt_broker_cafile, then a TLS connection is established to the
broker. Usually used with mqtt_broker_port=8883.


alfredh:
great, thanks!

Issue #1135, opened 2020-12-03
Baresip freeze at hangup


alberanid:
Hi, and thanks for the great project.

We have a problem that may be the same described in #885 

On an armv7l imx6ull single-core 528MHz system with just alsa (version 1.1.3, no pulseaudio), sometimes the **ua_hangup** call seems to freeze at `alsa: stopping playback thread `.
On this system a couple of alsa plugins are running to do echo cancellation; the CPU is loaded at about 30%.

Following your hint, we ve applied this patch to baresip 1.0.0:
```
Index: baresip-1.0.0/modules/alsa/alsa_play.c
===================================================================
--- baresip-1.0.0.orig/modules/alsa/alsa_play.c
+++ baresip-1.0.0/modules/alsa/alsa_play.c
@@ -39,6 +39,7 @@ static void auplay_destructor(void *arg)
 	if (st->run) {
 		debug("alsa: stopping playback thread (%s)\n", st->device);
 		st->run = false;
+		snd_pcm_drop(st->write);
 		(void)pthread_join(st->thread, NULL);
 	}
```

This seems to fix the problem (still testing, but a script with a quick call/hangup cycle is running fine since a day).
A thing that now happens is this:
```
alsa: write error: Input/output error
```

Do you think it's a problem? Despite this, everything seems to work fine.

Another question: do you think the same fix should be one in the `ausrc_destructor` function?

Thanks.

alfredh:
hi,

I think that adding a call to `snd_pcm_drop ` should fix the problem, yes.

Since the stream is closed in the destructor, it should not have any other implications.
You can also ignore the input/output error.

Can you please make a PR with your fix ? I can also help to test it on my Debian box.

Alfred

alberanid:
Done at #1145

Thanks for the feedback!

alfredh:
thanks for the PR. works also fine on Debian 10.

alberanid:
Thanks!

What about the idea to also add a similar line to `ausrc_destructor`?

Issue #1142, opened 2020-12-09
video source is activated even when stream is receive only


juha-h:
I did some /dialdir tests where one baresip calls another with command:
```
/dialdir test audio=sendrecv video=sendonly
```
When test answers using 'a', its video source (including selfview) gets activated even when stream is receive only:
```
sip:test@test.tutpro.com: Incoming call from:  sip:jh@test.tutpro.com - audio-video: sendrecv-recvonly - (press 'a' to accept)
sip:test@test.tutpro.com: Answering incoming call
call: answering call on line 1 from sip:jh@test.tutpro.com with 200
stream: update 'audio'
stream: update 'video'
audio: Set audio decoder: AMR-WB 16000Hz 1ch
audio: player started with sample format S16LE
audio: Set audio encoder: AMR-WB 16000Hz 1ch
webrtc_aec: creating shared state: [16000 Hz, 1 channels, subframe 160 samples, num_bands 1]
webrtc_aec: enabling extended_filter
audio: source started with sample format S16LE
audio tx pipeline:        alsa ---> webrtc_aec ---> AMR-WB
audio rx pipeline:        alsa <--- webrtc_aec <--- AMR-WB
Set video encoder: H264 packetization-mode=0 (500000 bit/s, 25.00 fps)
avcodec: h264 encoder activated
v4l2: /dev/video0: found valid V4L2 device (640 x 480) pixfmt=YUYV
v4l2: Current input: "Camera 1"
v4l2: current framerate is 30.00 fps
```
This looks like a bug to me, but don't know if it is in baresip core on in menu module.

cspiel1:
We will check this.

cHuberCoffee:
The commands (both acceptdir and dialdir) work on the stream and stream handling. We regulate the directions via the SDP media direction, affecting the send and receive path of the streams. Your intention was not to deactivate the sources, but to drop the stream data at `stream_send` and in the `rtp_handler`.
Not starting the video source makes absolutely sense, since it saves resources and prevents the self-view. We check and try to find a solution to stop (or not to start) the corresponding device for the different directions.

juha-h:
There was no video packets on the wire from the called party even when video source was active.  So that part is OK.  As you write (and what was my point), video source should not be started when it is not needed for the call.

alfredh:
please be aware, that if you are using nathelper+rtpproxy and are behind NAT,
if you dont send out any packets on the RTP socket, you will not receive anything either.

[...]

PR #1145, opened 2020-12-12
fix freeze on hangup (#1135)

alberanid:
See discussion at #1135


cspiel1:
1. I think that the snd_pcm_drop() could lead to an unclean end of the stream. That's why we added the snd_pcm_drain() call  in our commit 8e73f5beebe.
2. Maybe it is also a problem that snd_pcm_drop is done in a separate thread. I understand you had to solve the freeze. But does it really work always? See https://www.spinics.net/lists/alsa-devel/msg23556.html!
3. Then there is the new warning as mentioned in https://github.com/baresip/baresip/issues/1142:
```
alsa: write error: File descriptor in bad state
```

Did it freeze in snd_pcm_writei()?

PR #1146, opened 2020-12-14
debug_cmd: mem_deref of player fixes segfault

RobertMi21:


modules/debug_cmd/debug_cmd.c
```diff
@@ -253,6 +253,7 @@ static int module_close(void)
 {
 	cmd_unregister(baresip_commands(), debugcmdv);
 
+	mem_deref(g_play);
```
alfredh, line 256: to make it more robust, the pointer should also be set to null:

```c
g_play = mem_deref(g_play);
```

PR #1151, opened 2020-12-15
video: move video related start/stop/update into video file

cHuberCoffee:
A suggestion how #1142 could be solved only for the video part.
exclusive the NAT topic.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1249,10 +1249,9 @@ int  video_alloc(struct video **vp, struct list *streaml,
 		 video_err_h *errh, void *arg);
 int  video_encoder_set(struct video *v, struct vidcodec *vc,
 		       int pt_tx, const char *params);
-int  video_start_source(struct video *v, struct media_ctx **ctx);
-int  video_start_display(struct video *v, const char *peer);
+int  video_start(struct video *v, struct media_ctx **ctx, const char *peer);
+int  video_update(struct video *v, struct media_ctx **ctx, const char *peer);
 void video_stop(struct video *v);
-void video_stop_display(struct video *v);
 int   video_set_fullscreen(struct video *v, bool fs);
 void  video_vidsrc_set_device(struct video *v, const char *dev);
 int   video_set_source(struct video *v, const char *name, const char *dev);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1249,8 +1249,9 @@ int  video_alloc(struct video **vp, struct list *streaml,
 		 video_err_h *errh, void *arg);
 int  video_encoder_set(struct video *v, struct vidcodec *vc,
 		       int pt_tx, const char *params);
-int  video_start(struct video *v, struct media_ctx **ctx, const char *peer);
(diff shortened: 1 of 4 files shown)


alfredh:
is it possible to fix this without breaking the api ?

it makes baresip-webrtc fail to build:

```
 make
  CC      build-x86_64/src/main.o
  CC      build-x86_64/src/session.o
src/session.c:695:9: warning: implicit declaration of function 'video_start_source' is invalid in
      C99 [-Wimplicit-function-declaration]
                err = video_start_source(vid, &sess->ctx);
                      ^
src/session.c:695:9: warning: this function declaration is not a prototype [-Wstrict-prototypes]
2 warnings generated.
  CC      build-x86_64/src/demo.o
  CC      build-x86_64/src/util.o
  LD      baresip-webrtc
Undefined symbols for architecture x86_64:
  "_video_start_source", referenced from:
      _rtcsession_start_video in session.o
ld: symbol(s) not found for architecture x86_64
make: *** [baresip-webrtc] Error 1
```

https://github.com/baresip/baresip-webrtc

there is a use-case, where the application only wants to start the video source.

juha-h:
Alfred E. Heggestad writes:

> is it possible to fix this without breaking the api ?

It is breaking my app also:
```
 if (sdp_media_ldir(m) & SDP_SENDONLY) {
            err = video_start_source(v, ctx);
            if (err) {
                LOGE("video_start_source failed with error %d\n", err);
                re_thread_leave();
                return err;
            }
        }
        if (sdp_media_ldir(m) & SDP_RECVONLY) {
            err = video_start_display(v, NULL);
            if (err) {
                LOGE("video_start_display failed with error %d\n", err);
                re_thread_leave();
                return err;
            }
        }
```
but I'm not sure if these calls are still needed after this update.

alfredh:
should I merge this to master ?

cspiel1:
We should squash/fixup the commit history some how before the merge. API functions had been removed and added later again.

juha-h:
This merge removed
```
void video_stop_display(struct video *v);
```
from baresip API.  Was this intentional?

If so, I'll try to figure out how to replace its usage in my baresip
Android app.

cHuberCoffee:
It was the same thinking like for the start functions. `video_stop` is working on both, source and display.
I thought, the `video_update` takes care of enabling and disabling the video source and display with respect of the possible directions.
One more thing: the video_stop was a source stop only. which was kind of misleading. especially if there exist a separate start for the source and the display.
The functions exist in the video.c. Its not great (should thought about these two functions in here too), but bringing it back is only a matter of removing the static and adding it again in the baresip.h

cspiel1:
- video_stop() should be called to stop the video in both directions.
- video_update() uses the SDP media direction to correctly start/stop video.

If you want the video_stop_display() back in the API anyway, we could do this. Should we?

juha-h:
Christian Spielberger writes:

> If you want the video_stop_display() back in the API anyway, we could
> do this. Should we?

I let you know if I really need it.  Most likely not.

PR #1153, opened 2020-12-16
aufile: add audio player to write speaker data to wav file

cspiel1:
This works already:
```
audio_source  aufile,/usr/local/share/baresip/recorded.wav
```

This PR adds also the ouput to wav file:
```
audio_player		aufile,/tmp/output.wav
```

This can be used for debugging, audio analyzes or automatic tests.

### DIFF
--- a/modules/aufile/aufile.c
+++ b/modules/aufile/aufile.c
@@ -9,6 +9,7 @@
 #include <re.h>
 #include <rem.h>
 #include <baresip.h>
+#include "aufile.h"
 
 
 /**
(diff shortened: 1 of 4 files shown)


alfredh:
rem has an aufile module for read/writing wav.

is it possible to use rem instead of sndfile ?

https://github.com/baresip/rem/blob/master/include/rem_aufile.h#L26

cspiel1:
No sndfile dependency now. I will squash the commits before merge.

alfredh:
There is a button here called "Squash and Merge". When the PR is ready for merge, I press this button.
it will squash all commits into one single atomic commit, and then merge to master.

This system works fine for a long time, and it does not matter if the PR contains hundreds of small
commits, adding/removal etc. It is just the single squashed commit that matters.

I can merge this PR now, but it says that it is in a "Draft" state ..

cspiel1:
Ok, I understand.

cspiel1:
Thanks!

PR #1155, opened 2020-12-18
Play aufile extended support

cspiel1:
This PR adds support for playing ring tones and other signal tones with an ausrc with different sample rates without resampling.

- As example we use aufile. We made aufile returning the sample rate and channels number.
- Adds PCMA/PCMU support to aufile.
- Fixes read until the end of aubuf in play.c.
- If the ausrc does not support returning the sample rate and channels number, new config settings are used.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -60,6 +60,8 @@ rtp_stats		no
 
 # Play tones
 #file_ausrc		aufile
+#file_srate		16000
+#file_channels		1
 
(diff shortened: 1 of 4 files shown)


alfredh:
Amazing work from your team :)

it seems that you and your team are very productive these days ?
good work.

cspiel1:
Thanks! We do our best. ;-)

PR #1156, opened 2020-12-18
stream: add function to send a RTP dummy packet

cHuberCoffee:
The new function will open a NAT pinhole by sending a dummy
packet via the RTP port in case of no running video source.

This patch should fix the NAT pinhole topic in #1142

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1297,6 +1297,7 @@ void stream_set_secure(struct stream *strm, bool secure);
 bool stream_is_secure(const struct stream *strm);
 int  stream_start_mediaenc(struct stream *strm);
 int  stream_start(const struct stream *strm);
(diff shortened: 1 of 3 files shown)


juha-h:
Christoph Huber writes:

> The new function will open a NAT pinhole by sending a dummy
> packet via the RTP port in case of no running video source.
> 
> This patch should fix the NAT pinhole topic in #1142

Thanks for the code.  Looks like it sends one dummy packet when one way
stream starts.  I don't know if one dummy is enough or should they be
sent periodically.  I'll give it a try during the next few days.

cHuberCoffee:
As fare as i understood the answer of alfredh:
https://github.com/baresip/baresip/issues/1142#issuecomment-745281770
It should be enough to send one packet.

PR #1157, opened 2020-12-18
Audio file info

cspiel1:
This adds a command to debug_cmd that uses an ausrc (e.g. aufile) in order to count the samples and compute the length of the audio file in milli seconds.

The ausrc needs a new "blocking mode". As a first example we extend the aufile with a blocking mode.

Maybe later we also deliver our auogg (audio source for ogg files) or the gst module with a blocking mode.

### DIFF
--- a/modules/debug_cmd/debug_cmd.c
+++ b/modules/debug_cmd/debug_cmd.c
@@ -4,6 +4,7 @@
  * Copyright (C) 2010 - 2016 Creytiv.com
  */
 #include <stdlib.h>
+#include <string.h>
 #include <time.h>
 #ifdef USE_OPENSSL
 #include <openssl/crypto.h>
@@ -17,7 +18,6 @@
  *
  * Advanced debug commands
  */
-
 
 static uint64_t start_ticks;          /**< Ticks when app started         */
 static time_t start_time;             /**< Start time of application      */
@@ -160,6 +160,90 @@ static int cmd_play_file(struct re_printf *pf, void *arg)
 }
 
 
+static void fileinfo_readh(struct auframe *af, void *arg)
+{
+	size_t *sampc = arg;
+
+	if (!af || !arg)
+		return;
+
+	*sampc += af->sampc;
+}
+
+
+struct ausrc_st;
+
+
+/**
+ * Command fileinfo reads given audio file with ausrc that is specified in
+ * config file_ausrc and returns the length in milli seconds. The file has
+ * to be located in the path specified by audio_path.
+ *
+ * Usage:
+ * /fileinfo audiofile
+ *
+ * @param pf Print handler is used to return length.
+ * @param arg Command argument contains the file name.
+ *
+ * @return The length of the fileplay file in milli seconds.
+ */
+/* ------------------------------------------------------------------------- */
+static int cmd_fileinfo(struct re_printf *pf, void *arg)
+{
+	const struct cmd_arg *carg = arg;
+	size_t sampc = 0;
+	struct ausrc_prm prm;
+	int err = 0;
+	uint32_t ms = 0;
+	size_t len;
+	char *path;
+	char aumod[16];
+	struct ausrc_st *ausrc_st = NULL;
+
+	if (!str_isset(carg->prm)) {
+		re_hprintf(pf, "fileplay: filename not specified\n");
+		return EINVAL;
+	}
+
+	err = conf_get_str(conf_cur(), "file_ausrc", aumod, sizeof(aumod));
+	if (err) {
+		warning("debug_cmd: file_ausrc is not set\n");
+		return EINVAL;
+	}
+
+	len = str_len(conf_config()->audio.audio_path) + str_len(carg->prm)
+		+ 2;
+	path = mem_zalloc(len, NULL);
+	re_snprintf(path, len, "%s/%s",
+			conf_config()->audio.audio_path,
+			carg->prm);
+
+	prm.srate = 0;
+	prm.ch = 0;
+	prm.ptime = 0;
+	prm.fmt = 0;
+
+	/* prm->ptime == 0 means blocking mode for ausrc */
+	err = ausrc_alloc(&ausrc_st, baresip_ausrcl(),
+			NULL, aumod,
+			&prm, path,
+			fileinfo_readh, NULL, &sampc);
+	if (err)
+		warning("debug_cmd: %s - ausrc %s does not support blocking "
+				"mode or reading source %s failed. (%m)\n",
+				__func__, aumod, carg->prm, err);
+
+	if (prm.ch && prm.srate)
+		ms = sampc * 1000 / prm.ch / prm.srate;
+
+	re_hprintf(pf, "debug_cmd: length = %u ms\n", ms);
+
+	mem_deref(ausrc_st);
+	mem_deref(path);
+	return err;
+}
+
+
 static int cmd_sip_debug(struct re_printf *pf, void *unused)
 {
 	(void)unused;
@@ -226,6 +310,7 @@ static const struct cmd debugcmdv[] = {
 {"modules",     0,       0, "Module debug",           mod_debug           },
 {"netstat",    'n',      0, "Network debug",          cmd_net_debug       },
 {"play",        0, CMD_PRM, "Play audio file",        cmd_play_file       },
+{"fileinfo",    0, CMD_PRM, "Audio file info",        cmd_fileinfo        },
 {"sipstat",    'i',      0, "SIP debug",              cmd_sip_debug       },
 {"sysinfo",    's',      0, "System info",            print_system_info   },
 {"timers",      0,       0, "Timer debug",            tmr_status          },
--- a/modules/debug_cmd/debug_cmd.c
+++ b/modules/debug_cmd/debug_cmd.c
@@ -168,6 +168,7 @@ struct fileinfo_st {
 	struct ausrc_prm prm;
 	size_t sampc;
 	struct tmr tmr;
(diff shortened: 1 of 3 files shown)


alfredh:
please note that libre/baresip is designed around an async event-driven polling loop.
one thread can handle hundreds of I/O events and timers.
for this to work properly, each function must execute fast.

I am sceptical about blocking the call for N seconds, this will basically break the whole
design pattern.

If you run this command on a long audio file (say, 30 minutes) the polling loop will
be blocked until the whole file is read. During this time, no other commands can be executed.

cspiel1:
Good defence! I will try to find a better solution.

cspiel1:
- removed the "blocking mode" in aufile
- now aufile has a mode with ptime is zero which processes as fast as possible, but does not block the alloc_handler
- In debug_cmd the cmd_fileinfo() informs with a ua_event() about the result. Thus module cons, ctrl_tcp or ctrl_dbus pass the result to the user asynchronously.

alfredh:
I tested it locally on Debian 10, and it does not seem to work for me:

```
/fileinfo error.wav
.... dev '/usr/local/share/baresip/error.wav'
aufile: loading input file '/usr/local/share/baresip/error.wav'
aufile: /usr/local/share/baresip/error.wav: 8000 Hz, 1 channels, PCMU
aufile: audio ptime=0 sampc=320
aufile: read end of file
aufile: loaded 33436 bytes
debug_cmd: timeout, length > 1727557320 ms
```

modules/debug_cmd/debug_cmd.c
```diff
@@ -18,7 +19,6 @@
  * Advanced debug commands
  */
 
-
```
alfredh, line 21: unrelated change ..

modules/debug_cmd/debug_cmd.c
```diff
@@ -160,6 +160,143 @@ static int cmd_play_file(struct re_printf *pf, void *arg)
 }
 
 
+struct ausrc_st;
```
alfredh, line 163: is this forward declaration needed ?

modules/debug_cmd/debug_cmd.c
```diff
@@ -160,6 +160,143 @@ static int cmd_play_file(struct re_printf *pf, void *arg)
...
+{
+	struct fileinfo_st *st = arg;
+	size_t ms = 0;
+
+	if (st->prm.ch && st->prm.srate)
+		ms = st->sampc * 1000 / st->prm.ch / st->prm.srate;
+
+	if (st->finished) {
+		info("debug_cmd: length = %u ms\n", ms);
+		ua_event(NULL, UA_EVENT_MAX, NULL, "debug_cmd: length = %u ms",
```
alfredh, line 194: please dont use `UA_EVENT_MAX` for this purpose.

modules/debug_cmd/debug_cmd.c
```diff
@@ -160,6 +160,143 @@ static int cmd_play_file(struct re_printf *pf, void *arg)
...
+static void fileinfo_timeout(void *arg)
+{
+	struct fileinfo_st *st = arg;
+	size_t ms = 0;
+
+	if (st->prm.ch && st->prm.srate)
+		ms = st->sampc * 1000 / st->prm.ch / st->prm.srate;
+
+	if (st->finished) {
+		info("debug_cmd: length = %u ms\n", ms);
```
alfredh, line 193: I think that the unit _seconds_ is more user-friendly, use e.g. `double seconds`.

length: 30500 ms
length: 30.5 seconds

modules/debug_cmd/debug_cmd.c
```diff
@@ -160,6 +160,143 @@ static int cmd_play_file(struct re_printf *pf, void *arg)
...
+	else {
+		info("debug_cmd: timeout\n");
+		ua_event(NULL, UA_EVENT_MAX, NULL, "debug_cmd: timeout");
+	}
+
+	mem_deref(st);
+}
+
+
+static void fileinfo_readh(struct auframe *af, void *arg)
```
alfredh, line 211: the suffix for handlers should be `_handler`

same for below function.

modules/debug_cmd/debug_cmd.c
```diff
@@ -160,6 +160,143 @@ static int cmd_play_file(struct re_printf *pf, void *arg)
...
+		return;
+
+	st->sampc += af->sampc;
+}
+
+
+static void fileinfo_errh(int err, const char *str, void *arg)
+{
+	struct fileinfo_st *st = arg;
+	(void) err;
```
alfredh, line 225: no need to cast err, it is used.

modules/debug_cmd/debug_cmd.c
```diff
@@ -160,6 +160,143 @@ static int cmd_play_file(struct re_printf *pf, void *arg)
...
+ * config file_ausrc and returns the length in milli seconds. The file has
+ * to be located in the path specified by audio_path.
+ *
+ * Usage:
+ * /fileinfo audiofile
+ *
+ * @param pf Print handler is used to return length.
+ * @param arg Command argument contains the file name.
+ *
+ * @return The length of the fileplay file in milli seconds.
```
alfredh, line 244: is the return description correct ?

modules/debug_cmd/debug_cmd.c
```diff
@@ -160,6 +160,143 @@ static int cmd_play_file(struct re_printf *pf, void *arg)
...
+		warning("debug_cmd: file_ausrc is not set\n");
+		return EINVAL;
+	}
+
+	len = str_len(conf_config()->audio.audio_path) + str_len(carg->prm)
+		+ 2;
+	path = mem_zalloc(len, NULL);
+	re_snprintf(path, len, "%s/%s",
+			conf_config()->audio.audio_path,
+			carg->prm);
```
alfredh, line 272: this block can be replaced with `re_sdprintf`

modules/debug_cmd/debug_cmd.c
```diff
@@ -160,6 +160,143 @@ static int cmd_play_file(struct re_printf *pf, void *arg)
...
+	}
+
+	err = ausrc_alloc(&st->ausrc, baresip_ausrcl(),
+			NULL, aumod,
+			&st->prm, path,
+			fileinfo_readh, fileinfo_errh, st);
+	if (err)
+		warning("debug_cmd: %s - ausrc %s does not support zero ptime "
+				"or reading source %s failed. (%m)\n",
+				__func__, aumod, carg->prm, err);
```
alfredh, line 288: the error case should also goto out

modules/debug_cmd/debug_cmd.c
```diff
@@ -226,6 +363,7 @@ static const struct cmd debugcmdv[] = {
 {"modules",     0,       0, "Module debug",           mod_debug           },
 {"netstat",    'n',      0, "Network debug",          cmd_net_debug       },
 {"play",        0, CMD_PRM, "Play audio file",        cmd_play_file       },
+{"fileinfo",    0, CMD_PRM, "Audio file info",        cmd_fileinfo        },
```
alfredh, line 366: consider adding "au" prefix since it is audio related.

au - audio
vid - video

modules/gst/gst.c
```diff
@@ -427,9 +435,21 @@ static int gst_alloc(struct ausrc_st **stp, const struct ausrc *as,
 	err = setup_uri(st, device);
 	if (err) goto out;
 
-	st->prm   = *prm;
+	st->ptime = prm->ptime;
+	if (!st->ptime)
+		st->ptime = 20;
+
+	if (!st->ptime)
+		st->ptime = 20;
```
alfredh, line 443: duplicated code ?

alfredh:
would it be possible to read out the duration field of the WAV header and use this ?

cspiel1:
> would it be possible to read out the duration field of the WAV header and use this ?

Not with this approach. It works for general ausrc. E.g. module gst. The "zero ptime" extension in the ausrc modules let them process as fast as possible. And a timeout of 5 seconds stops if the file is to long or a stream was specified.

Thanks for the comments! I will update by adding commits.

Here is the output of my test with error.wav:
```/fileinfo error.wav
aufile: loading input file '/usr/local/share/baresip/error.wav'
aufile: /usr/local/share/baresip/error.wav: 8000 Hz, 1 channels, PCMU
aufile: audio ptime=0 sampc=320
aufile: read end of file
aufile: loaded 33436 bytes
aufile: end of file
debug_cmd: length = 2120 ms
```
I tried with gst and a very long mp3. Then I get this output:
```debug_cmd: timeout, length > 1272100 ms
```

modules/debug_cmd/debug_cmd.c
cspiel1: no.

modules/debug_cmd/debug_cmd.c
cspiel1: It would be nice in general to be able to send some kind of "general" events from any modules. 
- We could define for each purpose a constant e.g. UA_EVENT_FILEINFO, or
- we could define a constant in baresip.h called UA_EVENT_GENERIC.

modules/debug_cmd/debug_cmd.c
cspiel1: no it's not. Thanks!

modules/debug_cmd/debug_cmd.c
cspiel1: cool! ;-)

cspiel1:
I suggest to squash the debug_cmd commits, the aufile commits and the gst commits separately. Ok?

PR #1161, opened 2020-12-23
Makefile: set SYSROOT_LOCAL only if SYSROOT is the default /usr

cspiel1:
This fixes cross compilation builds without having to set both SYSROOT and
SYSROOT_LOCAL.

More concretely, it avoids that the fallback value of SYSROOT_LOCAL leads to:
```
 cc1: warning: include location "/usr/local/include/rem" is unsafe for cross-compilation [-Wpoison-system-directories]
```


.github/workflows/build.yml
```diff
@@ -79,5 +79,5 @@ jobs:
     - name: make baresip macOS
       if: ${{ runner.os == 'macOS' }}
       run: |
-        make V=1 CCACHE= info test modules
-        make clean; make CCACHE= STATIC=yes
+        make V=1 CCACHE= SYSROOT_LOCAL=/usr/local info info2 test modules
+        make clean; make CCACHE= SYSROOT_LOCAL=/usr/local STATIC=yes
```
sreimers, line 83: macOS should not require any special args to build. A simple `make` should be sufficient to build baresip on all supported platforms.

.github/workflows/build.yml
```diff
@@ -79,5 +79,5 @@ jobs:
     - name: make baresip macOS
       if: ${{ runner.os == 'macOS' }}
       run: |
-        make V=1 CCACHE= info test modules
-        make clean; make CCACHE= STATIC=yes
+        make V=1 CCACHE= SYSROOT_LOCAL=/usr/local info info2 test modules
```
sreimers, line 82: is `info2` only for debugging used?

.github/workflows/build.yml
cspiel1: But if one specifies SYSROOT for a cross compilation build then it is not common to include /usr/local into the header search paths.
e.g. yocto says:
```
cc1: warning: include location "/usr/local/include/rem" is unsafe for cross-compilation [-Wpoison-system-directories]
```

.github/workflows/build.yml
cspiel1: Yes. Like info. The info target is defined in re.mk. And the variables LIBREM_PATH, ... are not included. Should I remove it or rename it?

.github/workflows/build.yml
sreimers: SYSROOT for macOS is defined here https://github.com/baresip/re/blob/master/mk/re.mk#L23

cspiel1: Now I added /usr/local as fallback. Also no warning/error in our yocto build.

Makefile
```diff
@@ -329,3 +367,10 @@ src/static.c: $(BUILD) Makefile $(APP_MK) $(MOD_MK)
 	done
 	@echo "  NULL"  >> $@
 	@echo "};"  >> $@
+
+.PHONY: info2
+info2:
```
sreimers, line 372: You can now replace:
```
.PHONY: info2
info2:
```
by:

```
info::
```

https://www.gnu.org/software/make/manual/html_node/Double_002dColon.html#Double_002dColon

https://github.com/baresip/re/commit/432fa6a92f1b369e4e54ae0841c4808b686af025

alfredh:
I tested on OSX with GNU make 3.81 and got this error:

```
Makefile:371: *** target file `info' has both : and :: entries.  Stop.
```

sreimers:
> ```
> Makefile:371: *** target file `info' has both : and :: entries.  Stop.
> ```

Looks like this re commit https://github.com/baresip/re/commit/432fa6a92f1b369e4e54ae0841c4808b686af025 is missing?

alfredh:
If you want to use double-colon targets, it makes a hard depencency that baresip and re
must use an exact version of the makefile or later.

baresip (commit 29. dec) -> re (commit 29. dec)

it will for example make it difficult to use baresip from 29. dec +/- 1 day with libre 28. dec or older.

it is guaranteed that this will cause many support requests from users.

a simpler and more flexible solution is:

```
bareinfo: info
	@echo "  LIBREM_PATH:   $(LIBREM_PATH)"
	@echo "  LIBREM_INC:    $(LIBREM_INC)"
	@echo "  LIBREM_SO:     $(LIBREM_SO)"
```

cspiel1:
From my side it is ready to merge now. Tested in yocto:
```
export LIBRE_MK   = "${STAGING_DIR_TARGET}/usr/share/re/re.mk"
export SYSROOT    = "${STAGING_DIR_TARGET}/usr"
export DESTDIR    = "${D}"
```

alfredh:
thanks!

PR #1170, opened 2020-12-28
clang shorten-64-to-32 warnings

sreimers:
This fixes some remaining shorten-64-to-32 warnings. I leave `mixausrc` as is, so if @cspiel1 wants to provide a pull request for this.

### DIFF
--- a/modules/alsa/alsa_play.c
+++ b/modules/alsa/alsa_play.c
@@ -67,12 +67,12 @@ static void *write_thread(void *arg)
 
 		sampv = st->sampv;
 
-		n = snd_pcm_writei(st->write, sampv, samples);
+		n = (int)snd_pcm_writei(st->write, sampv, samples);
 
 		if (-EPIPE == n) {
 			snd_pcm_prepare(st->write);
 
-			n = snd_pcm_writei(st->write, sampv, samples);
+			n = (int)snd_pcm_writei(st->write, sampv, samples);
 			if (n != samples) {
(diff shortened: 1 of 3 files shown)


modules/alsa/alsa_play.c
```diff
@@ -67,12 +67,12 @@ static void *write_thread(void *arg)
 
 		sampv = st->sampv;
 
-		n = snd_pcm_writei(st->write, sampv, samples);
+		n = (int)snd_pcm_writei(st->write, sampv, samples);
```
alfredh, line 70: is it better to change the type of n to `long n` ?

in this case no casting is needed.

modules/oss/oss.c
```diff
@@ -187,7 +187,7 @@ static void *record_thread(void *arg)
 
 	while (st->run) {
 
-		n = read(st->fd, st->sampv, st->sampc*2);
+		n = (int)read(st->fd, st->sampv, st->sampc*2);
```
alfredh, line 190: I am not sure if we should maintain the oss module anymore, for modern linux alsa should be used.

is anyone using oss anymore ?

modules/alsa/alsa_play.c
sreimers: good suggestion. changed now.

modules/oss/oss.c
sreimers: I'm unsure if oss still a thing on unix systems (FreeBSD etc.).

Issue #1169, opened 2020-12-27
crash at avformat_video_alloc()


juha-h:
While testing my Android video app, I noticed that it _sometimes_ crashes at avformat_video_alloc() when video is started after it has first been stopped:
```
- start call with sendrecv video
- stop video by re-INVITE where video is inactive
- re-start video by re-INVITE where video is sendrecv
```
Below is two debugs logs starting from the last step, i.e., when video is re-started.  In the first one, restart works fine and in the second one there is crash on this line of avformat/video.c avformat_video_alloc():
```
	if (ctx && *ctx && (*ctx)->id && !strcmp((*ctx)->id, "avformat")) {
```
The logs include media. stream, and video debug logs and I have not not spotted any difference in them.  Since the crash does not happen every time, I suspect that it is somehow timing related and then (*ctx)->id does not point to a proper string.

Any ideas or suggestions on how I could further debug this?

Log where video re-start works OK:
```
18:49:00.475 13208 13253: ua event (CALL_RTCP) video
18:49:01.813 13208 13253: ua event (CALL_RTCP) audio
18:49:01.918 13208 13208: ua event (CALL_REMOTE_SDP) answer
18:49:01.918 13208 13208: video RTP/AVP
18:49:01.918 13208 13208:   local formats:
18:49:01.918 13208 13208:      96 VP9/90000/1 () *
18:49:01.918 13208 13208:      97 VP8/90000/1 () *
18:49:01.918 13208 13208:      98 H264/90000/1 () *
18:49:01.918 13208 13208:      99 H264/90000/1 ()
18:49:01.918 13208 13208:   remote formats:
18:49:01.918 13208 13208:       0
18:49:01.918 13208 13208:   local attributes:
18:49:01.918 13208 13208:     label='2'
18:49:01.918 13208 13208:     rtcp-rsize
18:49:01.918 13208 13208:     ssrc='3460127767 cname:sip:bar@test.tutpro.com'
18:49:01.918 13208 13208:     framerate='24.00'
18:49:01.918 13208 13208:     rtcp-fb='* nack pli'
18:49:01.918 13208 13208:     content='main'
18:49:01.918 13208 13208:   remote attributes:
18:49:01.918 13208 13208: 
18:49:01.918 13208 13208: local media dir 3
18:49:01.918 13208 13208: remote media dir 3
18:49:01.918 13208 13208:  video dir=sendrecv pt_enc=-1
[...]

PR #1173, opened 2020-12-30
Added ctx param to video_stop/video_stop_source and set ctx to null

juha-h:
when video source is stopped in order to solve issue https://github.com/baresip/baresip/issues/1169.

### DIFF
--- a/src/video.c
+++ b/src/video.c
@@ -1214,7 +1214,8 @@ static void video_stop_source(struct video *v, struct media_ctx **ctx)
 	debug("video: stopping video source ..\n");
 
 	v->vtx.vsrc = mem_deref(v->vtx.vsrc);
-	*ctx = NULL;
+	if (ctx)
+		*ctx = NULL;
 }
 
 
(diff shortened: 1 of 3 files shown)


src/video.c
```diff
@@ -1206,14 +1206,15 @@ int video_start_display(struct video *v, const char *peer)
...
-static void video_stop_source(struct video *v)
+static void video_stop_source(struct video *v, struct media_ctx **ctx)
 {
 	if (!v)
 		return;
 
 	debug("video: stopping video source ..\n");
 
 	v->vtx.vsrc = mem_deref(v->vtx.vsrc);
+	*ctx = NULL;
```
alfredh, line 1217: to make it more robust, please check if `ctx` is not null

alfredh:
can you confirm that this fixes the crash on your Android app ?

juha-h:
Alfred E. Heggestad writes:

> can you confirm that this fixes the crash on your Android app ?

Have tested tens of times to turn video on/off without crash.

alfredh:
thanks !

PR #1180, opened 2021-01-09
[WIP] ua refactoring

alfredh:
I have a proposal to refactor ua.c

1. make uag an allocated struct (e.g. struct ua_stack)
2. split ua.c into 2 files
3. make it fully re-entrant

for now I would like to find a good name to this struct (uag),
for example `struct ua_stack` or something like that.

any good suggestions ?

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -38,7 +38,7 @@ struct ua_xhdr_filter {
 	char *hdr_name;
 };
 
-static struct {
+static struct ua_stack {
 	struct config_sip *cfg;        /**< SIP configuration               */
 	struct list ual;               /**< List of User-Agents (struct ua) */
 	struct sip *sip;               /**< SIP Stack                       */


alfredh:
for now I need a good name for this guy:

```c
static struct ua_stack {
	struct config_sip *cfg;        /**< SIP configuration               */
	struct list ual;               /**< List of User-Agents (struct ua) */
	struct sip *sip;               /**< SIP Stack                       */
	struct sip_lsnr *lsnr;         /**< SIP Listener                    */
	struct sipsess_sock *sock;     /**< SIP Session socket              */
	struct sipevent_sock *evsock;  /**< SIP Event socket                */
	struct ua *ua_cur;             /**< Current User-Agent              */
	bool use_udp;                  /**< Use UDP transport               */
	bool use_tcp;                  /**< Use TCP transport               */
	bool use_tls;                  /**< Use TLS transport               */
	bool delayed_close;            /**< Module will close SIP stack     */
	sip_msg_h *subh;               /**< Subscribe handler               */
	ua_exit_h *exith;              /**< UA Exit handler                 */
	void *arg;                     /**< UA Exit handler argument        */
	char *eprm;                    /**< Extra UA parameters             */
#ifdef USE_TLS
	struct tls *tls;               /**< TLS Context                     */
#endif
} uag = {
```

GGGO:
I think ua_stack is fine, but I wonder if we could merge this uag struct in ua struct and remove the concept of current agent.
Right now we have ua_alloc, but soon we will have ua_stack_alloc too, right ?

Maybe we should add ua list in baresip struct ?

cspiel1:
The name ua_stack is preferable if we want to have the list of ua-s a specified order. A specified order could be beneficial for a automatic ua selection for incoming/outgoing calls. A short alternative would be to keep the name "uag".

Other suggestions
- ua_stack should stay "singleton" (so no ua_stack_alloc(struct ua_stack *pua_stack) instead a ua_stack_init(void)),
- ua_init() has to be moved to new file and renamed,
- also ua_closed() and of course all uag_... functions.

alfredh:
I agree that we should remove the concept of "current" ua.

this involves these functions:


```c
void         uag_current_set(struct ua *ua);
struct ua   *uag_current(void);
```

step 1 - remove usage of these functions
step 2 - remove the functions

alfredh:
please see my recent commits for some removal of uag_current usage.

modules left to fix:

- contact
- menu
- mqtt

for contact I suggest doing a lookup of all ua's and select the ua that matches
the domain of the target uri, with a fallback to the first ua.

cspiel1:
I see only one commit in baresip/baresip branch ua_refactor:
"struct ua_stack for global uag"

But what you describe is interesting. We did the ua selection similar in our private branch. I had no time to make an official PR so far. We also have commits that
- allow TCP/UDP/TLS peer-to-peer calls,
- allow IPv4/IPv6 peer-to-peer
all without baresip restart by specifying multiple local accounts with regint=0. We select the UA by matching host, transport protocol, and address family. We differ between local accounts and registration accounts.

We use the uag_current() only as fallback. But the fallback could also be the first UA or the first UA whose registration was successful.

Your plan fits perfectly to our changes. After you have your first version of UA selection I will check if and how we can extend it. Here is our uag_find_touri(). It uses a lot of new helper functions and also has other commits as dependency. So just to give an idea:
```
/**
 * Find a User-Agent (UA) from accounts that transport and address family.
 *
 * @param uri The uri that we want to connect to.
 *
 * @return User-Agent (UA) if found, otherwise NULL
 */
struct ua *uag_find_touri(const struct uri *uri)
{
	struct ua *ua;
	struct le *le;
	struct account *acc;
	struct ua *ret = NULL;
	char *buf;

	if (!uri)
		return NULL;

		ua = le->data;
		acc = ua->acc;

		/* Omit inactive proxies.  */
		if (acc_non_reg_proxy(acc))
			continue;

		/* Omit proxies with registration failures. */
		if (acc->regint && !ua_isregistered(ua))
			continue;

		/* Is only user given? In this case we take first registered.
		 * account. */
		if (uri_only_user(uri)) {
			if (acc->regint) {
				ret = ua;
				break;
			}
		}

		/* if user and host is given */
		if (uri_user_and_host(uri)) {
			/* If account is registered, then host has to match. */
			if (acc->regint) {
				if (0 != pl_cmp(&uri->host, &acc->luri.host)) {
					continue;
				} else {
					ret = ua;
					break;
				}
			}
		}

		/* Now we select a local account.
		 * For uri=IP / domain / user@domain. */

		/* address family match */
		if (uri_mismatch_af(&acc->luri, uri))
			continue;

		/* transport protocol match */
		if (!uri_match_transport(&acc->luri, uri, SIP_TRANSP_NONE))
			continue;

		if (!acc->regint && !uri_only_user(uri)) {
			/* Remember local account.
			 * But we prefer matching reg proxy. */
			ret = ua;
		}
	}

	if (!ret) {
		/* Ok, seems that matching account is missing. */
		ret = uag_current();
		if (!ret)
			return NULL;

		if (uri_only_user(uri)) {
			warning("ua: Could not find UA for %r", &uri->host);
			return NULL;
		}

		warning("ua: could not find UA for host %r, params %r."
				" We use current UA %s.\n", &uri->host,
				&uri->params, ua_account(ret)->buf);
	} else {
		buf = mem_zalloc(256, NULL);
		if (buf)
			info("ua: we choose UA %s\n", account_luri_anonym(
						ua_account(ret), buf, 256));

		mem_deref(buf);
	}

	return ret;
}
```

alfredh:
this is great, I think that it sounds like a good solution.

I will do some more changes, please feel free to submit patches with your lookup function.

for the ua_stack change, I was mainly looking for a good name for the struct.
examples:

- ua stack
- ua container
- ua list
- etc.

it is currently a singleton (global -> uag) and mainly contains:

- list of useragents
- the sip stack (sip, sip-session, event)
- flags to use UDP,TCP,TLS
- subscribe and exit handler
- extra parameters
- tls stack (for sip)


it is not so easy to merge ua and ua_stack, since there are multiple UAs and only
one single SIP-stack listening on a single port.

cspiel1:
I would prefer that the prefix for the "uag_" functions keep length three. But then ua_list is not good because uag already contains an ual. Suggestions:
struct | prefix
--- | ---
ua_stack | uas_
ua_set | uas_
ua_container | uac_
ua_selector | uas_
ua_sel | uas_

alfredh:
we should probably avoid "uac" and "uas" as it may conflict with terms defined in RFC 3261:

User Agent Client (UAC)
User Agent Server (UAS)

https://tools.ietf.org/html/rfc3261#section-6

one idea that comes to mind is uag where "g" is not defined (can also mean group).

```c
struct uag uag;
```

alfredh:
will make a new PR with some relevant code ..

alfredh:
@cspiel1 I moved the uag_current state to the menu module, where it can be deprecated.

it would be nice if you can integrate your function to automagically select the best ua
from the target uri.

PR #1188, opened 2021-01-13
Sip auto answer caller

cspiel1:
Adds SIP auto answer for outgoing calls. Used SIP header can be chosen by config "sip_autoanswer_method":

sip_autoanswer_method | SIP Header                                                   
------------ | ------------------------------------------------------------ 
rfc5373      | Answer-Mode: Auto                                            
call-info    | Call-Info: \<http://www.notused.com>;answer-after=0           
alert-info   | Alert-Info: <...>;info=alert-autoanswer;delay=0

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -205,6 +205,7 @@ video_selfview		window # {window,pip}
 #statmode_default	off
 #menu_clean_number	no
 #sip_autoanswer_beep	yes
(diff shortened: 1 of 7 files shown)


alfredh:
This is really good work, many thanks! Some people have also been asking for this.

Could you please add RFC 5373 to this list:

https://github.com/baresip/baresip#ietf-rfci-ds

also, please resolve the above conflicts.

PR #1199, opened 2021-01-25
Ua auto select

cspiel1:
- ua: improve auto selection for incoming calls (resp. any sip_msg)
- ua: auto selection of UA for outgoing calls
- menu: separate current call UA from dial UA and add an automatic mode for dial UA


cspiel1:
Will separate PR into incoming, outgoing calls PRs.

cspiel1:
Since some utility functions are used for both incoming and outgoing calls, could not separate the PR.

cspiel1:
Since some utility functions are used for both incoming and outgoing calls, could not separate the PR.

include/baresip.h
```diff
@@ -831,7 +832,7 @@ void uag_enable_sip_trace(bool enable);
 int  uag_reset_transp(bool reg, bool reinvite);
 void uag_set_sub_handler(sip_msg_h *subh);
 int  uag_set_extra_params(const char *eprm);
-struct ua   *uag_find(const struct pl *cuser);
+struct ua   *uag_find(const struct sip_msg *msg);
```
alfredh, line 835: I suggest adding a new function:

`uag_find_msg(const struct sip_msg *msg);`

and leave the current one unchanged.

this way we dont break the API :)

modules/menu/dynamic_menu.c
```diff
@@ -12,7 +12,7 @@
 static int set_audio_bitrate(struct re_printf *pf, void *arg)
 {
 	struct cmd_arg *carg = arg;
-	struct ua *ua = carg->data ? carg->data : menu_current();
+	struct ua *ua = carg->data ? carg->data : menu_uacur();
```
alfredh, line 15: I added a commit in master this, can you please rebase ?

modules/menu/static_menu.c
```diff
@@ -673,18 +699,15 @@ static int cmd_ua_next(struct re_printf *pf, void *unused)
 	(void)pf;
 	(void)unused;
 
-	if (!menu->le_cur)
```
alfredh, line 676: we should consider to remove this command (toggle ua / ua_next).

it would be nice if there is no "current" ua

src/config.c
```diff
@@ -228,6 +229,19 @@ static const char *jbuf_type_str(enum jbuf_type jbtype)
...
+{
+	switch (tr) {
+	case SIP_TRANSP_UDP: return "udp";
+	case SIP_TRANSP_TCP: return "tcp";
+	case SIP_TRANSP_TLS: return "tls";
+	case SIP_TRANSP_WS:  return "ws";
+	case SIP_TRANSP_WSS: return "wss";
+	default:             return "???";
+	}
+}
```
alfredh, line 242: this function already exist in re/src/sip:

```
const char *sip_transp_name(enum sip_transp tp)
```

src/ua.c
```diff
@@ -932,6 +934,192 @@ int ua_uri_complete(struct ua *ua, struct mbuf *buf, const char *uri)
...
+
+
+/**
+ * Find a User-Agent (UA) best fitting for an INVITE to given SIP uri
+ *
+ * @param req_uri The SIP uri for the INVITE
+ *
+ * @return User-Agent (UA) if found, otherwise NULL
+ */
+static struct ua *uag_find_connect(const struct pl *req_uri)
```
alfredh, line 1031: I think this function should be global, so that it can be used
by any module/app.

src/ua.c
```diff
@@ -954,18 +1142,33 @@ int ua_connect_dir(struct ua *ua, struct call **callp,
...
 
-	if (!ua || !str_isset(req_uri))
+	if (!str_isset(req_uri))
 		return EINVAL;
 
 	dialbuf = mbuf_alloc(64);
 	if (!dialbuf)
 		return ENOMEM;
 
+	if (!ua) {
```
alfredh, line 1152: this logic should be done outside ua.c

src/ua.c
```diff
@@ -1982,41 +2186,150 @@ struct sipevent_sock *uag_sipevent_sock(void)
...
+
+		return !pl_casecmp(&pl1, &pl2);
+	}
+
+	/* incoming calls: tp of SIP message */
+	return !pl_strcasecmp(&pl1, sip_transp_name(tp));
+}
+
+
+static bool match_port(uint16_t port1, uint16_t port2, bool tls)
```
alfredh, line 2210: perhaps "bool secure" is better ? what about WSS ?

alfredh:
we should also improve the accept incoming call logic:

1. iterate through all user-agents
2. for each ua, iterate through all calls.
3. if call is in state CALL_STATE_INCOMING then accept that call

cspiel1:
I separated the current call UA and the current dial UA.

Will try to remove these. Maybe the use-cases have to be handled differently. I have to check this.

src/ua.c
cspiel1: Ok, I'll move this to menu. But then gtk has to do it similar.

include/baresip.h
cspiel1: Ok, I but then I'll change the first commit. This is better to review. uag_find() will be unchanged and uag_find_msg() added.

Note: uag_find() will be un-used also.

src/ua.c
cspiel1: match_port() matches account port with uri port. If not set it uses the default port which is different for TLS. I will add a doxygen.

docs/examples/config
```diff
@@ -11,6 +11,7 @@ poll_method		epoll		# poll, select, epoll ..
 #sip_listen		0.0.0.0:5060
 #sip_certificate	cert.pem
 #sip_cafile		ca.crt
+#sip_trans_def		udp
```
alfredh, line 14: each user-agent has a preferable sip transport.

it is possible to re-use each user-agents transport instead of adding a global one?
if possible, this will reduce the complexity.

Example:

```
<sip:user@domain.com>
<sip:user@domain.com;transport=tcp>
<sip:user@domain.com;transport=tls>
```

include/baresip.h
```diff
@@ -832,8 +833,11 @@ int  uag_reset_transp(bool reg, bool reinvite);
 void uag_set_sub_handler(sip_msg_h *subh);
 int  uag_set_extra_params(const char *eprm);
 struct ua   *uag_find(const struct pl *cuser);
+struct ua   *uag_find_msg(const struct sip_msg *msg);
 struct ua   *uag_find_aor(const char *aor);
 struct ua   *uag_find_param(const char *name, const char *val);
+struct ua   *uag_find_connect(const struct pl *req_uri);
```
alfredh, line 839: perhaps rename "connect" to something else, for example "target".

this is a generic function, and should be possible to use for any outgoing request,
for example OPTIONS:

```c
static int options_command(struct re_printf *pf, void *arg)
{
	int err = 0;

	(void)pf;

	err = ua_options_send(ua, carg->prm,
			      options_resp_handler, NULL);
	if (err) {
		warning("menu: ua_options failed: %m\n", err);
	}

	return err;
}
```

modules/ctrl_tcp/ctrl_tcp.c
```diff
@@ -200,11 +202,15 @@ static bool command_handler(struct mbuf *mb, void *arg)
 
 	resp->pos = NETSTRING_HEADER_SIZE;
 
+	if (oe_dat)
+		sscanf(oe_dat->u.str, "%p", &data);
```
alfredh, line 206: is it possible to use `re_regex` instead ?

sscanf is not secure if I remember correctly.

modules/ctrl_tcp/ctrl_tcp.c
```diff
@@ -170,8 +170,9 @@ static bool command_handler(struct mbuf *mb, void *arg)
 	struct mbuf *resp = mbuf_alloc(2048);
 	struct re_printf pf = {print_handler, resp};
 	struct odict *od = NULL;
-	const struct odict_entry *oe_cmd, *oe_prm, *oe_tok;
+	const struct odict_entry *oe_cmd, *oe_prm, *oe_dat, *oe_tok;
```
alfredh, line 173: is it possible to make a separate patch out of the "ctrl_tcp" changes ?

the patch is already quite large and difficult to review

modules/menu/static_menu.c
```diff
@@ -626,13 +680,12 @@ static int ua_print_reg_status(struct re_printf *pf, void *unused)
 
 	(void)unused;
 
-	err = re_hprintf(pf, "\n--- User Agents (%u) ---\n",
+	err = re_hprintf(pf, "\n--- User Agents (%u)---\n",
```
alfredh, line 683: unrelated change ?

src/ua.c
```diff
@@ -126,7 +126,8 @@ void ua_printf(const struct ua *ua, const char *fmt, ...)
 		return;
 
 	va_start(ap, fmt);
-	info("%r@%r: %v", &ua->acc->luri.user, &ua->acc->luri.host, fmt, &ap);
+	info("%p %r@%r: %v", ua, &ua->acc->luri.user, &ua->acc->luri.host, fmt,
+			&ap);
```
alfredh, line 130: I think that printing a pointer to the console (read by an enduser) is a bit user-unfriendly ...

we should find a better way to identify each user-agent, for example index

src/ua.c
```diff
@@ -932,6 +935,192 @@ int ua_uri_complete(struct ua *ua, struct mbuf *buf, const char *uri)
...
+	sal = net_laddr_af(baresip_network(), AF_INET6);
+	if (sal && sa_cmp(sal, &sap, SA_ADDR))
+		return true;
+#endif
+
+	return false;
+}
+
+
+static bool acc_inactive(const struct account *acc)
```
alfredh, line 1003: consider inversing the negated logic here: acc_is_active()

I find it much easier to read code that has correct logic (i.e. not negated).

src/ua.c
```diff
@@ -932,6 +935,192 @@ int ua_uri_complete(struct ua *ua, struct mbuf *buf, const char *uri)
...
+	uri = &acc->luri;
+	if (!uri_host_local(uri))
+		return true;
+
+	return false;
+}
+
+
+static bool uri_mismatch_af(
+		const struct uri *accu, const struct uri *peeru);
```
alfredh, line 1019: same here, consider positive logic:

```c
static bool uri_match_af();
```

src/ua.c
alfredh: SIP over WS/WSS also has a default port:

```
In the absence of DNS SRV resource records or an explicit port, the
   default port for a SIP URI using the "sip" scheme and the "ws"
   transport parameter is 80, and the default port for a SIP URI using
   the "sips" scheme and the "ws" transport parameter is 443.
```

alfredh:
I added some more comments, and committed some hunks to master...

could you please re-base again ?

in general this patch is quite large, and takes some time to review.

docs/examples/config
cspiel1: We have the account on the one side and the peer URI on the other side. Then we try to match these.

In uri_match_transport() we use this config as default value for the peer-uri. Not for the account.

To simplify this we would have to suppose "UDP" always. Then the application/user will have to add always the URI parameter "transport" to the target URI if something different is wanted. Do you prefer this?

modules/ctrl_tcp/ctrl_tcp.c
cspiel1: ok.

src/ua.c
cspiel1: Yes, I agree! But then I didn't understand what was your intention with
```
struct ua *ua = carg->data;
```
How should the pointer carg->data be set?
Or, should we replace this and read an index from carg->prm? And in ctrl_tcp/ctrl_dbus set the index to carg->prm?

src/ua.c
cspiel1: Ok, will do this.
But for domains we can not decide the address family match:
- uri_mismatch_af() always correctly returns **true** (but not false)
- uri_match_af() always correctly returns **false** (but not true)

cspiel1:
Ok, thanks for your interest! The UA selection is one of our most tricky and most wanted extension to baresip. We should spend the time it needs.

juha-h:
Christian Spielberger writes:

> Ok, thanks for your interest! The UA selection is one of our most
> tricky and most wanted extension to baresip. We should spend the time
> it needs.

In my baresip Android app, if a user makes baresip call from Android
contacts app, I choose the UA whose domain matches the domain of the
called party.  If no match is found, I choose the default UA.

If the user is using baresip Android app, there is always one selected
UA that is shown on the main activity page.  If the user then goes from
main activity to baresip contacts activity and from there places a call
or sends a message, I choose the UA of the main activity from where the
user went to contacts activity.

It would also be possible that each UA has its own contacts, but to me
it feels too complicated.

juha-h:
Christian Spielberger writes:

> To simplify this we would have to suppose "UDP" always. Then the
> application/user will have to add always the URI parameter "transport"
> to the target URI if something different is wanted. Do you prefer
> this?

If the account has a outbound proxy (common case), then its transport
will be used.  The outbound proxy will then choose the next hop
transport based on DNS NAPTR/SRV records if no explicit transport is
specified in the Request-URI.

cspiel1:
> To simplify this we would have to suppose "UDP" always. Then the application/user will have to add always the URI parameter "transport" to the target URI if something different is wanted. Do you prefer this?
> If the account has a outbound proxy (common case), then its transport will be used. The outbound proxy will then choose the next hop transport based on DNS NAPTR/SRV records if no explicit transport is specified in the Request-URI.

You assume here that the account is selected already. But it is not.

For the auto selection of the UA/account the transport protocol of the account has to match the peer URI. If the transport protocol is not specified for the peer URI we use this config as default value. Or we simply use UDP. But maybe it is better to omit the transport protocol check in this case. What do you think?

cspiel1:
> Ok, thanks for your interest! The UA selection is one of our most tricky and most wanted extension to baresip. We should spend the time it needs.
> In my baresip Android app, if a user makes baresip call from Android contacts app, I choose the UA whose domain matches the domain of the called party. If no match is found, I choose the default UA. If the user is using baresip Android app, there is always one selected UA that is shown on the main activity page. If the user then goes from main activity to baresip contacts activity and from there places a call or sends a message, I choose the UA of the main activity from where the user went to contacts activity. It would also be possible that each UA has its own contacts, but to me it feels too complicated.

What I did here in this PR is to split the current UA into a current dial UA and a current call UA.
- Current dial UA can also be "auto" for the auto-selection, similar to what you describe in the first paragraph.
- Current call UA is used for answer a call, call on hold, call hangup, ..., similar to your "main activity" account.

@alfredh doesn't want a current UA, like me and maybe others. In this PR there is also a commit that removes the Current Dial UA.

I think the current call UA (or current activity UA) maybe makes sense in order to correctly select the UA for e.g. hangup().

cspiel1:
Since this PR is already very large and difficult, and contains questionable commits (e.g. the last four commits about the UA selection via pointer) I replace this PR by a new one today. Will also collect your ideas and comments and hopefully answers about my questions that will be added.

alfredh:
I am wondering if we should replace "current ua" with a "current call" instead,
identified by the call-id.

imagine that the current call-id string is empty by default, when there is active call it is set.

```c
struct menu {
    char *active_callid;
};
```

from this callid, you can look up "struct call"
from the call, you can look up "struct ua"

this also means removing "ua_cur" completely.

juha-h:
Christian Spielberger writes:

> You assume here that the account is selected already. But it is not.
> 
> For the auto selection of the UA/account the transport protocol of the
> account has to match the peer URI. If the transport protocol is not
> specified for the peer URI we use this config as default value. Or we
> simply use UDP. But maybe it is better to omit the transport protocol
> check in this case. What do you think? 

Transport protocol of peer URI does not need to match account's
transport protocol if outbound proxy has been specified for the account.
Then INVITE will be sent to outbound proxy whatever transport is used to
reach the ob proxy and it is then up to the ob proxy reach the peer with
the protocol specified in Request-URI.

Further, I consider it a special case that an account has a fixed
transport protocol.  Usually the account can use whatever transport the
peer URI want to be used based on DNS records if not explicitly given in
Request-URI.

juha-h:
Christian Spielberger writes:

> - Current call UA is used for answer a call, call on hold, call
> - hangup, ..., similar to your "main activity" account.

The common case is that UAs register and there is no need to select in
the app, which UA answers the call.

> @alfredh doesn't want a current UA, like me and maybe others. In this
> PR there is also a commit that removes the Current Dial UA.

It is OK with me to remove the current dial UA from the menu module.  In
my Android app the current UA is the one shown on main activity page,
but it is not necessarily the one that will be used when call is made
from contacts or messages activities or when call is made from some
other dialer.

> I think the current call UA (or current activity UA) maybe makes sense
> in order to correctly select the UA for e.g. hangup().

If there are several parallel calls, there needs to be some mechanism to
choose, which one will be hanged up by ESC.

cspiel1:
@juha-h I have to investigate why we used multiple local accounts with different transport protocol. Makes sense what you state here.

cspiel1:
@juha-h The difference is only for peer-to-peer connections (without SIP proxy). We use multiple local accounts with different transport protocol because we want to have peer-to-peer calls not only via UDP. Especially we want TCP, TLS. The user is able to choose between UDP/TCP/TLS for an outgoing call. Here the URI parameters, e.g. "transport", are added to the dialbuf in function ua_connect_dir():
```
	/* Append any optional URI parameters */
	err |= mbuf_write_pl(dialbuf, &ua->acc->luri.params);
	if (err)
		goto out;
```

And the user of our devices is able to enable UDP, TCP, TLS or "TLS only" for incoming calls via a web-interface. This configuration results in one or multiple local accounts.

I will make separate commits in the new PR with a meaningful commit message.

cspiel1:
> I am wondering if we should replace "current ua" with a "current call" instead,
> identified by the call-id.
> 
> imagine that the current call-id string is empty by default, when there is active call it is set.
> 
> ```c
> struct menu {
>     char *active_callid;
> };
> ```
> 
> from this callid, you can look up "struct call"
> from the call, you can look up "struct ua"
> 
> this also means removing "ua_cur" completely.

Currently working on this. Very good idea. It simplifies a lot.

modules/ctrl_tcp/ctrl_tcp.c
cspiel1: I think that we will not need the change in ctrl_tcp because we should not set the carg->data.

modules/ctrl_tcp/ctrl_tcp.c
cspiel1: Same here. Change in ctrl_tcp will be obsolete.

src/ua.c
cspiel1: Again was thinking about the checks in uag_find_msg(). Have to discuss with my team why we added the port check. There is only (at most) one port open for each transport, anyway.

PR #1200, opened 2021-01-26
Ua auto select incoming

cspiel1:
ua: improve UA selection for incoming calls

This avoids wrong selection especially if multiple local accounts are specified
with different address family or transport protocol.


cspiel1:
separation for incoming/outgoing is not possible

PR #1201, opened 2021-01-27
Sip message to application

cspiel1:
relay SIP MESSAGE to the application via:
- ctrl_tcp
- ctrl_dbus

The ctrl_dbus extension can be tested with:
```
dbus-monitor interface=com.github.Baresip
```

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -457,6 +457,8 @@ int  message_listen(struct message *message,
 void message_unlisten(struct message *message, message_recv_h *recvh);
 int  message_send(struct ua *ua, const char *peer, const char *msg,
 		  sip_resp_h *resph, void *arg);
+int message_encode_dict(struct odict *od, struct ua *ua, const struct pl *peer,
+			      const struct pl *ctype, struct mbuf *body);
 
 
 /*
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -457,8 +457,9 @@ int  message_listen(struct message *message,
 void message_unlisten(struct message *message, message_recv_h *recvh);
 int  message_send(struct ua *ua, const char *peer, const char *msg,
 		  sip_resp_h *resph, void *arg);
-int message_encode_dict(struct odict *od, struct ua *ua, const struct pl *peer,
-			      const struct pl *ctype, struct mbuf *body);
+int message_encode_dict(struct odict *od, struct account *acc,
+			const struct pl *peer, const struct pl *ctype,
+			struct mbuf *body);
 
 
 /*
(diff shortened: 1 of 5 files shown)


include/baresip.h
```diff
@@ -457,6 +457,8 @@ int  message_listen(struct message *message,
 void message_unlisten(struct message *message, message_recv_h *recvh);
 int  message_send(struct ua *ua, const char *peer, const char *msg,
 		  sip_resp_h *resph, void *arg);
+int message_encode_dict(struct odict *od, struct ua *ua, const struct pl *peer,
+			      const struct pl *ctype, struct mbuf *body);
```
alfredh, line 461: since only "account" is used inside the function, it might be better to
replace "struct ua *ua" with "struct account *acc" ...?

modules/ctrl_dbus/baresipbus.c
```diff
@@ -1,8 +1,9 @@
 /*
- * Generated by gdbus-codegen 2.56.4. DO NOT EDIT.
+ * This file is generated by gdbus-codegen, do not modify it.
```
alfredh, line 2: should we consider to remove the generated file from git source control,
and generate it as part of the build process ?

something like this in module/ctrl_dbus/module.mk:

```make
module/ctrl_dbus/baresipbus.c: dbus-source.foo
	dbus-generate <param1> <params2>
```

modules/ctrl_dbus/baresipbus.h
```diff
@@ -64,6 +72,13 @@ void dbus_baresip_emit_event (
     const gchar *arg_evtype,
     const gchar *arg_param);
 
+void dbus_baresip_emit_message (
+    DBusBaresip *object,
+    const gchar *arg_ua,
+    const gchar *arg_peer,
+    const gchar *arg_ctype,
+    const gchar *arg_body);
+
```
alfredh, line 81: the diff from the generated file is very large to review :)

modules/ctrl_dbus/baresipbus.c
cspiel1: Yes, I think it makes sense.

modules/ctrl_dbus/baresipbus.h
cspiel1: Yes, this is also generated. So it is a good suggestion to generate through the build.

alfredh:
I would like to get this one into master first.

I suggest that you make a separate PR that changes the dbus file
from stored in git to generated as part of the build, first.

after that, it will make this patch smaller and hence easier to review :)

alfredh:
could you please rebase this PR on top of master ?

alfredh:
merged.

again, many thanks for your contributions :)

cspiel1:
You're wellcome!

PR #1202, opened 2021-01-27
Fix interaction between CLI menu and GTK menu

thillux:
Fix parallel usage of CLI and GTK menu:
* Let notifications disappear, when accepting from CLI
* Fix race condition in call_window with mutex
* Fix use_after_free of vu enc/dec in call_window

### DIFF
--- a/modules/gtk/call_window.c
+++ b/modules/gtk/call_window.c
@@ -8,7 +8,7 @@
 #include <baresip.h>
 #include <gtk/gtk.h>
 #include "gtk_mod.h"
-
+#include <pthread.h>
 
 struct call_window {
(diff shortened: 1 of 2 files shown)


alfredh:
many thanks for this.

the gtk module needs a maintainer, could you volunteer to take care of it ?

I think we should also consider to move modules/gtk into a standalone application
that is using libbaresip. this is a bit more user-friendly for endusers.

the project can be called: baresip-gtk

thillux:
I don't know if I'm the right person to maintain the gtk module. I cannot guarantee, that I'll always have the spare time to deliver timely bug fixes. Nevertheless, thanks for this great software which I started using a week ago!

PR #1206, opened 2021-02-03
ua: improve UA selection for incoming calls

cspiel1:
This allows to deny certain transport protocols (e.g. only allow TLS) for
incoming peer-to-peer calls. Also checks for correct address family which is
important if there is no af hint in the SDP.

@juha-h Now loosened the transport check. If the account does not specify a transport protocol, the check is omitted.

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -686,7 +686,7 @@ static bool request_handler(const struct sip_msg *msg, void *arg)
 	if (pl_strcmp(&msg->met, "OPTIONS"))
 		return false;
 
-	ua = uag_find(&msg->uri.user);
+	ua = uag_find_msg(msg);
 	if (!ua) {
 		(void)sip_treply(NULL, uag_sip(), msg, 404, "Not Found");
 		return true;
@@ -930,6 +930,46 @@ int ua_uri_complete(struct ua *ua, struct mbuf *buf, const char *uri)
 
 	return err;
 }
+
+
+static bool uri_host_local(const struct uri *uri)
+{
+	const struct sa *sal;
+	struct sa sap;
+	int err;
+
+	if (!uri)
+		return false;
+
+	if (!pl_strcmp(&uri->host, "localhost"))
+		return true;
+
+	if (!pl_strcmp(&uri->host, "127.0.0.1"))
+		return true;
+
+	if (!pl_strcmp(&uri->host, "::1"))
+		return true;
+
+	sal = net_laddr_af(baresip_network(), AF_INET);
+
+	err = sa_set(&sap, &uri->host, 0);
+	if (err)
+		return false;
+
+	if (sal && sa_cmp(sal, &sap, SA_ADDR))
+		return true;
+
+#ifdef HAVE_INET6
+	sal = net_laddr_af(baresip_network(), AF_INET6);
+	if (sal && sa_cmp(sal, &sap, SA_ADDR))
+		return true;
+#endif
+
+	return false;
+}
+
+
+static bool uri_match_af(const struct uri *accu, const struct uri *peeru);
 
 
 /**
@@ -1528,7 +1568,7 @@ static void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
 	      sip_transp_name(msg->tp),
 	      &msg->src, &msg->dst);
 
-	ua = uag_find(&msg->uri.user);
+	ua = uag_find_msg(msg);
 	if (!ua) {
 		warning("ua: %r: UA not found: %r\n",
 			&msg->from.auri, &msg->uri.user);
@@ -1673,7 +1713,7 @@ static bool sub_handler(const struct sip_msg *msg, void *arg)
 
 	(void)arg;
 
-	ua = uag_find(&msg->uri.user);
+	ua = uag_find_msg(msg);
 	if (!ua) {
 		warning("subscribe: no UA found for %r\n", &msg->uri.user);
 		(void)sip_treply(NULL, uag_sip(), msg, 404, "Not Found");
@@ -1982,6 +2022,59 @@ struct sipevent_sock *uag_sipevent_sock(void)
 }
 
 
+static bool uri_match_transport(const struct uri *accu, enum sip_transp tp)
+{
+	struct pl pl;
+	int err;
+
+	err = msg_param_decode(&accu->params, "transport", &pl);
+	if (err)
+		return true;
+
+	return !pl_strcasecmp(&pl, sip_transp_name(tp));
+}
+
+
+static bool uri_match_af(const struct uri *accu, const struct uri *peeru)
+{
+#ifdef HAVE_INET6
+	struct sa sa1;
+	struct sa sa2;
+	int err;
+#endif
+
+	/* we list cases where we know there is a mismatch in af */
+#ifdef HAVE_INET6
+	if (accu->af == AF_INET && peeru->af == AF_INET6)
+		return false;
+
+	if (accu->af == AF_INET6 && peeru->af == AF_INET)
+		return false;
+
+	if (accu->af == AF_INET6 && peeru->af == AF_INET6) {
+		err =  sa_set(&sa1, &accu->host, 0);
+		err |= sa_set(&sa2, &peeru->host, 0);
+
+		if (err) {
+			warning("ua: No valid IPv6 URI %r, %r (%m)\n",
+					&accu->host,
+					&peeru->host, err);
+			return false;
+		}
+
+		return sa_is_linklocal(&sa1) == sa_is_linklocal(&sa2);
+	}
+
+	/* if peer uri is a domain, we simply return true */
+	if (peeru->af == AF_UNSPEC)
+		return true;
+#endif
+
+	/* both IPv4 or we can't decide if af will match */
+	return true;
+}
+
+
 /**
  * Find the correct UA from the contact user
  *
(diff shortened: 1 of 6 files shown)


cspiel1:
TODO: have to add the commit to fix the tests.

src/ua.c
```diff
@@ -932,6 +932,46 @@ int ua_uri_complete(struct ua *ua, struct mbuf *buf, const char *uri)
...
+		return true;
+
+#ifdef HAVE_INET6
+	sal = net_laddr_af(baresip_network(), AF_INET6);
+	if (sal && sa_cmp(sal, &sap, SA_ADDR))
+		return true;
+#endif
+
+	return false;
+}
```
alfredh, line 969: this code looks fine, but it is also possible to make an array of hosts/afs and compare it.

example code:

```c
static bool uri_host_local(const struct uri *uri)
{
	const char *hostv[] = {
		"localhost",
		"127.0.0.1",
		"::1"
	};
	int afv[2] = {AF_INET, AF_INET6};
	const struct sa *sal;
	struct sa sap;
	size_t i;
	int err;

	if (!uri)
		return false;

	for (i=0; i<ARRAY_SIZE(hostv); i++) {

		if (!pl_strcmp(&uri->host, hostv[i]))
			return true;
	}

	for (i=0; i<ARRAY_SIZE(afv); i++) {

		sal = net_laddr_af(baresip_network(), afv[i]);

		err = sa_set(&sap, &uri->host, 0);
		if (err)
			continue;

		if (sa_cmp(sal, &sap, SA_ADDR))
			return true;
	}

	return false;
}
```

it makes it easier in the future to add new host types should it
be needed, for example "127.0.0.2"

src/ua.c
```diff
@@ -965,7 +1005,6 @@ int ua_connect_dir(struct ua *ua, struct call **callp,
 
 	/* Append any optional URI parameters */
 	err |= mbuf_write_pl(dialbuf, &ua->acc->luri.params);
-
```
alfredh, line 968: this is an unrelated change

src/ua.c
```diff
@@ -1982,6 +2021,59 @@ struct sipevent_sock *uag_sipevent_sock(void)
...
+{
+	struct pl pl;
+	int err;
+
+	err = msg_param_decode(&accu->params, "transport", &pl);
+	if (err)
+		return true;
+
+	return !pl_strcasecmp(&pl, sip_transp_name(tp));
+}
```
alfredh, line 2034: I think here it is better to compare the sip transport tp instead of comparing the string.

1. decode transport parameter
2. convert from string to `enum sip_transp`
3. compare it with tp

src/ua.c
```diff
@@ -1982,6 +2021,59 @@ struct sipevent_sock *uag_sipevent_sock(void)
...
+
+
+static bool uri_match_af(const struct uri *accu, const struct uri *peeru)
+{
+#ifdef HAVE_INET6
```
alfredh, line 2039: I am wondering if we should get rid of the `ifdef HAVE_INET6` compile time check?

src/ua.c
```diff
@@ -1982,6 +2021,59 @@ struct sipevent_sock *uag_sipevent_sock(void)
...
+	struct sa sa1;
+	struct sa sa2;
+	int err;
+#endif
+
```
alfredh, line 2044: the uri has an `int af` parameter, that has the following meaning:

- AF_UNSPEC domain
- AF_INET IPv4
- AF_INET6 IPv6

if uri is a domain (i.e. af=AF_UNSPEC) I believe that this check should be skipped.

src/ua.c
```diff
@@ -1982,6 +2021,59 @@ struct sipevent_sock *uag_sipevent_sock(void)
...
+	/* we list cases where we know there is a mismatch in af */
+#ifdef HAVE_INET6
+	if (accu->af == AF_INET && peeru->af == AF_INET6)
+		return false;
+
+	if (accu->af == AF_INET6 && peeru->af == AF_INET)
+		return false;
```
alfredh, line 2051: perhaps these 5 lines could be replaced by:

```c
if (accu->af != peer->af)
    return false;
```

src/ua.c
```diff
@@ -1982,6 +2021,59 @@ struct sipevent_sock *uag_sipevent_sock(void)
...
+					&peeru->host, err);
+			return false;
+		}
+
+		return sa_is_linklocal(&sa1) == sa_is_linklocal(&sa2);
+	}
+
+	/* if peer uri is a domain, we simply return true */
+	if (peeru->af == AF_UNSPEC)
+		return true;
```
alfredh, line 2069: perhaps move this check a bit further up ?

src/ua.c
```diff
@@ -2020,6 +2112,69 @@ struct ua *uag_find(const struct pl *cuser)
...
+ * Find the correct UA from SIP message
+ *
+ * @param msg SIP message
+ *
+ * @return Matching UA if found, NULL if not found
+ */
+struct ua *uag_find_msg(const struct sip_msg *msg)
+{
+	struct le *le;
+	const struct pl *cuser = &msg->uri.user;
```
alfredh, line 2125: please note that this is a global function, and input parameters should be checked:

```
if (!msg)
    return NULL;
```

if msg is NULL then the program will crash

src/ua.c
cspiel1: Hm, we had to disable HAVE_INET6 for one project where the customer has HW + OS linux with disabled IPv6 in kernel. With HAVE_INET6 the network initialization failed. Anyway, it should be possible to fix this in baresip.

src/ua.c
cspiel1: I move these lines up to the top of the function.

```
	/* if peer uri is a domain, we simply return true */
	if (peeru->af == AF_UNSPEC)
		return true;
```

src/ua.c
cspiel1: If I add another exception for accu->af == AF_UNSPEC. I'll write the function differently.

src/ua.c
```diff
@@ -932,6 +932,48 @@ int ua_uri_complete(struct ua *ua, struct mbuf *buf, const char *uri)
...
+	if (!uri)
+		return false;
+
+	for (i=0; i<ARRAY_SIZE(hostv); i++) {
+
+		if (!pl_strcmp(&uri->host, hostv[i]))
+			return true;
+	}
+
+	for (i=0; i<ARRAY_SIZE(afv); i++) {
```
alfredh, line 958: at this point, the address family is already known in `uri->af`.

so there should be no need to check both address families, just use uri->af and lookup
the matching address.

src/ua.c
```diff
@@ -2020,6 +2113,73 @@ struct ua *uag_find(const struct pl *cuser)
...
+		struct account *acc = ua->acc;
+
+		if (!uri_match_transport(&acc->luri, msg->tp))
+			continue;
+
+		if (!uri_match_af(&acc->luri, &msg->uri))
+			continue;
+
+		if (uri_host_local(&acc->luri) != uri_host_local(&msg->uri))
+			continue;
```
alfredh, line 2155: I am wondering why we need to compare the IP-address of the incoming request?

the SIP stack is listening to an interface:

```
transports:
  10.0.1.13:56669 (UDP)
  10.0.1.13:56726 (TCP)
  10.0.1.13:56727 (TLS)
  10.0.1.13:9 (WS)
  10.0.1.13:9 (WSS)
```

an incoming request to 10.0.1.13 will be handled by the SIP-stack, and then
passed up to the application. incoming IP packets to different address/port will be
dropped by the kernel.

so I am not sure if I understand the usecase, say you have 3 uas:

```
<sip:user@10.0.1.13>
<sip:user@127.0.0.1>
<sip:user@localhost>
```

you want the incoming request to arrive at the correct UA?
is it perhaps enough to just use one UA?

alfredh:
There is something that I dont understand about this patch, perhaps you can explain it..

The optional transport parameter in the URI is mainly for outgoing requests.
There are 4 values, none plus the basic transport types (excl. WSS)

1. no transport
2. ;transport=udp
3. ;transport=tcp
4. ;transport=tls

The SIP stack is by default enabling IPv4+IPv6 and all transports.

If you have one useragent:

`<sip:user@domain;transport=tcp>`

means that TCP is used for outgoing+incoming requests.
A valid incoming SIP request via UDP or TLS will in this case be rejected.
I am not sure if this is the correct behaviour?

One more thing; We should also consider if a registrar is used or not:

1. using a registrar
2. not using a registrar

in case 1.) we want all incoming requests to arrive *from* the SIP registrar.
so the new checks that you have added are not needed for this case.

in case 2.) the incoming SIP requests can arrive from anywhere.

possibly one solution is to only perform the extra checks if `regint` is 0.

cspiel1:
Alfred, you are right. My understanding is, that for registration accounts the cuser match should work in 100% (for a RFC conform SIP proxy). So the second for loop should be reached mostly for peer-to-peer calls where a none-registrar account should match. This is what I added to your comment in line 2142.

In the commit message I wrote:
```
    This allows to deny certain transport protocols (e.g. only allow TLS) for
    incoming peer-to-peer calls. Also checks for correct address family which is
    important if there is no af hint in the SDP.
```
Hope this makes clear what this commit does. A limit to regint=0 accounts is okay for us. I will add a commit.

juha-h:
Christian Spielberger writes:

> Hope this makes clear what this commit does. A limit to regint=0
> accounts is okay for us. I will add a commit.

That limitation is good idea in order to avoid possible backwards
compatibility issues.

PR #1207, opened 2021-02-03
Remove uag next

cspiel1:
Get rid of the current UA which was used for dialing and also for answer call, hangup.

Instead menu holds the id of the active call.

TODO: For dialing an automatic UA selection will be implemented.

### DIFF
--- a/modules/menu/dynamic_menu.c
+++ b/modules/menu/dynamic_menu.c
@@ -43,27 +43,15 @@ static int call_audio_debug(struct re_printf *pf, void *arg)
 static int cmd_find_call(struct re_printf *pf, void *arg)
 {
 	struct cmd_arg *carg = arg;
-	struct le *le;
 	const char *id = carg->prm;
-	struct call *call;
-	int err;
-
-	for (le = list_head(uag_list()); le; le = le->next) {
-		struct ua *ua = le->data;
-		struct list *calls = ua_calls(ua);
-
-		call = call_find_id(calls, id);
-		if (call) {
-			err = re_hprintf(pf, "setting current call: %s\n", id);
-			call_set_current(calls, call);
-			menu_get()->ua_cur = call_get_ua(call);
-		}
-	}
+	struct call *call = menu_find_call(id);
 
 	if (!call)
-		err = re_hprintf(pf, "call not found (id=%s)\n", id);
+		(void)re_hprintf(pf, "menu: call not found (id=%s)\n", id);
+	else
+		(void)re_hprintf(pf, "menu: setting current call: %s\n", id);
 
-	return err;
+	return 0;
 }
 
 
@@ -90,10 +78,11 @@ static int set_current_call(struct re_printf *pf, void *arg)
 	if (call) {
 		err = re_hprintf(pf, "setting current call: line %u\n",
 				 linenum);
-		call_set_current(ua_calls(ua), call);
+		menu_selcall(call);
 	}
 	else {
-		err = re_hprintf(pf, "call not found\n");
+		err = re_hprintf(pf, "menu: call not found (ua=%s, line=%u)\n",
+				account_aor(ua_account(ua)), linenum);
 	}
 
 	return err;
@@ -135,10 +124,11 @@ static int call_reinvite(struct re_printf *pf, void *arg)
 
 static int cmd_call_resume(struct re_printf *pf, void *arg)
 {
+	struct cmd_arg *carg = arg;
+	struct ua *ua = carg->data ? carg->data : menu_uacur();
 	(void)pf;
-	(void)arg;
 
-	return call_hold(ua_call(menu_uacur()), false);
+	return call_hold(ua_call(ua), false);
 }
 
 
@@ -203,20 +193,23 @@ static int call_video_debug(struct re_printf *pf, void *arg)
 static int set_video_dir(struct re_printf *pf, void *arg)
 {
 	const struct cmd_arg *carg = arg;
-	struct ua *ua = menu_uacur();
+	struct call *call = menu_callcur();
 	int err = 0;
 
+	if (!call)
+		return EINVAL;
+
 	if (0 == str_cmp(carg->prm, sdp_dir_name(SDP_INACTIVE))) {
-		err = call_set_video_dir(ua_call(ua), SDP_INACTIVE);
+		err = call_set_video_dir(call, SDP_INACTIVE);
 	}
 	else if (0 == str_cmp(carg->prm, sdp_dir_name(SDP_SENDONLY))) {
-		err = call_set_video_dir(ua_call(ua), SDP_SENDONLY);
+		err = call_set_video_dir(call, SDP_SENDONLY);
 	}
 	else if (0 == str_cmp(carg->prm, sdp_dir_name(SDP_RECVONLY))) {
-		err = call_set_video_dir(ua_call(ua), SDP_RECVONLY);
+		err = call_set_video_dir(call, SDP_RECVONLY);
 	}
 	else if (0 == str_cmp(carg->prm, sdp_dir_name(SDP_SENDRECV))) {
-		err = call_set_video_dir(ua_call(ua), SDP_SENDRECV);
+		err = call_set_video_dir(call, SDP_SENDRECV);
 	}
 	else {
 		(void)re_hprintf(pf, "Invalid video direction %s"
@@ -232,16 +225,9 @@ static int set_video_dir(struct re_printf *pf, void *arg)
 static int digit_handler(struct re_printf *pf, void *arg)
 {
 	const struct cmd_arg *carg = arg;
-	struct call *call;
-	int err = 0;
+	(void) pf;
 
-	(void)pf;
-
-	call = ua_call(menu_uacur());
-	if (call)
-		err = call_send_digit(call, carg->key);
-
-	return err;
+	return call_send_digit(menu_callcur(), carg->key);
 }
 
 
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -588,7 +588,7 @@ struct menu *menu_get(void)
 struct call *menu_find_call(const char *id)
 {
 	struct le *le;
-	struct call *call = NULL;
+	struct call *call;
 
 	if (!str_isset(id))
 		return NULL;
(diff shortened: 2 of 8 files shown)


cspiel1:
@sreimers This fails:
```
sudo apt-get update && sudo apt-get install -y libssl-dev valgrind
```
Should I solve this somehow?

sreimers:
> @sreimers This fails:
> ```
> sudo apt-get update && sudo apt-get install -y libssl-dev valgrind
> ```
> Should I solve this somehow?

I just re-run and looks good now. Rarely the used apt mirrors seems to be broken.

cspiel1:
TODO: The call state has to be taken into account somehow for the "active call". Relates also to @alfredh-s comment https://github.com/baresip/baresip/pull/1199#issuecomment-769947402

alfredh:
the ua context can be passed in from the command input, for example mqtt:

https://github.com/baresip/baresip/blob/master/modules/mqtt/subscribe.c#L83

it is therefore important that the ua selection logic be kept.

modules/menu/dynamic_menu.c
```diff
@@ -12,76 +12,49 @@
 static int set_audio_bitrate(struct re_printf *pf, void *arg)
 {
 	struct cmd_arg *carg = arg;
-	struct ua *ua = carg->data ? carg->data : menu_uacur();
-	struct call *call;
+	struct call *call = menu_getcall(pf);
```
alfredh, line 15: we should keep this logic:

1. if carg->data is set, use ua from there
2. if carg->data is NULL, lookup ua from current call

modules/menu/menu.c
```diff
@@ -580,18 +582,77 @@ struct menu *menu_get(void)
...
+ */
+int menu_selcall(struct re_printf *pf, const char *id)
 {
-	if (!menu.ua_cur)
-		menu.ua_cur = list_ledata(list_head(uag_list()));
+	struct call *call = menu_find_call(id);
+	struct list *calls = ua_calls(call_get_ua(call));
+
+	if (call) {
+		menu.callid = id;
```
alfredh, line 617: please note that the call-id string is dynamically allocated, and should be
copied using e.g. `str_dup`.

otherwise if the call object is destructed, the callid string will point to freed memory

modules/menu/menu.h
```diff
@@ -17,8 +17,7 @@ struct menu{
 	struct tmr tmr_stat;          /**< Call status timer              */
 	struct play *play;            /**< Current audio player state     */
 	struct mbuf *dialbuf;         /**< Buffer for dialled number      */
-	struct le *le_cur;            /**< Current User-Agent (struct ua) */
-	struct ua *ua_cur;
+	const char *callid;           /**< Call-id of active call         */
```
alfredh, line 20: should be `char *callid`

also remember to mem_deref in module_close().

alfredh:
I can suggest some simple code for setting/getting active call:

```c
void menu_selcall(struct call *call)
{
	menu.callid = mem_deref(menu.callid);

	if (call)
		str_dup(&menu.callid, call_id(call));
}


struct call *menu_getcall(void)
{
	if (menu.callid)
		return menu_find_call(menu.callid);
	else
		return NULL;
}


struct ua *menu_uacur(void)
{
	return call_get_ua(menu_getcall());
}
```

for `CALL_CLOSED` event you can call menu_setcall(NULL);

if you keep the menu_uacur() signature, it will also make the diff smaller :)

modules/menu/dynamic_menu.c
cspiel1: Oh, I see. Did'nt saw how cmd_process_long() is used in mqtt. In ctrl_tcp and ctrl_dbus we should do this in the same way. 

But here is a problem. The parameter is a void*. And you allow all commands. Suppose one writes a command "do_foo" and wants to get a foo* in the data parameter and maybe he writes also an ui module which correctly passes a foo* in the data parameter. Then a user may also call do_foo by means of mqtt and specify the accountaor. Then ua* is cast to foo*.

cspiel1:
In menu_selcall() you omitted
```
call_set_current(calls, call);
```
We don't need the reorder of the calls in the UA, right?

modules/menu/dynamic_menu.c
cspiel1: For now I'll keep the logic.

cspiel1:
Because you want to keep the ua selection logic per data parameter (https://github.com/baresip/baresip/pull/1207#discussion_r573220048) we also have to call call_set_current(). In order to let the diff be small like you stated, I will replace the last commit.

cspiel1:
The last commit is a little bit off topic. But came up to me while integrating the new menu_find_call(). Should I make a separate PR for this?

cspiel1:
Also what fits to this PR is what @alfredh already suggested: cmd_answer() should operate on the last incoming call.

- first hold the last established call
- then answer the last incoming call

alfredh:
I tested calling a number and it is not working:

```
/dial music
menu: ua_connect failed: Invalid argument
```

modules/menu/dynamic_menu.c
```diff
@@ -15,19 +15,19 @@ static int set_audio_bitrate(struct re_printf *pf, void *arg)
...
 		audio_set_bitrate(call_audio(call), bitrate);
 	}
 	else {
-		err = re_hprintf(pf, "call not found\n");
+		(void)re_hprintf(pf, "call not found\n");
+		return EINVAL;
 	}
 
-	return err;
+	return 0;
```
alfredh, line 30: can you please make a separate PR for this block ?

it is not related to the main topic of this PR.

modules/menu/static_menu.c
```diff
@@ -116,7 +115,7 @@ static int cmd_answerdir(struct re_printf *pf, void *arg)
 			"[^ ]*", &argdir[0]);
 
 	if (err) {
-		warning("%s", usage);
+		(void) re_hprintf(pf, "%s", usage);
```
alfredh, line 118: please also make a separate PR for this and similar changes.

alfredh:
can you please rebase this PR on top of master HEAD ?

also, is it possible for you to squash all the commits into one single commit,
since you are doing a forced push ? it will make things a bit cleaner.

juha-h:
Sorry, but I haven't really followed the details of "Remove uag next"
effort.

What is the replacement of it?

Let's say that I have two accounts sip:foo@example.com and
sip:bar@another.com.  Then I want to make a call to PSTN:

/dial +3582345670

How can I choose, which of the two accounts is selected to make the
call, since there is nothing that could make baresip to automatically
choose the right one?

alfredh:
I would like to suggest that the UA selection is done on a per-command basis.

example:

```
/dial +3582345670 ua=0
```

juha-h:
Alfred E. Heggestad writes:
> I would like to suggest that the UA selection is done on a per-command basis.
> 
> example:
> 
> ```
> /dial +3582345670 ua=0
> ```

Then I would need to remember, which one is ua 0.

And how about if I chat with someone?  Would I need in every message
also tell the ua?

cspiel1:
- PR https://github.com/baresip/baresip/pull/1217 should do the automatic selection. An PSTN would be handled like a user only SIP address. This would select the first registered UA.
- A manual selection like @alfredh suggests as second command parameter would replace the uag_next() if the application wants to select a specific UA. --> TODO: another PR.

A chat application may invoke:
- /message a@proxy1
- /message a ua=0

A pre-selection could be handled in the application.

cspiel1:
> I tested calling a number and it is not working:
> 
> ```
> /dial music
> menu: ua_connect failed: Invalid argument
> ```

This PR removes uag_next() and introduces the active call for commands that belong to the active call.
The UA selection for dial is provided by https://github.com/baresip/baresip/pull/1217.

juha-h:
Currently I can re-dial previous destination just by pressing `d`.  If it is menu module that remembers the previous destination, could it do the same for ua?

juha-h:
Just checked and there exists le_cur, ua_cur, and redial_aor fields in `struct menu`.  So if I select one of the uas as current in menu module, there is no need for current ua related API calls and user experience would not need to change.  It should thus be possible to keep the /ua_next command (if there ever was intention to remove it).

cspiel1:
The ua_cur was the UA of the active call. And this UA also was used for future dials. But I think this are two different use-cases.

In the closed PR https://github.com/baresip/baresip/pull/1199#discussion_r566977088 @alfredh suggested to remove the current UA. The plan is to:
- Remove the current UA for dialing. Instead remember the active call for commands like answer, hold, resume, ...
- Add an automatic UA selection for dial, options and any outgoing requests.
- Add an optional manual UA selection for outgoing commands that overrides the automatic UA selection. It should be more convenient then ua_next().
- The redial_aor is used for scode 701. But we could also use it for a manual redial and link the 'd' in menu to a new function.

This PR is only the first part. The second is https://github.com/baresip/baresip/pull/1217. And I see that it was not a good idea to make separate PRs. I will add the commits of https://github.com/baresip/baresip/pull/1217 here. Maybe it will make clearer that we want to improve the comfort for the user, instead of removing functionality. Then I will also add a manual override.

But I suggest to review the commits one-by-one. This should be easier.

cspiel1:
@alfredh @juha-h Now this PR includes also the automatic UA selection and the manual override for dial and options. The indices for the manual selection are printed with command reginfo.

The manual override (function menu_ua_carg) is a first draft. Your suggestions are welcome.

For future PRs we may use the manual override also for other commands.

@juha-h You mentioned the UA selection for cmd_message. This is part of module contact. Nothing changes for this command, because it uses the unchanged ua_lookup_domain() in module contact.

juha-h:
Christian Spielberger writes:

> @juha-h You mentioned the UA selection for cmd_message. This is part
> of module contact. Nothing changes for this command, because it uses
> the unchanged ua_lookup_domain() in module contact.

Good, since I consider >/< (contact_next/contact_prev) very useful.
After all the enhancement, will there exist similar ua_next/ua_prev menu
commands?

cspiel1:
uag_next() is not used for cmd_message. Instead ua_lookup_domain().

In an intermediate state of PR https://github.com/baresip/baresip/pull/1199 I proposed an "auto" mode which could be selected with uag_next() by iterating over the UAs. If "auto" was selected then the automatic UA selection was used. But only for dial. @alfredh and me want to get rid of uag_next(). I think that the user should not have to deal with selecting an adequate UA. He wants to specify a SIP target address, not more. Then it should just work.

The advanced user may use the manual per command UA selection, which I added here shortly as a first draft.

First of all, we should distinguish the active call UA from the outgoing request UA (like dial). @alfredh Please could you make a decision?
- Should we keep uag_next() for outgoing requests? This would be ok for me, if we do this in addition to the automatic selection.
- Should we also keep the proposed manual per command UA selection, like you suggest above?

PR #1208, opened 2021-02-04
ctrl_dbus: generate dbus interface during build

cspiel1:


cspiel1:
It seems that in Ubuntu 16.04 the DBus C generator produces a gcc warning
```
modules/ctrl_dbus/baresipbus.c:338:62: error: ISO C does not allow extra ‘;’ outside of a function [-Werror=pedantic]
G_DEFINE_INTERFACE (DBusBaresip, dbus_baresip, G_TYPE_OBJECT);
```
With -Werror this lets the build fail.

How should we handle this?

modules/ctrl_dbus/gen.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+gdbus-codegen --generate-c-code baresipbus --c-namespace DBus \
+    --interface-prefix com.creytiv. com.creytiv.Baresip.xml
```
alfredh, line 3: please note that creytiv.com domain has been closed.
you should probably use a different domain

modules/ctrl_dbus/module.mk
```diff
@@ -15,5 +15,7 @@ $(MOD)_CFLAGS	+= -Wno-unused-parameter -Wno-declaration-after-statement
 
 $(MOD)_CCHECK_OPT	= -e baresipbus.h -e baresipbus.c
 
-include mk/mod.mk
+modules/ctrl_dbus/baresipbus.c: modules/ctrl_dbus/com.creytiv.Baresip.xml
+	@cd $(dir $@) && ./gen.sh
```
alfredh, line 19: I suggest also adding the header file to the makefile rule, since the output is .h and .c

example:

```
modules/ctrl_dbus/baresipbus.c modules/ctrl_dbus/baresipbus.h: modules/ctrl_dbus/com.creytiv.Baresip.xml
        cd $(dir $@) && ./gen.sh
```

alfredh:
the extra warning is because of this flag: -pedantic

could you try to disable it and see if you get any warnings ?

modules/ctrl_dbus/gen.sh
cspiel1: com.baresip.Baresip.xml ?

cspiel1: Have no better idea. If you have, I'll change it later.

sreimers: maybe: `com.github.Baresip.xml` ?

cspiel1: ok. definitely better.

modules/ctrl_dbus/module.mk
```diff
@@ -5,15 +5,19 @@
 #
 
 MOD		:= ctrl_dbus
-$(MOD)_SRCS	+= ctrl_dbus.c baresipbus.c
+$(MOD)_SRCS	+= baresipbus.c ctrl_dbus.c
```
alfredh, line 8: this change should not be needed.

the order of .c files in SRCS should not matter

cspiel1:
Missing: Add the generated files to the clean target. How this could be solved?

modules/ctrl_dbus/module.mk
cspiel1: - Tried on my dev host and found you must be right.
- Changed this in the commit.
- But travis tells:
```
modules/ctrl_dbus/ctrl_dbus.c:65:10: fatal error: baresipbus.h: No such file or directory
 #include "baresipbus.h"
          ^~~~~~~~~~~~~~
```
It is plausible, since the Makefile target baresipbus.h is never referenced.

With the reorder baresipbus.h is generated with baresipbus.c. But it is not a beautiful solution. How should I reference the header file correctly?

alfredh:
merged, thank you!

cspiel1:
Thanks. If someone knows a solution for putting the generated files to the "clean" target, then let me know!

The most important Makefile automatism is that if the xml is updated, the DBus C code is re-generated. I hope so.

Issue #1212, opened 2021-02-07
Use a v4l2 usb camera with mjpeg pixformat


viric:
I can't find how would one use a v4l2 vidsrc with mjpeg pixformat. Is it supported?

Many usb webcams can send not only YUYV pixfmt but also MJPG in v4l2, and that's effectively what allows 30fps at more than 640x480.

Let's say I want 1280x720x30fps... that can only be achieved with mjpeg v4l2 frames.

viric:
This issue https://github.com/baresip/baresip/issues/639 about IP MJPEG cameras faces the same challenge as for v4l2 MJPEG cameras.

I think that the avformat code can be extended to allow mjpeg acquisition from v4l2 but we lack the pixfmt conversion. mjpeg decoding provides yuvj422p.

I'll try if I can extend avformat with swscale to perform the color conversion although I'm new in this field.

alfredh:
MJPEG is not a pixel format, but a codec:

```
$ ffmpeg -hide_banner -decoders | grep -i mjpeg
 V....D mjpeg                MJPEG (Motion JPEG)
 V....D mjpegb               Apple MJPEG-B
 A....D adpcm_ima_smjpeg     ADPCM IMA Loki SDL MJPEG
```

to use a v4l2 camera you have two options:

1. v4l2.so
2. avformat.so

v4l2.so can also use libv4l2, it is possible that libv4l2 may be able to decode mjpeg to raw frames.
please try it and see if it works.

for avformat.so you can use this config:

```
video_source           avformat,v4l2,/dev/video0
```

it will also decode any input frames, and send a raw pixel format to the core (i.e. yuv720p)

please try it and see if it works.

if you have tried something that did not work, it would be very useful if you can
paste the exact config and matching log-file here.

viric:
Check how I address this in the PR https://github.com/baresip/baresip/pull/1216

alfredh:
tracked in PR

Issue #639, opened 2019-01-20
avformat MJPEG stream avformat: decode: bad pixel format (13) (yuvj422p)


daanpape:
Hello,

I'm working a project and I want to take in an MJPEG camera stream and use it as a video input for a SIP call. I have compiled baresip with the VP8 encoder and while this works perfectly with a USB webcam it doesn't work with the avformat input. I'm probably setting up something wrong but I can't find the right documentation. This is my configuration:

```
#
# baresip configuration
#

#------------------------------------------------------------------------------

# Core
poll_method             epoll           # poll, select, epoll ..

# SIP
sip_trans_bsize         128
#sip_listen             0.0.0.0:5060
#sip_certificate        cert.pem

# Audio
audio_player            alsa,default
audio_source            alsa,default
audio_alert             alsa,default
audio_srate             
audio_channels          1-2
#ausrc_srate            48000
#auplay_srate           48000
#ausrc_channels         0
#auplay_channels                0

# Video
video_source            avformat,http://127.0.0.1:8080/?action=stream
#video_display          x11,nil
video_size              352x288
#video_bitrate          262144
#video_fps              10

# AVT - Audio/Video Transport
rtp_tos                 184
#rtp_ports              10000-20000
#rtp_bandwidth          512-1024 # [kbit/s]
rtcp_enable             yes
rtcp_mux                no
jitter_buffer_delay     0-5             # frames
rtp_stats               no

# Network
#dns_server             10.0.0.1:53
#net_interface          eth0.1

# BFCP
#bfcp_proto             udp

#------------------------------------------------------------------------------
# Modules

module_path             /usr/lib/baresip/modules

# UI Modules
module                  stdio.so
#module                 cons.so
#module                 evdev.so
#module                 httpd.so

# Audio codec Modules (in order)
module                  opus.so
#module                 silk.so
#module                 amr.so
#module                 g7221.so
module                  g722.so
#module                 g726.so
#module                 g711.so
#module                 gsm.so
#module                 l16.so
#module                 speex.so
#module                 bv32.so

# Audio filter Modules (in encoding order)
module                  vumeter.so
#module                 sndfile.so
#module                 speex_aec.so
#module                 speex_pp.so
#module                 plc.so

# Audio driver Modules
module                  alsa.so
#module                 portaudio.so

# Video codec Modules (in order)
#module                 avcodec.so
#module                 v4l2_codec.so
module                  vp8.so
module                  vp9.so

# Video filter Modules (in encoding order)
[...]

PR #1215, opened 2021-02-08
RFC mjpeg 30fps from v4l2 avformat

viric:
Using swsscale to do the pixel conversion from YUV422P -> anything else, like YUYV422.

How do you see using swsscale? Or you would rather have this conversion implemented in librem?

Related to #639 and #1212.

### DIFF
--- a/modules/avformat/audio.c
+++ b/modules/avformat/audio.c
@@ -11,6 +11,7 @@
 #include <libavutil/opt.h>
 #include <libavformat/avformat.h>
 #include <libavcodec/avcodec.h>
+#include <libswscale/swscale.h>
 #include <libswresample/swresample.h>
 #include "mod_avformat.h"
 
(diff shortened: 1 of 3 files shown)


viric:
In fact I don't see much future for my needs in this swsscale path... I want to use hwaccel for mjpeg decoding (mmal on pi, vaapi on intel), and both hwaccel can produce already yuv420p. I should have rather spent my avformat time introducing ffmpeg's hwaccel.

alfredh:
can you please try to use the `swscale.so` module ?

I am not sure what kind of problem you are trying to solve.
it would greatly help if you can attach a full verbose log of the failing case.

BTW: we might consider to remove decoding from avformat, and pass the raw bitstream to the core

viric:
The problem I try to solve is to use v4l2 with mjpeg encoding, with hw decoding (and hw encoding to h264) if possible. In many usb webcams mjpeg is the only way to get 30fps for sizes >640x480.

I started with this RFC with software decoding and I missed there was a swscale.so module. Then I succesfully used hw decoding in https://github.com/baresip/baresip/pull/1216

I think I better close this PR and focus on #1216

PR #1216, opened 2021-02-09
RFC: Make avformat decode mjpeg v4l2 with vaapi

viric:
This makes mjpeg decoding on avformat+v4l2 with hwaccel vaapi (common in GNU/Linux specially with Intel CPUs). It works fine here with almost no cpu use on vidloop at 720p 30fps.

All is hardcoded so... how do you see adding some config values to handle this?

Something like:
avformat_hwaccel vaapi           # ffmpeg -hwaccel vaapi ....
avformat_inputformat mjpeg     # ffmpeg -input_format -f v4l2 ...

### DIFF
--- a/modules/avformat/avformat.c
+++ b/modules/avformat/avformat.c
@@ -14,7 +14,9 @@
 #include <libavformat/avformat.h>
 #include <libavcodec/avcodec.h>
 #include <libavdevice/avdevice.h>
+#if LIBAVUTIL_VERSION_MAJOR >= 56
 #include <libavutil/hwcontext.h>
+#endif
 #include "mod_avformat.h"
 
 
@@ -38,7 +40,9 @@
 static struct ausrc *ausrc;
 static struct vidsrc *mod_avf;
 
+#if LIBAVUTIL_VERSION_MAJOR >= 56
 static enum AVHWDeviceType avformat_hwdevice = AV_HWDEVICE_TYPE_NONE;
+#endif
 static char avformat_inputformat[64];
 static AVCodec *avformat_decoder;
 
@@ -182,14 +186,15 @@ static int open_codec(struct stream *s, const struct AVStream *strm, int i,
 		return ENOMEM;
 	}
 
+#if LIBAVUTIL_VERSION_MAJOR >= 56
 	if (avformat_hwdevice != AV_HWDEVICE_TYPE_NONE)
 	{
 		AVBufferRef *hwctx;
 		ret = av_hwdevice_ctx_create(&hwctx, avformat_hwdevice,
 				NULL, NULL, 0);
 		if (ret < 0) {
-			warning("avformat: error opening hw device vaapi (%i)\n",
-					ret);
+			warning("avformat: error opening hw device vaapi"
+                                       " (%i)\n", ret);
 			return ENOMEM;
 		}
 
@@ -197,6 +202,7 @@ static int open_codec(struct stream *s, const struct AVStream *strm, int i,
 
 		av_buffer_unref(&hwctx);
 	}
+#endif
 
 	s->time_base = strm->time_base;
 	s->ctx = ctx;
@@ -298,8 +304,8 @@ int avformat_shared_alloc(struct shared **shp, const char *dev,
 	}
 
 	if (strnlen(avformat_inputformat, sizeof avformat_inputformat)	> 0) {
-		ret = av_dict_set(&format_opts, "input_format", avformat_inputformat,
-				0);
+		ret = av_dict_set(&format_opts, "input_format",
+                                avformat_inputformat, 0);
 		if (ret != 0) {
 			warning("avformat: av_dict_set(input_format) failed"
 					" (ret=%s)\n", av_err2str(ret));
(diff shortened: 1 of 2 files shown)


viric:
About your comment on using the module swscale for non-hwaccel to decode mjpeg... I need to go from yuvj422p to the dst... But librem doesn't know yuvj422p even to be able to name it.

Maybe we could use swscale internally in avformat to automatically convert to videnc_format, without requiring a filter?

alfredh:
modules/avformat/avformat.c:17:33: fatal error: libavutil/hwcontext.h: No such file or directory
561

perhaps you need to add some compile-time version check.

alfredh:
yuvj422p is yuv422p with JPEG colorspace:

https://github.com/baresip/baresip/blob/master/modules/avcodec/decode.c#L67

viric:
> yuvj422p is yuv422p with JPEG colorspace:

Do you think I could treat yuvj422p as if it was yuv422p to the effect of librem pixfmt conversion, and that would fit?

viric:
> modules/avformat/avformat.c:17:33: fatal error: libavutil/hwcontext.h: No such file or directory
> 561
> 
> perhaps you need to add some compile-time version check.

Would it be fair to disable the hw decoding completely, unless some version is available? Because I don't know the old API to achieve this.

alfredh:
I tested with hwaccel videotoolbox on OSX and it crashes:

```
/vidloop
Enable video-loop on avformat,lavfi,testsrc2: 320 x 240
vidloop: open video display (.)
vidloop: avformat,lavfi,testsrc2: open video source: 320 x 240 at 30.00 fps
Segmentation fault: 11 (core dumped)
```

```
Thread 13 Crashed:
0   libavutil.56.dylib            	0x0000000111d294ce av_hwframe_transfer_data + 94
1   avformat.so                   	0x00000001175f97ac read_thread + 812
2   libsystem_pthread.dylib       	0x00007fffa59e793b _pthread_body + 180
3   libsystem_pthread.dylib       	0x00007fffa59e7887 _pthread_start + 286
4   libsystem_pthread.dylib       	0x00007fffa59e708d thread_start + 13
```

config:

```
avformat_hwaccel        videotoolbox
```

viric:
If you use 'vaapi' but the source is not in an apropriate codec for it to decode, it will crash... Right, I have to handle that!

For example, vaapi and YUY2 from usb webcam, crash. vaapi and MJPG, ok.

juha-h:
viric writes:

> > modules/avformat/avformat.c:17:33: fatal error: libavutil/hwcontext.h: No such file or directory
> > 561
> > 
> > perhaps you need to add some compile-time version check.
> 
> Would it be fair to disable the hw decoding completely, unless some version is available? Because I don't know the old API to achieve this.

When I build baresip avformat module for Android, I do not get that
error.

modules/avformat/avformat.c
```diff
@@ -173,6 +179,21 @@ static int open_codec(struct stream *s, const struct AVStream *strm, int i,
 		return ENOMEM;
 	}
 
+    if (avformat_hwdevice != AV_HWDEVICE_TYPE_NONE)
+    {
```
alfredh, line 183: please move curly bracket to end of previous line.

also please check the indentation, should be TAB(s)

modules/avformat/avformat.c
```diff
@@ -173,6 +179,21 @@ static int open_codec(struct stream *s, const struct AVStream *strm, int i,
...
+        ret = av_hwdevice_ctx_create(&hwctx, avformat_hwdevice,
+                NULL, NULL, 0);
+        if (ret < 0) {
+            warning("avformat: error opening hw device vaapi (%i)\n", ret);
+            return ENOMEM;
+        }
+
+        ctx->hw_device_ctx = av_buffer_ref(hwctx);
+
+        av_buffer_unref(&hwctx);
```
alfredh, line 194: perhaps these two lines can be replaced by:

```c
        ctx->hw_device_ctx = hwctx;
```

modules/avformat/avformat.c
```diff
@@ -368,7 +399,19 @@ void avformat_shared_set_video(struct shared *sh, struct vidsrc_st *st)
 
 static int module_init(void)
 {
-	int err;
+    int err;
+    char hwaccel[64] = "";
+
+    conf_get_str(conf_cur(), "avformat_hwaccel", hwaccel, sizeof(hwaccel));
+    if (strnlen(hwaccel, sizeof(hwaccel)) > 0) {
```
alfredh, line 406: you can also use `str_isset` to check if the string is set, or
check return value of conf_get_str.

alfredh:
regarding the compile-time check for ffmpeg, we have a similar check in avcodec.so:

https://github.com/baresip/baresip/blob/master/modules/avcodec/avcodec.c#L48

alfredh:
hi @viric do you have time to address my review comments ?

viric:
I will I will. I guess there is no hurry.

I'm waiting for a new computer to arrive and engage again refining this.

On Wed, Feb 17, 2021 at 03:38:51AM -0800, Alfred E. Heggestad wrote:
> hi @viric do you have time to address my review comments ?
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/baresip/baresip/pull/1216#issuecomment-780498281

-- 
(Escriu-me xifrat si saps PGP / Write ciphered if you know PGP)
PGP key 7CBD1DA5 - https://emailselfdefense.fsf.org/

modules/avformat/avformat.c
```diff
@@ -105,10 +116,10 @@ static void *read_thread(void *data)
 				sys_msleep(1000);
 
 				ret = av_seek_frame(st->ic, -1, 0,
-						    AVSEEK_FLAG_BACKWARD);
+							AVSEEK_FLAG_BACKWARD);
```
alfredh, line 119: unrelated change

alfredh:
please note that your patch
contains many whitespace changes.

please review your own patch

viric:
oh sorry. I didn't originally care and it seems I used a very inappropriate editor configuration for this project.
Tab characters and 8-column tabs, right?

viric:
Sorry for the inconveniences! I think I fixed the space crippling.

PR #1218, opened 2021-02-12
Intelligent call answer

cspiel1:
A first small step toward an intelligent call answer.

It is bound to the menu_uacur(). After merge of https://github.com/baresip/baresip/pull/1207 this is the UA of the active call.

For further improvements, should we consider an overall UA call answer strategy?


webstean:
Nice!

PR #1219, opened 2021-02-12
message: reply only once

cspiel1:
baresip sends a reply for a SIP MESSAGE for each message listener.

How often we process a SIP MESSAGE is not interesting for the remote who sent the MESSAGE. In our case “menu” prints the message and ctrl_tcp forwards the message.

### DIFF
--- a/src/message.c
+++ b/src/message.c
@@ -38,20 +38,22 @@ static void listener_destructor(void *data)
 
 
 static void handle_message(struct message_lsnr *lsnr, struct ua *ua,
-			   const struct sip_msg *msg)
+			   const struct sip_msg *msg, bool hdld)
 {
 	static const char ctype_text[] = "text/plain";
 	struct pl ctype_pl = {ctype_text, sizeof(ctype_text)-1};
 	(void)ua;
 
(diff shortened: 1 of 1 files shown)


cspiel1:
@sreimers A problem with TLS?
[ RUN      ] test_message
262
tls: connect: error (r=-1, ssl_err=1)

I can't reproduce this on my dev host.

sreimers:
Should work now (just restarted the tests), i reverted the last baresip/re change and make tls verify opt-in, until we make it configurable.

cspiel1:
Ah, I see. I forgot to update libre. Thanks for the hint!

cspiel1:
Should I make the PR for the accounts config?
We will have to pass it to libre somehow in sip_transp_send() and conn_send(). Do you have an suggestion?

sreimers:
> Should I make the PR for the accounts config?

That would be nice.

> We will have to pass it to libre somehow in sip_transp_send() and conn_send(). Do you have a suggestion?

Maybe throw sip_transp_add and the `Optional transport parameters such as TLS context` https://github.com/baresip/re/blob/master/src/sip/transp.c#L1021

But I didn't look into it any further. Maybe there is a better way.

PR #1221, opened 2021-02-18
Disable verify server cert

cspiel1:
Allows to disable the SIP TLS server verification if configured in accounts. By default the server verification is enabled for SIP TLS.

TODO: testing

### DIFF
--- a/docs/examples/accounts
+++ b/docs/examples/accounts
@@ -28,6 +28,7 @@
 #    ;regq=0.5
 #    ;sipnat={outbound}
 #    ;stunserver=stun:[user:pass]@host[:port]
+#    ;verify_server_cert=yes
 #    ;video_codecs=h264,h263,...
 #
 # Examples:
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -108,6 +108,7 @@ const char *account_mwi(const struct account *acc);
 const char *account_call_transfer(const struct account *acc);
 const char *account_extra(const struct account *acc);
 bool account_sip_autoanswer(const struct account *acc);
+bool account_verify_server(const struct account *acc);
 
 
 /*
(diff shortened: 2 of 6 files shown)


cspiel1:
Per account disable of SIP TLS server verification is to complicated in libre. The benefit is too low compared to a config flag.

replaced by https://github.com/baresip/baresip/pull/1224

juha-h:
I build baresip from master and then started it using existing configuration/account. I got to console:
```
config: no sip_cafile defined and sip_verify_server is enabled, sip tls connections maybe won't work
```
and
```
tls: verify_handler: err          = 19
tls: verify_handler: error_string = self signed certificate in certificate chain
tls: verify_handler: depth        = 3
tls: connect: error (r=-1, ssl_err=1)
reg: sip:foo@bar.com (prio 0): Register: Protocol error
```
Was it intentional to break backwards compatibility?

cspiel1:
yes, previously SIP TLS was not secure. The remote certificate was not verified. You could use a real certificate and specify a CA-file or disable sip_verify_server in the accounts file.

juha-h:
Christian Spielberger writes:

> yes, previously SIP TLS was not secure. The remote certificate was not
> verified. You could use a real certificate and specify a CA-file or
> disable sip_verify_server in the accounts file.

Yes, I know I can do that, but still, I don't consider breaking existing
baresip configurations a good idea. You could have created new default
configs with

sip_verify_server	yes

and keep 'no' as the default.

-- Juha

cspiel1:
Yes, ok. I will make a PR.

cspiel1:
@juha-h Please vote for https://github.com/baresip/baresip/pull/1236 if it is ok for you?

PR #1224, opened 2021-02-18
disable_verify_server_general

cspiel1:
Adds a config flag sip_verfiy_server that disables the SIP TLS server verification.

By default server verification is turned on.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -12,6 +12,7 @@ poll_method		epoll		# poll, select, epoll ..
 #sip_certificate	cert.pem
 #sip_cafile		ca.crt
 #sip_trans_def		udp
+#sip_verify_server	yes
 
 # Call
 call_local_timeout	120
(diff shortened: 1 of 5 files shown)


docs/examples/config
```diff
@@ -12,6 +12,7 @@ poll_method		epoll		# poll, select, epoll ..
 #sip_certificate	cert.pem
 #sip_cafile		ca.crt
 #sip_trans_def		udp
+#sip_verfiy_server	yes
```
sreimers, line 15: typo ->sip_verify_server

cspiel1: thanks!

include/baresip.h
```diff
@@ -285,6 +285,7 @@ struct config_sip {
 	char cert[256];         /**< SIP Certificate                */
 	char cafile[256];       /**< SIP CA-file                    */
 	enum sip_transp transp; /**< Default outgoing SIP transport protocol */
+	bool sverify;           /**< Enable SIP TLS verify server   */
```
sreimers, line 288: maybe we should align with libre `verify_server`

cspiel1: may I squash the commit?

sreimers (approved):

PR #1227, opened 2021-02-19
event: encode also remote audio direction

cspiel1:
This allows the application to detect if the remote set us on hold/resume call.
The event CALL_REMOTE_SDP already exists. We have to add only the remote audio
direction to the encoded event data.

```
remote audio dir | application
-----------------|----------------------------
recvonly         | The remote set us on hold.
sendonly         | We set the call on hold.
sendrecv         | The call is active (again).
```


cspiel1:
Thanks Alfred!

PR #1235, opened 2021-02-24
event: add string for UA_EVENT_CUSTOM

cspiel1:


juha-h:
Looks good to me.

cspiel1:
thanks!

PR #1236, opened 2021-02-24
config: disables server verification if sip_verify_server is missing

cspiel1:
This commit disables the server verification for old TLS configurations.

Requested by: @juha-h 

Please vote for or against this: @juha-h,  @alfredh, @sreimers , others?

I am not 100% for this commit, because we have to ensure for our products that a config line is added. If it is forgotten, then we believe that SIP TLS is safe. But it isn't.

This has to be added.
```
sip_verify_server	yes
```

Of course our testing have to do negative tests also for SIP TLS.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -12,7 +12,7 @@ poll_method		epoll		# poll, select, epoll ..
 #sip_certificate	cert.pem
 #sip_cafile		ca.crt
 #sip_trans_def		udp
-#sip_verify_server	yes
+sip_verify_server	yes
 
 # Call
 call_local_timeout	120
--- a/src/config.c
+++ b/src/config.c
@@ -31,7 +31,7 @@ static struct config core_config = {
 		"",
 		"",
 		SIP_TRANSP_UDP,
-		true,
+		false,
 	},
 
 	/** Call config */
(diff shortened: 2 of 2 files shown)


juha-h:
I don't know how to vote, but looked at the changes and they are OK.

cspiel1:
I also don't know. Maybe with emoijs? ;-)

cspiel1:
I also change this to "Ready for review". @alfredh If you agree with @juha-h you may merge this.

juha-h:
I did some testing on `sip_verify_server	yes` and found that there does not seem to exist any ua event that would tell my baresip app that registration failed because server certificate verification  failed.

I got this to log:
```
02-28 09:43:17.988  8272  8405 D Baresip Lib: tls: verify_handler: err          = 20
02-28 09:43:17.988  8272  8405 D Baresip Lib: tls: verify_handler: error_string = unable to get local issuer certificate
02-28 09:43:17.988  8272  8405 D Baresip Lib: tls: verify_handler: depth        = 0
02-28 09:43:17.989  8272  8405 D Baresip Lib: tls: connect: error (r=-1, ssl_err=1)
02-28 09:43:17.993  8272  8404 D Baresip Lib: ua event (REGISTER_FAIL) Network is unreachable
```
The ua event is misleading, since network is not unreachable.

Any suggestions on what needs to be done to get a proper ua event?

PR #1242, opened 2021-03-03
Fix multiple resumed calls

cHuberCoffee:
Prevent barsip from having multiple "onhold == false" calls. Resuming a call will replace the active call to the resumed one.
The handling of UA_EVENT_CALL_REMOTE_SDP is added to ensure that the menu does get the resumed call as active call.
Otherwise cmd_hangup will trigger the UA_EVENT_CALL_CLOSED and the menu.callid is removed. All commands with menu_uacur() will not work anymore.


juha-h:
Since this request introduces a new API call `uag_hold_resume`, there should be a comment about what it does on top of the function in ua.c.

Issue #1243, opened 2021-03-04
Should ua.c really resume a call when any call terminates?


cspiel1:
Was thinking about this again, because we (@cHuberCoffee) shortly replaced the
```
resume_call(ua);
```
by the new
```
uag_hold_resume(NULL);
```
This ensures that not two calls are established at the same time.

But found this additional problems:
1. It is not correct in some situation to do so. Suppose:
- There are three calls. One is established (E), one is on-hold (H), the third is an outgoing call (O).
- Then (O) is rejected by the peer. So it terminates.
- As result (E) and (H) will be exchanged, which makes no sense.
2. But now assume (E) terminates. I think that it is not always wanted that (H) is resumed. The application should decide this. E.g. module menu, gtk, ...

So I suggest to
- add a check if the call that terminates was the established call,
- let menu decide if uag_hold_resume() should be invoked, maybe in the ua_event_handler(). Maybe also gtk to stay backward compatible for gtk users.
- Add also a config for menu if the resume is wanted if a call terminates.

I need your opinion.

juha-h:
Is it really necessary to have such complex scenarios, where during an established call another call is placed?

cspiel1:
Yes, you can have multiple incoming calls while having an established call. Think about a control desk application!
Also, placing an outgoing call while there is an established call is a normal scenario. And it is possible in baresip right now.

Anyway, the simple use cases where you have only one call should not be affected by the mentioned problem. Here I suggest to simplify the behavior to remove the resume in ua_hangup() and the call_event_handler(). This would solve problem 1. and 2. mentioned above.

cspiel1:
Ok, this is a different subject. Should be discussed also. But maybe in a different issue.

Tomorrow I will make a PR for the mentioned problems. Any other suggestions how to handle this?

juha-h:
So what happens when there is established call and ongoing call and the ongoing call is answered?

cspiel1:
Currently then you have two established calls. This is another issue for which I plan to make another PR. @cHuberCoffee and me think that bareship should ensure that there is only one established call. There are two possible solutions:
- ua.c ensures that there is only one call, or
- the application module (e.g. menu) ensures this. I think this is the right solution to be more flexible if anyone does'nt use menu.

cspiel1:
### To sum up
- Resume a call if the established call terminates is business logic and should be done in module menu. Currently it is done in ua.c and it is done in-correct in some situations.
[...]

PR #1247, opened 2021-03-05
Menu ensure only one established call

cspiel1:
Related to https://github.com/baresip/baresip/issues/1243 (second point).

- Adds a new function uag_hold_others(call).
- For outgoing calls that become established uag_hold_others() is called in the ua_event_handler().
- For incoming calls the command accept and acceptdir have to use uag_hold_others().

@alfredh: Should we split uag from ua.c --> uag.c? Maybe in a soon next PR?

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -832,6 +832,7 @@ int  ua_init(const char *software, bool udp, bool tcp, bool tls);
 void ua_close(void);
 void ua_stop_all(bool forced);
 int  uag_hold_resume(struct call *call);
+int  uag_hold_others(struct call *call);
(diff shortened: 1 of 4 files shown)


alfredh:
the idea with ua.c/uag.c split is the following:

1. ua.c contains functions for struct ua
2. uag.c contains functions for the "global" struct uag


I am currently not looking into this.

if you want to have a go at it, please feel free :)

the work can also be done in stages ...

cspiel1:
Yes, I understood this idea exactly the same. Maybe I start on Monday with it.

Next I have also commits for an optional call-id Parameter for menus accept, hangup, hold, resume. This maybe does not touch ua.c. We will see.

Thanks for review and merge!

PR #1251, opened 2021-03-07
added multicast enabled message

webstean:


modules/multicast/multicast.c
```diff
@@ -514,6 +514,9 @@ static int module_init(void)
 	err = cmd_register(baresip_commands(), cmdv, ARRAY_SIZE(cmdv));
 	err |= module_read_config();
 
+	if (!err)
+		info("multicast: enabled\n");
```
cspiel1, line 518: "enable" has a different meaning in this module. Please write:
```
if (!err)
    info("multicast: module init\n");
```

PR #1255, opened 2021-03-09
Ua dnd feature

cHuberCoffee:
Global DnD functionality which rejects all incoming calls with the errorcode 480 Temporarily Unavailable.


cHuberCoffee:
Nonetheless, it implements the functionality to directly reply to incoming calls with scode 480. The multicast module may use this feature later on. I close this PR and open another one for the multicast module where it is more clear why i decided to add it in that way.

PR #1260, opened 2021-03-10
menu: on call termination select another active call

cspiel1:


modules/menu/menu.c
```diff
@@ -626,12 +626,30 @@ struct call *menu_find_call(const char *id)
...
+	};
+
 	menu.callid = mem_deref(menu.callid);
 
 	if (call) {
 		str_dup(&menu.callid, call_id(call));
 		call_set_current(ua_calls(call_get_ua(call)), call);
 	}
+	else {
+		for (i = sizeof(state)-1; i > 0; --i) {
```
alfredh, line 645: size(state) assumes that each element size is 1.

it is probably better to use `ARRAY_SIZE`

PR #1264, opened 2021-03-11
menu: correct hangupall command for parallel call feature

cspiel1:
A wrong filter function was used to hangup all outgoing calls.

When multiple outgoing calls are started to different peers, the first who
answers the call wins. Then we want to hangup all other outgoing calls. The
application can do this by invoking:
```
hangupall dir=out
```
or simpler
```
hangupall out
```
The flag call->outgoing also includes the established call. Correct is to use
call_state() for the filter.

### DIFF
--- a/modules/menu/static_menu.c
+++ b/modules/menu/static_menu.c
@@ -660,7 +660,7 @@ static int cmd_hangup(struct re_printf *pf, void *arg)
 }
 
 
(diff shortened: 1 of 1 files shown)


alfredh:
many thanks

btw, do you have someone in your team that wants to maintain the "menu" module ?

Issue #1261, opened 2021-03-10
alsa: could not open auplay device 'default' (Device or resource busy)


dannymarosi:
Running BARESIP v1.0.0 on raspberry pi zero with buster distribution.
No audio when uncommenting “**ringback_disabled	no**” in config file.
![2021_03_10 RingBack issue](https://user-images.githubusercontent.com/62627034/110678053-b958b300-818a-11eb-9972-8ccffc51e911.png)
Thanks for your help.

cspiel1:
I also use ALSA on pi zero. Will try to reproduce this.

But by checking our baresip branch I found that we have already patches that may solve this. With "normal" ALSA configuration the problem is, that the a device can only be opened by one instance. The ringback tone has to be stopped before the audio is opened for the established call.

When looking in master HEAD:
In call.c:1560 the audio is started for the call and then in line 1578 the CALL_EVENT_ESTABLISHED is passed to ua.c which forwards a UA-event to the menu. Then menu stops the ringback. In our branch we
- first stop the ringback, then
- start the audio for the call.

Another solution could be to configure ALSA differently. But I am not an expert. Maybe dmix and dsnoop plugins for ALSA help.

alfredh:
there are also 4 180 ringing

is the audio player started 4 times?

cspiel1:
With default configuration auplay_alloc() is called only once in play_tone().

The problem is that mem_deref(menu.play) is done too late. I will check our commits and create a PR if I am right.

cspiel1:
@alfredh Maybe you saw my try to solve this in the mentioned PR. Should we live with this ALSA restriction?
Maybe add in module alsa doxygen a note?

On the other side, the current behavior also is not very good:
If the event handler terminates the call, we this scenario:
- Audio devices are opened (for the call).
- Call is terminated.
- Audio devices are closed.

alfredh:
another option is to add a new event that is called before ESTABLISHED.

the menu module can then use this event to stop any ongoing tones

cspiel1:
Right! And ESTABLISHED then could be sent by a timer with zero delay.

UA_EVENT_CALL_PROGRESS also stops the ringback tone. Outgoing call states are already more complicated than incoming.

OUTGOING --> RINGING or PROGRESS --> ESTABLISHED

PROGRESS (Early media) normally also starts audio (only the mic?) before the event is passed to upper layers. I am not sure but think this opens the Mic and should not concur with the ringback which opens playback device.


[...]

PR #1266, opened 2021-03-11
call: send CALL_EVENT_ESTABLISHED before call_stream_start

cspiel1:
Related to: https://github.com/baresip/baresip/issues/1261

This prevents that ALSA playback device is opened for the call before it is
closed for the ringback tone. This is only possible if dmix is enabled in the
ALSA configuration.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -1557,6 +1557,14 @@ static void sipsess_estab_handler(const struct sip_msg *msg, void *arg)
 
 	set_state(call, CALL_STATE_ESTABLISHED);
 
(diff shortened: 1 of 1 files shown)


cspiel1:
Just realized this comment:
```
/* must be done last, the handler might deref this call */
```
This seems to be not a good solution.

cspiel1:
I close this, because the call to the event handler should stay in last line. The alternative would be an ugly if statement to check if the call pointer is still tracked by any User-Agent.

PR #1267, opened 2021-03-11
call: start streams asynchronously

cspiel1:
Related to: https://github.com/baresip/baresip/issues/1261

This fixes a problem with ALSA. The ALSA playback device can not be opened
because the ringback tone is not stopped before.


alfredh:
great, thanks!

PR #1268, opened 2021-03-11
opengl: remove deprecated module

alfredh:
proposal to remove opengl module using deprecated OSX api:

https://apple.stackexchange.com/questions/379326/is-opengl-deprecated-or-removed-in-macos-catalina-10-15


cspiel1:
It's fine to remove this.

Maybe an opengl module for linux would make sense, but don't care.

PR #1272, opened 2021-03-12
Resume call on termination

cspiel1:
This PR simplifies the check if a call is established and not put on hold locally.

E.g. When a call terminates that was established, resume another call that was on-hold locally.

Fixes also the call resume if the call termination happens during local stream error in stream_error_handler().


juha-h:
Does this mean that a call that is established but is on hold is not
established?
```
bool call_is_established(struct call *call)
{
	return  call ?
		call_state(call) == CALL_STATE_ESTABLISHED &&
		!call_is_onhold(call) : false;
}
```
If so, I consider it confusing.  According to RFC3261:

Dialog: A dialog is a peer-to-peer SIP relationship between two
  UAs that persists for some time.  A dialog is established by
  SIP messages, such as a 2xx response to an INVITE request.

Putting a call on hold does not de-establish it.

juha-h:
Why not just have two separate API calls `call_is_established()`and `call_is_onhold()`?

juha-h:
We have these:
```
/** Call States */
enum call_state {
	CALL_STATE_IDLE = 0,
	CALL_STATE_INCOMING,
	CALL_STATE_OUTGOING,
	CALL_STATE_RINGING,
	CALL_STATE_EARLY,
	CALL_STATE_ESTABLISHED,
	CALL_STATE_TERMINATED,
	CALL_STATE_UNKNOWN
};
```
and there already exists API functions
```
bool          call_is_onhold(const struct call *call);
bool          call_is_outgoing(const struct call *call);
```
So it remains to add `call_is_establised()` or `call_state()` that would return the enum value.

cspiel1:
call_is_onhold(), call_is_outgoing() and the new call_is_established() has nothing to do with the call->state. Thus, we are completely free to implement a function that is called "call_is_established". 

More interesting is to compare how call_is_established() and call_was_established() are implemented.

juha-h:
Christian Spielberger writes:

> call_is_onhold(), call_is_outgoing() and the new call_is_established()
> has nothing to do with the call->state. Thus, we are completely free
> to implement a function that is called "call_is_established".

If call_is_established() has nothing to with call state, then it really is
confusing.  But I don't think what you write is true, since its
implementation refers to call state:
```
bool call_is_established(struct call *call)
{
	return  call ?
		call_state(call) == CALL_STATE_ESTABLISHED &&
		!call_is_onhold(call) : false;
}
```
API functions should be general purpose.  Why can't you just remove the
&& part from the test?

src/call.c
```diff
@@ -2504,3 +2509,17 @@ void call_start_answtmr(struct call *call, uint32_t ms)
...
 	tmr_start(&call->tmr_answ, ms, delayed_answer_handler, call);
 }
+
+
+bool call_is_established(struct call *call)
+{
+	return  call ?
+		call_state(call) == CALL_STATE_ESTABLISHED &&
+		!call_is_onhold(call) : false;
+}
```
alfredh, line 2519: perhaps a better name is `call_is_active` ?

also the input param should be `const`

src/call.c
```diff
@@ -2504,3 +2509,17 @@ void call_start_answtmr(struct call *call, uint32_t ms)
...
+
+
+bool call_was_established(struct call *call)
```
alfredh, line 2522: input param should be `const`

src/call.c
```diff
@@ -94,6 +95,10 @@ static const char *state_name(enum call_state st)
 
 static void set_state(struct call *call, enum call_state st)
 {
+	call->was_established =
+		st == CALL_STATE_TERMINATED &&
+		call_is_established(call);
```
alfredh, line 100: another option is to set this flag in the established_handler

src/call.c
juha-h: Also a call that is in early state can be onhold and there may be need to be able to test it.  If we introduce a new API call for each combination, we'll end up with 6 new API calls, which does not make sense to me.  It would be cleaner to just have an API call that tells the state and another that tells if the call is (or was) onhold.

src/call.c
cspiel1: @juha-h I understand your arguments. But the combination CALL_STATE_ESTABLISHED and not on hold locally is used very often. Thus it reduces the code.

In order to avoid the confusion with "ESTABLISHED" the suggestion of @alfredh is good. I will rename the functions to:
```
bool call_is_active(const struct call *call);
bool call_was_active(const struct call *call);
```

juha-h:
Christian Spielberger writes:

> @juha-h I understand your arguments. But the combination
> CALL_STATE_ESTABLISHED and not on hold locally is used very
> often. Thus it reduces the code.

There should be some kind of policy regarding introduction of new API
functions to baresip in order to prevent bloating the API.

My proposal is that a new API function needs to be introduced if the
task in hand cannot be done using current API functions or it would be
too complex to do so.

In this particular case neither of the two is true.  It would be much
more intuitive and self documenting to write in the code

  if (call_state(call) == CALL_STATE_ESTABLISHED && !call_is_onhold(call))

than

  if (call_is_active(call))

The former is not too complex to write and everyone immediately
understands what the means. Whereas in the latter case, the reader needs
to check the sources.

If some app needs to make the above check many times in different parts
of the code, nothing prevents the app write from creating its own
utility function for it.

> In order to avoid the confusion with "ESTABLISHED" the suggestion of @alfredh is good. I will rename the functions to:
> ```
> bool call_is_active(const struct call *call);
> bool call_was_active(const struct call *call);
> ```

What would the API function be called that tests if call in early state
is onhold?

juha-h:
Regarding possible use cases of `call_was_active`, isn't it usually already known at that point that the call is not established anymore, i.e., has been terminated?

cspiel1:
Use case: On UA_EVENT_CALL_CLOSED resume another call.

```
		if (call_is_active(call) || call_was_active(call))
			uag_hold_resume(NULL);
```

It was already there before my changes, but in UA and not implemented correctly in all situations. We could also drop it completely, then the application has to do a resume. With application I means something that communicates e.g. with menu via ctrl_tcp or ctrl_dbus. This what we would prefer. But I tried to keep the behavior backward compatible.

If we let the uag_hold_resume(NULL) like this PR does, than I suggest to add a configure flag that turns off the resume on call termination.

cspiel1:
> What would the API function be called that tests if call in early state
is onhold?

We don't need such functions.

juha-h:
Christian Spielberger writes:

> Use case: On UA_EVENT_CALL_CLOSED resume another call.
> 
> ```
> 		if (call_is_active(call) || call_was_active(call))
> 			uag_hold_resume(NULL);
> ```

That too could be written w/o the new API functions.  B I'll give up if
the policy that each time someone needs a convenience function, it can
be added to baresip.h.

cspiel1:
Call was active can not be written w o.

juha-h ***@***.***> schrieb am So., 14. März 2021, 10:22:

juha-h:
Christian Spielberger writes:

> Call was active can not be written w o.

Then the logical choice is to create a new API function
`call_was_onhold`.

alfredh:
in general the public baresip api should be as slim and straight-forward as possible.

it should be easy to use, the less functions the better

sometimes it is nice to add a convenience function, in order to reduce code duplication

my approach to this is currently:

1. when working on code, I add a local convenience function
2. if that function has 1 user, keep it local
3. later, if it turns out that the function is very useful for many parts of the code,
consider moving it to a common place

juha-h:
Alfred E. Heggestad writes:

> my approach to this is currently:
> 
> 1. when working on code, I add a local convenience function
> 2. if that function has 1 user, keep it local
> 3. later, if it turns out that the function is very useful for many parts of the code,
> consider moving it to a common place

Sounds good to me.

juha-h:
Would call_previous_state() be a good generalization of this?  For example, if current state is CALL_STATE_TERMINATED, call_previous_state() would tell how we got there.  For example, if previous state was CALL_STATE_ESTABLISHED, would call_is_onhold() then tell if the call was previously "active"?

juha-h:
I did some more reading of this pull request.  Is it so that if "active" call was closed then at UA_EVENT_CALL_CLOSED  _some_ call that is onhold is resumed automatically:
```
if (call_is_active(call) || call_was_active(call))
			uag_hold_resume(NULL);
```
If so, I don't think that it is a good idea.  There may be more than one call onhold and the user might want to select., which call to resume, or not to resume any call and go to have coffee instead.

cspiel1:
Yes, I also thought on the previous state solution. But it makes the code where it is used more complicated.



> in general the public baresip api should be as slim and straight-forward as possible.
> 
> it should be easy to use, the less functions the better
> 
> sometimes it is nice to add a convenience function, in order to reduce code duplication
> 
> my approach to this is currently:
> 
> 1. when working on code, I add a local convenience function
> 2. if that function has 1 user, keep it local
> 3. later, if it turns out that the function is very useful for many parts of the code,
>    consider moving it to a common place

Sounds also good for me. E.g. call_is_outgoing(c): Could be realized also with
```
call_state(c) == CALL_STATE_OUTGOING ||
call_state(c) == CALL_STATE_RINGING ||
call_state(c) == CALL_STATE_EARLY ||
call_prev_state(c) == CALL_STATE_OUTGOING ||
call_prev_state(c) == CALL_STATE_RINGING ||
call_prev_state(c) == CALL_STATE_EARLY
```
But call_is_outgoing(c) is much shorter. And is used at two different locations: event.c and menu.c

Could remove the call_was_active() because it included also call_is_active(). I renamed "call_was_active" to "call_is_active" and dropped the other function. So now we have a function call_is_active() that is the same as:
```
(call_state(c) == CALL_STATE_ESTABLISHED && call_is_onhold()) ||
(call_prev_state(c) == CALL_STATE_ESTABLISHED && call_was_onhold())
```

Just mentioned that this solution would need another new function call_was_onhold(). The call_is_active() can replace the also longer
```
call_state(c) == CALL_STATE_ESTABLISHED && call_is_onhold()
```
in ua.c at two locations. In menu.c we need the call_is_active() like it is now.

cspiel1:
I think the changes are now minimal and reduces code complexity. The most benefit is that it fixes the "resume call on call termination" feature.

But another option for me would be to completely remove the "resume call on call termination" feature. In my opinion this is business logic and should not be done by baresip automatically.

juha-h:
Christian Spielberger writes:

> Could remove the call_was_active() because it included also call_is_active(). I renamed "call_was_active" to "call_is_active" and dropped the other function. So now we have a function call_is_active() that is the same as:
> ```
> (call_state(c) == CALL_STATE_ESTABLISHED && call_is_onhold()) ||
> (call_prev_state(c) == CALL_STATE_ESTABLISHED && call_was_onhold())
> ```

When call is terminated, does the value of its on_hold attribute change
or would call_is_onhold() still tell what the onhold status of the call
was?  That is, do we really need call_was_onhold()?

juha-h:
Christian Spielberger writes:

> But another option for me would be to completely remove the "resume
> call on call termination" feature. In my opinion this is business
> logic and should not be done by baresip automatically.

Exactly!  That was my point in my "go to have coffee" message.

cspiel1:
:-) I said it before. But I tried to keep backward compatible. But does'nt matter. Made a second PR with the option. Please decide which one we take!

https://github.com/baresip/baresip/pull/1288

juha-h:
Should we also prepare in the API for the conference call case, i.e., when there is more that one "active" call?  If it is not done now, there may be trouble later.

juha-h:
juha-h writes:

> Should we also prepare in the API for the conference call case, i.e.,
> when there is more that one "active" call?  If it is not done now,
> there may be trouble later.

And it is not just the conference call case, but also apps like echo and
announcement service.

cspiel1:
Was replaced by #1288.

juha-h:
I don't know is there some policy on passing functions as arguments to API functions, but such arguments could make API simpler.  For example, uag_find_call_state() could be replaced by uag_find_calls(struct list * (*f)(struct call *))  that takes the filter function as argument.

PR #1273, opened 2021-03-12
Menu fixes for play tones

cspiel1:
- ringback tone was not played
- fixes when and what to play for incoming calls
- resume right tone if a call terminates
- fix for alert_stop (terminal beep)

Has to be rebased after merge of https://github.com/baresip/baresip/pull/1272.

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -595,27 +595,6 @@ struct menu *menu_get(void)
 }
 
 
-struct call *menu_find_call(const char *id)
-{
-	struct le *le;
-	struct call *call;
-
-	if (!str_isset(id))
-		return NULL;
-
-	for (le = list_head(uag_list()); le; le = le->next) {
-		struct ua *ua = le->data;
-		struct list *calls = ua_calls(ua);
-
-		call = call_find_id(calls, id);
-		if (call)
-			return call;
-	}
-
-	return NULL;
-}
-
-
 /**
  * Selects the active call.
  *
@@ -657,7 +636,7 @@ void menu_selcall(struct call *call)
  */
 struct call *menu_callcur(void)
 {
-	return menu_find_call(menu.callid);
+	return uag_call_find(menu.callid);
 }
 
 
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -47,31 +47,6 @@ static int menu_set_incall(bool incall)
 }
 
 
-static void alert_start(void *arg)
-{
-	(void)arg;
-
-	if (!menu.bell)
-		return;
-
-	ui_output(baresip_uis(), "\033[10;1000]\033[11;1000]\a");
-
-	tmr_start(&menu.tmr_alert, 1000, alert_start, NULL);
-}
-
-
-static void alert_stop(void)
-{
-	if (!menu.bell)
-		return;
-
-	if (tmr_isrunning(&menu.tmr_alert))
-		ui_output(baresip_uis(), "\r");
-
-	tmr_cancel(&menu.tmr_alert);
-}
-
-
 static void tmrstat_handler(void *arg)
 {
 	struct call *call;
@@ -216,9 +191,6 @@ static void play_incoming(const struct ua *ua, bool waiting)
 			/* Alert user */
 			menu_play("ring_aufile", "ring.wav", -1);
 		}
-
-		if (menu.bell)
-			alert_start(0);
 	}
 }
 
@@ -466,8 +438,6 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 		/* stop any ringtones */
 		menu_stop_play();
 
-		alert_stop();
-
 		/* We must stop the re-dialing if the call was
 		   established */
 		redial_reset();
@@ -477,7 +447,6 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 	case UA_EVENT_CALL_CLOSED:
 		menu_play_closed(call);
 
-		alert_stop();
 		play_resume(call);
 
 		/* Activate the re-dialing if:
@@ -752,7 +721,6 @@ static int module_init(void)
 	int err;
 
 	memset(&menu, 0, sizeof(menu));
-	menu.bell = true;
 	menu.redial_attempts = 0;
 	menu.redial_delay = 5;
 	menu.ringback_disabled = false;
@@ -764,7 +732,6 @@ static int module_init(void)
 	/*
 	 * Read the config values
 	 */
-	conf_get_bool(conf_cur(), "menu_bell", &menu.bell);
 	conf_get_bool(conf_cur(), "ringback_disabled",
 		      &menu.ringback_disabled);
 	conf_get_bool(conf_cur(), "menu_clean_number", &menu.clean_number);
@@ -791,7 +758,6 @@ static int module_init(void)
 		return ENOMEM;
 
 	menu.start_ticks = tmr_jiffies();
-	tmr_init(&menu.tmr_alert);
 
 	if (0 == conf_get(conf_cur(), "statmode_default", &val) &&
 	    0 == pl_strcasecmp(&val, "off")) {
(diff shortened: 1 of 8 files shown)


modules/menu/menu.c
```diff
@@ -134,6 +134,14 @@ static char *errorcode_key_aufile(uint16_t scode)
 }
 
 
+static void menu_stop_play(void)
+{
+	menu.play = mem_deref(menu.play);
+	menu.ringback = false;
+	alert_stop();
```
alfredh, line 141: unrelated comment; should we consider to remove the "alert" code ?

cspiel1: Yes, I think so. Another PR or may add here a commit?

cspiel1:
The alert beep:
- It is somehow related to this PR and already touched here.
- E.g. Does not work on default Ubuntu systems because beep seems to be deactivated.
- Alert is obsolete, since a WAV file can be played everywhere with baresip. So ringtones should work anyway.

Will add a commit to remove the alert code. Let me know if you have counter-arguments!

cspiel1:
Appending a call to a custom list is not possible because calls are already tracked in the list of its UA. I tried a solution with this new API functions:
```
void call_append(struct list *calls, struct call *call);
int ua_filter_calls(struct ua *ua, struct list *calls, enum call_state st);
int uag_filter_calls(struct list *calls, enum call_state st);
```
But call_append() is not possible. It results in this error:
```
list: append: le linked to 0x55b50f9b5c80
```

I try to find a different solution for this ringtone PR. But you are right, that a call filter would be helpful in general.

juha-h:
Christian Spielberger writes:

> Appending a call to a custom list is not possible because calls are
> already tracked in the list of its UA.

So a call object can only be in one list (ua's struct list calls)?  If
so, can it be relaxed somehow or does re make it impossible?

cspiel1:
Yes, exactly. No it can't be relaxed. A list-element (struct le) has one pointer prev and one next.

juha-h:
Christian Spielberger writes:

> Yes, exactly. No it can't be relaxed. A list-element (struct le) has
> one pointer prev and one next.

Then how about making a copy of the list elements?

juha-h:
Christian Spielberger writes:

> Yes, exactly. No it can't be relaxed. A list-element (struct le) has
> one pointer prev and one next.

List element has `void *data`.  How about creating a new list with new
elements and take the data to element from the ua calls list?

cspiel1:
> Then how about making a copy of the list elements?
The call->le is hidden. A copy of a call, is too much.

But look, here I have two implemenations of menu_find_active_call(). The first one uses the call filter (which currently does not work). The second makes everything of its own.

```
static struct call *menu_find_active_call(void)
{
	struct list calls;
	struct le *le;
	int err;
	struct call *call = NULL;

	list_init(&calls);
	err = uag_filter_calls(&calls, CALL_STATE_ESTABLISHED);
	if (err)
		return NULL;

	for (le = calls.head; le; le = le->next) {
		call = le->data;

		if (!call_is_onhold(call))
			goto out;
	}

out:
	list_clear(&calls);
	return call;
}
```
Here without the call filter uag_filter_calls():
```
static struct call *menu_find_active_call(void)
{
	struct le *leu;
	struct le *lec;
	struct call *call = NULL;
	int err;

	for (leu = list_head(uag_list()); leu; leu = leu->next) {
		struct ua *ua = leu->data;
		for (lec = list_head(ua_calls(ua)); lec; lec = lec->next) {
			call = lec->data;

			if (call_state(call) == CALL_STATE_ESTABLISHED &&
					!call_is_onhold(call))
				goto out;
		}
	}

out:
	return call;
}
```

The second is shorter. And currently we have no other use-case for the call filter.

cspiel1:
> List element has `void *data`. How about creating a new list with new elements and take the data to element from the ua calls list?

We could add a function that calls a callback function for each call:

```
typedef void (call_process_h)(struct call *call, void *arg);
int uag_process_calls(call_process_h, void *arg);
int uag_filter_calls(call_process_h, enum call_state st, void *arg);
```

But also here the benefit for this PR would be minimal. And what use cases do we have else?

cspiel1:
Please look at the newest commit! If you prefer a solution with
```
int uag_filter_calls(call_process_h processh, enum call_state st, void *arg);
```
let me know!

juha-h:
The more general one would be better if it is not possible to return the call list.

cspiel1:
The two latest commits:
- The first adds the filter function uag_filter_calls() requested by @juha-h.
- The second adds a generic find function uag_find_call() similar to the filter which simplified menu by 13 lines.

Should we keep both functions?

cspiel1:
Should we replace the uag_find_call_state() we added shortly also by the generic uag_find_call()? Here are the uses:
```
$ cppgrep uag_find_call_state
./modules/menu/static_menu.c:106:		call = uag_find_call_state(CALL_STATE_INCOMING);
./modules/menu/static_menu.c:179:		call = uag_find_call_state(CALL_STATE_INCOMING);
./modules/menu/menu.c:577:			call = uag_find_call_state(state[i]);
```

juha-h:
Christian Spielberger writes:
> The two latest commits:

> - The first adds the filter function uag_filter_calls() requested by @juha-h.
> - The second adds a generic find function uag_find_call() similar to the filter which simplified menu by 13 lines.
> 
> Should we keep both functions?

The latter does not look like a good API call to me, since if there is
more than one call that matches the filter then it is not well defined,
which one is returned.

cspiel1:
Same for uag_find_call_state(). So I will replace both by use of uag_filter_calls(). But please let me traverse the calls of one User-Agent in reverse order, like I changed in last commit!

juha-h:
Christian Spielberger writes:

> Same for uag_find_call_state(). So I will replace both by use of
> uag_filter_calls(). But please let me traverse the calls of one
> User-Agent in reverse order, like I changed in last commit!

I noticed the commit "ua: reverse order of the calls", which changed the
order in which calls are processed/looked for.

The order should not matter in case all matching calls are processed,
but if only one is returned like in `uag_find_call` then the right order
may depend on the application that is using the function.  And as I
mentioned earlier, I don't consider such an API call a good idea.

I'm still not convinced, why a new call list cannot be returned.  Call
objects themselves would not need to be copied, just pointers to them.

One solution to avoid this problem is just to have ua_calls() in the
API.  Apps could then call it on all relevant UAs and process the calls
of each.

cspiel1:
In re/baresip a list of a struct X always is implemented by adding a list-element field to the struct.
```
struct X {
    struct le le;
    ...
};


```
Then an append looks like this:
```
struct X *x;
x = mem_zalloc(sizeof(x), NULL);
list_append(&list, &x->le, x);
```

A le can only be appended to one list. If you want a second independent list, you need another le in the struct.

juha-h:
Christian Spielberger writes:

> In re/baresip a list of a struct X always is implemented by adding a
> list-element field to the struct.

OK, got it.

Is call_id of call unique?  If yes, then the list could be list of
call_ids, since we already have these:
```
const char   *call_id(const struct call *call);
struct call  *call_find_id(const struct list *calls, const char *id);
```

cspiel1:
Then we would need a new struct.
```
struct Idelem {
    struct le le;
    char callid;
}
```
And code in call.c or ua.c to build such lists. Belief me, the current variant with the callback function is much cleaner.

Have to add another commit that makes use of uag_filter_calls() again in menu.c. Ist not used currently. That was a mistake.

juha-h:
Christian Spielberger writes:

> Then we would need a new struct.
> ```
> struct Idelem {
>     struct le le;
>     char callid;
> }
> ```

> And code in call.c or ua.c to build such lists. Belief me, the current
> variant with the callback function is much cleaner.

Callback based function is OK when something is done on all matching
elements.  But if more than one element matches and something is done on
only one arbitrary one, it is not a clean, generic solution.

> Have to add another commit that makes use of uag_filter_calls() again
> in menu.c. Ist not used currently. That was a mistake.

I'm already a bit lost with all the commits.  It would be easier to
digest if there would a separate pull request for API changes that would
not include any module related changes.

cspiel1:
> I'm already a bit lost with all the commits.  It would be easier to
digest if there would a separate pull request for API changes that would
not include any module related changes.

You're right. I very cleanly did the first 8 commits. The rest where a lot of back-forward steps. I will do a rebase to make it easier to review this PR commit-by-commit.

cspiel1:
This is the API change. https://github.com/baresip/baresip/pull/1293

cspiel1:
This PR is replaced by: https://github.com/baresip/baresip/pull/1294

PR #1274, opened 2021-03-12
ausine: mono support and stereo_left/right option

sreimers:


alfredh:
great stuff!

I am also wondering if we should use this module (ausine) in selftest,
instead of the fake/mock ausrc.

sreimers:
> I am also wondering if we should use this module (ausine) in selftest,
> instead of the fake/mock ausrc.

good idea , i will have a look at it.

PR #1277, opened 2021-03-12
Optional avformat decoder

viric:
In my prev PR I missed this and made this parameter mandatory.

I think it should not be; it should figure out the decoder automatically, if not set.

### DIFF
--- a/modules/avformat/avformat.c
+++ b/modules/avformat/avformat.c
@@ -303,7 +303,7 @@ int avformat_shared_alloc(struct shared **shp, const char *dev,
 		}
 	}
 
-	if (strnlen(avformat_inputformat, sizeof avformat_inputformat)	> 0) {
+	if (str_isset(avformat_inputformat)) {
 		ret = av_dict_set(&format_opts, "input_format",
                                 avformat_inputformat, 0);
 		if (ret != 0) {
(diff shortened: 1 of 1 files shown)


modules/avformat/avformat.c
```diff
@@ -433,10 +433,12 @@ static int module_init(void)
...
-	avformat_decoder = avcodec_find_decoder_by_name(decoder);
-	if (!avformat_decoder) {
-		warning("avformat: decoder not found (%s)\n", decoder);
-		return ENOENT;
+	if (strnlen(decoder, sizeof(decoder)) > 0) {
+		avformat_decoder = avcodec_find_decoder_by_name(decoder);
+		if (!avformat_decoder) {
+			warning("avformat: decoder not found (%s)\n", decoder);
+			return ENOENT;
+		}
```
alfredh, line 441: great, could you please use `str_isset` instead, it is more efficient

modules/gst/gst.c
```diff
@@ -13,6 +13,7 @@
 #include <re.h>
 #include <rem.h>
 #include <baresip.h>
+#include <unistd.h>
```
alfredh, line 16: this is an unrelated change, please make a separate PR for it.
stating the problem would also be good, i.e. why the change is needed

modules/avformat/avformat.c
```diff
@@ -303,7 +303,7 @@ int avformat_shared_alloc(struct shared **shp, const char *dev,
 		}
 	}
 
-	if (strnlen(avformat_inputformat, sizeof avformat_inputformat)	> 0) {
+	if (str_isset(avformat_inputformat)	> 0) {
```
alfredh, line 306: `str_isset` returns bool. the code should be:

```c
if (str_isset(avformat_inputformat))
```

viric: ooops

PR #1279, opened 2021-03-13
Fix gst.c missing header

viric:


modules/gst/gst.c
```diff
@@ -13,6 +13,7 @@
 #include <re.h>
 #include <rem.h>
 #include <baresip.h>
+#include <unistd.h>
```
cspiel1, line 16: I think normally this is put before include re.h.

cspiel1:
@viric Did it for you. Thanks for the patch!

This PR is replaced by https://github.com/baresip/baresip/pull/1297.

PR #1280, opened 2021-03-13
conf: add call_hold_other_calls config option

sreimers:
With the new on hold changes in `menu`, use cases like handling multiple separate calls are not possible anymore (like providing multiple test instances with `echo` or `ausine`). I think putting calls on hold should be optional.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -300,6 +300,7 @@ struct config_sip {
 struct config_call {
 	uint32_t local_timeout; /**< Incoming call timeout [sec] 0=off    */
 	uint32_t max_calls;     /**< Maximum number of calls, 0=unlimited */
+	bool hold_other_calls;  /**< Hold other calls */
 };
 
 /** Audio */
--- a/src/config.c
+++ b/src/config.c
@@ -37,7 +37,8 @@ static struct config core_config = {
 	/** Call config */
 	{
 		120,
-		4
+		4,
+		false
 	},
 
 	/** Audio */
@@ -301,6 +302,8 @@ int config_parse_conf(struct config *cfg, const struct conf *conf)
 			   &cfg->call.local_timeout);
 	(void)conf_get_u32(conf, "call_max_calls",
 			   &cfg->call.max_calls);
+	(void)conf_get_bool(conf, "call_hold_other_calls",
+			   &cfg->call.hold_other_calls);
 
 	/* Audio */
 	(void)conf_get_str(conf, "audio_path", cfg->audio.audio_path,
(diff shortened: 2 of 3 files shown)


cspiel1:
Good idea!

Before the changes call_hold_other changes, holding other calls was only done for the current UA. Which I though could lead to unwanted behavior. This config makes sense. But with "false" as default it changes the default behavior. For me this is ok. @sreimers, @alfredh, @juha-h  What do you think?

juha-h:
I'm not using `uag_hold_others()` API function, but in general it would be good to preserve backwards compatibility.  So in this case the default value of the config var should be 'yes'.

cspiel1:
Sorry, I was confused somehow. I would suggest a config flag for call resume on call termination (the call to uag_hold_resume()), which is much more related to business logic.

If you set the call_hold_other_calls=no you allow to have multiple calls on different UAs. What is your intention? A conference feature?

sreimers:
> If you set the call_hold_other_calls=no you allow to have multiple calls on different UAs. What is your intention? A conference feature?

Right, as mentioned I have many use cases, ausine module you have multiple test accounts like `400hz@sip` or with echo like `echo@`. Or I use icecast streaming services with a [custom audio](https://github.com/Studio-Link/app/blob/v21.xx.x/src/modules/auicecast/auicecast.c) module. I don`t how many others rely on multiple separate calls and menu. Its fine for me if yes is the default. But maybe this new menu behavior breaks other production setups too.

cspiel1:
Ok. A normal setup is to register to multiple SIP proxies. Then one would have multiple UAs,  e.g. A and B.

Suppose you have:
- an established call at A and
- there is an incoming call at B.
- Then you answer the call at B.

With call_hold_other_calls=no you would have two established calls. But it is not a  conference. It would be something mystical. The two peers could not talk to each other, but we could talk to both.

Somebody already mentioned the interest for a conference feature. We should treat it separately.

juha-h:
Christian Spielberger writes:

> Ok. A normal setup is to register to multiple SIP proxies. Then one
> would have multiple UAs,  e.g. A and B. 
> 
> Suppose you have:
> - an established call at A and
> - there is an incoming call at B.
> - Then you answer the call at B.
> 
> With call_hold_other_calls=no you would have two established
> calls. But it is not a  conference. It would be something
> mystical. The two peers could not talk to each other, but we could
> talk to both.

Should the user be given a choice by the user interface on what to do
when the second call comes in, e.g. 'answer' or 'hold and answer' or
perhaps in the future 'conference answer'?

cspiel1:
I think, that menu should ensure that there is only one established call that is not on-hold. In the future there could also be a conference, to which all established calls (that are not on-hold) belong.

Now we allow the exception for this with the call_hold_other_calls=no setting. I think this is a good solution.

capncybo:
Thank you ever-so much for incorporating the "no" option... without it baresip_v1.1.0 had broken my SIP-based Security Camera System.  In my use case... @ select locations I used a raspberry pi 4 to display multiple Pi cameras (at the same time) which have dialed & connected to a conference-call on an asterisk server (The Asterisk server is always running via Entware/Asus Router).
Good Catch... I didn't even have to write-in & complain

PR #1283, opened 2021-03-14
alsa_play.c add suggestion to use dmix

webstean:


modules/alsa/alsa_play.c
```diff
@@ -136,7 +136,8 @@ int alsa_play_alloc(struct auplay_st **stp, const struct auplay *ap,
 	if (err < 0) {
 		warning("alsa: could not open auplay device '%s' (%s)\n",
 			st->device, snd_strerror(err));
-		goto out;
+		info("consider using dmix as your default alsa device");
```
cspiel1, line 139: Good idea to add an info if the error occurs. Anyway we should solve this.

Please add a '\n' at end of line!

modules/alsa/alsa_play.c
```diff
@@ -136,7 +136,8 @@ int alsa_play_alloc(struct auplay_st **stp, const struct auplay *ap,
...
+		out;
```
cspiel1, line 140: "out;" ?

Should stay "goto out;".

modules/alsa/alsa_play.c
webstean: oops - sorry, fixed.

alfredh:
@webstean many thanks for your improvements :)

PR #1288, opened 2021-03-15
menu: remove call resume on termination

cspiel1:
Resume another call on termination of a call  does not work correctly if the
call was hangup locally. Anyway, this feature is business logic and should not
be done by baresip automatically.

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -496,9 +496,6 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 		if (!str_cmp(call_id(call), menu.callid))
 			menu_selcall(NULL);
 
-		if (call_state(call) == CALL_STATE_ESTABLISHED &&
-				!call_is_onhold(call))
-			uag_hold_resume(NULL);
 		break;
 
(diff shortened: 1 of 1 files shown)


juha-h:
I think this can be safely merged when checks pass.

cspiel1:
Thanks for merge!

The drawback is that there may be users that are used to this feature. They would have to invoke command "resume" manually now.

But this command now has the benefit that the user may also specify a call-id if there are multiple calls that are on hold.

juha-h:
Christian Spielberger writes:

> The drawback is that there may be users that are used to this
> feature. They would have to invoke command "resume" manually now.

That is right, but I consider this a bug fix, because it forced a
certain behavior to everyone.

> But this command now has the benefit that the user may also specify a
> call-id if there are multiple calls that are on hold.

Yes, I would imagine that UI has some way to list all calls on hold and
then allow the user to choose which one to unhold.  Or some other app,
would always unhold the one that has waited longest, etc.

cspiel1:
Ah, that's why you want a filter function for calls. Makes sense.

PR #1291, opened 2021-03-15
Multicast call priority

cHuberCoffee:
Introduce a new config parameter "multicast_call_prio" to allow the multicast module having receivers with higher priority than a call. To prevent barsip from incoming and outgoing calls two new flags are added in the uag. On the incoming side, the DnD  reject all requests with the error 480 "Temporarily Unavailable". On the other side, the noDial flag prevents users from initiating calls.
Removing "multicast_call_prio" from the config or assigning it to 0 will ensure that a call will always be prioritized.

In cases where a multicast does have a higher priority than a call, the module takes advantage of the DnD and noDial flags. If all streams with high priority are done, the module switches back the flags. For established calls, these calls are set on_hold until the multicast is finished and resume it.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -846,6 +846,8 @@ int  uag_hold_others(struct call *call);
 void uag_set_nodial(bool nodial);
 bool uag_nodial(void);
 void uag_set_exit_handler(ua_exit_h *exith, void *arg);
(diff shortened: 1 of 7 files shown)


juha-h:
The new API functions included in this PR look good to me.  The rest I don't have  competence to comment.

PR #1295, opened 2021-03-16
ua: add possibility of openssl capath

cHuberCoffee:
adds an config option to add capath
requires [re/#84](https://github.com/baresip/re/pull/84)


juha-h:
If ca_path is not defined, is it OK that ca_file also includes the path as it is now?

juha-h:
Another question.  If ca_path is defined, are certificates concatenated
from all files in the ca_path directory.  Please clarify this in the
documentation.

PR #1300, opened 2021-03-17
tls: more generic function to set cafile and capath

cHuberCoffee:
Same PR like #1295. Renaming of the branch for automated testing.

adds an config option to add capath
requires re/#84

The ca_path is does have the same meaning like in [openSSL/SSL_CTX_load_verify_locations](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_load_verify_locations.html).  The ca_file defines a file with (maybe) many certificates in one file. While the ca_path defines a directory containing (maybe) many certificates as single files.  The verification checks will check first the one defined in ca_file, then go on to the ca_path certificates.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -297,6 +297,7 @@ struct config_sip {
 	char local[64];         /**< Local SIP Address              */
 	char cert[256];         /**< SIP Certificate                */
 	char cafile[256];       /**< SIP CA-file                    */
+	char capath[256];       /**< SIP CA-path                    */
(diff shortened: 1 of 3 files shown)


juha-h:
Is there some place other than config file comments where the meaning of config params can be explained?  Or can doxygen comments be included in the examples/config file?

Without an explanation like the one in this PR, the user would need to start reading source code in order to learn how these two config vars are used.

alfredh:
this PR looks fine, please merge to master

We could add a new Wiki page that describes each config/account parameter in details.

PR #1303, opened 2021-03-19
Update config.c

alfredh:
increase default resolution and framerate

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -65,9 +65,9 @@ static struct config core_config = {
 	{
 		"", "",
 		"", "",
-		352, 288,
-		500000,
-		25,
+		640, 480,
+		1000000,
+		30,
 		true,
 		VID_FMT_YUV420P,
 	},


cspiel1:
ok. Will later rebase my PR #1302, if you agree with this.

juha-h:
Christian Spielberger writes:

> ok. Will later rebase my PR #1302, if you agree with this.

I read #1302 and it looked OK, except why do we need this at all:
```
call_hold_other_calls	yes
```
I check from code and found only this usage:
```
/**
 * Put all established calls on hold, except the given one
 *
 * @param call  Excluded call, or NULL
 *
 * @return 0 if success, otherwise errorcode
 */
int uag_hold_others(struct call *call)
{
	int err = 0;
	struct le *le = NULL;
	struct call *acall = NULL;

	if (!conf_config()->call.hold_other_calls) {
		return 0;
	}
```
If app calls `uag_hold_other()` when why doesn't it mean it?  Why does
this call need to be nullified based on config var?  Why isn't this
option in the app itself?

PR #1304, opened 2021-03-19
ua,menu: remove uag_find_call_state

cspiel1:
This PR should replace the uag_find_call_state() by the new uag_filter_calls(). But the uag_filter_calls() seems not to be useful for all use-cases of uag_find_call_state() in module menu.

@juha-h I couldn't find a simpler solution for the use case in menu_selcall() which selects a call when a call terminates.

It looks first for a CALL_STATE_ESTABLISHED, then CALL_STATE_EARLY, then CALL_STATE_RINGING, ...

If you like this PR anyway, you may merge it.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -871,7 +871,6 @@ struct tls  *uag_tls(void);
 struct sipsess_sock  *uag_sipsess_sock(void);
 struct sipevent_sock *uag_sipevent_sock(void);
 struct call *uag_call_find(const char *id);
-struct call *uag_find_call_state(enum call_state st);
 void uag_filter_calls(call_list_h *listh, call_match_h *matchh, void *arg);
 
 
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -125,6 +125,21 @@ static void find_first_call(struct call *call, void *arg)
 }
 
 
+struct call *menu_find_call_state(enum call_state st)
+{
+	struct le *le;
+
+	for (le = list_head(uag_list()); le; le = le->next) {
+		struct ua *ua = le->data;
+		struct call *call = ua_find_call_state(ua, st);
+
+		if (call)
+			return call;
+	}
+
+	return NULL;
+}
+
 /**
  * Search all User-Agents for a call that matches
  *
@@ -578,7 +593,7 @@ void menu_selcall(struct call *call)
 	if (!call) {
 		/* select another call */
 		for (i = ARRAY_SIZE(state)-1; i >= 0; --i) {
-			call = uag_find_call_state(state[i]);
+			call = menu_find_call_state(state[i]);
 
 			if (!str_cmp(call_id(call), menu.callid))
 				call = NULL;
--- a/modules/menu/menu.h
+++ b/modules/menu/menu.h
@@ -58,3 +58,4 @@ void dial_menu_unregister(void);
 void menu_update_callstatus(bool incall);
 int  menu_param_decode(const char *prm, const char *name, struct pl *val);
 struct call *menu_find_call(call_match_h *matchh);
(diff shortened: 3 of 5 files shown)


juha-h:
This PR is OK to me, but I'm not sure if it makes sense to automatically select a call when another call terminates.  Instead the user should be given possibility to see the list and make the choice explicitly.  For example, if I have two calls on hold and the currently active call terminates, I would want to choose which of the two I want to unhold or none of them.

juha-h:
Christian Spielberger writes:

> The user has always the possibility to override the call selected in menu.
> The auto select if a call terminates has to be done, because the
> second "hangup" without parameter would not work.

If a call is not selected, then the second hangup would not do anything.
Same as it is when I press ESC when there are no calls.

juha-h:
When I start barasip two accounts got registered.  I then tried without luck to find a menu command that would allow me to choose one of the accounts or toggle between the accounts to make a call.  Is there one?

My idea was that in order to make/answer/reject a call, the commands to do so would apply the currently selected account.  Regarding incoming calls, the call that is answered or rejected could also be done based on /listcalls output in which case the current account would be automatically switched to that of the selected call.  If there is only one incoming call and no established calls, the answer/reject command would automatically apply to the only call.

I think we should first agree on these kind of general principles and then work on the implementation.

cspiel1:
It was the wish of @alfread that we remove the principle of the current UA (account). And I immediately agreed with him. Now we have the selected call which is set in ua_event_handler() and can be changed by the user with command "callfind". Choosing the UA is hidden from the user and is done automatically. We started to work on this in January:
```
commit cce11c5e6838768d27bfbcd7a7c94ccd47cc1d89
Author: Alfred E. Heggestad <alfred.heggestad@gmail.com>
Date:   Thu Jan 14 22:10:12 2021 +0100

    ua: remove ua_cur
```

The commands that operate on a call use the selected call and now also are able to use an optional parameter to directly address a call.

When a call terminates and there are more calls, another call has to be selected. Then the function play_resume() also is coupled on this selected call. With "listcalls" the selected call is marked with a '>'.

Anyway, this PR has a different purpose. It removes uag_find_call_state() to reduce the size of the API. The result is that I had to move the function to menu. If you like it, then you could merge it. It is also ok for me if we close this PR.

But the behavior of the selected call in menu has to stay. Our application developer are already depending on.

Maybe @alfredh could help here?

juha-h:
It is fine to merge this PR, but I still think that the current ua/call handling is not easy to use/understand and also does not give user enough choice.  There is no need for current ua concept in core.  menu module can and should implement it.

cspiel1:
Thanks! In January we had also for a few days a hybrid version. The user was able to choose an UA. If not chosen the automatic UA selection was applied. But then @alfredh and me decided to remove the UA selection completely.

Now I remember that you already had some reservations. Maybe we find a solution that's is acceptable.

juha-h:
Christian Spielberger writes:

> But the behavior of the selected call in menu has to stay. Our
> application developer are already depending on.

The menu module included in baresip project should implement an easy to
use and general purpose UI for common user.  If someone needs something
special for their own use case, nothing prevents writing a custom UI
module that optimizes behavior for their application.

I somehow missed the earlier commits regarding call handling or didn't
notice their importance.

juha-h:
I found out how to select the account for making outbound call: I need to add the index of the account after the called uri:
```
d jh 1
```
That is not convenient if I need to frequently use account with some other index than 0.  I still haven't figured out how to send messages using an account that does not have index 0.

There should be possibility in menu module to choose/toggle the current account and that selection should only be know to menu module, not the core.

juha-h:
`struct menu` has `char *callid` field.  Is there some reason why it could not also have  `struct ua *ua` field that tells the current ua and allows getting the account of the ua?

PR #1306, opened 2021-03-21
multicast: allow odd port number

webstean:
Plenty of products do allow odd port numbers and in the end it is a only a strong recommendation that the multicast port number be even.

### DIFF
--- a/modules/multicast/multicast.c
+++ b/modules/multicast/multicast.c
@@ -43,7 +43,7 @@ static int decode_addr(struct pl *pladdr, struct sa *addr)
 
 
 	if (sa_port(addr) % 2) {
-		err = EINVAL;
+		/* err = EINVAL; -- issue warrning but allow */
 		warning("multicast: address port for RTP should be even"
 			" (%d)\n" , sa_port(addr));
 	}


cHuberCoffee:
The sender uses `rtp_open` which allows to send on any port number.
The receiver uses `rtp_listen` which does not allow an odd rtp port. It will replace the port number with the next  lower free even port.
See my debug output directly after the `rtp_listen`:  `DEBUG_INFO("local addr %J\n", rtp_local(mcreceiver->rtp));`

> /mcreg addr=224.0.0.1:8005 prio=5
> multicast: address port for RTP should be even (8005)
> mcreceiver: local addr 224.0.0.1:8004

To allow odd port numbers, one have to replace the `rtp_sock` object in the receiver with a plain `udp_sock` object.

cspiel1:
RTP should be on even port. This is what the RTP implementation of re ensures. The only purpose of the multicast module is to send and receive RTP. The multicast module should not use the udp API of re directly.

So we should first clarify if we really want RTP on odd port numbers in libre.

Which products allow RTP on odd port numbers?

alfredh:
the re RTP stack was tested for many years and we always had a very high interop.

one of the things we made sure, is that RTP listens on an even port and RTCP is port + 1
as is written in the specification.

the re RTP stack has also been subject to many SIP-it interop testing events
with many different vendors, with very high success rate.
many conservative vendors actually assumes that remote peer uses an even port for RTP.

cspiel1:
I agree. Should we close this?

webstean:
A lot of small SIP speakers don't enforce the even port number, for example: https://www.algosolutions.com/wp-content/uploads/2019/11/Multicast-Guide-1.pdf
As discussed above, this PR won't solve that, since this is enforced by libre, so this should be closed.
I'll add it to the roadmap for consideration.

PR #1312, opened 2021-03-24
ua: allow also IP:port as dial URI

cspiel1:
The check for uri_is_ip was not correct if also a port number is specified.
```
x.x.x.x:5060
```

Also if the scheme was prepended:
```
sip:x.x.x.x
```

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -1076,6 +1076,9 @@ int ua_uri_complete(struct ua *ua, struct mbuf *buf, const char *uri)
 	struct sa sa_addr;
 	size_t len;
 	bool uri_is_ip;
+	char *uridup;
+	char *host;
+	char *p;
 	int err = 0;
 
 	if (!buf || !uri)
@@ -1099,7 +1102,21 @@ int ua_uri_complete(struct ua *ua, struct mbuf *buf, const char *uri)
 	/* Append domain if missing and uri is not IP address */
 
 	/* check if uri is valid IP address */
-	uri_is_ip = (0 == sa_set_str(&sa_addr, uri, 0));
+	err = str_dup(&uridup, uri);
+	if (err)
+		return err;
+
+	if (!strncmp(uridup, "sip:", 4))
+		host = uridup + 4;
+	else
+		host = uridup;
+
+	p = strchr(host, ':');
+	if (p)
+		*p = 0;
+
+	uri_is_ip = (0 == sa_set_str(&sa_addr, host, 0));
+	mem_deref(uridup);
 	acc = ua->acc;
 
 	if (0 != re_regex(uri, len, "[^@]+@[^]+", NULL, NULL) &&


juha-h:
I have a general comment about this.  In my opinion, baresip core should always assume that it gets an rfc3261 compliant sip uri and should reject everything else with en error.  It is the task of the app to do whatever "fixing" is needed to user input.

cspiel1:
IP:Port is RFC compliant. This PR only tries to fix the completion of user by user@host where host is account->luri.host.

Without this fix an URI IP:Port is replaced by IP:Port@host which is nonsense.

juha-h:
Yes, the PR is fine.  My comment was a general one, i.e., should this function at all be in ua.c.  There could instead be a general purpose utility function that takes user input and domain as arguments.

cspiel1:
Good point! In the last weeks we already moved application specific stuff from ua.c to menu. This would be another task. Do you think that ua_uri_complete() should be called from menu? Then also module gtk and baresip-studio will have to call this function. Right?

 For the next view days I would suggest first to split ua.c into ua.c and uag.c. I plan to make a PR. Anyway, I hope this does not generate much merge conflicts.

juha-h:
Christian Spielberger writes:

> Good point! In the last weeks we already moved application specific
> stuff from ua.c to menu. This would be another task. Do you think that
> ua_uri_complete() should be called from menu? Then also module gtk and
> baresip-studio will have to call this function. Right?

That is what I was thinking.

cspiel1:
Can be done simpler with sa_decode(). --> convert to draft

juha-h:
Just checked and an IPv4 address like 192.168.43.160 is syntactically valid SIP URI userpart.

juha-h:
juha-h writes:

> Just checked and an IPv4 address like 192.168.43.160 is syntactically
> valid SIP URI userpart.

So would it make sense to require that user includes scheme `sip:` if
request is made to an IP address?

cspiel1:
No. The function ua_uri_complete() should be able to handle this. But in the future (maybe also in this PR) we should call this from menu. Especially the pre-pending of "sip:" is the first thing ua_uri_complete() does and did already for a long time.

alfredh:
one solution is to create a new helper .c file that can contain this function:

```c
/**
 * Auto complete a SIP uri, add scheme and domain if missing
 *
 * @param ua  User-Agent
 * @param buf Target buffer to print SIP uri
 * @param uri Input SIP uri
 *
 * @return 0 if success, otherwise errorcode
 */
int ua_uri_complete(struct ua *ua, struct mbuf *buf, const char *uri)
{
	struct account *acc;
	struct sa sa_addr;
	size_t len;
	bool uri_is_ip;
	int err = 0;

	if (!buf || !uri)
		return EINVAL;

	/* Skip initial whitespace */
	while (isspace(*uri))
		++uri;

	len = str_len(uri);

	/* Append sip: scheme if missing */
	if (0 != re_regex(uri, len, "sip:"))
		err |= mbuf_printf(buf, "sip:");

	err |= mbuf_write_str(buf, uri);

	if (!ua)
		return 0;

	/* Append domain if missing and uri is not IP address */

	/* check if uri is valid IP address */
	uri_is_ip = (0 == sa_set_str(&sa_addr, uri, 0));
	acc = ua->acc;

	if (0 != re_regex(uri, len, "[^@]+@[^]+", NULL, NULL) &&
		1 != uri_is_ip) {
#if HAVE_INET6
		if (AF_INET6 == acc->luri.af)
			err |= mbuf_printf(buf, "@[%r]",
					   &acc->luri.host);
		else
#endif
			err |= mbuf_printf(buf, "@%r",
					   &acc->luri.host);

		/* Also append port if specified and not 5060 */
		switch (acc->luri.port) {

		case 0:
		case SIP_PORT:
			break;

		default:
			err |= mbuf_printf(buf, ":%u", acc->luri.port);
			break;
		}
	}

	return err;
}
```

and the input param can be `struct account`.

I think the auto-complete logic should be done outside ua.c (i.e. in menu or app).
i.e. the design is that ua.c does not know anything about its user.

juha-h:
@alfredh Good proposal.  Since the function only uses acc of ua, ua parameter can be replaced by acc parameter.

cspiel1:
Shouldn't we put it into account.c and name to:
```
int account_uri_complete(struct account *acc, struct mbuf *buf, const char *uri)
```

Since the logic is to convert the input in a complete SIP URI as good as possible. This is general enough to be useful for menu, gtk, baresip-studio, ...

account_uri_complete() will be called from menu and not directly from ua.c.

Or we move the function to menu and it has to be implemented in other applications again. Since we use menu as API this is also ok for us.

juha-h:
Now that also port is added, then account.c would be the best place for the function.

juha-h:
I pushed a new branch uri_complete, but didn't create pull request yet.  It moved ua_uri_complete to account_uri_complete and used that in menu module.  I'm not sure if I got all the mbuf stuff correct.

juha-h:
I think this PR was now included in https://github.com/baresip/baresip/pull/1317.

cspiel1:
Yes, looks good now. Except that sa_decode() could simplify the check for an IP-address in account_uri_complete(). Maybe will make a small PR later.

juha-h:
Christian Spielberger writes:

> Yes, looks good now. Except that sa_decode() could simplify the check
> for an IP-address in account_uri_complete(). Maybe will make a small
> PR later.

OK.  By the way, just noticed that now ua index is always required in d
command.  Seems that there is no current UA anymore.

juha-h:
juha-h writes:

> OK.  By the way, just noticed that now ua index is always required in d
> command.  Seems that there is no current UA anymore.

Also d without any arguments that used to do repeat the previous call,
has stopped working,

cspiel1:
Ok, will try to solve this.

Please remember: current UA was replaced by current call. There is no "current UA" any longer.

juha-h:
Christian Spielberger writes:

> Please remember: current UA was replaced by current call. There is no
> "current UA" any longer.

And in my opinion that is not a good thing.  It made baresip cli much
harder to use.

I guess I have no other choice, but to roll my own version of menu
module.

cspiel1:
The opposite is the case. You don't have to select a UA. Selecting an UA to make a call is for expert users. The auto UA selection works with the fix for the account_uri_complete().

For the expert user, we have a per command UA selection that overrides the auto selection. Implemented by menu_ua_carg().

The PR is coming soon. There is also a memleak. That's new.  I want to solve this also.

cspiel1:
Everything works again. Also redial. https://github.com/baresip/baresip/pull/1318

juha-h:
Christian Spielberger writes:

> Everything works again. Also redial. https://github.com/baresip/baresip/pull/131

Yes, I tested too and it worked. So let's commit this PR.

juha-h:
But if I have two accounts like this:
```
baresip is ready.
***@***.***: (prio 0) {1/TLS/v4} 200 OK (OpenSIPg SIP Proxy (5.5.0-0b34 (x86_64/linux))) [1 binding]
***@***.***: (prio 0) {1/TLS/v4} 200 OK (OpenSIPg SIP Proxy (5.5.0-0b34 (x86_64/linux))) [2 bindings]
```

there seems to be no way to toggle between them and choose, which one to
use when I make call to `jh`.

The only way to choose ***@***.*** seems to be using index 1:

d jh 1

After I have made a few calls, its easy to forget, which account had
which index.  So I don't understand, why menu module cannot have both
concepts: current call and current account, and a way to toggle between
the accounts.

PR #1315, opened 2021-03-25
docs: change video settings to match the default values

cspiel1:


juha-h:
Isn't video_fps a floating point number?
```
 "video_fps\t\t%.2f\n"
```

PR #1317, opened 2021-03-27
replaced ua_uri_complete with account_uri_complete

juha-h:
I'll create this PR so that it is easier to compare to master if someone has time to take a look.  It works in my tests, but as I mentioned, I'm not sure if I have used the mbuf stuff correctly.


src/account.c
```diff
@@ -1420,6 +1421,91 @@ const char *account_extra(const struct account *acc)
...
+	struct sa sa_addr;
+	size_t len;
+	bool uri_is_ip;
+	char *uridup;
+	char *host;
+	char *p;
+	int err = 0;
+
+	if (!buf || !uri)
+		return EINVAL;
```
alfredh, line 1445: also check `acc` here

src/ua.c
```diff
@@ -1212,7 +1142,7 @@ int ua_connect_dir(struct ua *ua, struct call **callp,
 	if (!dialbuf)
 		return ENOMEM;
 
-	err |= ua_uri_complete(ua, dialbuf, req_uri);
+	mbuf_write_str(dialbuf, req_uri);
```
alfredh, line 1145: check return value

src/ua.c
```diff
@@ -1413,9 +1343,7 @@ int ua_options_send(struct ua *ua, const char *uri,
 	if (!dialbuf)
 		return ENOMEM;
 
-	err = ua_uri_complete(ua, dialbuf, uri);
-	if (err)
-		goto out;
+	mbuf_write_str(dialbuf, uri);
```
alfredh, line 1346: perhaps `uri` can be used directly here ?

src/ua.c
```diff
@@ -2501,16 +2428,14 @@ struct ua *uag_find_requri(const char *requri)
 	if (!mb)
 		return NULL;
 
-	err = ua_uri_complete(NULL, mb, requri);
-	if (err)
-		goto out;
+	account_uri_complete(NULL, mb, requri);
```
alfredh, line 2431: first parameter is NULL ..

src/account.c
juha-h: `acc` is checked later, because it can be `NULL`.  see usage in `uag_find_requri()`, where `acc` is not known. if `acc` is `NULL`. missing "sip:" is appended anyway.

src/ua.c
juha-h: Will do.

src/ua.c
juha-h: Yes, will use.  Looks much cleaner then.

PR #1318, opened 2021-03-29
Fixes for account uri complete

cspiel1:


juha-h:
My reply disappeared somewhere, so I'll repeat it here for the record.

If I have two accounts like this:
```
baresip is ready.
test@test.tutpro.com: (prio 0) {1/TLS/v4} 200 OK (OpenSIPg SIP Proxy (5.5.0-0b34 (x86_64/linux))) [1 binding]
foo@test.tutpro.com: (prio 0) {1/TLS/v4} 200 OK (OpenSIPg SIP Proxy (5.5.0-0b34 (x86_64/linux))) [2 bindings]
```

there seems to be no way to toggle between them and choose, which one to use when I make call to `jh`.

The only way to choose foo@test.tutpro.com seems to be using index 1:

d jh 1

After I have made a few calls, its easy to forget, which account had which index.  So I don't understand, why menu module  cannot have both concepts: current call and current account, and a way to toggle between the accounts.

cspiel1:
Thanks for merge!

Ok, I think we could find an improvement for the usability. The auto UA selection filters out accounts which make no sense for a dial URI (because the call would fail). If multiple UAs are possible for a given URI, then it simply takes the first one.

So changing the order of the UAs would be best solution in my opinion.

We could add a command the puts a given UA to the top of the UA list. Then if the user uses this "uatop" multiple times the result would be a order of UAs that makes most sense. Please give me an advise for a good name for "uatop"!

juha-h:
Christian Spielberger writes:

> We could add a command the puts a given UA to the top of the UA
> list. Then if the user uses this "uatop" multiple times the result
> would be a order of UAs that makes most sense. Please give me an
> advise for a good name for "uatop"!

We already have /contact_next and /contact_prev. Thus /ua_next and
/ua_prev would be the logical choice.

Issue #1290, opened 2021-03-15
Network handover blocks socket


jurjen-van-dijk:
Not sure if this is baresip or re, but I'll put it here now.

I have a working iOS version of baresip, but, when I do network handover from wifi to cell or the other way around, for some reason it becomes blocked.

This is what I see:
I have baresip running and it is registered. I turn off wifi, so 'net_change_handler' is called and does a 'uag_reset_transp'.
Which goes without a problem. But when the re-register is triggered it get's stuck in the 'dnsc_query'.
It call's 'udp_send_internal' and that sends 40 bytes with 'sendto' and I get the 40 as a return value. But there is no reply, so the reply handler is not triggered and it just get's stuck there.

When I change back to the original network,  I see the same flow, but the reply handler is triggered, and move's on to do the actual register and it all starts running again.
And it is not related to the network type, because I have the same if I start on cell and change to wifi (wifi gets stuck), or the other way around and start with wifi and change to cell (cell gets stuck).

Any ideas on what can cause this?

alfredh:
can you please provide a config and log file ? also exact version number is useful.

is it possible to reproduce the issue on a Mac with 2 network interfaces (e.g. Wifi and Ethernet) ?

perhaps check if the DNS client binds to a specific network interface, or to a wildcard address.

NOTE: The network code is currently polling for network changes, this is not optimal. Both OSX and iOS have
OS hooks/callbacks for when a network interfaces goes up/down. Perhaps consider to add support for this.

jurjen-van-dijk:
That last thing I already did. I added 'void ua_resetnetwork(void);' to ua.c and exposed it in baresip.h. It is a small function that calls 'uag_reset_transp(true, true)'. When the Swift side get's a network change event from the OS, it call's that function to do the reset.

As for running on a Mac, I will have a look at it, will get back on that.

It seems like the DNS client sticks to the IP it has from the start of the app, even after the changeover.
I printed the config in the log. Also had to remove some overhead of iOS log messages. This is what I see the I start with wifi and switch to cell.

```
main: set: [fd=0, flags=1] kevent: Invalid argument
main: fd_listen: fd=0 flags=0x01 (Invalid argument)
dl: mod: ./stdio.so (dlopen(//stdio.so, 6): image not found)
module stdio.so: No such file or directory
aucodec: opus/48000/2
aucodec: PCMU/8000/1
aucodec: PCMA/8000/1
[...]

PR #1319, opened 2021-03-29
Changed name and made public

jurjen-van-dijk:
I changed the name 'dns_refresh' to 'net_dns_refresh' and made it public.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -696,6 +696,7 @@ void net_force_change(struct network *net);
 bool net_check(struct network *net);
 bool net_af_enabled(const struct network *net, int af);
 int  net_set_af(struct network *net, int af);
+void net_dns_refresh(struct network *net);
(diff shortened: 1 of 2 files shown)


cspiel1:
net_use_nameserver() already is in the API.
Did you check if it does what you want if you call it like this:
```
err = net_use_nameserver(net, NULL, 0);
```

I am not sure because it does also:
```
net->nsn = (uint32_t)min(ARRAY_SIZE(net->nsv), srvc);
```

jurjen-van-dijk:
I made this PR for this issue.
https://github.com/baresip/baresip/issues/1290

I tried what you suggest, and in this case that works too. 
However using it that way will hide the functionality I was looking for (refreshing of the dns servers) in the list of public functions.

alfredh:
thanks, this PR looks good :)

PR #1320, opened 2021-03-29
menu: return call-id for dial and dialdir

cspiel1:
E.g. module ctrl_dbus will return the call id in the dbus response.
```
$ qdbus com.github.Baresip /baresip com.github.Baresip.invoke "dial 10.1.0.215"
call id: bde55d2528902248
```
Can be used for following call specific commands e.g. hangup, without the need
to query calls with "listcalls".


juha-h:
This looks OK to me, since it just gets access to the call object.

PR #1323, opened 2021-04-01
ua: remove prev call

alfredh:
this is a proposal to remove previous call hold

it is not clear if this is a useful feature, or if anyone is using it


cspiel1:
- I was also confused when I saw this commands, because it is not clear which use-case it should solve.
- It is not clear what exactly the previous call is, if there are multiple calls on different UAs.

Most likely uag_hold_others(call) called when a call gets established makes the "previous call hold" feature obsolete.
Use-Case: Ensure that there is only one established call.

Thus this PR is a good idea to reduce the complexity of the API and module menu.

alfredh:
thanks @cspiel1 

we should try to keep the codebase nice and slim, and remove unused features

PR #1330, opened 2021-04-07
Nw event UA_EVENT_SNAPSHOT to tell to app when snapshot has been written

juha-h:
It was clearer to add new event rather than use UA_CUSTOM_EVENT.  Did not see use for new event category.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -759,7 +759,7 @@ enum ua_event {
 	UA_EVENT_AUDIO_ERROR,
 	UA_EVENT_CALL_LOCAL_SDP,      /**< param: offer or answer */
 	UA_EVENT_CALL_REMOTE_SDP,     /**< param: offer or answer */
-	UA_EVENT_SNAPSHOT,
+	UA_EVENT_MODULE,
 	UA_EVENT_CUSTOM,
 
 	UA_EVENT_MAX,
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -759,6 +759,7 @@ enum ua_event {
 	UA_EVENT_AUDIO_ERROR,
 	UA_EVENT_CALL_LOCAL_SDP,      /**< param: offer or answer */
 	UA_EVENT_CALL_REMOTE_SDP,     /**< param: offer or answer */
+	UA_EVENT_SNAPSHOT,
 	UA_EVENT_CUSTOM,
 
 	UA_EVENT_MAX,
--- a/modules/snapshot/png_vf.c
+++ b/modules/snapshot/png_vf.c
@@ -121,7 +121,7 @@ int png_save_vidframe(const struct vidframe *vf, const char *path)
 
 	info("png: wrote %s\n", path);
 
-	ua_event(NULL, UA_EVENT_SNAPSHOT, NULL, "%s", path);
+	ua_event(NULL, UA_EVENT_MODULE, NULL, "snapshot,wrote,%s", path);
 
  out:
 	/* Finish writing. */
--- a/modules/snapshot/png_vf.c
+++ b/modules/snapshot/png_vf.c
@@ -121,6 +121,8 @@ int png_save_vidframe(const struct vidframe *vf, const char *path)
 
 	info("png: wrote %s\n", path);
 
+	ua_event(NULL, UA_EVENT_SNAPSHOT, NULL, "%s", path);
+
  out:
 	/* Finish writing. */
 	mem_deref(f2);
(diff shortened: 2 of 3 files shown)


cspiel1:
The UA_CUSTOM_EVENT was introduced exactly for module specific events.

juha-h:
In that case, there should be some useful format for the module events.  For example, module name, event name, params, and the name of the module event should be UA_EVENT_MODULE.  Because there wasn't, I decided to use a new event.  Also, mwi module has its own event.

cspiel1:
In my opinion mwi is a good example how we should NOT do it. Implementing modules that send there specific events should be possible without adding a new UA_EVENT_.

A UA_EVENT_MODULE with a general format module name, event name, params sounds good for me.

Then we could also replace the UA_EVENT_MWI by UA_EVENT_MODULE and maybe also UA_EVENT_CUSTOM.

juha-h:
Christian Spielberger writes:

> A UA_EVENT_MODULE with a general format module name, event name,
> params sounds good for me. 
> 
> Then we could also replace the UA_EVENT_MWI by UA_EVENT_MODULE and
> maybe also UA_EVENT_CUSTOM.

Fine, let's do that.

cspiel1:
Looks good. Should we add now a function to the API
```
int module_event(module,event,params)
{
return ua_event(NULL, UA_EVENT_MODULE, NULL, "%s,%s,%s", module, event, params);
}
```

Then replace
```
ua_event(NULL, UA_EVENT_MODULE, NULL, "snapshot,wrote,%s", path);
```
by
```
module_event("snapshot", "wrote", path);
```

juha-h:
May not be a good idea to fix the format (e.g. "%s,%s,%s") by defining module_event() function, since there may be many params with different types.  Calling ua_event() directly does not feel too complicated to me.

cspiel1:
You're right, but then the UA_EVENT_MODULE format is completely free like it is for UA_EVENT_CUSTOM. So where is the benefit of UA_EVENT_MODULE?
Only a renaming makes not much sense.

What about:
```c
void module_event(const char *module, const char *event, const char *fmt, ...);
```

juha-h:
Christian Spielberger writes:

> You're right, but then the UA_EVENT_MODULE format is completely free
> like it is for UA_EVENT_CUSTOM. So where is the benefit of
> UA_EVENT_MODULE? Only a renaming makes not much sense.

It is not just renaming, since it defines module,event,params format.
Current UA_EVENT_CUSTOM usage does not conform to that.

> What about:
> ```c
> void module_event(const char *module, const char *event, const char *fmt, ...);
> ```

With that we would loose possibility to include ua and call params in
module events.

cspiel1:
> It is not just renaming, since it defines module,event,params format.
This is only your convention. UA_EVENT_MODULE does the same like UA_EVENT_CUSTOM. It has no extra rules so far.

We have these options:
- Ensure the module,event,params format somehow for UA_EVENT_MODULE. Then we have a useful difference to UA_EVENT_CUSTOM.
- Rename UA_EVENT_CUSTOM to UA_EVENT_MODULE.
- Do not rename UA_EVENT_CUSTOM.

juha-h:
Christian Spielberger writes:

> > It is not just renaming, since it defines module,event,params format.
> 
> This is only your convention. UA_EVENT_MODULE does the same like
> UA_EVENT_CUSTOM. It has no extra rules so far.

That was my proposal and would have meant that all modules that now use
UA_EVENT_CUSTOM would have to be modified to match the format.

> We have these options:
> - Ensure the module,event,params format somehow for UA_EVENT_MODULE. Then we have a useful difference to UA_EVENT_CUSTOM.
> - Rename UA_EVENT_CUSTOM to UA_EVENT_MODULE.
> - Do not rename UA_EVENT_CUSTOM.

module_event() function is fine with me if it also includes ua and call
params so that we don't loose functionality and can also migrate mwi
module to use it.

cspiel1:
Good, then module_event() will ensure the format "module,event,params" and have also ua and call. It will exist in parallel to ua_event() that we have already.

>That was my proposal and would have meant that all modules that now use
UA_EVENT_CUSTOM would have to be modified to match the format.

ok.

juha-h:
In which file would module_event() API function be defined?

cspiel1:
beside ua_event() in event.c

juha-h:
Better that you write the function, since I don't know how to handle the format arguments.

cspiel1:
Ok, I could do this. But will take some time because I am very busy today.

PR #1339, opened 2021-04-10
gtk with account_uri_complete

mbattista:
GTK was no longer able to dial since the URI was not auto completed.

### DIFF
--- a/modules/gtk/gtk_mod.c
+++ b/modules/gtk/gtk_mod.c
@@ -641,13 +641,35 @@ static gboolean status_icon_on_button_press(GtkStatusIcon *status_icon,
 }
 
 
-void gtk_mod_connect(struct gtk_mod *mod, const char *uri)
+int gtk_mod_connect(struct gtk_mod *mod, const char *uri)
 {
-	if (!mod)
-		return;
+	struct mbuf *uribuf = NULL;
+	char *uri_copy = NULL;
+	int err = 0;
 
-	mqueue_push(mod->mq, MQ_CONNECT, (char *)uri);
+	if (!mod)
+		return ENOMEM;
+
+	uribuf = mbuf_alloc(64);
+	if (!uribuf)
+		return ENOMEM;
+
+	err = account_uri_complete(ua_account(mod->ua_cur), uribuf, uri);
+	if (err)
+		return EINVAL;
+
+	uribuf->pos = 0;
+	err = mbuf_strdup(uribuf, &uri_copy, uribuf->end);
+	if (err)
+		goto out;
+
+	err = mqueue_push(mod->mq, MQ_CONNECT, uri_copy);
+
+out:
+	mem_deref(uribuf);
+	return err;
 }
+
 
 bool gtk_mod_clean_number(struct gtk_mod *mod)
 {
@@ -711,6 +733,7 @@ static void mqueue_handler(int id, void *data, void *arg)
 		if (err) {
 			ua_hangup(ua, call, 500, "Server Error");
 		}
+		mem_deref(data);
 		break;
 
 	case MQ_HANGUP:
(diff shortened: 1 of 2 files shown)


modules/gtk/gtk_mod.c
```diff
@@ -641,14 +641,36 @@ static gboolean status_icon_on_button_press(GtkStatusIcon *status_icon,
...
+
+	uribuf->pos = 0;
+	err = mbuf_strdup(uribuf, &uri_copy, uribuf->end);
+	if (err)
+		goto out;
+
+	err = mqueue_push(mod->mq, MQ_CONNECT, uri_copy);
+
+out:
+	mem_deref(uribuf);
```
cspiel1, line 669: mem_deref(uri_copy) is missing

cspiel1:
Looks good.

alfredh:
tested on OSX 10.14, dialing a contact gives a segfault:

```
/insmod gtk
aufilt: gtk_vumeter
loaded module gtk

(process:37338): Gtk-WARNING **: 10:40:32.253: Locale not supported by C library.
	Using the fallback 'C' locale.
dbus[37338]: Dynamic session lookup supported but failed: launchd did not provide a socket path, verify that org.freedesktop.dbus-session.plist is loaded!
gtk_menu starting
/gtk
ua: using best effort AF: af=AF_INET
call: connecting to '"Anti Echo" <sip:echo@sip.antisip.com>;transport=tcp'..
mem: mem_deref: magic check failed 0x00000000 (0x10119aee0)
Process 37338 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BREAKPOINT (code=EXC_I386_BPT, subcode=0x0)
    frame #0: 0x000000010018961f libre.dylib`mem_deref(data=0x000000010119aee0) at mem.c:314:8
   311 	
   312 		MAGIC_CHECK(m);
   313 	
-> 314 		if (--m->nrefs > 0)
   315 			return NULL;
   316 	
   317 		if (m->dh)
Target 0: (baresip) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BREAKPOINT (code=EXC_I386_BPT, subcode=0x0)
  * frame #0: 0x000000010018961f libre.dylib`mem_deref(data=0x000000010119aee0) at mem.c:314:8
    frame #1: 0x00000001110d1df3 gtk.so`mqueue_handler(id=1, data=0x000000010119aee0, arg=0x00000001110d8748) at gtk_mod.c:736:3
    frame #2: 0x000000010018bf70 libre.dylib`event_handler(flags=1, arg=0x00000001005b7e60) at mqueue.c:86:2
    frame #3: 0x00000001001885b1 libre.dylib`fd_poll(re=0x00000001001a3af8) at main.c:892:4
    frame #4: 0x0000000100187d7c libre.dylib`re_main(signalh=(baresip`signal_handler at main.c:21)) at main.c:1026:9
    frame #5: 0x000000010002ec7a baresip`main(argc=3, argv=0x00007ffeefbff978) at main.c:309:8
    frame #6: 0x00007fff67fa53d5 libdyld.dylib`start + 1
    frame #7: 0x00007fff67fa53d5 libdyld.dylib`start + 1
(lldb) 
```

mbattista:
Very good catch.
I reuse the connect function now in case of dialing contact.

I wish I had a solution for "cleaner" memory cleanup, but no matter how I tried it, the `mem_deref` function was called before I could get an copy of the object in the `mqueue_push` function. I hope the proposed solution is still ok.

alfredh:
we are still looking for a maintained for the gtk.so module:

feel free to grab the spot :)

mbattista:
> we are still looking for a maintained for the gtk.so module:
> 
> https://github.com/baresip/baresip/wiki/Modules
> 
> feel free to grab the spot :)

I will think about it. Feel free to tag me with issues of the gtk module.

PR #1340, opened 2021-04-10
DTMF tones for A B C D

mbattista:
Added DTMF tones for A B C D created via https://www3.nd.edu/~dthain/courses/cse20211/fall2013/wavfile/

I had a look at DTMF tones for the menu, or at least, why A B C D are not sendable.
From my perspective those would only sendable if there was a dedicated `dial_dtmf_handler` function, similar to the `dial_handler` function.


cspiel1:
Looks good for me.

The user has also the option to use command "sndcode" of menu.

PR #1343, opened 2021-04-11
Document and enable rtcpsummary and snapshot modules

### LINKED ISSUE #1341: Module status of av1, rtcpsummary and snapshot?
As of writing, the following modules exist in the code of HEAD, but are not enabled by any defaults or mechanisms:

  * `av1`: Not mentioned in `mk/modules.mk` at all
  * `rtcpsummary`: Mentioned in `mk/modules.mk`, but not enabled by default
  * `snapshot`: Not mentioned in `mk/modules.mk` at all

So what's the status of these modules? As a downstream packager (for Fedora, RHEL/CentOS), should I build them? Or not?

robert-scheck:
An attempt to clarify the unknown status of these modules (fixes #1341).

### DIFF
--- a/mk/modules.mk
+++ b/mk/modules.mk
@@ -44,6 +44,7 @@
 #   USE_PULSE         Pulseaudio audio driver
 #   USE_RTCPSUMMARY   RTCP summary output after calls
 #   USE_SDL           libSDL2 video output
+#   USE_SNAPSHOT      Snapshot video module
 #   USE_SNDFILE       sndfile wav dumper
 #   USE_SNDIO         sndio audo driver
 #   USE_SPEEX_PP      Speex preprocessor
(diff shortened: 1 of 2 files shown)


robert-scheck:
Oh, the `snapshot` module uses `-Wno-clobbered`, which isn't supported by `clang`:

```
error: unknown warning option '-Wno-clobbered' [-Werror,-Wunknown-warning-option]
```

robert-scheck:
So, Mac OS `clang` fails now like this:

```
error: unknown warning option '-Wno-aggressive-loop-optimizations' [-Werror,-Wunknown-warning-option]
```

mk/modules.mk
```diff
@@ -10,6 +10,7 @@
 #   USE_AMR           Adaptive Multi-Rate (AMR) audio codec
 #   USE_APTX          aptX audio codec
 #   USE_AUDIOUNIT     AudioUnit audio driver for OSX/iOS
+#   USE_AV1           AV1 video codec module
```
alfredh, line 13: I would suggest to not enable av1, at least until it has been interop tested

PR #1344, opened 2021-04-12
autotest: add a module for automatic call testing

cspiel1:
* Supports automatic repeated dialing and hangup via timers.
 * The commands for dialing and hangup can be specified by means of the registered commands.

### DIFF
--- a/mk/modules.mk
+++ b/mk/modules.mk
@@ -10,6 +10,7 @@
 #   USE_AMR           Adaptive Multi-Rate (AMR) audio codec
 #   USE_APTX          aptX audio codec
 #   USE_AUDIOUNIT     AudioUnit audio driver for OSX/iOS
+#   USE_AUTOTEST      Automatic test module
 #   USE_AVCAPTURE     AVFoundation video capture for OSX/iOS
 #   USE_AVCODEC       avcodec video codec module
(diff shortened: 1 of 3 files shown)


alfredh:
I think we should discuss putting features like this outside the baresip tree.
For example a new project `baresip-tools` or `baresip-apps` can be created.
This project can use the libbaresip api.

Some other modules that fits into this category:

* auloop
* vidloop

The reason is that the number of modules is growing quite high.
For a new enduser it can be quite challenging to figure out which module to use or not.

this is just an idea for now, feel free to discuss :)

apart from this comment, the code in the PR looks fine to me.

cspiel1:
I like this idea. You mean a new repository github.com/baresip, right? Do you want to create such a new repo?

alfredh:
something like this:

```
github.com/baresip/baresip-apps
```

the binaries can include auloop, vidloop, autotest and more.
the binaries will use the public libbaresip api.

cspiel1:
Should I create this?

- Named: `baresip-app`
- Makefile similar to baresip that includes re.mk?
- a first simple github workflow

alfredh:

yes, please move the autotest module to this new baresip-apps repo.

cspiel1:
Replaced by https://github.com/baresip/baresip-apps/pull/1

PR #1350, opened 2021-04-14
GTK caller history

mbattista:
A simple caller history on the gtk module.

Tracks up to 20 missed / outgoing / incoming calls.

### DIFF
--- a/modules/gtk/gtk_mod.c
+++ b/modules/gtk/gtk_mod.c
@@ -201,7 +201,8 @@ static void add_history_menu_item(struct gtk_mod *mod, const char *uri,
 
 	if (mod->call_history_length < 20) {
 		mod->call_history_length++;
-	} else {
+	}
+	else {
 		/* Remove old call history */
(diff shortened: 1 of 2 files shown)


modules/gtk/gtk_mod.c
```diff
@@ -150,6 +156,23 @@ static void menu_on_dial_contact(GtkMenuItem *menuItem, gpointer arg)
...
+static void menu_on_dial_history(GtkMenuItem *menuItem, gpointer arg)
+{
+	struct gtk_mod *mod = arg;
+	const char *label = gtk_menu_item_get_label(menuItem);
+	char *label_1;
+	char buf[256];
+	char *uri;
+
+	str_ncpy(buf, label, sizeof(buf));
+	label_1 = strchr(buf, '[');
```
alfredh, line 168: can strchr return NULL ?

mbattista: while all labels are inserted with an `[`, this theoretically should never be NULL, but since it can return NULL a check against NULL should still be done.

PR #1349, opened 2021-04-13
ua,uag: split ua.c and uag.c

cspiel1:
- uag.c contains code for User-Agent Group. E.g. functions that operate on the
  list of User-Agents.
- ua.c contains code for User-Agent. E.g. functions that operate on one User-
  Agent.

PR #1355, opened 2021-04-15
menu: command uafind raises UA to head

cspiel1:
- The UA at head will be used for a dial to a user only URI.
- Command `uafind` also switches the active call if there is one at this UA.

This PR is based on https://github.com/baresip/baresip/pull/1349 and has to be rebased after #1349 was merged.

This solution replaces the already removed `uanext` ("current UA") solution. The benefit is that `uag_find_requri()` selects an adequate UA regarding these rules: If URI looks like
- user --> selects the head UA
- user@domain --> selects the UA whose account host matches domain

### DIFF
--- a/src/core.h
+++ b/src/core.h
@@ -377,6 +377,7 @@ struct uag {
 struct config_sip *uag_cfg(void);
 const char *uag_eprm(void);
 bool uag_delayed_close(void);
+int uag_raise(struct ua *ua, struct le *le);
 
 
 /*
--- a/src/core.h
+++ b/src/core.h
@@ -342,6 +342,41 @@ struct ua;
 void         ua_printf(const struct ua *ua, const char *fmt, ...);
 
 int ua_print_allowed(struct re_printf *pf, const struct ua *ua);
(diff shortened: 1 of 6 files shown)


src/core.h
```diff
@@ -342,6 +342,42 @@ struct ua;
...
+	char *eprm;                    /**< Extra UA parameters             */
+#ifdef USE_TLS
+	struct tls *tls;               /**< TLS Context                     */
+#endif
+};
+
+struct config_sip *uag_cfg(void);
+const char *uag_eprm(void);
+bool uag_delayed_close(void);
+int uag_raise(struct ua *ua, struct le *le);
```
alfredh, line 380: this patch contains a subset of another patch.

I suggest you close this one and wait until the base patch has been merged.

alfredh:
there is now around 10 patches that depends on a base patch to be merged first.

this is not an ideal situation.

it would be better if you can submit 1-2 patches and then wait for them to be merged.

I will close this one for now, please reopen when the base patch is merged.

PR #1356, opened 2021-04-15
gst: Use GstBus sync handler instead of GMainLoop

RobertMi21:
This PR removes the use of GMainLoop in module gst.

The module ctrl_dbus instantiates also a GMainLoop which might result in problems like dead-lock. We observed a dead-lock during stop of ring tone.

### DIFF
--- a/modules/gst/gst.c
+++ b/modules/gst/gst.c
@@ -8,7 +8,6 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
-#include <pthread.h>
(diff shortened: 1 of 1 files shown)


alfredh:
this looks like a bugfix to me ?

should we merge it now and include it in the 1.1.0 release ?

cspiel1:
Absolutely! We had this deadlock after 200-400 iterations at the pthread_join() which could be removed in this PR.

alfredh:
great stuff, thanks!

Issue #1357, opened 2021-04-15
/tlsissuer: unable to get certificate issuer (Invalid argument)


alfredh:
tested with commit 2987b4e4198b70ba75d8cad0d90e806914055e57 on OSX 10.14
OpenSSL 1.1.1k 

```
/tlsissuer
unable to get certificate issuer (Invalid argument)
```

cspiel1:
Will have a look on this tomorrow.

juha-h:
If I remember correctly, it printed certificate during call, but not
that of the server.

juha-h:
I checked that managed to get these:
```
/tlsissuer
TLS Cert Issuer:  CN=CAcert Class 3 Root,OU=http://www.CAcert.org,O=CAcert Inc.
/tlssubject
TLS Cert Subject:  CN=test.tutpro.com
```
But what is the point, because the same info can be got using openssl command?  It would be much more useful to get tls info of the server.

cspiel1:
You are right. It prints the information of the certificate configured with `sip_certificate`. Right, a TLS info of the peer certificate during a call would be useful. We could
- add new commands to print this
- or rewrite the two commands. I have to check if, and how we are using already these commands. Most likely no one else uses them so far, because they where added not long ago.

alfredh:
in my case, the sip_certificate option was empty.

my feedback is, that in such a case it should print a more userfriendly error-message
instead of EINVAL ... :)

alfredh:
fixed in 2983f38df084c02c41af11922fc089932512f3f0

PR #1359, opened 2021-04-16
menu: print useful responses for tlsissuer and tlssubject (#1357)

cspiel1:
This PR depends on libre PR: https://github.com/baresip/re/pull/94

### DIFF
--- a/modules/menu/static_menu.c
+++ b/modules/menu/static_menu.c
@@ -1154,11 +1154,20 @@ static int cmd_tls_issuer(struct re_printf *pf, void *unused)
 		return ENOMEM;
 
 	err = tls_get_issuer(uag_tls(), mb);
-	if (err){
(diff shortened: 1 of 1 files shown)


alfredh:
with this patch, I still get the same error message

perhaps it is better to check that the sip certificate has been configured ?

```c
if (str_isset(config.sip.cert)) {
    tls_get_issuer();
}
else {
    re_printf("no local client certificate\n");
}
```

cspiel1:
> with this patch, I still get the same error message
> 
> perhaps it is better to check that the sip certificate has been configured ?
> 
> ```c
> if (str_isset(config.sip.cert)) {
>     tls_get_issuer();
> }
> else {
>     re_printf("no local client certificate\n");
> }
> ```

First the re PR has to be merged. Sorry forgot to mention in description. 
For the re PR the retest PRs have to be merged first.

Issue #1358, opened 2021-04-15
ctrl_dbus: ignoring return value of write


alfredh:
Ubuntu 20.04 with gcc 9.3.0

```make
$ make RELEASE=yes

modules/ctrl_dbus/ctrl_dbus.c: In function âon_handle_invokeâ:
modules/ctrl_dbus/ctrl_dbus.c:160:2: error: ignoring return value of âwriteâ, declared with attribute warn_unused_result [-Werror=unused-result]
  160 |  write(st->fd[1], buf, sizeof(buf));
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
modules/ctrl_dbus/ctrl_dbus.c: In function âcommand_handlerâ:
modules/ctrl_dbus/ctrl_dbus.c:142:2: error: ignoring return value of âreadâ, declared with attribute warn_unused_result [-Werror=unused-result]
  142 |  read(st->fd[0], buf, sizeof(buf));
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
modules/ctrl_dbus/ctrl_dbus.c: At top level:
cc1: error: unrecognized command line option â-Wno-shorten-64-to-32â [-Werror]
cc1: all warnings being treated as errors
make: *** [mk/mod.mk:49: build-x86_64/modules/ctrl_dbus/ctrl_dbus.o] Error 1
```

cspiel1:
will fix this tomorrow.

alfredh:
fixed in b62f9fc0871684f1a94c1e7da266cc67ac561c9a

PR #1360, opened 2021-04-16
ctrl_dbus: ensure save synchronization (#1358)

cspiel1:
- The dbus call to invoke writes exactly one byte.
- This wakes command_handler() which reads exactly one byte.
- If byte could not be written, do not wait for the command_handler().

This constraints should avoid deadlocks.

### DIFF
--- a/modules/ctrl_dbus/ctrl_dbus.c
+++ b/modules/ctrl_dbus/ctrl_dbus.c
@@ -107,6 +107,7 @@ static void command_handler(int flags, void *arg)
 {
 	struct ctrl_st *st = arg;
 	char buf[1];
+	ssize_t n = 0;
 
 	if (!st->command)
 		goto out;
@@ -140,8 +141,16 @@ static void command_handler(int flags, void *arg)
 out:
 	pthread_mutex_lock(&st->wait.mutex);
 	pthread_cond_signal(&st->wait.cond);
-	(void)read(st->fd[0], buf, sizeof(buf));
+	n = read(st->fd[0], buf, sizeof(buf));
+
 	pthread_mutex_unlock(&st->wait.mutex);
+
+	if (n != 1) {
+		warning("ctrl_dbus: detected a pipe error during read\n");
+		info("ctrl_dbus: stopping here\n");
+		st->run = false;
+		g_main_loop_quit(st->loop);
+	}
 }
 
 
@@ -310,7 +319,6 @@ static void *thread(void *arg)
 	struct ctrl_st *st = arg;
 	int err;
 
-	st->run = true;
 	if (pipe(st->fd) == -1) {
 		warning("ctrl_dbus: could not create pipe (%m)\n", errno);
 		return NULL;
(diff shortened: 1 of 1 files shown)


modules/ctrl_dbus/ctrl_dbus.c
```diff
@@ -137,9 +137,10 @@ static void command_handler(int flags, void *arg)
 	mbuf_set_pos(st->mb, 0);
 	st->command = mem_deref(st->command);
 
+out:
 	pthread_mutex_lock(&st->wait.mutex);
 	pthread_cond_signal(&st->wait.cond);
-	read(st->fd[0], buf, sizeof(buf));
+	(void)read(st->fd[0], buf, sizeof(buf));
```
alfredh, line 143: if read fails, the content of `buf` will be uninitialized.

it is probably good to init the buffer and to check the return value of read.

modules/ctrl_dbus/ctrl_dbus.c
```diff
@@ -152,19 +153,29 @@ on_handle_invoke(DBusBaresip *interface,
 {
 	char *response;
 	struct ctrl_st *st = arg;
-	char buf[8] = {0,0,0,0,0,0,0,1};
+	char buf[1] = {1};
+	size_t n;
```
alfredh, line 157: write returns type `ssize_t` (signed)

modules/ctrl_dbus/ctrl_dbus.c
```diff
@@ -152,19 +153,29 @@ on_handle_invoke(DBusBaresip *interface,
...
+		pthread_cond_wait(&st->wait.cond, &st->wait.mutex);
+
 	pthread_mutex_unlock(&st->wait.mutex);
 
-	mbuf_strdup(st->mb, &response, mbuf_get_left(st->mb));
-	dbus_baresip_complete_invoke(interface, invocation, response);
-	mem_deref(response);
-	st->mb = mem_deref(st->mb);
+	if (st->mb) {
+		mbuf_strdup(st->mb, &response, mbuf_get_left(st->mb));
```
alfredh, line 169: also check return value here ...

alfredh:
I tested again on Ubuntu 20.04 with "make RELEASE=yes".

one warning has been fixed, but one warning is still there.

modules/ctrl_dbus/ctrl_dbus.c
cspiel1: ok. Will print a warning if `mbuf_strdup()` fails and return an empty string for dbus call.

cspiel1:
Ah, I see. Thanks! Will also handle the return value of `read()`.

cspiel1:
In case of a read failure which should never happen, the module should simply stop the thread and give up.

Additionally played around with
```
/rmmod ctrl_dbus
/insmod ctrl_dbus
```
Reloading does not work due to some strange Glib errors.
```
(baresip:667026): GLib-GObject-WARNING **: 12:25:40.681: cannot register existing type 'DBusBaresip'
```
Should we create an issue? Seams to be complex.

Issue #1074, opened 2020-08-29
Release baresip 1.0.0


alfredh:
We are planning to make a new release of baresip.

The exact date is not set yet, but should be when the code is properly tested
and when it has no known critical bugs.

Please help with testing!

### Platforms

| Platform | Arch | Compiler | OpenSSL | Status |
| --- | --- | --- | --- | --- |
| Android r21d | armeabi-v7a and arm64-v8a | clang | ? | ok (thanks @juha-h ) |
| CentOS 8.2 | x86_64 | gcc 8.3.1 | OpenSSL 1.1.1c-fips | ok |
| Debian 9 | x86_64 | gcc 6.3.0 | OpenSSL 1.1.0l | ok |
| Debian 10 | x86_64 | gcc 8.3.0 | OpenSSL 1.1.1d | ok |
| Fedora 32 | x86_64 | ? | ? | ok |
| OpenBSD 6.6 | x86_64 | gcc 4.2.1 | LibreSSL 3.0.2 | ok |
| OSX 10.12 | x86_64 | clang 9.0.0 | OpenSSL 1.1.1g | ok |
| Ubuntu 20.04 | x86_64 | gcc 9.3.0 | OpenSSL 1.1.1f | ok |
| Win10 (mingw) | x86_64 | gcc 5.3.1 | OpenSSL 1.1.0e | ok (thanks @nicotyze ) |

### SIP Servers

| Provider | SIP Server | Test service | Status |
| --- | --- | --- | --- |
| iptel.org | ser 3.3.0-pre1 | music@iptel.org | ok |
| antisip.com | kamailio 5.2.2 | echo@antisip.com | ok  |
| avm | ? | avm-HD-test <sip:550@hd-telefonie.avm.de> | ok |

### SIP Endpoints

| SIP Endpoint | Audio | Video | Status |
| --- | --- | --- | --- |
| Linphone OSX v4.1.1 | opus | h264 |  |
| Chrome 85.0.4183.83 | opus | vp8 |  ok |
| Firefox 80.0 | opus | vp8 | ok |
| Cisco SX10 | opus | h264 |  ok |
| Cisco C20 | pcmu | h264 |  ok |
| FreeSWITCH Version 1.10.2-release-13 | ? | ? | ok |

### Modules:

| Module | Status |
| ---    | ---    |
| aac | ok |
| account | ok |
| alsa | ok |
| amr |  ok |
| aptx | ok |
| aubridge | ok |
| audiounit | ok |
[...]

PR #1367, opened 2021-04-17
ci: add extra checks

sreimers:


mbattista:
How do you think about even more checks with different environments?

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on
Offers mac and windows, which could make future releases simpler (https://github.com/baresip/baresip/issues/1074)

alfredh:
this looks great!

should we merge it to master for the 1.1.0 release ?

.github/workflows/extra.yml
```diff
@@ -0,0 +1,61 @@
...
+
+    - name: make HAVE_PTHREAD= baresip
+      run: |
+        make V=1 CCACHE= EXTRA_CFLAGS=-Werror HAVE_PTHREAD=
+        make clean
+
+    - name: make RELEASE=1 baresip
+      run: |
+        make V=1 CCACHE= EXTRA_CFLAGS=-Werror RELEASE=1
+        make clean
```
cspiel1, line 61: no newline at end of file

cspiel1: yes, looks great! Will not hurt in 1.1.0.

Issue #96, opened 2015-12-18
Gstreamer 1.0 audio source spamming console


nshopik:
I've tried to build baresip with Gstreamer 1.0 and found out that gstreamer spamming console instead updating same line like it was in 0.10. Here is how console output looks like when I generate sound from file. 
Building 0.4.16 on ubuntu 15.10 with libgstreamer1.0-dev  1.6.0-1

```
  handoff handler/0 (bit/s)
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler/0 (bit/s)
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler/0 (bit/s)
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler/0 (bit/s)
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler/0 (bit/s)
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler/0 (bit/s)
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler/0 (bit/s)
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler/0 (bit/s)
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler4000/54186 (bit/s)
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler4000/54186 (bit/s)
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler4000/54186 (bit/s)
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler
    format: rate=8000, channels=1, width=16, sign=1
  handoff handler
[...]

PR #1376, opened 2021-04-19
Tls sipcert per acc

cHuberCoffee:
This PR depends on [#96](https://github.com/baresip/re/pull/96)

It allows the user to specify different client certificates for different
accounts in the accounts file.

SIP proxies environments may not allow installing new trusted certificates. In 
the same time they may have a certificate installed which is used to issue
new SIP certificates. For a single server it is no problem, since a user could
configure the server generated certificate as the SIP_certificate in the
config file. Having multiple such servers (maybe as fallback servers)
leads to a problem where baresip is currently not able to handle a variety of
client certificates.

This PR overcomes exactly the described problem.
In case a user has more than one account on the same SIP proxy. The connection
TLS handshake is done once. Therefore the client certificate check of the
server is done once. Configuring different client certificates for the same 
server configuration makes no sense and should be avoided.

Since the certificate option in the accounts file is optional, baresip uses
the SIP_certificate of the config file as fallback.

### DIFF
--- a/docs/examples/accounts
+++ b/docs/examples/accounts
@@ -15,6 +15,7 @@
 #    ;auth_user=username
 #    ;auth_pass=password
 #    ;call_transfer=no
+#    ;cert=cert.pem
 #    ;mediaenc={srtp,srtp-mand,srtp-mandf,dtls_srtp,zrtp}
 #    ;medianat={stun,turn,ice}
 #    ;mwi=no
--- a/src/ua.c
+++ b/src/ua.c
@@ -977,6 +977,7 @@ int ua_alloc(struct ua **uap, const char *aor)
 	if (err)
 		goto out;
 
+
 	/* generate a unique contact-user, this is needed to route
 	   incoming requests when using multiple useragents */
 	err = re_sdprintf(&ua->cuser, "%r-%p", &ua->acc->luri.user, ua);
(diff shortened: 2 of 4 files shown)


docs/examples/accounts
```diff
@@ -15,6 +15,7 @@
 #    ;auth_user=username
 #    ;auth_pass=password
 #    ;call_transfer=no
+#    ;cert=cert.pem
```
alfredh, line 18: what kind of certificate is this ?

is it Client cert, or Server cert ?

src/ua.c
```diff
@@ -977,6 +977,7 @@ int ua_alloc(struct ua **uap, const char *aor)
 	if (err)
 		goto out;
 
+
```
alfredh, line 980: this is an unrelated change ...

src/ua.c
```diff
@@ -1000,6 +1001,18 @@ int ua_alloc(struct ua **uap, const char *aor)
 			  ua->acc->menc->id);
 	}
 
+	if (ua->acc->cert) {
+		err = sip_transp_add_ccert(uag.sip,
```
alfredh, line 1005: does ccert mean client certificate ?

alfredh:
if this patch adds support for per-account certificates, should we then consider to
remove support for global certificate ?

having only one way to do things makes it easier to use and reduces the codesize.

juha-h:
Alfred E. Heggestad writes:

> if this patch adds support for per-account certificates, should we
> then consider to remove support for global certificate ?

I would prefer to keep the global one for backwards compatibility.
Also, it can be considered as the device certificate.

May be config certificate could be considered as an account certificate
for those accounts that don't have one.

alfredh:
@sreimers can you please merge the re patch ?

sreimers:
> @sreimers can you please merge the re patch ?

Done

PR #1385, opened 2021-04-20
dtls_srtp: use elliptic curve cryptography

cHuberCoffee:
The used curve is configured in the config file via "dtls_srtp_use_ec".

### DIFF
--- a/modules/dtls_srtp/dtls_srtp.c
+++ b/modules/dtls_srtp/dtls_srtp.c
@@ -474,6 +474,10 @@ static struct menc dtls_srtp = {
 static int module_init(void)
 {
 	struct list *mencl = baresip_mencl();
+	struct pl plec = PL("");
+	char *ec = NULL;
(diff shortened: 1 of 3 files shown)


modules/dtls_srtp/dtls_srtp.c
```diff
@@ -483,11 +487,31 @@ static int module_init(void)
 		return err;
 	}
 
-	err = tls_set_selfsigned_rsa(tls, "dtls@baresip", 2048);
-	if (err) {
-		warning("dtls_srtp: failed to self-sign certificate (%m)\n",
-			err);
-		return err;
+	(void)conf_get(conf_cur(), "dtls_srtp_use_ec", &plec);
+	if (pl_isset(&plec)) {
```
alfredh, line 491: you can also use `conf_get_str` to get the config item directly into a string.

alfredh:
it would be interesting to know how the EC has been tested

PR #1391, opened 2021-04-21
config: enable sip_verify_server if sip_cafile or sip_capath is set

cHuberCoffee:
in case no sip_cafile or sip_capath is configured sip_verify_server is false
the sip_verify_server option in the config file overrules the
decision via sip_cafile and sip_capath.
the standard config  shows the possibility of a sip_verify_server
option but it is not active


cHuberCoffee:
is replaced by #1392

PR #1393, opened 2021-04-21
ToS for video and sip

cspiel1:
Allows to configure ToS differently for RTP video/audio and SIP

Was not sure which default value we should use for SIP.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -13,6 +13,7 @@ poll_method		epoll		# poll, select, epoll ..
 sip_cafile		/etc/ssl/certs/ca-certificates.crt
 #sip_trans_def		udp
 sip_verify_server	yes
+sip_tos			104
 
 # Call
 call_local_timeout	120
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -48,7 +48,7 @@ videnc_format		yuv420p
 
 # AVT - Audio/Video Transport
 rtp_tos			184
-rtpv_tos		104
+rtp_video_tos		104
 #rtp_ports		10000-20000
 #rtp_bandwidth		512-1024 # [kbit/s]
 rtcp_mux		no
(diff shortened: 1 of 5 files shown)


docs/examples/config
```diff
@@ -47,6 +48,7 @@ videnc_format		yuv420p
 
 # AVT - Audio/Video Transport
 rtp_tos			184
+rtpv_tos		104
```
alfredh, line 51: perhaps `v` should be written as `video` instead ? it is easier to understand

docs/examples/config
```diff
@@ -13,6 +13,7 @@ poll_method		epoll		# poll, select, epoll ..
 sip_cafile		/etc/ssl/certs/ca-certificates.crt
 #sip_trans_def		udp
 sip_verify_server	yes
+sip_tos			104
```
alfredh, line 16: is TOS for SIP really needed ?

TOS/QoS is mainly about the real-time media traffic that gets higher priority in a network.

alfredh:
perhaps the TOS field bit-fields can be expanded and documented in the sample config ?

alfredh:
@sreimers can you please merge the re patch ?

docs/examples/config
cspiel1: I think so. It seems that different manufacturer also set it for SIP with different default values. https://kb.smartvox.co.uk/ip-phones/sip-configuration/voip-qos-settings-part-2/

We need it also for our products.

cspiel1:
Added a commit for TOS field explanation. Tried to understand the values for "precedence" field. The answer seems not to be simple enough to put it there. A lot of RFC upgrades. Useful values for the whole ToS byte are listed here: https://linuxreviews.org/Type_of_Service_(ToS)_and_DSCP_Values

cspiel1:
rebased to master

PR #1395, opened 2021-04-22
account: add accounts parameter to force media address family

cspiel1:
```diff
+#    ;mediaaf={ipv4,ipv6}
```

This makes most sense for local accounts (regint=0). The logic is already in ua.c. This PR adds only the setting in accounts file.

### DIFF
--- a/src/account.c
+++ b/src/account.c
@@ -177,6 +177,11 @@ static int af_decode(struct account *acc, const struct pl *prm)
 	acc->maf = !pl_strcasecmp(&pl, "ipv6") ? AF_INET6 :
 		   !pl_strcasecmp(&pl, "ipv4") ? AF_INET : AF_UNSPEC;
 
+	if (acc->maf == AF_UNSPEC) {
+		warning("account: invalid address family '%r'\n", &pl);
+		return EINVAL;
+	}
+
 	return 0;
 }
 
--- a/src/account.c
+++ b/src/account.c
@@ -483,6 +483,7 @@ int account_alloc(struct account **accp, const char *sipaddr)
 	if (!acc)
 		return ENOMEM;
 
+	acc->maf = AF_UNSPEC;
 	err = str_dup(&acc->buf, sipaddr);
 	if (err)
 		goto out;
--- a/src/account.c
+++ b/src/account.c
@@ -791,16 +791,13 @@ int account_set_stun_pass(struct account *acc, const char *pass)
  *
  * @param acc      User-Agent account
  * @param mediaaf  Media address family
- *
- * @return 0 if success, otherwise errorcode
  */
-int account_set_mediaaf(struct account *acc, int mediaaf)
+void account_set_mediaaf(struct account *acc, int mediaaf)
 {
 	if (!acc)
-		return EINVAL;
+		return;
 
 	acc->maf = mediaaf;
-	return 0;
 }
 
 
--- a/src/account.c
+++ b/src/account.c
@@ -1430,7 +1430,7 @@ static const char *dtmfmode_str(enum dtmfmode mode)
  *
  * @param acc User-Agent account
  *
- * @return The address family
+ * @return Address family
  */
 int account_mediaaf(const struct account *acc)
 {
--- a/src/account.c
+++ b/src/account.c
@@ -167,6 +167,7 @@ static int extra_decode(struct account *acc, const struct pl *prm)
 static int af_decode(struct account *acc, const struct pl *prm)
 {
 	struct pl pl;
+	int maf;
 
 	if (!acc || !prm)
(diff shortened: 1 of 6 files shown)


juha-h:
There already exists ua_set_media_af() that can be used to set preferred media af.  What is the rationale for forcing it?

cspiel1:
Yes, this PR uses `ua_set_meda_af()`. We need config for that, per account.

juha-h:
Christian Spielberger writes:

> Yes, this PR uses `ua_set_media_af()`. We need config for that, per
> account.

In my Android app, there is per account config option "Prefer IPv6 for
Media" implemented using existing API functions.

juha-h:
I checked and turned out that I store "Prefer IPv6 for Media" in account's extra parameter. So it may be a good idea to have a dedicated parameter for it.

juha-h:
... but then there also needs to be a function to set the parameter.

cspiel1:
It needs no setter/getter. Because in ua.c the `ua->acc->maf` can be accessed. But if you prefer to have one, I could add it. I already added a getter.

cspiel1:
Also moved the getter function to the other account_media...() functions and adjusted its doxygen.

juha-h:
Christian Spielberger writes:

> It needs no setter/getter. Because in ua.c the `ua->acc->maf` can be
> accessed. But if you prefer to have one, I could add it. I already
> added a getter.

It must be possible to set account parameters on the fly, not just when
baresip is started.  struct account is defined in core.h, not in
baresip.h.

Two comment edits:

Get the preferred address for media of an account =>
  Get the preferred address family for media of an account

# Force IPv6 for media =>
  # Prefer IPv6 for media

src/account.c
```diff
@@ -138,6 +138,24 @@ static int extra_decode(struct account *acc, const struct pl *prm)
...
+	struct pl pl;
+
+	if (!acc || !prm)
+		return EINVAL;
+
+	if (msg_param_decode(prm, "mediaaf", &pl))
+		return 0;
+
+	acc->maf = !pl_strcasecmp(&pl, "ipv6") ? AF_INET6 :
+		   !pl_strcasecmp(&pl, "ipv4") ? AF_INET : AF_UNSPEC;
```
alfredh, line 153: perhaps a bogus string such as `ipv5` should give a parse error instead ?

src/account.c
```diff
@@ -729,6 +748,24 @@ int account_set_stun_pass(struct account *acc, const char *pass)
...
 
+/**
+ * Set the preferred media address family of a SIP account
+ *
+ * @param acc      User-Agent account
+ * @param mediaaf  Media address family
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int account_set_mediaaf(struct account *acc, int mediaaf)
```
alfredh, line 759: consider returning `void` as the function cannot fail

src/ua.c
```diff
@@ -983,6 +983,7 @@ int ua_alloc(struct ua **uap, const char *aor)
 	if (err)
 		goto out;
 
+	ua_set_media_af(ua, ua->acc->maf);
```
alfredh, line 986: now we have duplicated info in both ua and account about `mediaaf`.

perhaps consider to remove it from ua and store it only in `struct account` ?

alfredh:
one more comment:

consider adding this option:

```
;mediaaf=auto

auto -- use same AF as SIP registration
```

alfredh:
unrelated comment:

in theory the whole `struct account` class can be replaced by a general dictionary.
example:


```json
{
  "aor" : "sip:username@domain.com".
  "regint" : 3600,
  "auth_pass" : "xxxxxx"
}
```

the dictionary can be read directly from a json file.

cspiel1:
> one more comment:
> 
> consider adding this option:
> 
> ```
> ;mediaaf=auto
> 
> auto -- use same AF as SIP registration
> ```

Ok. Thus, this is like unspecified.

cspiel1:
> unrelated comment:
> 
> in theory the whole `struct account` class can be replaced by a general dictionary.
> example:
> 
> ```json
> {
>   "aor" : "sip:username@domain.com".
>   "regint" : 3600,
>   "auth_pass" : "xxxxxx"
> }
> ```
> 
> the dictionary can be read directly from a json file.

Indeed. But this would mean replacing C-API by usage conventions or a table of supported dictionary keys.

src/ua.c
cspiel1: Right, removing from ua seems to be the right solution. Although it changes the API.

juha-h:
Two comments on the PR code:

1) account module account_write_template() does not include the parameter
2) account_set_mediaaf does not check if af value is valid

PR #1399, opened 2021-04-22
Support for s16 playback in jack. Needed for play tones

srperens:
For busy tones and menu command "/play" the normal resample/reformat routine in src/audio.c is not used, therefore these commands do not work with jack since the files (busy.wav for example) are s16 and jack normally needs floats. This PR fixes that and allows jack playback to receive s16 directly; it can resample (up/down) and convert to floats. 

I have tried this with mono and stereo files in 2, 8, 16, 48, 96, 192 kHz files with valgrind without errors. (jack server running at 48kHz)

### DIFF
--- a/modules/jack/jack_play.c
+++ b/modules/jack/jack_play.c
@@ -21,6 +21,11 @@ struct auplay_st {
 	jack_client_t *client;
 	jack_port_t **portv;
 	jack_nframes_t nframes;       /* num frames per port (channel) */
+
+	struct auresamp *resamp;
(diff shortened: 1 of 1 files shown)


alfredh:
thanks Per!

would you consider to help with maintaining the jack module:

regarding sample-rates and sample-formats, I think we should try to build
a more flexible audio pipeline with aufilt modules for resampling and conversion.

example:

```
[ausrc] ---> [ auconv ] ---> [ libsamplerate ] ---> [core]
```

PR #1401, opened 2021-04-24
Check that account ;sipnat param has valid value

juha-h:


alfredh:
./src/account.c:597: else: ending if bracket should be on previous line
5

alfredh:
thanks! merged.

PR #1403, opened 2021-04-27
multicast: add source and player API calls

cHuberCoffee:
This adds some API calls which are useful for the exchangeable multicast source/player.

### DIFF
--- a/modules/multicast/player.c
+++ b/modules/multicast/player.c
@@ -565,10 +565,11 @@ void mcplayer_stop(void)
 /**
  * Initialize everything needed for the player beforhand
  *
+ * @return 0 if success, otherwise errorcode
  */
-void mcplayer_init(void)
+int mcplayer_init(void)
 {
-	return;
+	return 0;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


modules/multicast/player.c
```diff
@@ -560,3 +560,22 @@ void mcplayer_stop(void)
 {
 	player = mem_deref(player);
 }
+
+
+/**
+ * Initialize everything needed for the player beforhand
+ *
+ */
+void mcplayer_init(void)
```
alfredh, line 569: in general an init function should return `int` (errno) because it can potentially fail.

alfredh:
this patch adds some empty functions ...

what is the purpose of that ?

cHuberCoffee:
The multicast source and player are exchangeable. It may be the case that somebody want to implement its own source and player. The API calls are useful if somebody want to initialize one or both with something which must be accessible right at the beginning of the lifetime of the module. The standard implementation does not need it but i want to open the support for such sources and players.

cspiel1:
Replacing of player and source needs only to patch `multicast/module.mk`. The file multicast.c can stay untouched. We have such alternative player and source in one of our products.

PR #1406, opened 2021-04-27
Account media af template

cspiel1:


juha-h:
In my baresip Android app, I'm also using value AF_UNSPEC, when you does not want to prefer any protocol. Is that value not supported anymore?

cspiel1:
Oh, yes. I forgot this for the check.

juha-h:
Looks good now.

PR #1411, opened 2021-04-29
mixminus: add conference feature

sreimers:


cspiel1:
- RFCs? https://tools.ietf.org/html/rfc4353
- Mixer: mixausrc mixes encode/decode stream with an audio file. Maybe it can help to find some ideas for a conference mixer.

src/audio.c
```diff
@@ -2562,6 +2563,38 @@ void audio_set_hold(struct audio *au, bool hold)
...
+/**
+ * Is audio on conference?
+ *
+ * @param au    Audio object
+ *
+ * @return true if on conference, false if not
+ */
+bool audio_is_conference(const struct audio *au)
+{
+	return au->conference ? true : false;
```
alfredh, line 2594: suggested code:

```c
return au ? au->conference : false;
```

modules/mixminus/mixminus.c
```diff
@@ -0,0 +1,401 @@
...
+				return EINVAL;
+			}
+		}
+		else {
+			aubuf_read_samp(mix->ab, sampv_mix, af->sampc);
+		}
+
+		for (i = 0; i < af->sampc; i++) {
+			sampv[i] = sampv[i] + sampv_mix[i];
+		}
```
alfredh, line 277: did you consider saturation of added samples here ?

the range is 16-bit signed, if two added samples have e.g. value 32000 and 32000,
the sum will be 64000 and outside the range.

it is possible to add the samples using a 32-bit integer and then soft clip at +/- 32768.

src/call.c
```diff
@@ -1243,6 +1243,25 @@ int call_hold(struct call *call, bool hold)
...
+ */
+int call_conference(struct call *call, bool conference)
+{
+	struct audio *au = call_audio(call);
+
+	if (!au)
+		return EINVAL;
+
+	return audio_set_conference(au, conference);
+}
```
alfredh, line 1262: also consider using the `call_audio` function to fetch the audio object from the call object.
then there is no need for an explicit wrapper function.

alfredh:
did you consider using the `aumix` module in rem:

```c
int aumix_alloc(struct aumix **mixp, uint32_t srate,
		uint8_t ch, uint32_t ptime);
```

Btw, when you mention "High Latency" how many milliseconds/seconds are we talking about ?

modules/mixminus/mixminus.c
sreimers: Yes I`m aware of this, still in proof of concept stage, will fix this.

sreimers:
> did you consider using the `aumix` module in rem:
> 
> ```c
> int aumix_alloc(struct aumix **mixp, uint32_t srate,
> 		uint8_t ch, uint32_t ptime);
> ```

Had a look at it, but I think its not flexible enough and adds a extra thread per call to keep in sync.


> Btw, when you mention "High Latency" how many milliseconds/seconds are we talking about ?

I have fixed latency and syncing problems now.

alfredh:
should we update the patch to use the original aufilt api ?

looks like there is an error from the build:

```
modules/mixminus/mixminus.c:456:18: error: nested extern declaration of ‘baresip_au_pre_filtl’ [-Werror=nested-externs]
516
```

PR #1420, opened 2021-05-01
Update menu.c

alfredh:
the second parameter is an index for AORs.

with the current code it is possible to write "abc" and get ua at index 0.

the proposal is to check for numerical index

@cspiel1

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -662,7 +662,7 @@ struct ua   *menu_ua_carg(struct re_printf *pf, const struct cmd_arg *carg,
 	if (ua)
 		return ua;
 
-	err = re_regex(carg->prm, str_len(carg->prm), "[^ ]+ [^ ]+", word1,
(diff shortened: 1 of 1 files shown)


modules/menu/menu.c
```diff
@@ -662,7 +662,7 @@ struct ua   *menu_ua_carg(struct re_printf *pf, const struct cmd_arg *carg,
 	if (ua)
 		return ua;
 
-	err = re_regex(carg->prm, str_len(carg->prm), "[^ ]+ [^ ]+", word1,
+	err = re_regex(carg->prm, str_len(carg->prm), "[^ ]+ [0-9]+", word1,
```
cspiel1, line 665: Looks good.

And maybe we should rename parameter `word2` to `uaid`. Or should we remove it? It is used only internally.

cspiel1: Can we be sure that `word1` is set correctly if no second word is specified?

alfredh:
thanks for the review, lets make a new PR for this

PR #1421, opened 2021-05-01
menu: add command /uareg

alfredh:
add `/uareg` command register/unregister useragent.

examples:

```
/uareg 0
sip:alfredh@sip.antisip.com: selected for request
unregistering sip:alfredh@sip.antisip.com
...
/uareg 300
sip:alfredh@sip.antisip.com: selected for request
registering sip:alfredh@sip.antisip.com with interval 300 seconds
alfredh@sip.antisip.com: (prio 0) {0/TCP/v4} 200 OK (kamailio (5.4.4 (x86_64/linux))) [1 binding]
```

@jose1711

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -655,23 +655,34 @@ struct ua   *menu_ua_carg(struct re_printf *pf, const struct cmd_arg *carg,
 		struct pl *word1, struct pl *word2)
 {
 	int err;
-	struct le *le;
+	struct le *le = NULL;
 	uint32_t i;
 	struct ua *ua = carg->data;
 
 	if (ua)
 		return ua;
 
-	err = re_regex(carg->prm, str_len(carg->prm), "[^ ]+ [^ ]+", word1,
-			word2);
+	err = re_regex(carg->prm, str_len(carg->prm), "[^ ]+[ ]*[0-9]*",
+		       word1, NULL, word2);
 	if (err)
 		return NULL;
 
-	i = pl_u32(word2);
+	if (pl_isset(word2)) {
+		i = pl_u32(word2);
 
-	le = uag_list()->head;
-	while (le && i--)
-		le = le->next;
+		le = uag_list()->head;
+		while (le && i--)
+			le = le->next;
+	}
+	else {
+		if (list_count(uag_list()) == 1) {
+			info(">> SINGLE UA <<\n");
+			le = list_head(uag_list());
+		}
+		else {
+			info("ambiguous result.\n");
+		}
+	}
 
 	if (le) {
 		ua = le->data;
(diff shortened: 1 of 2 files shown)


modules/menu/menu.c
```diff
@@ -655,23 +655,34 @@ struct ua   *menu_ua_carg(struct re_printf *pf, const struct cmd_arg *carg,
...
 	uint32_t i;
 	struct ua *ua = carg->data;
 
 	if (ua)
 		return ua;
 
-	err = re_regex(carg->prm, str_len(carg->prm), "[^ ]+ [^ ]+", word1,
-			word2);
+	err = re_regex(carg->prm, str_len(carg->prm), "[^ ]+[ ]*[0-9]*",
+		       word1, NULL, word2);
```
cspiel1, line 666: Ok, looks better than the other PR. Maybe we should remove the pointer `word2 ` in the parameter list of this function, because it is not needed.

modules/menu/menu.c
```diff
@@ -655,23 +655,34 @@ struct ua   *menu_ua_carg(struct re_printf *pf, const struct cmd_arg *carg,
...
+			le = le->next;
+	}
+	else {
+		if (list_count(uag_list()) == 1) {
+			info(">> SINGLE UA <<\n");
+			le = list_head(uag_list());
+		}
+		else {
+			info("ambiguous result.\n");
+		}
```
cspiel1, line 684: The else path will break the automatic UA selection. If no second word is given in `carg->prm` then the function has to return NULL.

modules/menu/static_menu.c
```diff
@@ -1054,6 +1054,43 @@ static int create_ua(struct re_printf *pf, void *arg)
...
+{
+	const struct cmd_arg *carg = arg;
+	struct pl word[2] = {PL_INIT, PL_INIT};
+	struct ua *ua = menu_ua_carg(pf, carg, &word[0], &word[1]);
+	struct account *acc;
+	uint32_t regint;
+	int err;
+
+	if (!ua) {
+		warning("uareg: ua not found (%s)\n", carg->prm);
```
cspiel1, line 1067: The warning is confusing, because the first word will be `regint`.
```
warning("uareg: ua index not specified or ua not found\n");
```

modules/menu/menu.c
alfredh: in re_regex a start (*) means optional (zero or more matches).

in any case the `struct pl` parameter will be initialized to PL_INIT or the match

cspiel1: > in re_regex a start (*) means optional (zero or more matches).

Yes, I know.

> in any case the struct pl parameter will be initialized to PL_INIT or the match

Good to know.

 But what I mean is that the parameter `word2` in function `menu_ua_carg()` is never used. Ok, it is not really related to this PR.

alfredh:
thanks for the review! lets make a separate commit for the menu_ua_carg function

the 2nd word can be removed from the API, and into the function

PR #1423, opened 2021-05-03
AMR Bandwidth Efficient mode support

srperens:
Basic support for AMR Bandwidth Efficient mode. Original code hade support for 'octet aligned' mode only; with this code added both modes are supported. This code changes the default proposed mode from Baresip from octet aligned to bandwidth efficient.

Interop testing is OK with:
1. MicroSIP client (which seems to run only octet-aligned') 
2. PSTN carrier supporting both modes with bandwidth efficient being default
3. Mitel hardware SIP phone
4. MX-ONE PBX

### DIFF
--- a/modules/amr/amr.c
+++ b/modules/amr/amr.c
@@ -33,9 +33,6 @@
  * This module supports both AMR Narrowband (8000 Hz) and
  * AMR Wideband (16000 Hz) audio codecs.
  *
- * NOTE: only octet-align mode is supported.
- *
- *
  * Reference:
  *
  *     http://tools.ietf.org/html/rfc4867
@@ -60,11 +57,13 @@ enum {
 struct auenc_state {
 	const struct aucodec *ac;
 	void *enc;                  /**< Encoder state            */
+	bool aligned;
 };
 
 struct audec_state {
 	const struct aucodec *ac;
 	void *dec;                  /**< Decoder state            */
+	bool aligned;
 };
 
 
@@ -117,7 +116,6 @@ static int encode_update(struct auenc_state **aesp,
 	struct auenc_state *st;
 	int err = 0;
 	(void)prm;
-	(void)fmtp;
 
 	if (!aesp || !ac)
 		return EINVAL;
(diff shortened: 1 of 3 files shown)


alfredh:
thanks!

could you please also update the Interop page with your results:

https://github.com/baresip/baresip/wiki/Interoperability


do you want to become a maintainer of the amr module:

if yes, please add your name to the module or let me know.

modules/amr/amr.c
```diff
@@ -248,7 +292,15 @@ static int decode_wb(struct audec_state *st,
 	if (fmt != AUFMT_S16LE)
 		return ENOTSUP;
 
-	IF2D_IF_decode(st->dec, &buf[1], sampv, 0);
+	if (amr_ac->aligned) {
+		IF2D_IF_decode(st->dec, &buf[1], sampv, 0);
+	}
+	else {
+		uint8_t temp[len];
```
alfredh, line 299: we should try to avoid variable-length arrays, this is considered to be insecure coding.

sreimers: Maybe we can add this with `-Wvla` as compiler warning flag? I found only one other issue:

```
modules/gst/gst.c:160:13: warning: variable length array used [-Wvla]
        int16_t buf[st->sampc];
```

alfredh: I think this is a great idea :)

thanks for fixing the gst module

srperens: The max size of len is already protected by checks for size NB_SERIAL_MAX. See rows 289 and 356. 

Should I commit change where these temp arrays are always allocated at the max allowed size, i.e. NB_SERIAL_MAX and +1 ? 

thanks

alfredh: it would better to declare a fixed-size buffer that is big enough to hold a frame.

the buffer can be located either on the stack, or in the instance.

srperens: see new commit using fixed-size buffer. tested ok.

alfredh:
Thank you :)

PR #1429, opened 2021-05-05
sip: allow sip authentification only for registered proxies

cHuberCoffee:
This PR depends on [re/101](https://github.com/baresip/re/pull/101)

### DIFF
--- a/src/sipreq.c
+++ b/src/sipreq.c
@@ -150,7 +150,8 @@ int sip_req_send(struct ua *ua, const char *method, const char *uri,
 	if (err)
 		goto out;
 
-	err = sip_auth_alloc(&sr->auth, auth_handler, ua_account(ua), true);
+	err = sip_auth_alloc(&sr->auth, uag_sip(), auth_handler,
+		ua_account(ua), true);
 	if (err)
 		goto out;
 


src/sipreq.c
```diff
@@ -150,7 +150,8 @@ int sip_req_send(struct ua *ua, const char *method, const char *uri,
 	if (err)
 		goto out;
 
-	err = sip_auth_alloc(&sr->auth, auth_handler, ua_account(ua), true);
+	err = sip_auth_alloc(&sr->auth, uag_sip(), auth_handler,
+		ua_account(ua), true);
```
alfredh, line 154: is it possible to implement this without breaking the sip_auth_alloc API ?

cspiel1: Yes, with a setter. This was also my argument at first moment.

We discussed and found that because this solves a major security problem it maybe is better that anyone who uses this API will have to fix his/her build to solve this security problem.

But if you prefer the setter, it will be also okay for us. In this case we can decide between the old behavior. So the attack still will be possible. Or if `sip` is not set, the authorization will be canceled by default in `sip_auth_authenticate()`.

Again we are discussing about "security by default" or not. ;-)

Which would you prefer?

juha-h:
Could you clarify, what this is all about?  If I have defined one or two outbound proxies for an account and use TLS transport, I assume that initial requests are sent to them and if I have asked to verify server's certificate, then it will be done.  What is the security problem?

cspiel1:
> Could you clarify, what this is all about? If I have defined one or two outbound proxies for an account and use TLS transport, I assume that initial requests are sent to them and if I have asked to verify server's certificate, then it will be done. What is the security problem?

you find the description in the re PR.

juha-h:
How about making it configurable if peer-to-peer calls are allowed at
all when an outbound proxy is defined for an account?

Also, as I suggested earlier that it should be possible to force also
all in-dialog requests via outbound proxy (like it is in twinkle).

With these two, the security issue discussed here should be gone.

-- Juha

juha-h:
I just checked and if all in-dialog requests are forced via outbound proxy, the problem is solved, since any correctly configured outbound proxy removes Proxy-Authorization headers and the attacker thus cannot access the digest.  Forcing in-dialog requests via outbound proxy most likely will also solve other security issues.

juha-h:
In Kamailio this is used:

4.1. consume_credentials()

This function removes previously authorized credential headers from the
message being processed by the server. That means that the downstream
message will not contain credentials there were used by this
server. This ensures that the proxy will not reveal information about
credentials used to downstream elements and also the message will be a
little bit shorter. The function must be called after www_authorize,
proxy_authorize, www_authenticate or proxy_authenticate.

cspiel1:
That does not solve the security issue, because you can't force the user to configure an outbound. Also peer-to-peer calls should be possible always. It is a common use-case.

We found that what the attack makes possible is that the registered UA can be selected for a peer-to-peer call because of the lines:
```
		if (0 == pl_casecmp(cuser, &acc->luri.user)) {
			ua_printf(ua, "account match for %r\n", cuser);
			return ua;
		}
```
in function `uag_find_msg()`. This match makes sense for registrar less accounts (`regint=0`).

For registered accounts this match is only useful if the SIP registrar does not use the user in the Contact header of the REGISTER for the INVITE. Thus, it is only a fallback selection for un-typical cases. This line persists since the initial commit. We should not remove or change this to `regint=0` accounts.

Anyway, the main PR is the libre PR. Where the security leak is located. It has to be solved there. The only question is if we should change the API or use a setter `sip_auth_set_sip()` like was asked by @alfredh. And my question was, if we use a setter what should be the default behavior if the setter is not called?

- Prevent authorization to be secure, or
- always send authorization like it is done now?

juha-h:
Christian Spielberger writes:

> That does not solve the security issue, because you can't force the
> user to configure an outbound. Also peer-to-peer calls should be
> possible always. It is a common use-case.

If there is no outbound proxy, how is it configured from which peers the
UAC accepts 401 Unauthorized responses and from which proxies 407 Proxy
Authentication Required responses?

cspiel1:
1.) User Agent A registers at Proxy X. (Now forget X! It is doesn't matter here.)
2.) Attacker calls User Agent A directly by a peer-to-peer call.
3.) Baresip selects User Agent A.
4.) Then the user or the application of baresip is accepting the call. After a while the call is hangup by the user.
5.) Baresip sends a BYE to the attacker.
6.) The attacker answers with a 407.
7.) Then baresip sends the Digest response to the attacker.
8.) The attacker is now able to offline crack the password.

Note 3.) The selection of the UA is not correct in this case. But the Attacker is able to manipulate the SIP-Request in order to select any UA he wants.

This is an issue found by sipvicious.

juha-h:
Christian Spielberger writes:

> 1.) User Agent A registers at Proxy X. (Now forget X! It is doesn't
> matter here.)

It should matter.  UA registers for the purpose that its SIP proxy knows
how to reach it.

Once registered the UA should only accepts incoming request using the
very same TCP/TLS connection it used for the registration.  In most
cases due to NATs and firewalls that in fact is the only possible way to
reach the UA.

If UA does not register and allows peer-to-peer incoming requests, there
must be a way to configure which ones.

cspiel1:
The configuration
```
sip_listen		0.0.0.0:5060
```
enables peer-to-peer calls from every where. We should not break this. Just solve this security leak explicitly.

alfredh:
after reading more about this, I dont think that it is the right solution for the problem.

It is the first time I have heard about this security issue, and it seems to be an ad-hoc issue
because there are no IETF RFCs that is mentioning the problem and/or a solution.

there is one thing that we have to understand first, that the Register client is not related
to other SIP components like outbound requests or authentication/challenge.

A SIP User-agent has N number of Register clients that can be:

1. Enabled (periodic registering to a Registrar)
2. Disabled

The SIP transport between the client and the server can be:

1. UDP
2. TCP
3. TLS

for TCP and TLS, the TCP connection goes between the client and the server.
Inbound TCP connection is rare for a normal user.

That leaves only the UDP transport which by nature allowes inbound connections.

a simpler and more flexible solution would be to add configuration:

1. Enable/Disable UDP transport
2. Allow inbound TCP/TLS connections (sip_listen).

Peer-to-peer calls should still be allowed and not related to register on/off.

juha-h:
Alfred E. Heggestad writes:

> a simpler and more flexible solution would be to add configuration:
> 
> 1. Enable/Disable UDP transport
> 2. Allow inbound TCP/TLS connections (sip_listen).
>

I like these config options.  I would definitely disable UDP transport,
since it is unreliable, subject to SIP request splitting to more than
one packet, requires keepalive requests to keep NAT binding open, and is
insecure (source IP address can be faked).

> Peer-to-peer calls should still be allowed and not related to register
> on/off.

Their allowance could be related to outbound proxy definition.  If
outbound proxy has been defined, I don't see need to for peer-to-peer
calls.

cHuberCoffee:
juha-h writes:
>Once registered the UA should only accepts incoming request using the
very same TCP/TLS connection it used for the registration.  In most
cases due to NATs and firewalls that in fact is the only possible way to
reach the UA.

I think this is the way to go. 

A registered UA should only answer requests from its configured server. This does not necessarily mean to allow only requests from the very same TCP/TLS connection. But allowing only connections from its configured server. Only UAs with `acc->regint=0` should accept peer-to-peer connections.

Fixing a problem like this in re is possible but out of the context of re. Therefore i close the relating PRs and come up with a baresip fix linking against this PR.

oej:
Only accepting requests on the same TCP/TLS connection breaks the standard-compliance, unless SIP Outbound is used. Has to be configurable.

I may want to have multiple TCP/TLS connections with mutual TLS certs.

juha-h:
Olle E. Johansson writes:

> Only accepting requests on the same TCP/TLS connection breaks the
> standard-compliance, unless SIP Outbound is used. Has to be
> configurable.

Yes configurable if outbound is not in use.

It should also be possible to configure not to listen at any ip/port.
Perhaps it is already using listen config variable, but I'm not sure.

PR #1432, opened 2021-05-05
Work on Intercom module

cspiel1:
call: add a setter function for answer delay

Can be used to override the SIP auto answer delay from an external application/module. E.g. Set it back to -1, which suppresses the auto answer.

The baresip-apps PR https://github.com/baresip/baresip-apps/pull/2 introduces an intercom module with privacy mode. If privacy is on, the SIP auto answer is suppressed.

TODO: override ringtones for module menu, or prevent menu from ringing for intercom calls


cspiel1:
This PR is only a setter. Think its ok if I merge this now.

Issue #1433, opened 2021-05-05
wrong account selected


juha-h:
I just noticed that my baresip Android app has got broken (using latest master).  I have four accounts and only one of them (foo@tutpro.com)  is registered.  When call comes in from test@tutpro.com to foo@tutpro.com, account foo@lab.tutpro.com gets selected:
```
05-05 21:33:17.381  7259  7373 D Baresip Lib: foo@tutpro.com: (prio 0) {1/TLS/v4} 200 OK (OpenSIPg SIP Proxy (5.4.4-13 (x86_64/linux))) [1 binding]
05-05 21:33:17.381  7259  7372 D Baresip Lib: ua event (REGISTER_OK) 200 OK
...
05-05 21:34:11.371  7259  7372 D Baresip Lib: ua event (CALL_INCOMING) sip:test@tutpro.com
05-05 21:34:11.372  7259  7373 D Baresip Lib: ua: sipsess connect via TLS 192.26.111.38:8003 --> 10.145.233.30:47057
05-05 21:34:11.372  7259  7373 D Baresip Lib: foo@lab.tutpro.com: selected
```
Why is an account that did not register the contact used in the Request-URI of the INVITE get selected? I don't remember changing anything in the app that could have caused this issue.

juha-h:
Here is another example, where test@test.tutpro.com is calling foo@test.tutpro.com and foo@lab.tutpro.com gets selected.
```
05-05 22:03:24.111  8351  8380 D Baresip Lib: TCP 192.168.43.50:37418 -> 192.168.43.160:5060
05-05 22:03:24.112  8351  8380 D Baresip Lib: REGISTER sip:test.tutpro.com SIP/2.0
05-05 22:03:24.112  8351  8380 D Baresip Lib: Via: SIP/2.0/TCP 10.145.233.30:60376;branch=z9hG4bKc4845faa117db836;rport
05-05 22:03:24.112  8351  8380 D Baresip Lib: Contact: <sip:foo-0x77296452d0@10.145.233.30:60376;transport=tcp>;expires=3600;+sip.instance="<urn:uuid:314b31b1-d978-5ef1-4135
05-05 22:03:24.112  8351  8380 D Baresip Lib: -f3305e7fe7b3>";q=0.5;reg-id=1
05-05 22:03:24.114  8351  8380 D Baresip Lib: From: "Foo" <sip:foo@test.tutpro.com>;tag=e4dd832f458c0efd
...
[...]

PR #1434, opened 2021-05-06
uag: use local cuser for selecting user-agent (#1433)

cspiel1:


juha-h:
I tested and this PR solved the issue.

PR #1435, opened 2021-05-06
Attended Transfer on GTK

mbattista:
I implemented an attended transfer for gtk.

It still has some rough edges, e.g. you have to set the call on hold manually to enable attended transfer, but it is functional (at least it worked on a `starface` telephony system.

Edit: I also fixed a bug with clean number in this merge request, which includes `ctype.h` on the `menu` and `gtk` module. It seems that either a contact formatted entry returns a errorcode from `re_regex` or I missunderstood how the function should be used.

On a sidenote, I could segfault re_regex with `int err = re_regex(str, str_len(str), "[0-9]+");` (I wanted to check "[0-9\\-\\.\\+\\(\\)"]+, but it to showcase the error, this was the shortest regex I tested) and the following string: `"Doe, John" <sip:75@192.0.2.55>`, which is why i now check the string with ctype instead.
Should I make an issue on `baresip/re` for this?

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -198,6 +198,7 @@ int  call_send_digit(struct call *call, char key);
 bool call_has_audio(const struct call *call);
 bool call_has_video(const struct call *call);
 int  call_transfer(struct call *call, const char *uri);
+int  call_replace_transfer(struct call *target_call, struct call *source_call);
 int  call_status(struct re_printf *pf, const struct call *call);
 int  call_debug(struct re_printf *pf, const struct call *call);
 int  call_notify_sipfrag(struct call *call, uint16_t scode,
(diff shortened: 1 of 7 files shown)


mbattista:
I could probably make a similar pull request for the menu module if wanted.

juha-h:
mbattista writes:

> I could probably make a similar pull request for the menu module if
> wanted.

Please do.  I have reported problem with the current transfer feature

  https://groups.google.com/g/baresip/c/NUHJeMqAyTw

and attended version would be an improvement.

alfredh:
should we move gtk module to baresip-apps ?

since gtk module is the only GUI module, I have a feeling that it does not really belong
in the main baresip repo ...

alfredh:
for `re_regex` each pattern matching must have a corresponding `struct pl *` parameter (can be NULL).

if the parameter is missing, it is a user error and not a bug in libre.
this is not possible to fix since it is using `va_args`.

PR #1437, opened 2021-05-07
menu: allow other modules to reject a call

cspiel1:
Other modules loaded before menu may also have an UA event handler. If e.g. an
incoming call is rejected there, then menu should process its
UA_EVENT_CALL_CLOSED updates instead of processing UA_EVENT_CALL_INCOMING.


negbie:
What a perfect timing. Noticed this 1 hour ago and was just about to open an issue for this.

cspiel1:
Has to be done differently. The question is how to reject the call? `call_hangup()` or `ua_hangup()`. The later also derefs the call object. Then the other event handler have a problem when accessing the call object. But somebody has to deref the call object.
 --> setting to draft

mbattista:
The GTK module uses `ua_hangup` [(GTK)](https://github.com/baresip/baresip/blob/master/modules/gtk/gtk_mod.c#L856) but since GTK gets loaded after the menu module, I do not know if this comment is of any help.

cspiel1:
Yes it helps. But the order of loading depends on the configuration. Modules menu and gtk both should be loaded with module_app. In the example gtk.so is commented out but before menu.so.

I think normally the user has the choice to use gtk or menu. It should also be possible to load and use both.

cspiel1:
The other module (gtk, intercom) should use this to reject an incoming call:
```
call_hangup(call, 406, "Not Acceptable");
ua_event(ua, UA_EVENT_CALL_CLOSED, call, "Rejected by ...");
tmr_start(tmr, 0, delayed_call_deref, call);
```

juha-h:
Christian Spielberger writes:

> The other module (gtk, intercom) should use this to reject an incoming call:
> ```
> call_hangup(call, 406, "Not Acceptable");
> ua_event(ua, UA_EVENT_CALL_CLOSED, call, "Rejected by ...");
> tmr_start(tmr, 0, delayed_call_deref, call);
> ```

My Android app has been using ua_hangup to reject an incoming call:
```
Api.ua_hangup(uap, callp, 486, "Busy Here")
```
Is that wrong thing to do and if so why?

cspiel1:
> The other module (gtk, intercom) should use this to reject an incoming call: ``` call_hangup(call, 406, "Not Acceptable"); ua_event(ua, UA_EVENT_CALL_CLOSED, call, "Rejected by ..."); tmr_start(tmr, 0, delayed_call_deref, call); ```
> My Android app has been using ua_hangup to reject an incoming call: ``` Api.ua_hangup(uap, callp, 486, "Busy Here") ``` Is that wrong thing to do and if so why?

Its ok.

mbattista:
> Yes it helps. But the order of loading depends on the configuration. Modules menu and gtk both should be loaded with module_app. In the example gtk.so is commented out but before menu.so.
> 
> I think normally the user has the choice to use gtk or menu. It should also be possible to load and use both.

You just blew my mind. It never occurred to me to disable menu, since gtk has not all options the menu has.
Since https://github.com/baresip/baresip/commit/84220df98861c91a3779b3d86985f246b446dc3c#diff-8620cc97c7678e5d452c85a56257ea700ab041ad3cb7c6ab41600fbcfd7278fa I have no problems running them in parallel.

cspiel1:
@mbattista Updated the commit, the description and the example in baresip-apps.

Baresip-apps contains a candidate for a new re-API function `mem_deref_later()`.

alfredh:
the event handler can be called many times, and if one handler deref call
then the next handler will get a corrupt object:

```
ua_event:
-> handler1
-> handler2 mem_deref(call)
-> handler3
-> handler4
```

perhaps it is better to add `call_close` that would do the same as the destructor
except destroying the object.

in this case it should be safe to pass the (closed) call object to multiple handlers

modules/menu/menu.c
```diff
@@ -379,6 +379,10 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 #endif
 
 
+	if (call_state(call) == CALL_STATE_TERMINATED &&
+			ev != UA_EVENT_CALL_CLOSED)
+		return;
+
```
alfredh, line 385: this looks like a hack. there are also events that are not call related

cspiel1:
> the event handler can be called many times, and if one handler deref call
> then the next handler will get a corrupt object:
> 
> ```
> ua_event:
> -> handler1
> -> handler2 mem_deref(call)
> -> handler3
> -> handler4
> ```
> 
> perhaps it is better to add `call_close` that would do the same as the destructor
> except destroying the object.
> 
> in this case it should be safe to pass the (closed) call object to multiple handlers

`call_hangup()` already does this. And I use it now in intercom module.
```
static int reject_call(struct call *call, uint16_t scode, const char *reason)
{
	struct ua *ua = call_get_ua(call);
	call_hangup(call, scode, reason);

	ua_event(ua, UA_EVENT_CALL_CLOSED, call, reason);
	return mem_deref_later(call);
}
```

What do you think about `mem_deref_later()`? We could put it to `re`.

modules/menu/menu.c
cspiel1: Added a commit. Check if call is set, doesn't hurt and maybe makes it clearer.

alfredh:
hi,

lets take a step back and look at the problem again.

if we look at the event handler and the code calling the event handlers:


```c
ua_event:

	/* send event to all clients */
	le = ehl.head;
	while (le) {
		struct ua_eh *eh = le->data;
		le = le->next;

		eh->h(ua, ev, call, buf, eh->arg);
	}

```


and the event handler:


```c
static void ua_event_handler(struct ua *ua, enum ua_event ev,
			     struct call *call, const char *prm, void *arg)
{

	switch (ev) {
	}
}

```

and other event handlers in other modules have similar code.

the code calling the event handlers does not know anything
about the event handlers, nor should it.

one event handler does not know anything about what the
other event handler is doing.

event is just an event, and struct call is just an object.

if the problem is that one event handler destroys the call object,
that we should fix that. we cannot pass destruced objects
around the system.

is it a problem that you have loaded both the intercom module
and the menu module ? if so, is it possible to not load menu at all ?

cspiel1:
As I tried to explain above we want menu to be the main application that handles SIP calls. I don't want to copy the menu functionality to another module. As @mbattista already stated gtk and menu also are loaded both for his application.

Another module e.g. baresip-apps/modules/intercom may reject an incoming call before menu handles the UA_EVENT_CALL_INCOMING event.

Calling `call_hangup()` to reject the call seems to fit perfectly. The object is not destroyed.

What I would expect from an event-handler that has a call object as parameter is that it should check for a call event if the call was already rejected.

cspiel1:
Instead of the proposed solution we could also add a return value to the event handler (similar to glib, qt, ...):
```
ua_event:

	/* send event to all clients */
	le = ehl.head;
	while (le) {
		struct ua_eh *eh = le->data;
		le = le->next;

		if (eh->h(ua, ev, call, buf, eh->arg))
			break;
	}
```

juha-h:
Christian Spielberger writes:

> Instead of the proposed solution we could also add a return value to
> the event handler (similar to glib, qt, ...): 
> ```c
> 	/* send event to all clients */
> 	le = ehl.head;
> 	while (le) {
> 		struct ua_eh *eh = le->data;
> 		le = le->next;
> 
> 		if (eh->h(ua, ev, call, buf, eh->arg))
> 			break;
> 	}
> ```

`struct call` has

struct ua *ua;            /**< SIP User-agent

Can that ua be null, i.e. that the call does not belong to any ua?  If
not, why would an ua to which the call does not belong to need to handle
the event?

cspiel1:
> Instead of the proposed solution we could also add a return value to the event handler (similar to glib, qt, ...): ```c /* send event to all clients */ le = ehl.head; while (le) { struct ua_eh *eh = le->data; le = le->next; if (eh->h(ua, ev, call, buf, eh->arg)) break; } ```
> `struct call` has struct ua *ua; /**< SIP User-agent Can that ua be null, i.e. that the call does not belong to any ua? If not, why would an ua to which the call does not belong to need to handle the event?

If call->ua is always set. If call is set, then ua is set.

alfredh:
it may be possible to have 2 "menu" modules loaded at the same time.
but even if it is possible, does not mean that you have to do it.

for example if you are `gtk.so` you should not have to load `menu.so`.
they are mutually exclusive.

of course, you may load both gtk and menu and see how it goes.
but we should not try to solve all possible usecases and combinations.

the source code should be clean and straightforward and easy to read.
if you have to make many exceptions and workarounds, then that is a hint
that the design is wrong.

instead of adding tweaks and hacks, we should fix the design.

is there a problem with rejecting an incoming call twice from 2 modules ?

modules/menu/menu.c
alfredh: I disagree with this. The event handler should handle the event, even if the call state is closed.

The `struct call` code has to robust enough to check its own state

alfredh:
how about this simple solution:


```diff
diff --git a/modules/menu/menu.c b/modules/menu/menu.c
index c1b6d3e..c761261 100644
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -477,6 +477,9 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 
        case UA_EVENT_CALL_INCOMING:
 
+               if (call_state(call) != CALL_STATE_INCOMING)
+                       break;
+
                /* set the current User-Agent to the one with the call */
                menu_selcall(call);
                menu_stop_play();
```

cspiel1:
Ok, I agree. It should be kept simple. Your simple solution will be enough for rejecting incoming calls in other modules. And this is the only example I have in the moment. --> I replace the commits by one with a modified commit message.

Anyway you wrote about changing the design: What do you think about the proposition here?https://github.com/baresip/baresip/pull/1437#issuecomment-842255189

PR #1442, opened 2021-05-11
gst: audio playback not correct for some WAV files.

RobertMi21:
Some WAV files are played much too fast especially if the depth is not 16 bit and the sampling rate is not 16000. 
The behavior can be tested with the file pcm0822m.wav attached:
[pcm0822m.wav.gz](https://github.com/baresip/baresip/files/6457570/pcm0822m.wav.gz)

### DIFF
--- a/modules/gst/gst.c
+++ b/modules/gst/gst.c
@@ -128,16 +128,13 @@ sync_handler(
 
 static void format_check(struct ausrc_st *st, GstStructure *s)
 {
-	int rate, channels, width;
-	gboolean sign;
+	int rate, channels;
 
 	if (!st || !s)
 		return;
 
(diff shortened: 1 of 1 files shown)


modules/gst/gst.c
```diff
@@ -147,12 +144,6 @@ static void format_check(struct ausrc_st *st, GstStructure *s)
 		warning("gst: expected %d channels (got %d)\n",
 			st->prm.ch, channels);
 	}
-	if (16 != width) {
```
RobertMi21, line 150: Not needed anymore. The audio format must be S16LE as defined in set_caps function.

alfredh:
please consider to also add support for FLOAT audio sample format

PR #1444, opened 2021-05-11
ua selection checks for UDP transport

cHuberCoffee:
Depends on [re/#102](https://github.com/baresip/re/pull/102)
Followup to #1429

We do not allow peer-to-peer connections via UDP for registered UAs neither for registrar-less accounts which does have authentication properties configured.

For any other transport protocol nothing changes.

### DIFF
--- a/src/core.h
+++ b/src/core.h
@@ -194,6 +194,7 @@ int  reg_debug(struct re_printf *pf, const struct reg *reg);
 int  reg_json_api(struct odict *od, const struct reg *reg);
 int  reg_status(struct re_printf *pf, const struct reg *reg);
 int  reg_af(const struct reg *reg);
+struct sipreg *reg_sipreg(const struct reg *reg);
 
 
 /*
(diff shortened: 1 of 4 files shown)


juha-h:
* Verify the received SIP msg IP address against the registrar of the UA.
  * This check only has to be done if the transport protocol is UDP.

If transport protocol is UDP, then source IP address can be faked. It
cannot be relied on at least in the initial request, where response is
sent to Contact URI.

cHuberCoffee:
You are right, faking the source IP address is possible. But the current behavior is no check and this PR at least adds a barrier. 
If you think the benefit is too less for the change, then I will close it.

juha-h:
Christoph Huber writes:

> You are right, faking the source IP address is possible. But the
> current behavior is no check and this PR at least adds a barrier.  
> If you think the benefit is too less for the change, then I will close
> it.

I think the benefit is not big enough to justify the amount of extra
code that would be need.

PR #1449, opened 2021-05-17
call answer media dir

cspiel1:
- In combination with re PR https://github.com/baresip/re/pull/103 fixes audio only calls. There was an RTP timeout for the video stream.
- Fixes also one-directional `acceptdir` for incoming early-video calls.


alfredh:
FYI, 

the RTP timeout checking should not be started if the stream is not active.
you can check this like this:


```c
  if (sdp_media_has_media(m->sdpm)) {
  }
```

cspiel1:
Right! Then we don't need the re PR https://github.com/baresip/re/pull/103. I will close this.

PR #1451, opened 2021-05-17
Use line number 0 if user did not provide any line number

negbie:


negbie:
@alfredh I choosed 0 so that the user won't pick line 1 accidentally. Do you think line number 1 as default would be better?

modules/menu/dynamic_menu.c
```diff
@@ -99,7 +99,11 @@ static int set_current_call(struct re_printf *pf, void *arg)
 	struct cmd_arg *carg = arg;
 	struct ua *ua = carg->data ? carg->data : menu_uacur();
 	struct call *call;
-	uint32_t linenum = atoi(carg->prm);
+	uint32_t linenum = 0;
+	
+	if (str_len(carg->prm)!=0) {
```
alfredh, line 104: it is probably better to use `str_isset` or just check the pointer

alfredh:
there is a small ccheck error in the patch:

```
./modules/menu/dynamic_menu.c:103: has trailing tab(s)
```

PR #1456, opened 2021-05-20
ctrl_dbus: generate dbus code and documentation in makefile

cspiel1:
More general usable script gen.sh


alfredh:
would it be possible to invoke gdbus-codegen directly from the makefile ?


perhaps something like this:


```diff
diff --git a/modules/ctrl_dbus/module.mk b/modules/ctrl_dbus/module.mk
index 0e4c437..b543d3d 100644
--- a/modules/ctrl_dbus/module.mk
+++ b/modules/ctrl_dbus/module.mk
@@ -16,8 +16,15 @@ $(MOD)_CFLAGS        += -Wno-unused-parameter -Wno-declaration-after-statement \
 
 $(MOD)_CCHECK_OPT      = -e baresipbus.h -e baresipbus.c
 
+modules/ctrl_dbus/ctrl_dbus.c: modules/ctrl_dbus/baresipbus.h
+modules/ctrl_dbus/baresipbus.c:        modules/ctrl_dbus/baresipbus.h
+
+
 modules/ctrl_dbus/baresipbus.h modules/ctrl_dbus/baresipbus.c: \
        modules/ctrl_dbus/com.github.Baresip.xml
-       @cd $(dir $@) && ./gen.sh
+       @echo "GEN $<"
+       @gdbus-codegen --output-directory $(dir $@) --generate-c-code baresipbus \
+               --c-namespace DBus --interface-prefix com.github. $<
+
 
 include mk/mod.mk
```

cspiel1:
This looks very good. I will try also to make a target for the DBUS documentation.

Do you have also an idea how to add baresipbus.h and baresipbus.c to the clean target?

alfredh:
this looks very good.

there is a merge conflict, could you please have a look ?

cspiel1:
Ok, did rebase. But found that path of object files was wrong. Replaced wit your solution. Then GEN is called twice. I don't know why, but it is acceptable.

alfredh:
many thanks, great stuff!

PR #1462, opened 2021-05-21
serreg: unregister correct User-Agents on registration failure

cspiel1:
For the "Serial registration mode" stop the registration not before all
accounts with same prio failed. It is best to do this in register_curprio().


alfredh:
thanks!

PR #1468, opened 2021-05-23
auframe: always set srate and ch

janh:
https://github.com/baresip/baresip/commit/1ef3a96412667581ac35547c53394426a8a261c1 added `srate` and `ch` to `auframe`. However there are several cases where the values are not actually set. If I haven't missed anything, this should fix the remaining cases.

### DIFF
--- a/modules/aubridge/device.c
+++ b/modules/aubridge/device.c
@@ -109,12 +109,14 @@ static void *device_thread(void *arg)
 		}
 
 		if (dev->ausrc->rh) {
-			struct auframe af = {
-				.fmt   = dev->ausrc->prm.fmt,
-				.sampv = sampv_in,
-				.sampc = sampc_in,
-				.timestamp = ts * 1000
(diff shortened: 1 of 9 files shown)


modules/aubridge/device.c
```diff
@@ -113,7 +113,9 @@ static void *device_thread(void *arg)
 				.fmt   = dev->ausrc->prm.fmt,
 				.sampv = sampv_in,
 				.sampc = sampc_in,
-				.timestamp = ts * 1000
+				.timestamp = ts * 1000,
+				.srate = dev->auplay->prm.srate,
+				.ch    = dev->auplay->prm.ch
 			};
```
sreimers, line 119: We should use auframe_init, its easier for future refactorings.

janh:
`auframe_init` is now used everywhere for initialization.

sreimers:
Thanks

PR #1472, opened 2021-05-28
account: fix IPv6 only URI for account_uri_complete()

cspiel1:
```
/dial [IPv6]
```
lead to wrong append of the account host
```
sip:[IPv6]@account-host
```


alfredh:
is it possible to also add a test for this usecase ?

I think that for lowlevel functions like this, it is very good to have a couple of testcases for it.

there are already some tests for account in `test/account.c`.

alfredh:
thanks! there is a tiny ccheck warning:

```
./test/test.h:66: line is too wide (81 - max 79)
5
```

PR #1473, opened 2021-05-28
IPv6 link local support

cspiel1:
This PR adds IPv6 link local addresses as SIP transport.

This adds support for outgoing calls to IPv6 link local addresses and answering incoming calls from IPv6 link local peers.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -689,6 +689,7 @@ const char *menc_event_name(enum menc_event event);
 struct network;
 
 typedef void (net_change_h)(void *arg);
+typedef int (net_laddr_h)(const struct sa *sa);
 
 int  net_alloc(struct network **netp, const struct config_net *cfg);
 int  net_use_nameserver(struct network *net,
@@ -704,6 +705,8 @@ void net_dns_refresh(struct network *net);
 int  net_dns_debug(struct re_printf *pf, const struct network *net);
 int  net_debug(struct re_printf *pf, const struct network *net);
 const struct sa *net_laddr_af(const struct network *net, int af);
+int net_laddr_apply(const struct network *net, net_laddr_h *laddrh);
+bool net_is_laddr(const struct network *net, struct sa *sa);
 struct dnsc     *net_dnsc(const struct network *net);
 
 
--- a/src/net.c
+++ b/src/net.c
@@ -10,10 +10,8 @@
 
 struct network {
 	struct config_net cfg;
-	struct sa laddr;
-#ifdef HAVE_INET6
-	struct sa laddr6;
-#endif
+	struct list laddrs;           /**< List of local addresses           */
+	struct list laddrs_tmp;       /**< Temp. list of local addresses     */
 	struct tmr tmr;
 	struct dnsc *dnsc;
 	struct sa nsv[NET_MAX_NS];/**< Configured name servers           */
@@ -23,6 +21,12 @@ struct network {
 	uint32_t interval;
 	net_change_h *ch;
 	void *arg;
+};
+
+
+struct laddr {
+	struct le le;
+	struct sa sa;
 };
 
 
@@ -41,6 +45,7 @@ static void net_destructor(void *data)
 
 	tmr_cancel(&net->tmr);
 	mem_deref(net->dnsc);
+	list_flush(&net->laddrs);
 }
 
 
@@ -185,6 +190,22 @@ static int net_dns_srv_get(const struct network *net,
 }
 
 
+static int list_add_laddr(struct list *list, const struct sa *sa)
+{
+	struct laddr *laddr;
+	if (!sa_isset(sa, SA_ADDR))
+		return 0;
+
+	laddr = mem_zalloc(sizeof(*laddr), NULL);
+	if (!laddr)
+		return ENOMEM;
+
+	laddr->sa = *sa;
+	list_append(list, &laddr->le, laddr);
+	return 0;
+}
+
+
 /*
  * Check for DNS Server updates
  */
@@ -223,6 +244,60 @@ static void ipchange_handler(void *arg)
 }
 
 
+static bool ifaddr_filter(const char *ifname, const struct sa *sa, void *arg)
+{
+	struct network *net = arg;
+	const struct config_net *cfg = &net->cfg;
+
+	if (str_isset(cfg->ifname) && str_cmp(cfg->ifname, ifname))
+		return false;
+
+	if (!net_af_enabled(net, sa_af(sa)))
+		return false;
+
+	if (sa_is_loopback(sa))
+		return false;
+
+	list_add_laddr(&net->laddrs_tmp, sa);
+	return false;
+}
+
+
+static bool laddr_cmp(struct le *le, void *arg)
+{
+	struct laddr *laddr = le->data;
+	struct sa *sa = arg;
+
+	return sa_cmp(&laddr->sa, sa, SA_ADDR);
+}
+
+
+static bool net_misses_laddr(struct le *le, void *arg)
+{
+	struct network *net = arg;
+	struct laddr *laddr = le->data;
+	struct sa *sa = &laddr->sa;
+
+	return NULL == list_apply(&net->laddrs, true, laddr_cmp, sa);
+}
+
+
+static int net_dst_is_source_addr(const struct sa *dst, const struct sa *ip)
+{
+	struct sa src;
+	int err;
+
+	err = net_dst_source_addr_get(dst, &src);
+	if (err)
+		return err;
+
+	if (!sa_cmp(ip, &src, SA_ADDR))
+		return ECONNREFUSED;
+
+	return 0;
+}
+
+
 /**
  * Check if local IP address(es) changed
  *
@@ -232,58 +307,30 @@ static void ipchange_handler(void *arg)
  */
 bool net_check(struct network *net)
 {
-	struct sa laddr;
-#ifdef HAVE_INET6
-	struct sa laddr6;
-#endif
-	bool change = false;
+	const struct config_net *cfg = &net->cfg;
+	struct sa sa;
 
 	if (!net)
 		return false;
 
-	laddr = net->laddr;
-#ifdef HAVE_INET6
-	laddr6 = net->laddr6;
-#endif
+	if (str_isset(cfg->ifname) && 0 == sa_set_str(&sa, cfg->ifname, 0)) {
 
-	if (str_isset(net->cfg.ifname)) {
-
-		if (net_af_enabled(net, AF_INET))
-			net_if_getaddr(net->cfg.ifname, AF_INET, &net->laddr);
-
-#ifdef HAVE_INET6
-		if (net_af_enabled(net, AF_INET6))
-			net_if_getaddr(net->cfg.ifname, AF_INET6,
-				       &net->laddr6);
-#endif
+		info("Binding to IP address '%j'\n", &sa);
+		list_add_laddr(&net->laddrs_tmp, &sa);
 	}
 	else {
-		if (net_af_enabled(net, AF_INET))
-			net_default_source_addr_get(AF_INET, &net->laddr);
-
-#ifdef HAVE_INET6
-		if (net_af_enabled(net, AF_INET6))
-			net_default_source_addr_get(AF_INET6, &net->laddr6);
-#endif
+		net_if_apply(ifaddr_filter, net);
 	}
 
-	if (sa_isset(&net->laddr, SA_ADDR) &&
-	    !sa_cmp(&laddr, &net->laddr, SA_ADDR)) {
-		change = true;
-		info("net: local IPv4 address changed: %j -> %j\n",
-		     &laddr, &net->laddr);
+	if (list_count(&net->laddrs_tmp) != list_count(&net->laddrs) ||
+	    list_apply(&net->laddrs_tmp, true, net_misses_laddr, net)) {
+
+		list_flush(&net->laddrs);
+		net->laddrs = net->laddrs_tmp;
+		return true;
 	}
 
-#ifdef HAVE_INET6
-	if (sa_isset(&net->laddr6, SA_ADDR) &&
-	    !sa_cmp(&laddr6, &net->laddr6, SA_ADDR)) {
-		change = true;
-		info("net: local IPv6 address changed: %j -> %j\n",
-		     &laddr6, &net->laddr6);
-	}
-#endif
-
-	return change;
+	return false;
 }
 
 
@@ -349,7 +396,7 @@ int net_alloc(struct network **netp, const struct config_net *cfg)
 	struct network *net;
 	struct sa nsv[NET_MAX_NS];
 	uint32_t nsn = ARRAY_SIZE(nsv);
-	char buf4[128] = "", buf6[128] = "";
+	struct le *le;
 	int err;
 
 	if (!netp || !cfg)
@@ -414,102 +461,21 @@ int net_alloc(struct network **netp, const struct config_net *cfg)
 		goto out;
 	}
 
-	sa_init(&net->laddr, AF_INET);
-
-	if (str_isset(cfg->ifname)) {
-
-		struct sa temp_sa;
-		bool got_it = false;
-
-		info("Binding to interface or IP address '%s'\n", cfg->ifname);
-
-		/* check for valid IP-address */
-		if (0 == sa_set_str(&temp_sa, cfg->ifname, 0)) {
-
-			switch (sa_af(&temp_sa)) {
-
-			case AF_INET:
-				net->laddr = temp_sa;
-				break;
-
-#ifdef HAVE_INET6
-			case AF_INET6:
-				net->laddr6 = temp_sa;
-				break;
-#endif
-
-			default:
-				err = EAFNOSUPPORT;
-				goto out;
-			}
-
-			goto print_network_data;
-		}
-
-		if (net_af_enabled(net, AF_INET)) {
-
-			err = net_if_getaddr(cfg->ifname, AF_INET,
-					     &net->laddr);
-			if (err) {
-				info("net: %s: could not get IPv4 address"
-				     " (%m)\n",
-				     cfg->ifname, err);
-			}
-			else
-				got_it = true;
-		}
-
-#ifdef HAVE_INET6
-		if (net_af_enabled(net, AF_INET6)) {
-
-			err = net_if_getaddr(cfg->ifname, AF_INET6,
-					     &net->laddr6);
-			if (err) {
-				info("net: %s: could not get IPv6 address"
-				     " (%m)\n",
-				     cfg->ifname, err);
-			}
-			else
-				got_it = true;
-		}
-#endif
-		if (got_it)
-			err = 0;
-		else {
-			warning("net: %s: could not get network address\n",
-				cfg->ifname);
-			err = EADDRNOTAVAIL;
-			goto out;
-		}
+	(void)net_check(net);
+	if (!list_count(&net->laddrs)) {
+		warning("net: %s: could not get network address\n",
+			cfg->ifname);
+		err = EADDRNOTAVAIL;
 	}
 	else {
-		if (net_af_enabled(net, AF_INET))
-			net_default_source_addr_get(AF_INET, &net->laddr);
-
-#ifdef HAVE_INET6
-		sa_init(&net->laddr6, AF_INET6);
-
-		if (net_af_enabled(net, AF_INET6))
-			net_default_source_addr_get(AF_INET6, &net->laddr6);
-#endif
+		info("Local network addresses:\n");
+		LIST_FOREACH(&net->laddrs, le) {
+			struct laddr *laddr = le->data;
+			info("  %H\n", print_addr, &laddr->sa);
+		}
 	}
-
-print_network_data:
-
-	if (sa_isset(&net->laddr, SA_ADDR)) {
-		re_snprintf(buf4, sizeof(buf4), " IPv4=%H",
-			    print_addr, &net->laddr);
-	}
-#ifdef HAVE_INET6
-	if (sa_isset(&net->laddr6, SA_ADDR)) {
-		re_snprintf(buf6, sizeof(buf6), " IPv6=%H",
-			    print_addr, &net->laddr6);
-	}
-#endif
 
 	(void)dns_srv_get(NULL, 0, nsv, &nsn);
-
-	info("Local network address: %s %s\n", buf4, buf6);
 
  out:
 	if (err)
@@ -564,22 +530,7 @@ int net_set_address(struct network *net, const struct sa *ip)
 	if (!net)
 		return EINVAL;
 
-	switch (sa_af(ip)) {
-
-	case AF_INET:
-		sa_cpy(&net->laddr, ip);
-		break;
-
-#ifdef HAVE_INET6
-	case AF_INET6:
-		sa_cpy(&net->laddr6, ip);
-		break;
-#endif
-	default:
-		return EAFNOSUPPORT;
-	}
-
-	return 0;
+	return list_add_laddr(&net->laddrs, ip);
 }
 
 
@@ -687,6 +638,46 @@ static bool if_debug_handler(const char *ifname, const struct sa *sa,
 }
 
 
+enum laddr_check {
+	LADDR_NOLINKLOCAL = 1,
+	LADDR_INTERNET = 2
+};
+
+
+static const struct sa *find_laddr_af(const struct network *net, int af,
+		enum laddr_check lc)
+{
+	struct le *le;
+	struct sa dst;
+
+	if (!net)
+		return NULL;
+
+	sa_init(&dst, af);
+	if (af == AF_INET6)
+		sa_set_str(&dst, "1::1", 53);
+	else
+		sa_set_str(&dst, "1.1.1.1", 53);
+
+	LIST_FOREACH(&net->laddrs, le) {
+		struct laddr *laddr = le->data;
+		if (sa_af(&laddr->sa) != af)
+			continue;
+
+		if ((lc & LADDR_NOLINKLOCAL) && sa_is_linklocal(&laddr->sa))
+			continue;
+
+		if ((lc & LADDR_INTERNET) &&
+				net_dst_is_source_addr(&dst, &laddr->sa))
+			continue;
+
+		return &laddr->sa;
+	}
+
+	return NULL;
+}
+
+
 /**
  * Get the local IP Address for a specific Address Family (AF)
  *
@@ -697,17 +688,50 @@ static bool if_debug_handler(const char *ifname, const struct sa *sa,
  */
 const struct sa *net_laddr_af(const struct network *net, int af)
 {
-	if (!net)
-		return NULL;
+	const struct sa *sa = NULL;
 
-	switch (af) {
+	sa = find_laddr_af(net, af, LADDR_NOLINKLOCAL | LADDR_INTERNET);
+	if (sa)
+		return sa;
 
-	case AF_INET:  return &net->laddr;
-#ifdef HAVE_INET6
-	case AF_INET6: return &net->laddr6;
-#endif
-	default:       return NULL;
+	sa = find_laddr_af(net, af, LADDR_NOLINKLOCAL);
+	if (sa)
+		return sa;
+
+	sa = find_laddr_af(net, af, 0);
+	return sa;
+}
+
+
+int net_laddr_apply(const struct network *net, net_laddr_h *laddrh)
+{
+	struct le *le;
+	int err = 0;
+
+	if (!net || !laddrh)
+		return EINVAL;
+
+	LIST_FOREACH(&net->laddrs, le) {
+		struct laddr *laddr = le->data;
+
+		err |= laddrh(&laddr->sa);
 	}
+
+	return err;
+}
+
+
+/**
+ * Checks if given IP address is a local address.
+ *
+ * @param net Network intance
+ * @param sa  IP address to check
+ *
+ * @return true if sa is a local address, false if not
+ */
+bool net_is_laddr(const struct network *net, struct sa *sa)
+{
+	return NULL != list_apply(&net->laddrs, true, laddr_cmp, sa);
 }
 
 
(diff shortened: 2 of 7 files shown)


src/net.c
```diff
@@ -224,6 +225,20 @@ static void ipchange_handler(void *arg)
 }
 
 
+static void set_default_ipv6ll(struct network *net)
+{
+	char ifname[32] = "";
```
sreimers, line 230: `ifname[32]` is a bit small, would prefer `ifname[64]`, on windows i see interface names like this:
`{B3B5F4C1-231D-4878-8882-2F81737EEA97}`

alfredh:
ipv6 linklocal address is also an IPv6 address, so I dont see the reason to add a new field here.

instead of storing each address for each address family, I think we should
change it into an array which makes the code more flexible.

`struct network`:

Before:

```c
	struct sa laddr;
#ifdef HAVE_INET6
	struct sa laddr6;
#endif
```

After:

```c
#define MAX_ADDR 8
	struct sa laddrv[MAX_ADDR];
```

this would also fit into the plan of supporting multiple IPv6 addresses.

cspiel1:
From baresip point of view this should be possible.

I have to check if the libre PR https://github.com/baresip/re/pull/106 is also possible with multiple IPv6 addresses.

The crucial point will be how to implement function
```
static const struct sip_transport *transp_find(struct sip *sip,
					       enum sip_transp tp,							       enum sip_transp tp,
					       int af, const struct sa *dst)							       int af, const struct sa *dst)
```
It is called from `sip_transp_laddr()`, which is used to get the local address to copy the IPv6 scope ID to the `dst` address on different locations where a destination address could be an IPv6 link local address. Otherwise the IPv6 packet can't be routed correctly.

alfredh:
keep in mind that net.c and ua.c are two different subsystems.

net.c is for the whole application (incl SIP,RTP,ICE etc)

ua.c is for SIP only

so we can still add support for multiple addresses in net.c

cspiel1:
- Now net.c adds all local addresses (except loopback) to a list.
- Instead of net_laddr_af() new function net_laddr_dst() should be used wherever it is possible.
- This version allows any IPv4, IPv6, IPv6ll address to be used for peer-to-peer calls. It should not make a difference if there is a route to the internet.

Future work:
- Without a change in the API (`ua_call_alloc()`) the destination address is not available where `net_laddr_af()` is called. :-(
- Is net.c already obsolete? The information could be retrieved from the system with libre-s net API on demand.

cspiel1:
@sreimers The build fails with the re version check.
```
ld: malformed 32-bit x.y.z version number: 1.0.0-dev
230
231
make: *** [librem.dylib] Error 1
232
Error: Process completed with exit code 2.
```

sreimers:
I think you need to rebase the libre PR (it uses the old mk/re.mk).

juha-h:
Does this work now solve the issue https://github.com/juha-h/baresip-studio/issues/54 that my baresip Android app has had from the very beginning, i.e., that baresip can use only one IP address for outbound requests?  For example, if the device has mobile network IP address, WiFi IP address and a VPN IP address, then baresip will choose the outbound source IP address based on OS routing table?

cspiel1:
Yes, I think so because the new function `net_check_conn()` in the libre PR should test the connection with a temporary UDP socket.

Have to check/fix that network changes are detected and applied to active calls and the net object. Then will activate this PR again.

cspiel1:
Now changes in the `laddrs` list work. The PR is ready for your tests and review.

As I already wrote above in the future TODOs:
- The use of `net_laddr_af()` in ua.c should be replaced by `net_laddr_dst()` if the call destination is an IP-address. But we have no destination address in `ua_call_alloc()`.
- The result is that if there are multiple interfaces the IP address in SDP might be a different than for SIP. 
- Also it is possible that e.g. SIP uses IPv6 and RTP IPv4. I know that this is normally ok. But maybe can lead to troubles.

juha-h:
Christian Spielberger writes:

> Yes, I think so because the new function `net_check_conn()` in the
> libre PR should test the connection with a temporary UDP socket.

It is not possible to test SIP connectivity with UDP socket, since
firewall somewhere between client and server could block UDP SIP
traffic.

sreimers:
> It is not possible to test SIP connectivity with UDP socket,

There is no real UDP Datagram send, only local udp socket is connected for test.

From `man connect`
```
    If the initiating socket is not connection-mode, then connect() shall set the socket's peer address, and no connection is made.  For  SOCK_DGRAM  sockets,  the
       peer  address  identifies  where  all  datagrams  are sent on subsequent send() functions, and limits the remote sender for subsequent recv() functions. If the
       sa_family member of address is AF_UNSPEC, the socket's peer address shall be reset. Note that despite no connection being made, the term ``connected'' is  used
       to describe a connectionless-mode socket for which a peer address has been set.
```

cspiel1:
> > It is not possible to test SIP connectivity with UDP socket,
> 
> There is no real UDP Datagram send, only local udp socket is connected for test.
> 
> From `man connect`
> 
> ```
>     If the initiating socket is not connection-mode, then connect() shall set the socket's peer address, and no connection is made.  For  SOCK_DGRAM  sockets,  the
>        peer  address  identifies  where  all  datagrams  are sent on subsequent send() functions, and limits the remote sender for subsequent recv() functions. If the
>        sa_family member of address is AF_UNSPEC, the socket's peer address shall be reset. Note that despite no connection being made, the term ``connected'' is  used
>        to describe a connectionless-mode socket for which a peer address has been set.
> ```

For `net_check_conn()` I took the code from `net_dst_source_addr_get()`. Maybe I made a to strong statement, that the connection is tested. Only the route is checked what should be enough to select the correct local IP address, right? Anyway I will have to write a unit-test for `net_check_conn()`. Then we will see.

sreimers:
> Only the route is checked what should be enough to select the correct local IP address, right?

I think so.

juha-h:
I tested baresip built with pull requests:

https://github.com/baresip/re/pull/106
https://github.com/baresip/baresip/pull/1473.

on host with the following following interfaces:

$ ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: enp0s3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 08:00:27:a9:81:ac brd ff:ff:ff:ff:ff:ff
    inet 10.0.2.15/24 brd 10.0.2.255 scope global dynamic enp0s3
       valid_lft 84989sec preferred_lft 84989sec
    inet6 fe80::a00:27ff:fea9:81ac/64 scope link 
       valid_lft forever preferred_lft forever
3: enp0s8: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 08:00:27:b5:50:84 brd ff:ff:ff:ff:ff:ff
    inet 192.26.134.12/24 brd 192.26.134.255 scope global enp0s8
       valid_lft forever preferred_lft forever
    inet6 fe80::a00:27ff:feb5:5084/64 scope link 
       valid_lft forever preferred_lft forever

and the following routing table:

$ route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         10.0.2.2        0.0.0.0         UG    0      0        0 enp0s3
10.0.2.0        0.0.0.0         255.255.255.0   U     0      0        0 enp0s3
169.254.0.0     0.0.0.0         255.255.0.0     U     1000   0        0 enp0s3
192.26.134.0    0.0.0.0         255.255.255.0   U     0      0        0 enp0s8
192.168.43.0    192.26.134.1    255.255.255.0   UG    1      0        0 enp0s8

In the first test, destination was ***@***.*** where tutpro.com is
reachable via default route gateway 10.0.2.2.  In that test INVITE was
sent like this:

TLS 10.0.2.15:40128 -> 192.26.111.38:8003

and the local IP address in INVITE headers and SDP correctly was
10.0.2.15.

Then I made another test where destination was ***@***.*** where
test.tutpro.com IP address 192.168.43.160 is reachable via gateway
192.26.134.1.  In that test the local IP address in INVITE headers and
SDP was again 10.0.2.15:

TLS 192.26.134.12:34080 -> 192.168.43.160:5061
INVITE ***@***.*** SIP/2.0
Via: SIP/2.0/TLS 10.0.2.15:5061;branch=z9hG4bKc737df1a879c24cf;rport
Contact: ***@***.***:5061;transport=tls>
Max-Forwards: 70
To: ***@***.***>
From: ***@***.***>;tag=9806c1a997cebd11
Call-ID: 3eac1754e98c4821
CSeq: 48392 INVITE
User-Agent: baresip v1.1.0 (x86_64/linux)
Allow: INVITE,ACK,BYE,CANCEL,OPTIONS,NOTIFY,SUBSCRIBE,INFO,MESSAGE,REFER
Supported: gruu
Content-Type: application/sdp
Content-Length: 320

I think that is incorrect.  IP address in INVITE headers and SDP should
have been 192.26.134.12.

juha-h:
Looks like github censored my comment.  Let me know if you need the censored bits.

cspiel1:
Removed `net_check_conn()` in libre. It didn't check the route correctly. Instead use `net_dst_source_addr_get()` which provides a useful local address for a given destination.

@juha-h Thanks for the test. Please could you try again with current branches ipv6ll_support? I think, you are right with what you are expecting for your setup.

juha-h:
So it does not make sense for me to do another test?

I tried, but baresip didn't compile using latest pull request patched re and baresip:
```
 CC      build-x86_64/src/net.o
src/net.c: In function ‘find_laddr_af’:
src/net.c:656:32: warning: implicit declaration of function ‘net_check_conn’; did you mean ‘net_check’? [-Wimplicit-function-declaration]
   if ((lc & LADDR_INTERNET) && net_check_conn(&laddr->sa, &dst))
                                ^~~~~~~~~~~~~~
                                net_check
src/net.c:656:32: warning: nested extern declaration of ‘net_check_conn’ [-Wnested-externs]
```

cspiel1:
Maybe a new field in struct ua used temporary.
```
struct ua {
...
	struct sa dst;               /**< Current destination address        */
};
```

juha-h:
Christian Spielberger writes:

> Maybe a new field in struct ua used temporary.
> ```
> struct ua {
> ...
> 	struct sa dst;               /**< Current destination address        */
> };
> ```

How could ua have a current destination address, since I can have an
ongoing call from the ua with somebody and at the same time send
messages from the ua to other parties?

cspiel1:
`ua->dst` is set before `ua_call_alloc()` is called and directly used in it, then cleared immediately. Of course it is more convenient to add it to the parameter list of `ua_call_alloc()`. Should we?

sreimers:
```
[ RUN      ] test_call_answer
ua: SDP offer origin address 127.0.0.1 not reachable
selftest: re_main() loop timed out -- test hung..
TEST_ERR: test/call.c:364: (Operation timed out)
test_call_answer: test failed (Operation timed out)
test failed (Operation timed out)
```
Why test_call_answer with `127.0.0.1` fails on macos is a little strange (because `net_dst_source_addr_get` with 127.0.0.1 works in retest on macos). I can investigate this or is this related to `net_laddr_af` ?

juha-h:
Christian Spielberger writes:

> `ua->dst` is set before `ua_call_alloc()` is called and directly used in it, then cleared immediately. Of course it is more convenient to add it to the parameter list of `ua_call_alloc()`. Should we?

Is ua_call_alloc used also when MESSAGE, OPTIONS, and PUBLISH requests
are sent?

juha-h:
Kamailio SIP Proxy has this file that includes socket related stuff:

  https://github.com/kamailio/kamailio/blob/master/src/core/forward.c

For example, it has function get_out_socket() that

* return a socket_info_pointer to the sending socket; as opposed to
 * get_send_socket, which returns process's default socket, get_out_socket
 * attempts to determine the outbound interface which will be used;
 * it uses a temporary connected socket to determine it; it will
 * be very likely noticeably slower, but it can deal better with
 * multihomed hosts
 */

Then it has forward_request() function that calls

  build_req_buf_from_sip_req(msg, &len, send_info, 0);

which rebuilds the SIP message based on the sending socket.

Makes sense to me to handle the source IP/port stuff at low level where
there is no knowledge left about UAs and other high level SIP concepts.

cspiel1:
Now it works also "correct" for outgoing calls. That means if the destination is an IP-address:
- The address family (af) of the destination is used for RTP.
- The local IP address is chosen for SDP that has a route to the destination IP.
- If destination is a IPv6ll it simply works. Also in SDP we have the local IPv6ll address of the correct interface.

Your questions:
- @juha-h MESSAGE, OPTIONS and PUBLISH seems to have not a real call. Only handling of OPTIONS has a temp dummy call to encode the SDP answer. See `ua_handle_options()`! We should pass the `sip_msg` to `ua_call_alloc` and set af and laddr adequate.
- @juha-h The local address is needed in the `sdp_session`. Currently it is collected in baresip-s net object and passed from ua to call to the `sdp_session`. Maybe we could get rid of collecting laddrs in the net object and retrieve the right local address on the fly in the SDP is encoded. I am not sure if it is possible.
- @sreimers I have no idea about the problem on MAC-OS or the error "Can't assign requested address". Could you please debug this?

juha-h:
Christian Spielberger writes:

> Now it works also "correct" for outgoing calls. That means if the
> destination is an IP-address:

In my tests, destination was not an IP address, but a domain name that is
resolved via naptr/srv/a process.

Why not just mark in the message where the source address need to be
placed and let the low level part of the code replace the mark with real
address once the destination has been resolved and socket chosen?

That would be independent of what kind of message is being sent out.

There could also exist a "force send socket" account parameter if user
wants to override the routing table one.

sreimers:
> * @sreimers I have no idea about the problem on MAC-OS or the error "Can't assign requested address". Could you please debug this?

The destination port must not be zero, i just tested successfully this patch:

```diff
diff --git a/src/net/net.c b/src/net/net.c
index 70f7b25b..5cf928d5 100644
--- a/src/net/net.c
+++ b/src/net/net.c
@@ -70,6 +70,7 @@ int net_dst_source_addr_get(const struct sa *dst, struct sa *ip)
 {
        int err;
        struct udp_sock *us;
+       struct sa tmp_dst;
 
        if (sa_af(dst) == AF_INET6)
                err = sa_set_str(ip, "::", 0);
@@ -79,11 +80,15 @@ int net_dst_source_addr_get(const struct sa *dst, struct sa *ip)
        if (err)
                return err;
 
+       tmp_dst = *dst;
+       if (sa_port(&tmp_dst) == 0)
+               sa_set_port(&tmp_dst, 53);
+
        err = udp_listen(&us, ip, NULL, NULL);
        if (err)
                return err;
 
-       err = udp_connect(us, dst);
+       err = udp_connect(us, &tmp_dst);
        if (err)
                goto out;
```

Would you like to include this on your PR or should I make a separate PR and you rebase?

sreimers:
Maybe we simply fallback to `net_default_source_addr_get` if no destination address is available?

cspiel1:
> > * @sreimers I have no idea about the problem on MAC-OS or the error "Can't assign requested address". Could you please debug this?
> 
> The destination port must not be zero, i just tested successfully this patch:
> 
> ```diff
> diff --git a/src/net/net.c b/src/net/net.c
> index 70f7b25b..5cf928d5 100644
> --- a/src/net/net.c
> +++ b/src/net/net.c
> @@ -70,6 +70,7 @@ int net_dst_source_addr_get(const struct sa *dst, struct sa *ip)
>  {
>         int err;
>         struct udp_sock *us;
> +       struct sa tmp_dst;
>  
>         if (sa_af(dst) == AF_INET6)
>                 err = sa_set_str(ip, "::", 0);
> @@ -79,11 +80,15 @@ int net_dst_source_addr_get(const struct sa *dst, struct sa *ip)
>         if (err)
>                 return err;
>  
> +       tmp_dst = *dst;
> +       if (sa_port(&tmp_dst) == 0)
> +               sa_set_port(&tmp_dst, 53);
> +
>         err = udp_listen(&us, ip, NULL, NULL);
>         if (err)
>                 return err;
>  
> -       err = udp_connect(us, dst);
> +       err = udp_connect(us, &tmp_dst);
>         if (err)
>                 goto out;
> ```
> 
> Would you like to include this on your PR or should I make a separate PR and you rebase?

Thanks! I will include this.

cspiel1:
> In my tests, destination was not an IP address, but a domain name that is resolved via naptr/srv/a process. Why not just mark in the message where the source address need to be placed and let the low level part of the code replace the mark with real address once the destination has been resolved and socket chosen? That would be independent of what kind of message is being sent out. There could also exist a "force send socket" account parameter if user wants to override the routing table one.

I will have a look on it. One complication what I know currently is that the audio/video objects and their stream objects are allocated very early. The stream objects are opening their sockets. For this the address family has to be known.

juha-h:
Christian Spielberger writes:

> I will have a look on it. One complication what I know currently is
> that the audio/video objects and their stream objects are allocated
> very early. The stream objects are opening their sockets. For this the
> address family has to be known.

Address family is not known before dns lookup and the lookup may result
in both ipv4 and ipv6 addresses and more than one of those may need to
be tried before the server is reached.  So the both SIP headers and SDP
may need to be changed.

Perhaps the stream object has to have place holders for address and
address family related fields.

cspiel1:
@juha-h 
- It is possible to postpone the choice of laddr directly before the call of `call_sdp_get()` where the SDP is encoded.
- Then we have also to solve the complication with the address family for the RTP sockets. If the `af` does not match you get:
```
stream: rtp_send to open natpinholefailed (Bad file descriptor)
video: video stream error: Bad file descriptor
```
- Moving the choice to an even lower layer is not possible, because the target address for SIP is not available in the `sdp_session` or the `rtp` part. So we have to solve this in baresip call.c and ua.c. Directly before the calls to `call_sdp_get()`. Or in `call_sdp_get()` if we add a parameter `struct sa *dst`.

cspiel1:
> I will have a look on it. One complication what I know currently is that the audio/video objects and their stream objects are allocated very early. The stream objects are opening their sockets. For this the address family has to be known.
> Address family is not known before dns lookup and the lookup may result in both ipv4 and ipv6 addresses and more than one of those may need to be tried before the server is reached. So the both SIP headers and SDP may need to be changed. Perhaps the stream object has to have place holders for address and address family related fields.

Ok, maybe possible if the socket open can be postponed also. I will check this.

cspiel1:
> * The result of the DSN lookup should be available already.

The DNS lookup result is not available when the SDP is encoded. I mixed up DSN lookup with NAT resolution. So the SDP encoding would also have to be postponed.

cspiel1:
The current solution works if the destination address is already an IP-address. This is what we need.

An improvement for domain addresses seems to be not possible without changing the API in re/baresip.

I will solve the failed check now. And try to write a retest for more test coverage.

alfredh:
we should also discuss if `src/net.c` is obsolete.

currently it is a thin wrapper on top of `re/src/net`.

the main functions are:

- storing of local IP address(es)
- storing of DNS client
- timer to detect network changes

cspiel1:
> we should also discuss if `src/net.c` is obsolete.
> 
> currently it is a thin wrapper on top of `re/src/net`.
> 
> the main functions are:
> 
> * storing of local IP address(es)
> * storing of DNS client
> * timer to detect network changes

I agree, thanks for the list.
- local IP addresses. --> They can be retrieved on demand from the system.
- DNS clients. --> The same?
- A timer for local IP changes is only needed if there is currently a call that should be updated like it is done with `call_reset_transp()`.

cspiel1:
In order to detect any change in the list of local addresses, we need to store the list to compare it periodically. For this purpose it makes sense to have it stored in net.c.

Instead we can compare the systems local addresses with the laddr of the transport objects in `uag.sip`.  Right?

src/call.c
```diff
@@ -2104,6 +2104,19 @@ struct audio *call_audio(const struct call *call)
...
+/**
+ * Get the remote IP address for the given call
+ *
+ * @param call  Call object
+ *
+ * @return Remote IP address
+ */
+const struct sa *call_raddr(const struct call *call)
+{
+	return stream_raddr(audio_strm(call_audio(call)));
```
alfredh, line 2116: what about video stream ?

include/baresip.h
```diff
@@ -689,6 +690,7 @@ const char *menc_event_name(enum menc_event event);
 struct network;
 
 typedef void (net_change_h)(void *arg);
+typedef int (net_laddr_h)(const struct sa *sa);
```
alfredh, line 693: perhaps also add interface name ?

src/call.c
cspiel1: If there is a `c` line in the remote SDP then it is possible that raddr differ for audio and video. But both should be reachable from same local network interface and local address. The result for
```
net_dst_source_addr_get(call_raddr(call),
							     &laddr)
```
should be the same.

The name of the function might be confusing. But we don't need it. I will drop it and use the underlying functions directly.

include/baresip.h
cspiel1: Yes, this makes sense.

cspiel1:
Removal of baresip/src/net.c:
- Removal of the laddr list seems to be possible and makes sense to reduce redundancy in data.
- Removal of nsv, nsvf. This is also redundancy. It is duplicated from the config.
- Removal of dnsc. I am not sure if this would improve something. It makes sense that the DNS clients are allocated only once and re-used for each DNS lookup. Further we should not move this all to uag.c. Right?
- Check of network change could also stay inside net.c, if it is able to compare with the laddrs of the transp list.

sreimers:
> * Removal of nsv, nsvf. This is also redundancy. It is duplicated from the config.

Is not redundant because its a special use-case for nameserver fallbacks. I have many customers with defect nameservers, but I don't want enforce for privacy something like 8.8.8.8 for all customers (in some restricted networks they also not work, so overriding is not an option). If no `dns_server` is configured, baresip detects the system ones and with `dns_fallback` I can define additional ones to query (especially SRV querys are blocked sometimes). `dns_server` is used to ignore the system ones.

include/baresip.h
cspiel1: But if we drop the net->laddrs list, then this function pointer is obsolete.

juha-h:
Christian Spielberger writes:

> The current solution works if the destination address is already an
> IP-address. This is what we need.

I don't think it is a good idea to introduce solutions that work only in
some not so common special cases.

> An improvement for domain addresses seems to be not possible without
> changing the API in re/baresip.

It would be better to fix the API in re/baresip and have a solution that
works also in the very common case where server address is domain name.

juha-h:
Sebastian Reimers writes:

> Is not redundant because its a special use-case for nameserver
> fallbacks. I have many customers with defect nameservers, but I don't
> want enforce for privacy something like 8.8.8.8 for all customers (in
> some restricted networks they also not work, so overriding is not an
> option). If no `dns_server` is configured, baresip detects the system
> ones and with `dns_fallback` I can define additional ones to query
> (especially SRV querys are blocked sometimes). `dns_server` is used to
> ignore the system ones.

Just to note that Android does not allow applications to access system
name server list.

alfredh:
I would also like to suggest to split up this patch into logical parts.


for SIP multi-homing it should be quite easy to do it for TCP/TLS:

1. TCP connect to the SIP server
2. when connected, look at IP-address in `tcp_conn_local_get`
3. use this address in SIP via header and other headers for local IP

this should work if the device has multiple network interfaces.

The SDP part is not so easy, and should be handled separately.

juha-h:
Alfred E. Heggestad writes:

> for SIP multi-homing it should be quite easy to do it for TCP/TLS:
> 
> 1. TCP connect to the SIP server
> 2. when connected, look at IP-address in `tcp_conn_local_get`
> 3. use this address in SIP via header and other headers for local IP
> 
> this should work if the device has multiple network interfaces.

Sounds very promising.

cspiel1:
> I would also like to suggest to split up this patch into logical parts.

Ok. I will squash the commits and split them again into logical parts.

> 
> for SIP multi-homing it should be quite easy to do it for TCP/TLS:
> 
> 1. TCP connect to the SIP server
> 2. when connected, look at IP-address in `tcp_conn_local_get`
> 3. use this address in SIP via header and other headers for local IP
> 
> this should work if the device has multiple network interfaces.
> 
> The SDP part is not so easy, and should be handled separately.

This PR is more concerned to Server-less calls to an IP-address.
- It selects correct address family and local IP address for SDP.
- It sets the scope ID for destination address if it is an IPv6 link local address.

For registered UAs maybe we could use the local IP of the registration.

cspiel1:
@alfredh @juha-h @sreimers Would it be ok to change the API in order to set the `laddr` after the DNS resolution was done?
e.g.
- replace the `struct mbuf *desc` by a `struct struct sdp_session *desc` in function `sipsess_connect()`

juha-h:
Christian Spielberger writes:

> @alfredh @juha-h @sreimers Would it be ok to change the API in order
> to set the `laddr` after the DNS resolution was done? 
> e.g.
> - replace the `struct mbuf *desc` by a `struct struct sdp_session
> *desc` in function `sipsess_connect()`

Should be OK, but in theory also SUBSCRIBE requests can contain bodies.

cspiel1:
Split the PR into logical separated commits.

So far not added the suggested features discussed above.
- Select laddr after DNS resolution.
- Tcp connect before select laddr.

We should handle these in future PRs.

alfredh:
is there an easy way to test IPv6 link-local on the local machine ?

alfredh:
should we also support IPv6 address suffix?

```
fe80::b059:65f4:e877:c40%en0
```

https://superuser.com/questions/99746/why-is-there-a-percent-sign-in-the-ipv6-address

cspiel1:
> is there an easy way to test IPv6 link-local on the local machine ?

I test always with two machines in local network. Both with baresip.
```
echo "/dial [fe80::...]" | nc -N localhost 5555
```

cspiel1:
> should we also support IPv6 address suffix?
> 
> ```
> fe80::b059:65f4:e877:c40%en0
> ```
> 
> https://superuser.com/questions/99746/why-is-there-a-percent-sign-in-the-ipv6-address

This patch automatically selects right scope ID which is assigned to the interface. There is no need that the user specifies the network interface for now.

Anyhow it could make sense to support the address suffix in `sa_decode() or sa_set(), sa_set_str()`. We could postpone this until someone needs this.

alfredh:
regarding multiple addresses and IPv6-linklocal;

it looks like this patch can be split into 2 parts:

1. add support for multiple SIP listen interfaces
2. add support for IPv6 link local


regarding setting the scope-id on struct sa:

if you imagine the lifetime of a `struct sa` object, it is not very clean to
change the scopeid during the middle of the sa lifetime. I think it is more
clear to set the scopeid either:

1. when the sa is inited
2. right before the sa is used to send a packet out on a udp/tcp socket

is this possible to do ?

cspiel1:
- The split should be possible.
- You gave me an idea. When I slightly change `transp_find()` such that `dst` is not a const, it could "find" and set also the correct scope ID if `dst` is an IPv6ll.

cspiel1:
Now removed the scope ID stuff for IPv6ll.

The PR contains now only the commits to support all local network addresses and interfaces.

IPv6ll is already listed but does not work. Setting the scope ID will be re-based on top of this.

sreimers:
Makefile:148: *** bad libre version, required version is ">= 2.0.1-dev2". LIBRE_MK: /usr/local/share/re/re.mk.  Stop.

The libre PR needs a rebase, after this I can trigger a re-build.

cspiel1:
> `Makefile:148: *** bad libre version, required version is ">= 2.0.1-dev2". LIBRE_MK: /usr/local/share/re/re.mk. Stop.`
> 
> The libre PR needs a rebase, after this I can trigger a re-build.

Did the rebase already, this triggered the re-build. But the error is new:
```
Err:7 https://packages.microsoft.com/ubuntu/20.04/prod focal Release
14
  Could not handshake: Error in the pull function. [IP: 40.74.238.15 443]
```

sreimers:
triggered a re-build, seems to be fixed now.

juha-h:
Should this test https://github.com/baresip/baresip/pull/1473#issuecomment-857494756 now work with this PR or does it only handle ipv6ll case?

cspiel1:
> Should this test [#1473 (comment)](https://github.com/baresip/baresip/pull/1473#issuecomment-857494756) now work with this PR or does it only handle ipv6ll case?

Both no.
- DNS resolution before selecting laddr in ua.c would be another PR. But this is a first step.
- IPv6ll support will be added in https://github.com/baresip/baresip/pull/1500, which also depends on this PR.

alfredh:
please take a look at this test-program:

```c

#include <string.h>
#include <re.h>
#include <net/if.h>
#define __USE_POSIX 1  /**< Use POSIX flag */
#define __USE_XOPEN2K 1/**< Use POSIX.1:2001 code */
#include <sys/types.h>
#include <sys/socket.h>
#include <netdb.h>


static int resolve_address(struct sa *sa, const char *addr)
{
	struct addrinfo *res, *res0;
	int error;

	error = getaddrinfo(addr, NULL, NULL, &res0);
	if (error != 0) {
		fprintf(stderr, "error in getaddrinfo: %s\n",
			gai_strerror(error));
		return ENOENT;
	}

	for (res = res0; res; res = res->ai_next) {

		char ifname[64];

		sa_set_sa(sa, res->ai_addr);

		re_printf("scopeid: %u (%s)\n",
			  sa_scopeid(sa),
			  if_indextoname(sa_scopeid(sa), ifname));
		break;
	}

	return 0;
}


static void tcp_estab_handler(void *arg)
{
	re_printf("estab\n");
}


static void tcp_recv_handler(struct mbuf *mb, void *arg)
{
	re_printf("recv: %b\n", mbuf_buf(mb), mbuf_get_left(mb));
}


static void tcp_close_handler(int err, void *arg)
{
	re_printf("closed\n");
}


int main(int argc, char *argv[])
{
	struct tcp_conn *tc;
	struct sa srv;
	int err;

	if (argc < 2)
		return 2;

	err = resolve_address(&srv, argv[1]);
	if (err)
		return err;

	sa_set_port(&srv, 22);

	re_printf("connecting to: %J\n", &srv);

	err = tcp_connect(&tc, &srv, tcp_estab_handler,
			  tcp_recv_handler, tcp_close_handler, NULL);
	if (err) {
		re_printf("connect error (%m)\n", err);
		return err;
	}

	re_main(NULL);
}
```

when running the program locally on OSX:

```
$ ./test fe80::96c6:91ff:fead:cdc9%en0
scopeid: 5 (en0)
connecting to: [fe80::96c6:91ff:fead:cdc9]:22
estab
recv: SSH-2.0-OpenSSH_7.9p1 Debian-10+deb10u2

^C
```

`getaddrinfo` is able to resolve the IPv6 link local address and also the Scope-ID suffix.

perhaps we should try to get the HTTP client working first with this format?
example:

```
http://fe80::96c6:91ff:fead:cdc9%en0/index.html
```

cspiel1:
Yes, I know that `getaddrinfo` is able to decode the interface suffix. And `tcp_conn_alloc` already uses `getaddrinfo`. When trying  the HTTP client, I found that `sa_set_str` seems to deny the IPv6ll address with the suffix. It uses `net_inet_pton`. I will try to replace this by a `getaddrinfo` solution. This would be a different PR.

Note: This PR adds support for all local addresses which are collected by `getifaddr` which already sets the scope ID correctly. The goal should be that the user has not to specify the network interface for a IPv6ll target address. Instead it should be found by libre.

Anyway, I agree with you that the suffix should be supported. I will open another PR for this, if it works.

alfredh:
after reading more about IPv6 linklocal and testing the suffix solution,
I believe we are in a better shape to take better technical decisions.

I would like to propose the following:

1. we add support for IPv6 link local address suffix in HTTP
2. we add support for IPv6 link local address suffix in SIP
3. we add support for multiple interfaces in SIP


step 3 will require changes in re and baresip at the same time.
we should not add support for multiple addresses in baresip
before it is working properly in re.


this PR is getting rather large. could you please squash all the commits into
one single commit and then make a new PR for it ?

cspiel1:
will squash and create new PR.

include/baresip.h
cspiel1: I tried to add the interface name to `struct laddr`. But this enlarges the PR. And the benefit is low.

cspiel1:
> after reading more about IPv6 linklocal and testing the suffix solution,
> I believe we are in a better shape to take better technical decisions.
> 
> I would like to propose the following:
> 
> 1. we add support for IPv6 link local address suffix in HTTP
https://github.com/baresip/re/pull/116 is enough for this. No baresip change needed for HTTP.
> 2. we add support for IPv6 link local address suffix in SIP
In re, nothing in addition to the re PR is needed. As a wrote already, we need the correct link local address in baresip. So we need therefore already PR https://github.com/baresip/baresip/pull/1513.
> 3. we add support for multiple interfaces in SIP

Issue #1474, opened 2021-05-29
autoanswer issues


juha-h:
If I have in account config

;answermode=manual;answerdelay=1000

the call is answered automatically.

I would consider this a bug.  ;answerdelay should only have effect when answermode is auto.

In my opinion,  account's answer configuration is getting far too complicated:

;answermode={manual,early,auto,early-audio,early-video}
;answerdelay=0
;sip_autoanswer={yes, no}
;sip_autoanswer_beep={off, on, local}

There used to be only this

;answermode={manual,early,auto}

What is sip_autoanswer?

juha-h:
OK, I understood that sip_autoanswer relates to SIP request auto answer headers, but what still remains unclear to me is, what is the relationship between answermode and sip_autoanswer.  If sip_autoanswer is yes, does it override answermode and answerdelay values?

juha-h:
According to wiki, answerdelay values are seconds.  Based on my tests, they are milliseconds.  Sorry about writing all this under the same issue.

negbie:
Noticed the same. I think it's because of this:
https://github.com/baresip/baresip/blob/ae5f7ba18b3fe47a368d230fb52eff6c4f4867ca/src/ua.c#L424

negbie:
I would say that condition should be removed from ANSWERMODE_MANUAL case.

juha-h:
Eugen Biegler writes:

> I would say that condition should be removed from ANSWERMODE_MANUAL
> case.

I agree.

cspiel1:
- Sorry, it should be milli seconds. This is right. --> I will correct this in the wiki.
- I also once thought about in which answer mode adelay should be applicable. You can think about it, and we could discuss on Monday. Note, that menu.c decides when the ring-tone should be started or not! If we change something in ua.c then we have also do something in function `play_incoming)()` in menu.c.
- We have also the EARLY media answer modes. The adelay should also be applicable for these.

juha-h:
According to wiki, "user has to answer the call"  in all answermode cases except "auto".  That is how it should be independent of what answerdelay value is.  That allows user to change from auto to the other modes without need to change answerdelay value.

cspiel1:
Ok, but there are also these scenarios which are currently allowed:

- If adelay is a positive value (let's say a few seconds) then the ringtone will start and the user has the chance to accept the call before the answer delay elapsed.
- In addition also early media should be possible. Early media + adelay + manual accept if the user is faster than the adelay elapses.

[...]

PR #1475, opened 2021-05-31
ua, menu: move auto answer delay handling to menu (#1474)

cspiel1:
This simplifies auto answer mode and the handling for the answer delay.

- The account answer delay is handled in menu instead of ua.c.
- The `ANSWERMODE_AUTO` simply answer the call immediately.

After this PR is merged, I will update the wiki.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -129,6 +129,8 @@ const char *account_call_transfer(const struct account *acc);
 const char *account_extra(const struct account *acc);
 int account_uri_complete(const struct account *acc, struct mbuf *buf,
 			 const char *uri);
+int account_adelay(const struct account *acc);
+void account_set_adelay(struct account *acc, int adelay);
 bool account_sip_autoanswer(const struct account *acc);
 void account_set_sip_autoanswer(struct account *acc, bool allow);
 enum sipansbeep account_sipansbeep(const struct account *acc);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -129,8 +129,8 @@ const char *account_call_transfer(const struct account *acc);
 const char *account_extra(const struct account *acc);
 int account_uri_complete(const struct account *acc, struct mbuf *buf,
 			 const char *uri);
-int account_adelay(const struct account *acc);
-void account_set_adelay(struct account *acc, int adelay);
+int account_answerdelay(const struct account *acc);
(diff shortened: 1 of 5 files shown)


juha-h:
I gave this pr a try.  answermode/answerdelay settings seem work fine,
but then call comes in and answermode is manual, I don't hear ringing
sound.

cspiel1:
You're right! This was to less. I suggest you have an account like this:
```
"A" <sip:A@localhost>;regint=0;answermode=manual;answerdelay=2000
```
It does not ring.

I have also to unify the answer delay from account and SIP header to type `int32_t`. And have to use function `call_set_answer_delay()` to override the (unset) answer delay of the call by the accounts answer delay.

With this and the account above first the autoanswer.wav is played first, then the ringtone starts. The call is automatically answered after 2000 ms. If the answer delay is shorter than 1000ms, then no ringtone starts. It is too short for a ringtone. The user is able to manually accept the call before the answer delay elapsed.

modules/menu/menu.c
```diff
@@ -422,7 +422,7 @@ static void start_sip_autoanswer(struct call *call)
 
 		if (!beep)
 			beep = menu_play(call, "sip_autoanswer_aufile",
-					 "autoanswer.wav", 1);
+					 "autonswer.wav", 1);
```
alfredh, line 425: typo ?

include/baresip.h
```diff
@@ -129,6 +129,8 @@ const char *account_call_transfer(const struct account *acc);
 const char *account_extra(const struct account *acc);
 int account_uri_complete(const struct account *acc, struct mbuf *buf,
 			 const char *uri);
+int account_adelay(const struct account *acc);
```
alfredh, line 132: it is difficult to know what _a_ in _adelay_ means, without knowing the whole context.
perhaps better to write the full name, e.g. "answerdelay"

modules/menu/menu.c
cspiel1: Oh, no. Right! Didn't saw this.

include/baresip.h
cspiel1: Yes, I will change this to:
```
int account_answerdelay(const struct account *acc);
void account_set_answerdelay(struct account *acc, int adelay);
```
Because this are API functions.

The field name in `struct account` should stay short: `adelay`

PR #1479, opened 2021-06-02
ausine: support for multiple samplerates

alfredh:
support for multiple samplerates.

@sreimers what do you think ?


sreimers:
Looks good, thanks!

PR #1481, opened 2021-06-03
pkg-config version check

sreimers:
Depends on https://github.com/baresip/re/pull/107 and https://github.com/baresip/rem/pull/10

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,9 @@ ifeq ($(V),)
 HIDE=@
 endif
 
+LIBRE_MIN	:= 2.0.1-dev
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -18,6 +18,9 @@ ifeq ($(V),)
 HIDE=@
 endif
 
+MIN_LIBRE	:= 2.0.1-dev
+MIN_LIBREM	:= 1.0.0-dev
```
sreimers, line 22:
```suggestion
LIBREM_MIN	:= 1.0.0-dev
```
perhaps this looks better?

juha-h:
Looks good to me.

PR #1495, opened 2021-06-15
SIP redirect callbackfunction

cHuberCoffee:
In case of a automated call forwarding it is nice to inform baresip about the current status.
For outgoing calls a new callback function gets executed to get the information of a "blind" transfer into baresip.

Depends on PR [re#111](https://github.com/baresip/re/pull/111)


alfredh:
@sreimers could you please merge to re patch ?

sreimers:
Done

PR #1497, opened 2021-06-15
tls: verify server only if ca cert is provided

sreimers:


cspiel1:
You already know that we prefer security by default. TLS makes not much sense without certificate verification.

But we can live with this change.

juha-h:
Christian Spielberger writes:

> You already know that we prefer security by default. TLS makes not
> much sense without certificate verification. 
> 
> But we can live with this change.

I didn't understand the above.  In my test baresip didn't make any
complain when config had:

sip_certificate		/home/jh/.baresip/cert.pem
#sip_cafile		/home/jh/.baresip/ca_list.pem
sip_verify_server	yes

It should be an error when user asks to verify server, but does not give
ca information.

sreimers:
I close this PR, since the `re/sip/transp: ws_conn_send` and `re/http/client` needs in any case a way to disable server_verify first and for `re/sip/transp: ws_conn_send`  there is also no way to specify ca file/path.

PR #1500, opened 2021-06-17
ipv6ll scope id

cspiel1:
Second part for IPv6ll support. Setting the scope ID.

Has to be re-based after merge of https://github.com/baresip/baresip/pull/1473


alfredh:
this PR depends on another PR.

can you please close it and re-open after the original PR has been merged ?

I think it is bad practice to have a lot of PR's open that depends on eachother

PR #1502, opened 2021-06-18
uag,menu: add commands to enable/disable UDP/TCP/TLS

cspiel1:


juha-h:
Rather than having a separate API function for each transport, it would be better to have only one (enable_transport) with transport as argument.  Then new API functions would not be needed for ws and wss transports.  Same for menu commands.

Finally, transport enabling should be per account rather than globally.

cspiel1:
Right, will create another PR to have only one command with two parameters. This is a good idea.

Per account is not possible. The transport objects which opens listen sockets are system wide. An incoming call is detected on a listen socket which is related to a transport object which can be of type UDP, TCP, TLS, WS, WSS. The UA/account selection for e.g. an incoming call is done after the transport object is already known.

juha-h:
Christian Spielberger writes:

> Per account is not possible. The transport objects which opens listen
> sockets are system wide. An incoming call is detected on a listen
> socket which is related to a transport object which can be of type
> UDP, TCP, TLS, WS, WSS. The UA/account selection for e.g. an incoming
> call is done after the transport object is already known.

OK. I was thinking of situation where one account is used to
communicate, say with a PSTN gateway, that only supports UDP, and
another account which is used for secure (TLS) communication with work
peers.

PR #1504, opened 2021-06-20
attended transfer on menu

mbattista:
Here is a PR with the attended transfer in the menu module.

It worked for me with a Starface telephony system.

For two more points I need help (if they are needed)
1.) What should be the behavior if the telephony system does not support the replace header?
2.) Are there any more possibilities how a call can be ended, in which I need to clean the attended transfer in? As of now, I clean it when either person on the attended call hung up, or if the call is transferred.

### DIFF
--- a/modules/menu/dynamic_menu.c
+++ b/modules/menu/dynamic_menu.c
@@ -228,6 +228,52 @@ static int call_xfer(struct re_printf *pf, void *arg)
 }
 
 
+static int call_att_xfer(struct re_printf *pf, void *arg)
+{
+	struct menu *menu = menu_get();
+	struct call *call = uag_call_find(menu->callid);
+
+	if (menu->attended_callid) {
+		re_hprintf(pf, "attended transfer already in process\n");
+		return ENOENT;
+	} else {
+		if (!call_is_onhold(call)) {
+			re_hprintf(pf, "please hold your active call\n");
+			return ENOENT;
+		}
+		str_dup(&menu->attended_callid, menu->callid);
+		dial_handler(pf, arg);
+	}
+
+	return 0;
+}
+
+
+static int transfer_att_xfer(struct re_printf *pf, void *unused)
+{
+	struct menu *menu = menu_get();
+	struct call *call = uag_call_find(menu->callid);
+	struct call *attended_call;
+	(void)unused;
+
+	if (menu->attended_callid) {
+		attended_call = uag_call_find(menu->attended_callid);
+		if (!call_is_onhold(call)) {
+			re_hprintf(pf, "please hold your active call\n");
+			return ENOENT;
+		}
+		mem_deref(menu->attended_callid);
+		menu->attended_callid = NULL;
+		call_replace_transfer(attended_call, call);
+	} else {
+		re_hprintf(pf, "no attended to transfer call\n");
+		return ENOENT;
+	}
+
+	return 0;
+}
+
+
 static int call_video_debug(struct re_printf *pf, void *arg)
 {
 	const struct cmd_arg *carg = arg;
@@ -334,20 +380,22 @@ static int digit_handler(struct re_printf *pf, void *arg)
 /*Dynamic call menu*/
 static const struct cmd callcmdv[] = {
 
-{"aubitrate",    0,  CMD_PRM, "Set audio bitrate",    set_audio_bitrate    },
-{"audio_debug", 'A',       0, "Audio stream",         call_audio_debug     },
-{"callfind",     0,  CMD_PRM, "Find call <callid>",   cmd_find_call        },
-{"hold",        'x',       0, "Call hold",            cmd_call_hold        },
-{"line",        '@', CMD_PRM, "Set current call <line>", set_current_call  },
-{"mute",        'm',       0, "Call mute/un-mute",    call_mute            },
-{"reinvite",    'I',       0, "Send re-INVITE",       call_reinvite        },
-{"resume",      'X',       0, "Call resume",          cmd_call_resume      },
-{"sndcode",      0,  CMD_PRM, "Send Code",            send_code            },
-{"statmode",    'S',       0, "Statusmode toggle",    toggle_statmode      },
-{"transfer",    't', CMD_PRM, "Transfer call",        call_xfer            },
-{"video_debug", 'V',       0, "Video stream",         call_video_debug     },
-{"videodir",      0, CMD_PRM, "Set video direction",  set_video_dir        },
-{"medialdir",     0, CMD_PRM, "Set local media direction",  set_media_ldir },
+{"aubitrate",      0, CMD_PRM, "Set audio bitrate",        set_audio_bitrate },
+{"audio_debug",  'A',       0, "Audio stream",             call_audio_debug  },
+{"callfind",       0, CMD_PRM, "Find call <callid>",       cmd_find_call     },
+{"hold",         'x',       0, "Call hold",                cmd_call_hold     },
+{"line",         '@', CMD_PRM, "Set current call <line>",  set_current_call  },
+{"mute",         'm',       0, "Call mute/un-mute",        call_mute         },
+{"reinvite",     'I',       0, "Send re-INVITE",           call_reinvite     },
+{"resume",       'X',       0, "Call resume",              cmd_call_resume   },
+{"sndcode",        0, CMD_PRM, "Send Code",                send_code         },
+{"statmode",     'S',       0, "Statusmode toggle",        toggle_statmode   },
+{"transfer",     't', CMD_PRM, "Transfer call",            call_xfer         },
+{"att_transfer", 'T', CMD_PRM, "Attended call transfer",   call_att_xfer     },
+{"forw_transfer",'F',       0, "Forward attended call",    transfer_att_xfer },
+{"video_debug",  'V',       0, "Video stream",             call_video_debug  },
+{"videodir",       0, CMD_PRM, "Set video direction",      set_video_dir     },
+{"medialdir",      0, CMD_PRM, "Set local media direction", set_media_ldir   },
 
 /* Numeric keypad for DTMF events: */
 {NULL, '#',         0, NULL,                  digit_handler         },
--- a/modules/menu/dynamic_menu.c
+++ b/modules/menu/dynamic_menu.c
@@ -233,10 +233,10 @@ static int call_attended_xfer(struct re_printf *pf, void *arg)
 	const struct cmd_arg *carg = arg;
 	struct call *call = menu_callcur();
 	struct call *attended_call;
-	struct pl callid = PL_INIT;
-	struct pl attended_callid = PL_INIT;
-	char attended_call_string[140], call_string[140];
+	struct pl topl = PL_INIT, frompl = PL_INIT;
+	char *to, *from;
 	bool ok = false;
+	int err;
 
 	const char *usage = "usage: /transfercall"
 			" from=<from-id>"
@@ -245,43 +245,43 @@ static int call_attended_xfer(struct re_printf *pf, void *arg)
 
 	(void)pf;
 
-	ok |= 0 == menu_param_decode(carg->prm, "from", &attended_callid);
-	ok |= 0 == menu_param_decode(carg->prm, "to", &callid);
+	ok |= 0 == menu_param_decode(carg->prm, "from", &frompl);
+	ok |= 0 == menu_param_decode(carg->prm, "to", &topl);
 	if (!ok) {
 		ok = 0 == re_regex(carg->prm, str_len(carg->prm),
-			"[^ ]*[ \t\r\n]*[^ ]*", &attended_callid, NULL,
-								&callid);
+			"[^ ]*[ \t\r\n]*[^ ]*", &frompl, NULL, &topl);
 	}
 
 	if (!ok) {
 		(void) re_hprintf(pf, "%s\n", usage);
 		return EINVAL;
 	}
-	pl_strcpy(&attended_callid, attended_call_string,
-					sizeof(attended_call_string));
+	pl_strdup(&from, &frompl);
 
-	if (callid.l > 1) {
-		pl_strcpy(&callid, call_string, sizeof(call_string));
-		call = uag_call_find(call_string);
+	if (pl_isset(&topl)) {
+		pl_strdup(&to, &topl);
+		call = uag_call_find(to);
 		if (!call) {
-			warning("no call with call-id %s found\n",
-							call_string);
+			warning("menu: no call with call-id %s found\n", to);
 		}
 	}
 
 	if (!call_is_onhold(call)) {
-		warning("please hold your active call\n");
+		warning("menu: please hold call %s\n", call_id(call));
 		return ENOENT;
 	}
 
-	attended_call = uag_call_find(attended_call_string);
+	attended_call = uag_call_find(from);
 	if (!attended_call) {
-		warning("no call with call-id %s found\n",
-						attended_call_string);
+		warning("menu: no call with call-id %s found\n", from);
 		return ENOENT;
 	}
 
-	return call_replace_transfer(attended_call, call);
+	err = call_replace_transfer(attended_call, call);
+
+	mem_deref(from);
+	mem_deref(to);
+	return err;
 }
 
 
@@ -402,7 +402,7 @@ static const struct cmd callcmdv[] = {
 {"sndcode",      0,  CMD_PRM, "Send Code",            send_code            },
 {"statmode",    'S',       0, "Statusmode toggle",    toggle_statmode      },
 {"transfer",    't', CMD_PRM, "Transfer call",        call_xfer            },
-{"transfercall", 'T', CMD_PRM, "Transfers <call-id> to ative call",
+{"transferatt", 'T', CMD_PRM, "Transfers <call-id> to ative call",
 						      call_attended_xfer   },
 {"video_debug", 'V',       0, "Video stream",         call_video_debug     },
 {"videodir",      0, CMD_PRM, "Set video direction",  set_video_dir        },
--- a/modules/menu/dynamic_menu.c
+++ b/modules/menu/dynamic_menu.c
@@ -228,51 +228,60 @@ static int call_xfer(struct re_printf *pf, void *arg)
 }
 
 
-static int call_att_xfer(struct re_printf *pf, void *arg)
+static int call_attended_xfer(struct re_printf *pf, void *arg)
 {
-	struct menu *menu = menu_get();
-	struct call *call = uag_call_find(menu->callid);
-
-	if (menu->attended_callid) {
-		re_hprintf(pf, "attended transfer already in process\n");
-		return ENOENT;
-	}
-	else {
-		if (!call_is_onhold(call)) {
-			re_hprintf(pf, "please hold your active call\n");
-			return ENOENT;
-		}
-		str_dup(&menu->attended_callid, menu->callid);
-		dial_handler(pf, arg);
-	}
-
-	return 0;
-}
-
-
-static int transfer_att_xfer(struct re_printf *pf, void *unused)
-{
-	struct menu *menu = menu_get();
-	struct call *call = uag_call_find(menu->callid);
+	const struct cmd_arg *carg = arg;
+	struct call *call = menu_callcur();
 	struct call *attended_call;
-	(void)unused;
+	struct pl callid = PL_INIT;
+	struct pl attended_callid = PL_INIT;
+	char attended_call_string[140], call_string[140];
+	bool ok = false;
 
-	if (menu->attended_callid) {
-		attended_call = uag_call_find(menu->attended_callid);
-		if (!call_is_onhold(call)) {
-			re_hprintf(pf, "please hold your active call\n");
-			return ENOENT;
-		}
-		mem_deref(menu->attended_callid);
-		menu->attended_callid = NULL;
-		call_replace_transfer(attended_call, call);
+	const char *usage = "usage: /transfercall"
+			" from=<from-id>"
+			" [to=to-id]\n"
+			"/transfercall <from-id> [to-id]\n";
+
+	(void)pf;
+
+	ok |= 0 == menu_param_decode(carg->prm, "from", &attended_callid);
+	ok |= 0 == menu_param_decode(carg->prm, "to", &callid);
+	if (!ok) {
+		ok = 0 == re_regex(carg->prm, str_len(carg->prm),
+			"[^ ]*[ \t\r\n]*[^ ]*", &attended_callid, NULL,
+								&callid);
 	}
-	else {
-		re_hprintf(pf, "no attended to transfer call\n");
+
+	if (!ok) {
+		(void) re_hprintf(pf, "%s\n", usage);
+		return EINVAL;
+	}
+	pl_strcpy(&attended_callid, attended_call_string,
+					sizeof(attended_call_string));
+
+	if (callid.l > 1) {
+		pl_strcpy(&callid, call_string, sizeof(call_string));
+		call = uag_call_find(call_string);
+		if (!call) {
+			warning("no call with call-id %s found\n",
+							call_string);
+		}
+	}
+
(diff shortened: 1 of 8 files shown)


juha-h:
mbattista writes:

> For two more points I need help (if they are needed)

> 1.) What should be the behavior if the telephony system does not
> support the replace header?

From RFC 5589:

   The Transferor knows that both the Transferee and the Transfer Target
   support the Replaces header from the Supported: replaces header
   contained in the 200 OK responses from both.

My understanding of this is that if replaces is not supported, the
transferor cannot do this kind of attended transfer.

mbattista:
juha-h writes:
> My understanding of this is that if replaces is not supported, the transferor cannot do this kind of attended transfer.

This is my understanding as well. So I should check the `Supported` header in both calls before finalizing the transfer?
If so, there are more questions on my part.
3.) Where can I find the `Supported` header? Is it in the custom headers of the call?
4.) How should the notice look like if the `replace` header is not supported? _Attended transfer is not supported. Please use the normal transfer_?
5.) Should there be an option to overrule the `Supported` header? Like a force_attended_transfer, if someone knows that it works, but knows that sometimes the header is missing?

juha-h:
mbattista writes:

> This is my understanding as well. So I should check the `Supported`
> header in both calls before finalizing the transfer?

Yes.

> If so, there are more questions on my part.
> 3.) Where can I find the `Supported` header? Is it in the custom
> headers of the call?

I haven't tried, but looks like sip_msg_hdr() API call gives you access
to SIP message headers.

> 4.) How should the notice look like if the `replace` header is not
> supported? _Attended transfer is not supported. Please use the normal
> transfer_?

> 5.) Should there be an option to overrule the `Supported` header? Like a force_attended_transfer, if someone knows that it works, but knows that sometimes the header is missing?

I don't like the idea of trying to fix in baresip bugs that other SIP
UAs have.

modules/menu/dynamic_menu.c
```diff
@@ -334,20 +382,22 @@ static int digit_handler(struct re_printf *pf, void *arg)
...
+{"hold",         'x',       0, "Call hold",                cmd_call_hold     },
+{"line",         '@', CMD_PRM, "Set current call <line>",  set_current_call  },
+{"mute",         'm',       0, "Call mute/un-mute",        call_mute         },
+{"reinvite",     'I',       0, "Send re-INVITE",           call_reinvite     },
+{"resume",       'X',       0, "Call resume",              cmd_call_resume   },
+{"sndcode",        0, CMD_PRM, "Send Code",                send_code         },
+{"statmode",     'S',       0, "Statusmode toggle",        toggle_statmode   },
+{"transfer",     't', CMD_PRM, "Transfer call",            call_xfer         },
+{"att_transfer", 'T', CMD_PRM, "Attended call transfer",   call_att_xfer     },
+{"forw_transfer",'F',       0, "Forward attended call",    transfer_att_xfer },
```
cspiel1, line 397: - Pls no underline symbols in command names!
- A lot of unrelated changed lines.

modules/menu/menu.c
```diff
@@ -604,6 +604,8 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 		}
 
 		if (!str_cmp(call_id(call), menu.callid)) {
+			mem_deref(menu.attended_callid);
+			menu.attended_callid = NULL;
```
cspiel1, line 608: You can write:
```
menu.attended_callid = mem_deref(menu.attended_callid);
```
And it is displaced here.

modules/menu/dynamic_menu.c
```diff
@@ -228,6 +228,54 @@ static int call_xfer(struct re_printf *pf, void *arg)
 }
 
 
+static int call_att_xfer(struct re_printf *pf, void *arg)
+{
+	struct menu *menu = menu_get();
+	struct call *call = uag_call_find(menu->callid);
+
+	if (menu->attended_callid) {
+		re_hprintf(pf, "attended transfer already in process\n");
```
cspiel1, line 237: Use `warning()` if you return an error code.

modules/menu/menu.c
mbattista: Thanks for the reminder with `mem_deref`.

Can you elaborate why you think it is displaced?
If you are in the progress of transferring a call and the other side of your active call hung up, you are no longer in the progress of transferring a call.

modules/menu/dynamic_menu.c
```diff
@@ -228,6 +228,54 @@ static int call_xfer(struct re_printf *pf, void *arg)
...
+		return ENOENT;
+	}
+	else {
+		if (!call_is_onhold(call)) {
+			re_hprintf(pf, "please hold your active call\n");
+			return ENOENT;
+		}
+		str_dup(&menu->attended_callid, menu->callid);
+		dial_handler(pf, arg);
```
cspiel1, line 246: The function `dial_handler()` is very high-level. And it should not be used from `dynamic_menu.c`.

The command `att_transfer` resp. function `call_att_xfer` should be put to static_menu.c. But is it really needed? Further, do we really need variable `attended_callid`? It would be also possible that the user specifies the `callid` for the `attended_call` in function `transfer_att_xfer()`.

So the attended transfer could start by calling command `dial`. Then baresip prints/returns the call-id (see UA events!).
Then user calls `forw_transfer` with the call-id.

mbattista: So change the workflow to
- put active call on hold
- start a new call via dial
- transfer the calls with a call-id parameter

I wanted to keep it simple for the user but I am open for changes.

modules/menu/dynamic_menu.c
mbattista: I thought it is fine since there is `video_debug` and `audio_debug` already. Will remove them.

modules/menu/dynamic_menu.c
cspiel1: We have two options for the second call.
- Full flexibility by using command `dial` or `dialdir` from static_menu.c
   - This approach needs less changes.
   - All features for an outgoing call like selecting audio/video dir, SIP auto answer. But maybe this is not needed for an attended transfer and only confusing.
- Your `call_att_xfer` in dynamic_menu.c would allow more control for this use-case.
   - I agree that this should stay in the dynamic_menu.c, because another call (the first call) has to be present already.
   - This function should not use `dial_handler`.
   - Select of the current call is ok. But an optional argument `[callid=id]` would be advantageous.
   - The ua is then `call_ua(call)`.
   - Use `account_uri_complete()` and `ua_connect()` instead of `dial_handler()`!
   - Maybe we should also hold the call if it is not on-hold.

Note: `re_hprintf` is ok.

modules/menu/menu.c
cspiel1: `if (!str_cmp(call_id(call), menu.callid))` means "if the call terminated is the current call".

I think you need another if statement:
```c
if (!str_cmp(call_id(call), menu.attended_callid)) {
     menu.attended_callid = mem_deref(menu.attended_callid);
}
```

cspiel1:
Btw, what happens if we put an incoming call on hold?

And how could we make an attended transfer of an incoming call?

mbattista:
> Btw, what happens if we put an incoming call on hold?

Can you further elaborate which could be problematic in this case?

> And how could we make an attended transfer of an incoming call?

In my current workflow any call could be transferred to the active dialed call. With the new workflow I have to check what happens and what the RFC specifies. But since you replace your call with another call, I think this should be possible.


Thank you very much for your Input.

cspiel1:
Yes, for incoming, like you wrote:
- check RFC for attended call transfer of a call that is in state "incoming". (ringing locally, or playing early media)
- try what baresip does for peer-to-peer calls
- try also with a SIP server

Also for outgoing calls, that are not established already?

modules/menu/dynamic_menu.c
cspiel1: Some already said the same to a prev PR of mine.

modules/menu/dynamic_menu.c
mbattista: On my rewrite I will try the first option.

If you think that the callid could change unexpectedly should the ids of both calls be parameter of the `att_xfer` function?

cspiel1: Ok.

Yes, for the first option the variable `attended_callid` should be removed and given as parameter. The second callid could be optional. Add also a usage text like in `cmd_answerdir()`! It should be enough to support this variant:
```
/atransfer <from-id> [to-id]
```
which calls your `transfer_att_xfer()`.

mbattista:
Improvements:
- attended transfer now with parameters call-id of the transfer-call and optional call-id of the transfer-target
- both calls need to be in state established
- header is checked if replaces is under the supported headers (hopefully at the right spots)

Need help with
- [ ] try what baresip does for peer-to-peer calls (how to establish a peer-to-peer call?)
- [ ] try also with a SIP server (I just have a starface PBX and a fritz.box to connect to, I can only try on those)
- [ ] RFC compatibilities of the pull request

modules/menu/dynamic_menu.c
```diff
@@ -228,51 +228,60 @@ static int call_xfer(struct re_printf *pf, void *arg)
...
+	}
+	pl_strcpy(&attended_callid, attended_call_string,
+					sizeof(attended_call_string));
 
-	if (menu->attended_callid) {
-		attended_call = uag_call_find(menu->attended_callid);
-		if (!call_is_onhold(call)) {
-			re_hprintf(pf, "please hold your active call\n");
-			return ENOENT;
+	if (callid.l > 1) {
```
cspiel1, line 263: pls use `pl_isset()`!

modules/menu/dynamic_menu.c
```diff
@@ -228,51 +228,60 @@ static int call_xfer(struct re_printf *pf, void *arg)
...
+static int call_attended_xfer(struct re_printf *pf, void *arg)
 {
-	struct menu *menu = menu_get();
-	struct call *call = uag_call_find(menu->callid);
+	const struct cmd_arg *carg = arg;
+	struct call *call = menu_callcur();
+	struct call *attended_call;
+	struct pl callid = PL_INIT;
+	struct pl attended_callid = PL_INIT;
+	char attended_call_string[140], call_string[140];
```
cspiel1, line 238: Pls use a dynamic buffers, and functions pl_strdup() and mem_deref().
I would also prefer short local variables.
`cidatt, cid, cidattpl, cidpl`

modules/menu/dynamic_menu.c
```diff
@@ -228,51 +228,60 @@ static int call_xfer(struct re_printf *pf, void *arg)
...
-	if (menu->attended_callid) {
-		attended_call = uag_call_find(menu->attended_callid);
-		if (!call_is_onhold(call)) {
-			re_hprintf(pf, "please hold your active call\n");
-			return ENOENT;
+	if (callid.l > 1) {
+		pl_strcpy(&callid, call_string, sizeof(call_string));
+		call = uag_call_find(call_string);
+		if (!call) {
+			warning("no call with call-id %s found\n",
```
cspiel1, line 267: Start the warnings with "menu: "!

modules/menu/dynamic_menu.c
```diff
@@ -382,22 +391,22 @@ static int digit_handler(struct re_printf *pf, void *arg)
...
+{"callfind",     0,  CMD_PRM, "Find call <callid>",   cmd_find_call        },
+{"hold",        'x',       0, "Call hold",            cmd_call_hold        },
+{"line",        '@', CMD_PRM, "Set current call <line>", set_current_call  },
+{"mute",        'm',       0, "Call mute/un-mute",    call_mute            },
+{"reinvite",    'I',       0, "Send re-INVITE",       call_reinvite        },
+{"resume",      'X',       0, "Call resume",          cmd_call_resume      },
+{"sndcode",      0,  CMD_PRM, "Send Code",            send_code            },
+{"statmode",    'S',       0, "Statusmode toggle",    toggle_statmode      },
+{"transfer",    't', CMD_PRM, "Transfer call",        call_xfer            },
+{"transfercall", 'T', CMD_PRM, "Transfers <call-id> to ative call",
```
cspiel1, line 405: All commands in dynamic_menu.c are related to calls. Should we rename the command to "transferatt"?

src/call.c
```diff
@@ -790,8 +793,18 @@ int call_alloc(struct call **callp, const struct config *cfg, struct list *lst,
 	call->af     = prm->af;
 	call->ansadir = SDP_SENDRECV;
 	call->ansvdir = SDP_SENDRECV;
+	call->support_replaces = false;
 	call_decode_sip_autoanswer(call, msg);
 
+	/* check if replaces is supported on incoming calls */
+	hdr = sip_msg_hdr(msg, SIP_HDR_SUPPORTED);
+	if (hdr) {
+		if (!re_regex(hdr->val.p, hdr->val.l, "replaces", &v)) {
```
cspiel1, line 802:
```c
if (hdr && !re_regex(hdr->val.p, hdr->val.l, "replaces", &v)) {
```

src/call.c
```diff
@@ -790,8 +793,18 @@ int call_alloc(struct call **callp, const struct config *cfg, struct list *lst,
...
+			info("call %s supports attended transfer\n", call->id);
```
cspiel1, line 803: - Should be a `debug()` line, or remove it after your tests!
- Start the string with `call: `!

src/call.c
```diff
@@ -1606,6 +1620,15 @@ static void sipsess_estab_handler(const struct sip_msg *msg, void *arg)
...
 
+	/* check if replaces is supported on outgoing call */
+	hdr = sip_msg_hdr(msg, SIP_HDR_SUPPORTED);
+	if (hdr) {
+		if (!re_regex(hdr->val.p, hdr->val.l, "replaces", &v)) {
+			info("call %s supports attended transfer\n", call->id);
+			call->support_replaces = true;
+		}
+	}
+
```
cspiel1, line 1631: For outgoing calls we know that we support "replaces". Right?
Then you should call
```c
add_extension(ua, "replaces");
```
And initialize `support_replaces` with `true` for outgoing calls.

cspiel1 (changes requested):

cspiel1:
Please squash the commits and separate then two one commit for call.c and one for menu!

Please could you also refer the associated RFC in the commit message! Is it RFC-5589?

src/call.c
```diff
@@ -1606,6 +1620,15 @@ static void sipsess_estab_handler(const struct sip_msg *msg, void *arg)
 
 	call_stream_start(call, true);
 
+	/* check if replaces is supported on outgoing call */
+	hdr = sip_msg_hdr(msg, SIP_HDR_SUPPORTED);
```
alfredh, line 1624: please check if this api can be used:

```c
bool sip_msg_hdr_has_value(const struct sip_msg *msg, enum sip_hdrid id,
                           const char *value);
```

src/call.c
mbattista: I was not sure if I could add the extension replaces header for the baresip client.

If I understand https://datatracker.ietf.org/doc/html/rfc3891 correctly the client should send the header if he can send the `Replaces` header.
There will never be an call with an `Replaces` header send to the client because of it?
Will the PBX rewrite the headers if they dont support it?

modules/menu/dynamic_menu.c
mbattista: if I put one call on hold and directly call the transfer it seems that _sometimes_ there are problems with the transfer.

I cannot reproduce it constantly, so I leave the holding to the user.

src/call.c
mbattista: I changed it in favor of `sip_msg_hdr_has_value`

src/call.c
mbattista: Added the configuration entry if baresip should send the replaces header.

I still think that I need to know if the telephone system I am connected to supports replaces.
If you think the config value is enough I can remove it and just check the config value.

modules/menu/dynamic_menu.c
```diff
@@ -228,6 +228,63 @@ static int call_xfer(struct re_printf *pf, void *arg)
...
+
+	ok |= 0 == menu_param_decode(carg->prm, "from", &attended_callid);
+	ok |= 0 == menu_param_decode(carg->prm, "to", &callid);
+	if (!ok) {
+		ok = 0 == re_regex(carg->prm, str_len(carg->prm),
+			"[^ ]*[ \t\r\n]*[^ ]*", &attended_callid, NULL,
+								&callid);
+	}
+
+	if (!ok) {
```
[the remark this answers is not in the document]
mbattista: Are you sure?

I can call the function successful only with `from=call-id.`
https://github.com/baresip/baresip/blob/master/modules/menu/dynamic_menu.c#L260 here the optional value also is checked.

mbattista:
> Please squash the commits and separate then two one commit for call.c and one for menu!
> 
> Please could you also refer the associated RFC in the commit message! Is it RFC-5589?

Squashing the commits now or when all conversations are resolved?

src/call.c
mbattista: thanks, this works

modules/menu/dynamic_menu.c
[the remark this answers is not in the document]
cspiel1: yes. And `set_media_ldir()` also is not correct.

cspiel1:
> Need help with
> 
> * [ ]  try what baresip does for peer-to-peer calls (how to establish a peer-to-peer call?)
- You need the config line: https://github.com/baresip/baresip/wiki/Configuration#sip_listen
- A registrar-less account in file accounts.
```
<sip:alice@office>;regint=0
```
- then dial from other device to the IP of your baresip instance


> * [ ]  try also with a SIP server (I just have a starface PBX and a fritz.box to connect to, I can only try on those)
> * [ ]  RFC compatibilities of the pull request

src/call.c
```diff
@@ -2326,15 +2342,39 @@ int call_transfer(struct call *call, const char *uri)
 /**
  * Transfer the call to a target SIP uri and replace the source call
  *
- * @param call  Call object
- * @param uri   Target SIP uri
+ * @param call  Call object target
+ * @param call  Call object source
```
cspiel1, line 2346: wrong param name

src/ua.c
```diff
@@ -816,6 +816,9 @@ static int create_register_clients(struct ua *ua)
 	if (uag_cfg() && str_isset(uag_cfg()->uuid))
 		add_extension(ua, "gruu");
 
+	if (uag_cfg() && uag_cfg()->sip_replaces_hdr)
+		add_extension(ua, "replaces");
```
alfredh, line 820: if this feature is supported by us, it is probably good to always have it enabled.

it is nice to have many config options, but too many can make it harder to use for average Joe :)

src/call.c
cspiel1: As @alfredh wrote. You don't need a config option. Baresip should put "replaces" into the Supported header.

In line 1619 baresip checks from incoming SIP requests if the remote supports replaces.

src/call.c
mbattista: should work now

cspiel1:
Currently one of `from` or `to` could be specified. This is not correct. What about:
```c
	if (!str_isset(carg->prm)) {
		(void) re_hprintf(pf, "%s\n", usage);
		return EINVAL;
	}

	ok = 0 == menu_param_decode(carg->prm, "from", &frompl);
	(void)menu_param_decode(carg->prm, "to", &topl);
	if (!ok) {
		ok = 0 == re_regex(carg->prm, str_len(carg->prm),
			"[^ ]*[ \t\r\n]*[^ ]*", &frompl, NULL, &topl);
	}

	if (ok)
		err = pl_strdup(&from, &frompl);
	else
		err = str_dup(&from, carg->prm);

	if (err)
		return err;
```

mbattista:
> There are some examples in RFC-5589 with header
> 
> ```
> Refer-To: %s?Replaces=%s\r\n
> ```
> 
> like you implemented in function `call_replace_transfer()`. Where did you see this header?
> What is the difference to the `Replaces: ...` header?

I oriented on the https://www.tech-invite.com/fo-sip/tinv-fo-sip-service-05.html sheet, which looked like it was in line with the RFC 5359.
I wanted to handle only the "Bob" part of the attended transfer, mainly the [Refer-To part](https://www.tech-invite.com/fo-sip/tinv-fo-sip-service-05-15.html).
I did not handle the ["Alice" part](https://www.tech-invite.com/fo-sip/tinv-fo-sip-service-05-19.html), which actually sends the Replaces header, and ["Carol" part](https://www.tech-invite.com/fo-sip/tinv-fo-sip-service-05-20.html), which accepts the Replace Invite. If those feature should also be implemented in the same pull request, I can look into it.

mbattista:
> Currently one of `from` or `to` could be specified. This is not correct. What about:
> 
> ```c
> 	if (!str_isset(carg->prm)) {
> 		(void) re_hprintf(pf, "%s\n", usage);
> 		return EINVAL;
> 	}
> 
> 	ok = 0 == menu_param_decode(carg->prm, "from", &frompl);
> 	(void)menu_param_decode(carg->prm, "to", &topl);
> 	if (!ok) {
> 		ok = 0 == re_regex(carg->prm, str_len(carg->prm),
> 			"[^ ]*[ \t\r\n]*[^ ]*", &frompl, NULL, &topl);
> 	}
> 
> 	if (ok)
> 		err = pl_strdup(&from, &frompl);
> 	else
> 		err = str_dup(&from, carg->prm);
> 
> 	if (err)
> 		return err;
> ```

Thanks. Will do it like that. I wanted to debug it (hence the info line which will be removed) got distracted and forgot about it. Sorry.

juha-h:
This rfc has attended transfer call flow example:

https://datatracker.ietf.org/doc/rfc5359/

cspiel1:
@mbattista should we complete this now?

mbattista:
Sorry, I had no spare time in the last days. I will try to look at it again this weekend.

alfredh:
is there any progress here ?

if you dont have time to work on it, it is better to close the PR for now
and reopen it when the code has been updated.

cspiel1:
I already thought to grab this, since the feature seems to be useful for everybody.
The problem is that more RFC research has to be done. Which is the correct SIP header extension for "Refer-To"?  Mode concretely? https://github.com/baresip/baresip/pull/1504#issuecomment-870618204

Has somebody knowledge about this?

I`ll keep this on my queue. We should close it for now.

mbattista:
I am very sorry, that did not have the time to finish this at the hands right now. If I find time and I do not see any commits that handle it, I will try to contribute again as soon as I find a time.

https://www.tech-invite.com/fo-sip/tinv-fo-sip-service-05.html from this my understanding is the following:
The transferor tells the transferee via `Refer-To` header to which call he will be refered [1](https://www.tech-invite.com/fo-sip/tinv-fo-sip-service-05-15.html) The transferee then sends a `Replace` header to the transfer target [2](https://www.tech-invite.com/fo-sip/tinv-fo-sip-service-05-19.html) The transfer target then ack and byes the calls.

mbattista:
> Which is the correct SIP header extension for "Refer-To"?

From my understanding the support for refer-to is handled via the `Allow` header, which then should contain a `REFER`?
https://datatracker.ietf.org/doc/html/rfc3515#page-4

cspiel1:
Ah, thanks. But baresip/re does not add a `Replace` header for the final step [2](https://www.tech-invite.com/fo-sip/tinv-fo-sip-service-05-19.html), right? This would be also needed for this PR, if I am right.

mbattista:
> Ah, thanks! But baresip/re does not add a Replace header for the final step 2, right? This would be also needed for this PR, if I am right.

Exactly this. I wanted to cover step one since the PBX handles the incoming replaces (if it supports it). Then the merge request got bigger, to cover all the RFC parts which is a more correct handling and I understand that it is necessary since not all users have a PBX who supports it or a PBX at all.

I would have gladly implemented it, but since I am no longer in home-office the commute takes up the time that I used to develop and even limits the time available on the weekends.

cspiel1:
I think we need also the `Replace` header to be able to test the whole use-case "attended transfer" with two (or three) baresip instances. This should be a separate commit (but maybe in the same PR).

We would need at least your help, because of your test setup. ;-)

PR #1512, opened 2021-06-23
Fix I2S compile error, use auframe

andreaswatch:
This PR makes the I2S module compile again using auframe.
However, the microphone does not seem to work.

### DIFF
--- a/modules/i2s/i2s_src.c
+++ b/modules/i2s/i2s_src.c
@@ -70,6 +70,7 @@ static void convert_pcm(struct ausrc_st *st, size_t i, size_t n)
 static void *read_thread(void *arg)
 {
 	struct ausrc_st *st = arg;
+	struct auframe af;
 
 	while (st->run) {
 		size_t i;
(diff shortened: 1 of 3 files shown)


cspiel1:
The github workflow shows:
```
./modules/i2s/i2s_play.c:71: line is too wide (88 - max 79)
./modules/i2s/i2s_src.c:89: has trailing tab(s)
```

Please fix this and push force!

modules/i2s/i2s_src.c
```diff
@@ -86,8 +86,12 @@ static void *read_thread(void *arg)
 			convert_pcm(st, i, n / 4);
 			i += (n / 4);
 		}
-
-		st->rh(st->sampv, st->sampc, st->arg);
+		
+		struct auframe af = {
```
cspiel1, line 90: Please move declaration to the top of the function body!

cspiel1:
The PR seems to be technical correct and the I2S source and playback should work. I see no changes that could break something.

You could debug in `convert_pcm()`. Print values, make adjustments for the volume! Also it could be that your MICs deliver different "kind" of values. Do also have an

Adafruit I2S MEMS Microphone Breakout - SPH0645LM4H ?

modules/i2s/i2s_src.c
sreimers: And please use auframe_init like for play.

andreaswatch:
Thanks guys, and sorry for all the commits. This is my first pull request and I need to get familiar with it.
@cspiel1 : I'll try to take a closer look into convert_pcm() tonight. I only have an INMP441here though.

cspiel1:
It is already merged which is absolutely okay. If there might be some adjustments necessary for the mic, then please do this by use of config values. E.g.
```c
uint32_t v = 100;   /* default value */
conf_get_u32(conf_cur(), "i2s_mic_xxx", &v);
```
And create another PR!

PR #1513, opened 2021-06-28
net,ua,uag: support for multiple network interfaces

cspiel1:
net,ua,uag: support for multiple network interfaces
    
This commit adds support for multiple network interfaces and local
IP addresses.

IPv6ll support will be added in a following PR.

This PR replaces https://github.com/baresip/baresip/pull/1473.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -709,6 +709,7 @@ int  net_dns_debug(struct re_printf *pf, const struct network *net);
 int  net_debug(struct re_printf *pf, const struct network *net);
 void net_laddr_apply(const struct network *net, net_ifaddr_h *ifh, void *arg);
 const struct sa *net_laddr_af(const struct network *net, int af);
+bool net_is_laddr(const struct network *net, struct sa *sa);
 struct dnsc     *net_dnsc(const struct network *net);
 
 
--- a/src/net.c
+++ b/src/net.c
@@ -10,10 +10,8 @@
 
 struct network {
 	struct config_net cfg;
-	struct sa laddr;
-#ifdef HAVE_INET6
-	struct sa laddr6;
-#endif
+	struct list laddrs;           /**< List of local addresses           */
+	bool   changed;               /**< Changed flag for list laddrs      */
 	struct tmr tmr;
 	struct dnsc *dnsc;
 	struct sa nsv[NET_MAX_NS];/**< Configured name servers           */
@@ -26,12 +24,20 @@ struct network {
 };
 
 
+struct laddr {
+	struct le le;
+	struct sa sa;
+	bool found;
+};
+
+
 static void net_destructor(void *data)
 {
 	struct network *net = data;
 
 	tmr_cancel(&net->tmr);
 	mem_deref(net->dnsc);
+	list_flush(&net->laddrs);
 }
 
 
@@ -137,6 +143,22 @@ static int net_dns_srv_get(const struct network *net,
 }
 
 
+static int list_add_laddr(struct list *list, const struct sa *sa)
+{
+	struct laddr *laddr;
+	if (!sa_isset(sa, SA_ADDR))
+		return 0;
+
+	laddr = mem_zalloc(sizeof(*laddr), NULL);
+	if (!laddr)
+		return ENOMEM;
+
+	laddr->sa = *sa;
+	list_append(list, &laddr->le, laddr);
+	return 0;
+}
+
+
 /*
  * Check for DNS Server updates
  */
@@ -175,6 +197,98 @@ static void ipchange_handler(void *arg)
 }
 
 
+static bool net_ifaddr_filter(struct network *net, const char *ifname,
+			      const struct sa *sa)
+{
+	const struct config_net *cfg = &net->cfg;
+
+	if (str_isset(cfg->ifname) && str_cmp(cfg->ifname, ifname))
+		return false;
+
+	if (!net_af_enabled(net, sa_af(sa)))
+		return false;
+
+	if (sa_is_loopback(sa))
+		return false;
+
+	return true;
+}
+
+
+static bool ifaddr_add(const char *ifname, const struct sa *sa, void *arg)
+{
+	struct network *net = arg;
+
+	if (net_ifaddr_filter(net, ifname, sa))
+		list_add_laddr(&net->laddrs, sa);
+
+	return false;
+}
+
+
+static bool laddr_cmp(struct le *le, void *arg)
+{
+	struct laddr *laddr = le->data;
+	struct sa *sa = arg;
+
+	return sa_cmp(&laddr->sa, sa, SA_ADDR);
+}
+
+
+static bool net_misses_laddr(const char *ifname, const struct sa *sa,
+			     void *arg)
+{
+	struct network *net = arg;
+	struct sa sac;
+
+	net->changed = false;
+	if (!net_ifaddr_filter(net, ifname, sa))
+		return false;
+
+	sa_cpy(&sac, sa);
+	net->changed =
+		NULL == list_apply(&net->laddrs, true, laddr_cmp, &sac);
+
+	return net->changed;
+}
+
+
+static bool sys_laddr_cmp(const char *ifname, const struct sa *sa, void *arg)
+{
+	struct laddr *laddr = arg;
+	(void) ifname;
+
+	laddr->found = sa_cmp(&laddr->sa, sa, SA_ADDR);
+	return laddr->found;
+}
+
+
+static bool sys_removed_net_laddr(struct le *le, void *arg)
+{
+	struct laddr *laddr = le->data;
+	(void) arg;
+
+	net_if_apply(sys_laddr_cmp, laddr);
+	return !laddr->found;
+}
+
+
+static int net_dst_is_source_addr(const struct sa *dst, const struct sa *ip)
+{
+	struct sa src;
+	int err;
+
+	err = net_dst_source_addr_get(dst, &src);
+	if (err)
+		return err;
+
+	if (!sa_cmp(ip, &src, SA_ADDR))
+		return ECONNREFUSED;
+
+	return 0;
+}
+
+
 /**
  * Check if local IP address(es) changed
  *
@@ -184,58 +298,41 @@ static void ipchange_handler(void *arg)
  */
 bool net_check(struct network *net)
 {
-	struct sa laddr;
-#ifdef HAVE_INET6
-	struct sa laddr6;
-#endif
-	bool change = false;
+	const struct config_net *cfg = &net->cfg;
+	struct sa sa;
 
 	if (!net)
 		return false;
 
-	laddr = net->laddr;
-#ifdef HAVE_INET6
-	laddr6 = net->laddr6;
-#endif
+	if (str_isset(cfg->ifname) && 0 == sa_set_str(&sa, cfg->ifname, 0)) {
+		struct laddr *laddr;
+		if (list_count(&net->laddrs) == 1) {
+			laddr = net->laddrs.head->data;
+			net->changed = !sa_cmp(&laddr->sa, &sa, SA_ADDR);
+		}
+		else {
+			net->changed = true;
+		}
 
-	if (str_isset(net->cfg.ifname)) {
+		if (!net->changed)
+			return false;
 
-		if (net_af_enabled(net, AF_INET))
-			net_if_getaddr(net->cfg.ifname, AF_INET, &net->laddr);
-
-#ifdef HAVE_INET6
-		if (net_af_enabled(net, AF_INET6))
-			net_if_getaddr(net->cfg.ifname, AF_INET6,
-				       &net->laddr6);
-#endif
-	}
-	else {
-		if (net_af_enabled(net, AF_INET))
-			net_default_source_addr_get(AF_INET, &net->laddr);
-
-#ifdef HAVE_INET6
-		if (net_af_enabled(net, AF_INET6))
-			net_default_source_addr_get(AF_INET6, &net->laddr6);
-#endif
+		info("Binding to IP address '%j'\n", &sa);
+		list_flush(&net->laddrs);
+		list_add_laddr(&net->laddrs, &sa);
+		return true;
 	}
 
-	if (sa_isset(&net->laddr, SA_ADDR) &&
-	    !sa_cmp(&laddr, &net->laddr, SA_ADDR)) {
-		change = true;
-		info("net: local IPv4 address changed: %j -> %j\n",
-		     &laddr, &net->laddr);
-	}
+	net_if_apply(net_misses_laddr, net);
+	net->changed |= NULL != list_apply(&net->laddrs, true,
+					   sys_removed_net_laddr, NULL);
+	if (!net->changed)
+		return false;
 
-#ifdef HAVE_INET6
-	if (sa_isset(&net->laddr6, SA_ADDR) &&
-	    !sa_cmp(&laddr6, &net->laddr6, SA_ADDR)) {
-		change = true;
-		info("net: local IPv6 address changed: %j -> %j\n",
-		     &laddr6, &net->laddr6);
-	}
-#endif
+	list_flush(&net->laddrs);
+	net_if_apply(ifaddr_add, net);
 
-	return change;
+	return net->changed;
 }
 
 
@@ -301,7 +398,7 @@ int net_alloc(struct network **netp, const struct config_net *cfg)
 	struct network *net;
 	struct sa nsv[NET_MAX_NS];
 	uint32_t nsn = ARRAY_SIZE(nsv);
-	char buf4[128] = "", buf6[128] = "";
+	struct le *le;
 	int err;
 
 	if (!netp || !cfg)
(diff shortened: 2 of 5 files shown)


src/ua.c
```diff
@@ -493,23 +494,31 @@ static bool require_handler(const struct sip_hdr *hdr,
...
-	struct pl af;
+	struct pl pl1, pl2;
+	char *addr;
 	int err;
 
+	*af = AF_UNSPEC;
 	err = re_regex((char *)mbuf_buf(mb), mbuf_get_left(mb),
-		       "IN IP[46]+", &af);
+		       "o=[^ ]* [^ ]* [^ ]* IN IP[46]+ [^ \r\n]+",
+		       NULL, NULL, NULL, &pl1, &pl2);
```
alfredh, line 506: I am wondering if the o= field should not be used for any logic.
I think the c= field should be used instead ?

src/ua.c
```diff
@@ -493,23 +494,31 @@ static bool require_handler(const struct sip_hdr *hdr,
...
-	case '6': return AF_INET6;
+	case '4': *af = AF_INET;
+		  break;
+	case '6': *af = AF_INET6;
+		  break;
 	}
 
-	return AF_UNSPEC;
+	pl_strdup(&addr, &pl2);
+	err = sa_set_str(sa, addr, 5060);
```
alfredh, line 519: port 5060 is hardcoded here. The different SIP transports have different default ports,
e.g. 5061 for tls

src/ua.c
```diff
@@ -860,6 +886,22 @@ static const char *autoans_header_name(enum answer_method met)
...
+{
+	struct sip_addr addr;
+	int err;
+
+	sa_init(&ua->dst, AF_UNSPEC);
+	err = sip_addr_decode(&addr, pl);
+	err |= sa_set(&ua->dst, &addr.uri.host, addr.uri.port);
+
+	if (!sa_isset(&ua->dst, SA_PORT))
+		sa_set_port(&ua->dst, 5060);
```
alfredh, line 899: port 5060 is hardcoded ...

src/ua.c
```diff
@@ -1013,13 +1055,14 @@ int ua_connect_dir(struct ua *ua, struct call **callp,
 	if (err)
 		goto out;
 
+	pl.p = (char *)dialbuf->buf;
+	pl.l = dialbuf->end;
+
+	(void)ua_check_dst_ip(ua, &pl);
```
alfredh, line 1061: what is the purpose of checking the dst IP here ?

if you only want to check IP-address, then I think explicit logic is better:

```c
if (ip_address)
    ua_check_dst_ip();
```

when reading this code, it is not really possible to know what the function is doing
with checking the implementation

src/uag.c
```diff
@@ -677,17 +661,18 @@ int uag_reset_transp(bool reg, bool reinvite)
...
-				struct call *call = lec->data;
-				const struct sa *laddr;
+		for (lec = ua_calls(ua)->head; lec; lec = lec->next) {
+			struct call *call = lec->data;
+			struct sa laddr;
 
-				laddr = net_laddr_af(net, call_af(call));
+			if (net_dst_source_addr_get(stream_raddr(audio_strm(
+						call_audio(call))), &laddr))
+				continue;
```
alfredh, line 673: this code uses the audio stream address.

please note that in SDP the audio and video medialines can have different IP-addresses
example:

audio: 1.2.3.4
video: 5.6.7.8

src/net.c
```diff
@@ -26,6 +24,12 @@ struct network {
 };
 
 
+struct laddr {
+	struct le le;
+	struct sa sa;
+};
```
alfredh, line 30: I think it is a bit overkill to use a list as a container.
perhaps it is enough with a static array:

```c
struct sa sav[8];
```

how many IP-addresses can a system have ?

src/net.c
```diff
@@ -10,10 +10,8 @@
 
 struct network {
 	struct config_net cfg;
-	struct sa laddr;
-#ifdef HAVE_INET6
-	struct sa laddr6;
-#endif
+	struct list laddrs;           /**< List of local addresses           */
+	struct list laddrs_tmp;       /**< Temp. list of local addresses     */
```
alfredh, line 14: should we consider to skip the copy and only use this function:

```c
int net_if_apply(net_ifaddr_h *ifh, void *arg);
```

there is one usecase where the user can specify which interface to use:

```
	-n <net_if>      Specify network interface
```

src/core.h
```diff
@@ -333,6 +333,7 @@ void stream_enable_rtp_timeout(struct stream *strm, uint32_t timeout_ms);
 bool stream_is_ready(const struct stream *strm);
 int  stream_decode(struct stream *s);
 void stream_silence_on(struct stream *s, bool on);
+const struct sa *stream_raddr(const struct stream *s);
```
alfredh, line 336: I added this function in master, so you can try to rebase :)

alfredh:
I tested the patch locally, when I dial out it gives a warning:

```c
>                             echo
alfredh@sip.antisip.com: selected for request
call uri: sip:echo@sip.antisip.com
sa: sa_set_port: no af 0 (port 5060)
ua: using best effort AF: af=AF_INET
```

src/net.c
cspiel1: I started with an array, but found that `list` is simpler.

src/uag.c
cspiel1: Baresip/re puts only one laddr into SDP. So we have to make a decision.

src/ua.c
cspiel1: This is only a dummy port that makes `udp_connect()` work for OSX. Has nothing to do with SIP. It must not be zero for OSX.
Added:
```c
/* OSX/iOS needs a dummy port number for udp_connect() */
```

src/ua.c
cspiel1: same here.

src/ua.c
cspiel1: - I remove the function and put the body into `ua_connect_dir()`.
- Another question is, why we should store the destination address in `ua`? The only reason is that we don't have the request URI in `ua_call_alloc()`. And we don't want to change the API. Right?

alfredh:
tested the patch on OSX.

the SIP transports contain:


```
transports:
  [fe80::1]:5050 (UDP)
  [fe80::1]:5050 (TCP)
  [fe80::1]:5051 (TLS)
  [fe80::1]:9 (WS)
  [fe80::1]:9 (WSS)
  [fe80::8b:e7b4:8fe7:2240]:5050 (UDP)
  [fe80::8b:e7b4:8fe7:2240]:5050 (TCP)
  [fe80::8b:e7b4:8fe7:2240]:5051 (TLS)
  [fe80::8b:e7b4:8fe7:2240]:9 (WS)
  [fe80::8b:e7b4:8fe7:2240]:9 (WSS)
  10.0.1.14:5050 (UDP)
  10.0.1.14:5050 (TCP)
  10.0.1.14:5051 (TLS)
  10.0.1.14:9 (WS)
  10.0.1.14:9 (WSS)
  [fe80::40f7:f4ff:fe8f:20f0]:5050 (UDP)
  [fe80::40f7:f4ff:fe8f:20f0]:5050 (TCP)
  [fe80::40f7:f4ff:fe8f:20f0]:5051 (TLS)
  [fe80::40f7:f4ff:fe8f:20f0]:9 (WS)
  [fe80::40f7:f4ff:fe8f:20f0]:9 (WSS)
  [fe80::f7e:ea28:f0b7:2bf9]:5050 (UDP)
  [fe80::f7e:ea28:f0b7:2bf9]:5050 (TCP)
  [fe80::f7e:ea28:f0b7:2bf9]:5051 (TLS)
  [fe80::f7e:ea28:f0b7:2bf9]:9 (WS)
  [fe80::f7e:ea28:f0b7:2bf9]:9 (WSS)
  [fe80::d3eb:fcdb:8450:ee02]:5050 (UDP)
  [fe80::d3eb:fcdb:8450:ee02]:5050 (TCP)
  [fe80::d3eb:fcdb:8450:ee02]:5051 (TLS)
  [fe80::d3eb:fcdb:8450:ee02]:9 (WS)
  [fe80::d3eb:fcdb:8450:ee02]:9 (WSS)
```

should we add a config option to enable/disable link-local and loopback addresses ?

the default value should be disabled

src/ua.c
alfredh: what if the routing table has different rules for different ports ?

the port here should be the same as is used for the real traffic.

juha-h:
Alfred E. Heggestad writes:

> should we add a config option to enable/disable link-local and
> loopback addresses ?
> 
> the default value should be disabled

I support that, since I most use cases link-local and loopback addresses
have no use.

src/ua.c
sreimers: Like here? https://www.system-rescue.org/networking/Destination-port-routing/
I think we should test if it is even discoverable with `net_dst_source_addr_get` (since it is a udp based dummy routing check.).
Maybe for special and not detectable cases its better to use `net_interface IP` in config or a account option?

cspiel1:
loopback is disabled, isn't it?

IPv6 link-local is very useful. It works out of the box, without any network setup. And this is the reason why we need this for our products.

cspiel1:
> I tested the patch locally, when I dial out it gives a warning:
> 
> ```c
> >                             echo
> alfredh@sip.antisip.com: selected for request
> call uri: sip:echo@sip.antisip.com
> sa: sa_set_port: no af 0 (port 5060)
> ua: using best effort AF: af=AF_INET
> ```

Should be solved now by the `!err` check in `ua_connect_dir()`
```
if (!err && !sa_isset(&ua->dst, SA_PORT))
		sa_set_port(&ua->dst, 5060);
```

cspiel1:
@sreimers the new workflow tools.yml fails. Can I fix this somehow?

src/ua.c
cspiel1: In this case it is the IP for the RTP stream. As we know, we have only one `laddr` for SDP. So I suggest to select `laddr` from which we have a route to destination IP:port where
- IP is taken from (c) line and
- the port number is taken from (m) **audio** line

src/ua.c
cspiel1: `ua->dst` is also used as destination address to select `laddr` for the RTP traffic. But for outgoing calls we don't have an RTP remote port here.

When we want to use SIP default ports 5060, 5061, ...: We don't have the transport protocol at this point. It is chosen in request.c function `sip_request()` called indirectly via `call_connect()`. But we need  the destination address in `ua_call_alloc()`. Thus which port should we use here?

sreimers:
> @sreimers the new workflow tools.yml fails. Can I fix this somehow?

Looks like a pylint bug, in master I have fixed this now.

src/ua.c
cspiel1: Added a commit.

src/net.c
cspiel1: - Right, this should be possible and avoids redundancy. I will try to add a commit that removes `laddrs_tmp`.
- The option `-n` and setting `net_interface` still works. If it is set, all other interfaces are ignored. It can be the name of the interface or even an IP address.

cspiel1:
The tests bind the loopback address 127.0.0.1. So if `net_interface` is an IP address, then we should not check if the IP is available in the system, ignore also the loopback check.

alfredh:
```
.-------.                  .--------.
|       |                  |        |
|       o eth0 <---SIP---> | Server |
|       |                  |        |
| host  o eth1             '--------'
|       |
|       o eth2
|       |
'-------'

```

if a host has multiple network interfaces (multihoming) then which interface
should be used for RTP/SDP ?

ICE solves this problem by having different candidates for each interface.

for non-ICE, if the client registers to a SIP server via eth0, would it make sense
to also use the address of eth0 for RTP ?

baresip can also be used as a server, in this case does it make sense
to let the user configure which interface to use (explicit config) ?

juha-h:
RTP/SDP interface should be the same as SIP interface. For example, if one of the interfaces is VPN interface, then both signalling and media would need to use the VPN.

Also as server, baresip could have several interfaces.  So if interface configuration is available, it should be possible to configure several interfaces.

cspiel1:
All these questions are only applicable to outgoing calls where we create an initial SDP.

For registered UAs:
- The address family is taken from the registration.
- The local address for SDP/RTP is taken from `net_laddr_af()` which almost works the same like before. This could be improved to improvement (A): Search the laddr that has a route to the registration peer IP:port.

For registrar-less UAs:
- The address family is select by `best_effort_af()` which is not very intelligent.
- If the call destination URI is an IP address  `ua->dst`, then `laddr` is set by `net_dst_source_addr_get()` which select an laddr that has a route to `ua->dst`.
- If the call destination is `user@domain` where domain is none of the registration SIP proxy domains, then also `laddr` is set by `net_laddr_af()` which also is not very intelligent. Improvement B: A DNS lookup mnat could be used to set an `laddr` that has a route to the DNS result. But the address family is already selected by `best_effort_af()` which always sets IPv4 in this case.

For the case if the dial URI is not an IP address we could improve further:
Improvement (C): Perfect would be a DNS lookup at the beginning of a call. If there is a result for the DNS lookup it could be used for:
- Select correct UA if only a single word is given as dial address. If there is a DNS result, than select a registrar-less UA. Otherwise select the first registered UA (like it is done now).
- Select correct address family.
- Select an `laddr` which has a route to the DNS result.

Improvement (A): Maybe could be added to this PR.
Improvement (B) or even better (C): Should be delayed to another PR.

cspiel1:
For incoming calls we already get an SDP offer with (c)-line IP address and (m)-line port number. This PR selects an `laddr` that has a route to the RTP remote IP:port.

cspiel1:
> ICE solves this problem by having different candidates for each interface.

Any of the mnat modules replace the SDP laddr in the `mnat_handler()`.

> for non-ICE, if the client registers to a SIP server via eth0, would it make sense
> to also use the address of eth0 for RTP ?

Yes! See answer above "For registered UAs"! And improvement (A) that could solve this.

> 
> baresip can also be used as a server, in this case does it make sense
> to let the user configure which interface to use (explicit config) ?

I assume you mean an incoming peer-to-peer call. No config. See comment above "For incoming calls..."!

cspiel1:
Similar questions like from @alfredh:
> RTP/SDP interface should be the same as SIP interface. For example, if one of the interfaces is VPN interface, then both signalling and media would need to use the VPN.

Yes! Improvement (A) could ensure this.

> Also as server, baresip could have several interfaces. So if interface configuration is available, it should be possible to configure several interfaces.

Yes, this PR makes it possible to receive calls on all interfaces.

src/ua.c
```diff
@@ -30,6 +30,7 @@ struct ua {
 	struct list hdr_filter;      /**< Filter for incoming headers        */
 	struct list custom_hdrs;     /**< List of outgoing headers           */
 	char *ansval;                /**< SIP auto answer value              */
+	struct sa dst;               /**< Current destination address        */
```
alfredh, line 33: perhaps rename this to a more accurate description ?

src/ua.c
```diff
@@ -493,23 +494,36 @@ static bool require_handler(const struct sip_hdr *hdr,
...
+	case '4': *af = AF_INET;
+		  break;
+	case '6': *af = AF_INET6;
+		  break;
 	}
 
-	return AF_UNSPEC;
+	/* OSX/iOS needs a port number for udp_connect() */
+	err = re_regex((char *)mbuf_buf(mb), mbuf_get_left(mb),
+		       "m=audio [0-9]+ ", &pl1);
```
alfredh, line 520: not sure about this ? what if the SDP only has a video line ?

is it possible to use the re sdp module to fetch the already parsed fields ?

include/baresip.h
```diff
@@ -690,6 +690,7 @@ const char *menc_event_name(enum menc_event event);
 struct network;
 
 typedef void (net_change_h)(void *arg);
+typedef int (net_laddr_h)(const struct sa *sa);
```
alfredh, line 693: can we reuse this callback handler type from re ?

```c
/**
 * Defines the interface address handler - called once per interface
 *
 * @param ifname Name of the interface
 * @param sa     IP address of the interface
 * @param arg    Handler argument
 *
 * @return true to stop traversing, false to continue
 */
typedef bool (net_ifaddr_h)(const char *ifname, const struct sa *sa,
			    void *arg);
```

alfredh:
regarding multi-homing and linklocal, I am wondering if we should consider adding
some new config options:

```
net_linklocal    yes|no
```

Link-local address can only be used on the local network, they are not globally reachable.
The default behaviour of baresip is a simple SIP client that can be used to call other
SIP endpoints (on the internet). It is assumed that the enduser is sitting behind NAT.

regarding the network change detection (roaming), it is mainly useful for a client
and not a server. We should consider to move the code from net.c to
a module that can use OS specific hooks to detect when network goes up/down.
Suggested name: roaming.so

cspiel1:
I thought we agree that we replace
```
       struct sa laddr;
       struct sa laddr6;
```
by a list of all network addresses.
```
       struct list laddrs;           /**< List of local addresses           */
```
in order to support all local addresses and network interfaces instead. My opinion is that this would be a big improvement and a peer-to-peer call is the simplest use-case. Supporting only one `laddr` and one `laddr6` can lead to wrong address in the SDP. A fix should not be done by replacing net.c by a module.

Regarding IPv6ll: If the user dials to an IPv6ll address, it should simply work. If he don't, then the IPv6ll will not be included in SIP or SDP. Does a config option has any benefits?

As we already discussed. Maybe we could get rid of `net->laddrs` completely. But first I'll try to rebase this.

cspiel1:
Re-based to master.

Here I have an alternative PR draft that removes the need for the `net->laddrs` list: https://github.com/baresip/baresip/pull/1522

src/ua.c
```diff
@@ -493,23 +494,36 @@ static bool require_handler(const struct sip_hdr *hdr,
 }
 
 
-static int sdp_af_hint(struct mbuf *mb)
+static int sdp_connection(struct mbuf *mb, int *af, struct sa *sa)
```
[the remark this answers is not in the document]
cspiel1: Ok, I open a PR for re.

src/ua.c
cspiel1: Audio should be preferred. If no audio then the video line.

I will check this and extend the parsing.

include/baresip.h
cspiel1: Yes, thanks for your commit!

src/ua.c
cspiel1: Tried to remove `sdp_connection()` by a call to `sdp_decode()` in `call_alloc()`. The result is
```
call: no common audio codecs - rejected
```
Have to postpone this. We can do this in a separate PR.

src/ua.c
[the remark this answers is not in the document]
cspiel1: This would be obsolete if we can use the decoded `call->sdp` to get `raddr`. --> Another PR.

cspiel1:
@alfredh: It seems that with each iteration this PR becomes larger. Also I am not sure if removing the `laddrs` list in the draft #1522  makes the PR simpler. Should we split the PR somehow into smaller pieces?

First of all: Should I close the draft #1522, or this PR? Or both?

For our products we have to use our patches for now.  Which should not be the long term solution. At least we need IPv6 link local support. And as you already wrote selecting three local addresses `laddr, laddr6, laddr6ll` somehow should be replaced. A logical consequence is to support all local addresses.

juha-h:
Alfred E. Heggestad writes:

> regarding multi-homing and linklocal, I am wondering if we should
> consider adding some new config options:
> 
> ```
> net_linklocal    yes|no
> ```

I support this suggestion.  Perhaps it could help to keep the code
simple for the "no" case.

> The default behaviour of baresip is a simple SIP client that can be
> used to call other 
> SIP endpoints (on the internet). It is assumed that the enduser is
> sitting behind NAT. 
> 
> regarding the network change detection (roaming), it is mainly useful
> for a client 
> and not a server. We should consider to move the code from net.c to
> a module that can use OS specific hooks to detect when network goes up/down.
> Suggested name: roaming.so

My baresip Android app has registered network callback that gets called
whenever there is a change in network connectivity. When there is, the
app prefers VPN network if available.  If not, it prefers network with
Internet connectivity.  If not, it uses network that Android considers
to be the "active" one. Once the decision is made, the app uses
net_address() API call to set the network address followed by
uag_reset_transp() API call.

Hope that this (or something even simpler) keeps on working after the
changes.

cspiel1:
> Alfred E. Heggestad writes:
> regarding multi-homing and linklocal, I am wondering if we should consider adding some new config options: ``` net_linklocal yes|no ```
> I support this suggestion. Perhaps it could help to keep the code simple for the "no" case.
> The default behaviour of baresip is a simple SIP client that can be used to call other SIP endpoints (on the internet). It is assumed that the enduser is sitting behind NAT. regarding the network change detection (roaming), it is mainly useful for a client and not a server. We should consider to move the code from net.c to a module that can use OS specific hooks to detect when network goes up/down. Suggested name: roaming.so
> My baresip Android app has registered network callback that gets called whenever there is a change in network connectivity. When there is, the app prefers VPN network if available. If not, it prefers network with Internet connectivity. If not, it uses network that Android considers to be the "active" one. Once the decision is made, the app uses net_address() API call to set the network address followed by uag_reset_transp() API call. Hope that this (or something even simpler) keeps on working after the changes.

The drawback if this solution is that it supports only one interface and one local address. We want to support all interfaces and local addresses.

juha-h:
Christian Spielberger writes:

> > My baresip Android app has registered network callback that gets called whenever there is a change in network connectivity. When there is, the app prefers VPN network if available. If not, it prefers network with Internet connectivity. If not, it uses network that Android considers to be the "active" one. Once the decision is made, the app uses net_address() API call to set the network address followed by uag_reset_transp() API call. Hope that this (or something even simpler) keeps on working after the changes.
> 
> The drawback if this solution is that it supports only one interface and one local address. We want to support all interfaces and local addresses.

Yes, that is because currently there cannot be more than one.  In the
future, net_address() could be replaced by net_address_add() and
net_address_rm().

alfredh:
we should take all the requirements into consideration when doing this work.
from what I understand, the main requirements are:


1. add link-local for use in Christian's products
2. Juha's Android application
3. generic networking behaviour that will fit for most users


in order to support Juha's app we can make network.c a container (list) for
local network addresses, that can be manipulated via public API:

- add a network address
- remove a network address
- apply a function handler for all addresses

it would be nice if this work is split up and done in small stages.
each patch/PR can be a small change that can be reviewed and tested.

thanks for your work and patience Christian :)

cspiel1:
Thanks @alfredh! I will close this for now, keep my branch and then try to split it into small PRs.

PR #1514, opened 2021-06-29
uag,menu: simplify transport enable/disable and support also ws/wss

cspiel1:
Its better to have only one command for enabling transport. Requested by @juha-h.

Rework of: https://github.com/baresip/baresip/pull/1502


juha-h:
Christian Spielberger writes:

>   * uag,menu: simplify transport enable/disable and support also
>   * ws/wss

Thanks, I did some testing and everything worked fine.

juha-h:
Since transport enabling/disabling is global, would it make sense to have a config variable or variables to set initial values?

cspiel1:
> Since transport enabling/disabling is global, would it make sense to have a config variable or variables to set initial values?

Yes, I also though about this. I will add a commit.

cspiel1:
Note: added utility functions
```
void u32mask_enable(uint32_t *mask, uint8_t bit, bool enable);
bool u32mask_enabled(uint32_t *mask, uint8_t bit);
```
Can be moved to re. In next PR?

PR #1517, opened 2021-07-01
account: add setting audio payload type for telephone-event

cspiel1:
It makes sense to make the config option for telephone-event payload type per account.


cspiel1:
Should we remove the global config option `audio_telev_pt` ?

Currently, the account setting overrides the global setting.

PR #1522, opened 2021-07-05
net multiple interfaces no laddrs

cspiel1:
This is an alternative PR(draft) to PR https://github.com/baresip/baresip/pull/1513.

An additional commit removes the list of local addresses in struct network. Instead directly syncs the network addresses with supported transport objects in uag.

Which PR should be preferred?

Based on: https://github.com/baresip/re/pull/123

Note: For this I had to change `net_laddr_af()` API function.

Tested by adding/removing IP addressess. And by plugging/unplugging of a network cable.


cspiel1:
Buiild fails, because some of the workflows do not checkout correct libre branch.

cspiel1:
The `laddrs` list in net.c makes it simpler to filter and detect changes of the local addresses.
I close this, because nobody seems to be interested.

PR #1535, opened 2021-07-15
net: remove periodic check (experimental)

alfredh:
a PR for discussion ...

how to handle network changes when multiple address is supported.

if we remove this code, will it make the net.c code simpler ?

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -699,10 +699,6 @@ int  net_alloc(struct network **netp, const struct config_net *cfg);
 int  net_use_nameserver(struct network *net,
 			const struct sa *srvv, size_t srvc);
 int  net_set_address(struct network *net, const struct sa *ip);
-void net_change(struct network *net, uint32_t interval,
-		net_change_h *ch, void *arg);
-void net_force_change(struct network *net);
-bool net_check(struct network *net);
 bool net_af_enabled(const struct network *net, int af);
 int  net_set_af(struct network *net, int af);
 void net_dns_refresh(struct network *net);
--- a/modules/httpreq/httpreq.c
+++ b/modules/httpreq/httpreq.c
@@ -89,24 +89,6 @@ static void http_resph(int err, const struct http_msg *msg, void *arg)
 }
 
 
-static void net_handler(void *arg)
-{
-	const struct sa *sa;
-	(void) arg;
-
-	sa = net_laddr_af(d->net, AF_INET);
-	if (sa)
-		http_client_set_laddr(d->client, sa);
-	info("httpreq: network changed %j", sa);
-#ifdef HAVE_INET6
-	sa = net_laddr_af(d->net, AF_INET6);
-	if (sa)
-		http_client_set_laddr6(d->client, sa);
-	info("httpreq: network changed %j", sa);
-#endif
-}
-
-
 static int ensure_alloc(void)
 {
 	int err = 0;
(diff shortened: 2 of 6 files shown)


cspiel1:
Interesting idea. We need `laddr` for:
- putting in SDP
- adding transport objects with `sip_transp_add()`

For the first we don't need to check changes.

For the second we have to detect changes, and then call `uag_reset_transp()` which re-creates the transport objects and start the re-REGISTER.

alfredh:
consider this case:

1. the system has one IPv4 address and one IPv6 address

2. there is one active call using the IPv4 address

3. during the call, the system gets an extra IPv6 address


with the current code, it will trigger a network change event, and a re-invite.
but this is not really correct, since the IPv4 address used is still available.

this leads to my conclusion that the network detect has to be closer to the
code that is actually using the address.

alfredh:
polling for network changes can be done, but might leave the system w/o network
for many seconds.

instant network change is possible, using OS specific hooks:

https://stackoverflow.com/questions/41590562/detect-ethernet-wifi-network-change

anyways, this was meant as input to the discussion ... closing it for now

juha-h:
Alfred E. Heggestad writes:

> instant network change is possible, using OS specific hooks:
> 
> https://stackoverflow.com/questions/41590562/detect-ethernet-wifi-network-change

In Android callback can be registered that receives network change
events, for example:

```
       cm = getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
        val builder = NetworkRequest.Builder()
                .removeCapability(NetworkCapabilities.NET_CAPABILITY_NOT_VPN)
        cm.registerNetworkCallback(
                builder.build(),
                object : ConnectivityManager.NetworkCallback() {

                    override fun onAvailable(network: Network) {
                        super.onAvailable(network)
                        Log.i(TAG, "Network $network is available")
                        updateNetwork()
                    }

                    override fun onLost(network: Network) {
                        super.onLost(network)
                        Log.i(TAG, "Network $network is lost")
                        if (activeNetwork == "$network")
                            updateNetwork()
                    }

                    override fun onLinkPropertiesChanged(network: Network, props: LinkProperties) {
                        super.onLinkPropertiesChanged(network, props)
                        Log.i(TAG, "Network $network link properties changed")
                        if (activeNetwork == "$network")
                            updateNetwork()
                    }

                    override fun onCapabilitiesChanged(network: Network, caps: NetworkCapabilities) {
                        super.onCapabilitiesChanged(network, caps)
                        Log.i(TAG, "Network $network capabilities changed: $caps")
                    }
                }
        )
```

alfredh:
it would make sense to put this code in the application which has access to those apis.
in your case, the android app ...

juha-h:
Alfred E. Heggestad writes:

> it would make sense to put this code in the application which has
> access to those apis. 
> in your case, the android app ...

Yes, I just gave an example of application code, not proposed to put the
code in libbaresip.

cspiel1:
We could
- put a list `laddrs` into net.c which is filled at startup,
- add functions `net_laddr_add()` and `net_laddr_rm()` in to net API,
- remove the network change detect code (E.g. we could move it into a module).

With the new API functions we have simple constraints that make sense:
- `net_laddr_add()` will not change the current call or send a re-INVITE,
- `net_laddr_rm()` will only change the current call if the given address is the used `laddr`

juha-h:
Sounds good.  What function is then used to change the current call, for example, when WiFi or VPN network becomes available?

cspiel1:
> Sounds good. What function is then used to change the current call, for example, when WiFi or VPN network becomes available?

See uag.c function `uag_reset_transp()`! Here you see:
```
				laddr = net_laddr_af(net, call_af(call));

				err |= call_reset_transp(call, laddr);
```

This is how it is done currently. This could also be improved, because currently the choice of `laddr` does not include a check if there is a route from `laddr` to the peer.

PR #1536, opened 2021-07-16
account: add enable/disable flag for video

cspiel1:
This allows to disable video for specific accounts.

Examples to disable video:
```
<sip:user@example.com>;video_codecs=
<sip:user@example.com>;video_codecs=none
```


include/baresip.h
```diff
@@ -137,6 +137,8 @@ enum sipansbeep account_sipansbeep(const struct account *acc);
 void account_set_sipansbeep(struct account *acc, enum sipansbeep beep);
 void account_set_autelev_pt(struct account *acc, uint32_t pt);
 uint32_t account_autelev_pt(struct account *acc);
+void account_enable_video(struct account *acc, bool enable);
+bool account_video_enabled(struct account *acc);
```
alfredh, line 141: input argument should be const

cspiel1: yes, for the getter function.

PR #1547, opened 2021-07-21
outgoing calls early callid

cspiel1:
based on https://github.com/baresip/re/pull/135

The call-id should be returned to the application immediately. This is in the first event: CALL_EVENT_OUTGOING


alfredh:
looks good to me

sreimers:
https://github.com/baresip/re/pull/135 is merged now.

cspiel1:
thanks!

PR #1555, opened 2021-07-23
move network check to module

cspiel1:
Based on our discussions in #1535.

TODOs:
- Remove `laddr`, `laddr6` in net.c. The local address can be queried from transport list.
- Remove `net_laddr_apply()` and `handle_addr()` from net.c.
- Add to uag.c:
   - `uag_laddr_dst(struct sa *dst)`
   - `uag_laddr_af(int af)` replaces the `net_laddr_af()`. We are currently not able to get the destination IP for an outgoing request if the user specifies a SIP URI that does not contain an IP address.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -699,10 +699,6 @@ int  net_alloc(struct network **netp, const struct config_net *cfg);
 int  net_use_nameserver(struct network *net,
 			const struct sa *srvv, size_t srvc);
 int  net_set_address(struct network *net, const struct sa *ip);
-void net_change(struct network *net, uint32_t interval,
-		net_change_h *ch, void *arg);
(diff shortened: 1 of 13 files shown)


alfredh:
this looks really great!

some input:

- perhaps netroam.so is a better name for the module (Network Roaming) ?
- the list of local addresses should be available from net.c (sip agnostic)

here is a project that is using libbaresip without SIP:

https://github.com/baresip/baresip-webrtc

this project will use the net.c API to fetch local addresses.

alfredh:
hi, is there any updates on this PR ?

cspiel1:
Coming soon. We have release pressure currently.

alfredh:
close this for now .. please reopen when you have time to work on it

Issue #1558, opened 2021-08-01
New Defects reported by Coverity Scan


alfredh:
Hi,

Please find the latest report on new defect(s) introduced to alfredh/baresip found with Coverity Scan.

4 new defect(s) introduced to alfredh/baresip found with Coverity Scan.
3 defect(s), reported by Coverity Scan earlier, were marked fixed in the recent build analyzed by Coverity Scan.

New defect(s) Reported-by: Coverity Scan
Showing 4 of 4 defect(s)

** CID 333030:  Security best practices violations  (STRING_OVERFLOW)
/src/config.c: 260 in decode_sip_transports()

________________________________________________________________________________________________________
*** CID 333030:  Security best practices violations  (STRING_OVERFLOW)
/src/config.c: 260 in decode_sip_transports()
254     
255     	for (i = 0; i < SIP_TRANSPC; ++i) {
256     		bool en;
257     		char buf[9];
258     		struct pl e=PL_INIT;
259     
>>>     CID 333030:  Security best practices violations  (STRING_OVERFLOW)
>>>     You might overrun the 9-character fixed-size string "buf" by copying the return value of "sip_transp_name" without checking the length.
260     		strcpy(buf, sip_transp_name(i));
261     		strcat(buf, "[^,]+");
262     		en = 0 == re_regex(pl->p, pl->l, sip_transp_name(i)) &&
263     		     0 != re_regex(pl->p, pl->l, buf, &e);
264     		u32mask_enable(&cfg->transports, i, en);
265     	}

** CID 333029:  Null pointer dereferences  (REVERSE_INULL)
/modules/amr/amr.c: 339 in encode_nb()

________________________________________________________________________________________________________
*** CID 333029:  Null pointer dereferences  (REVERSE_INULL)
/modules/amr/amr.c: 339 in encode_nb()
333     		     size_t *len, int fmt, const void *sampv, size_t sampc)
334     {
335     	const struct amr_aucodec *amr_ac = (struct amr_aucodec *)st->ac;
336     	int r;
337     	(void)marker;
338     
>>>     CID 333029:  Null pointer dereferences  (REVERSE_INULL)
>>>     Null-checking "st" suggests that it may be null, but it has already been dereferenced on all paths leading to the check.
339     	if (!st || !buf || !len || !sampv || sampc != FRAMESIZE_NB)
340     		return EINVAL;
341     	if (*len < NB_SERIAL_MAX)
342     		return ENOMEM;
343     
[...]

PR #1560, opened 2021-08-03
config: replace strcpy by saver re_snprintf (#1558)

cspiel1:
Here is the fix.  How can I add the CID number of the Code Coverity Scan?

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -254,11 +254,13 @@ static void decode_sip_transports(struct config_sip *cfg,
 
 	for (i = 0; i < SIP_TRANSPC; ++i) {
 		bool en;
-		char buf[9];
+		char buf[16];
 		struct pl e=PL_INIT;
 
-		strcpy(buf, sip_transp_name(i));
-		strcat(buf, "[^,]+");
+		if (re_snprintf(buf, sizeof(buf), "%s[^,]+",
+					sip_transp_name(i)) <= 0)
+			break;
+
 		en = 0 == re_regex(pl->p, pl->l, sip_transp_name(i)) &&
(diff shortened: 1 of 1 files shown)


src/config.c
```diff
@@ -254,11 +254,15 @@ static void decode_sip_transports(struct config_sip *cfg,
...
 		bool en;
-		char buf[9];
+		char buf[16];
 		struct pl e=PL_INIT;
+		int err;
+
+		err = re_snprintf(buf, sizeof(buf), "%s[^,]+",
+				sip_transp_name(i));
+		if (err)
+			break;
```
alfredh, line 264: does this work ?

```c
/**
 * Print a formatted string to a buffer
 *
 * @param str  Buffer for output string
 * @param size Size of buffer
 * @param fmt  Formatted string
 *
 * @return The number of characters printed, or -1 if error
 */
int re_snprintf(char *str, size_t size, const char *fmt, ...)
{
        va_list ap;
        int n;

        va_start(ap, fmt);
        n = re_vsnprintf(str, size, fmt, ap);
        va_end(ap);

        return n;
}
```

alfredh:
the CID is:


CID 333030 (#1 of 1): Copy into fixed size buffer (STRING_OVERFLOW)
2. fixed_size_dest: You might overrun the 9-character fixed-size string buf by copying the return value of sip_transp_name without checking the length.

cspiel1:
Oops. Fixed the use of `re_snprintf()`.
And added the CID info to the commit message. Thanks!

How the Coverity Scan can be invoked?

alfredh:
many thanks for fixing this. A free account can be created here:

https://www.synopsys.com/software-integrity.html

PR #1569, opened 2021-08-11
core: move stream_enable_rtp_timeout to api

sreimers:
For RTP Connections like webrtc (module/app) its useful to make `stream_enable_rtp_timeout` available.


alfredh:
we should also align the stream api with webrtc api,

the stream consist of:

1. sender
2. receiver
3. transport

something like this:

1. transceiver (stream)
2. sender
3. receiver

if we can manage to make the stream api more modular, the application
can for example choose to use only the sender or receiver.

PR #1580, opened 2021-08-24
Fix static analyzer warnings

viordash:
p.s. I would like to write tests for case with `clean_number `. 
How do you look at adding the ability to run tests in the visual studio?
And to simplify access to the `clean_number` method, it would be well to remove the static attribute from it, otherwise it is so difficult to call through `dial_handler`. 
if you agree, I will prepare a new PR, in which I will add a new visual studio project, only for running tests, and the new test itself.


alfredh:
thanks!

you can move the function clean_number to its own .c file, and add the interface
in src/core.h or include/baresip.h

the testcode is in test/*.c and can then have some testcode that test the function.

the testcode that we have is really useful, and new testcode is also welcome :)

PR #1581, opened 2021-08-24
stream: RTP inbound telephone events should not lead to packet loss

cspiel1:
Telephone-event packets also increment the RTP sequence number and thus have to
be put into the jitter buffer. Otherwise missing packets are detected wrongly.
This leads to audio PLC (Packet loss concealment) or double playback of an
audio frame.
If a telephone event packet is read from jitter buffer, it has to be ignored
by reading a second packet. So the `stream_recv_handler()` needs a return
value which can be EAGAIN.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -942,6 +942,15 @@ static void handle_telev(struct audio *a, struct mbuf *mb)
 }
 
 
+static bool audio_is_telev(struct audio *a, int pt)
+{
+	const struct sdp_format *lc;
+
+	lc = sdp_media_lformat(stream_sdpmedia(a->strm), pt);
+	return  lc && !str_casecmp(lc->name, "telephone-event");
+}
+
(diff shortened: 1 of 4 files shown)


src/audio.c
```diff
@@ -1119,7 +1128,7 @@ static int aurx_stream_decode(struct aurx *rx, bool marker,
 
 
 /* Handle incoming stream data from the network */
-static void stream_recv_handler(const struct rtp_header *hdr,
+static int stream_recv_handler(const struct rtp_header *hdr,
 				struct rtpext *extv, size_t extc,
 				struct mbuf *mb, unsigned lostc, void *arg)
```
alfredh, line 1133: can we keep this to have return type void ?

I prefer void in callback handlers, because the handler code cannot fail
hence it is easier to write the code.

PR #1584, opened 2021-08-26
move network check to module

cspiel1:
proceeding with PR: https://github.com/baresip/baresip/pull/1555

- Renamed module netcheck to netroam.
- the local IP addresses are tracked now in net.c
- Adds also function for manual adding/removing of local addresses.

The net.c has the main logic for holding local addresses. Initially all local IP addresses are collected. Baresip works without the module and listens on all local IP addresses. Module netroam keeps the list up-to-date.

TODO for following PRs:
- Improve selection of address family / local address for outgoing/incoming calls. Currently the account parameter `mediaaf=ipv6` can be used. But still IPv6 link local will not work until we improve this selection.

Related to: https://github.com/baresip/re/pull/152

### DIFF
--- a/modules/netroam/netroam.c
+++ b/modules/netroam/netroam.c
@@ -1,5 +1,7 @@
 /**
- * @file net.c Network change detection module
+ * @file netroam.c Network roaming module
+ *
+ * Detects and applies changes of the local network addresses
  *
  * Copyright (C) 2021 Commend.com - c.spielberger@commend.com
  */
@@ -11,13 +13,13 @@
 
 
 /**
- * @defgroup netcheck netcheck
+ * @defgroup netroam netroam
  *
  * The network check structure
  *
  */
 
-struct netcheck {
+struct netroam {
 	const struct config_net *cfg;
 	struct network *net;
 	uint32_t interval;
@@ -26,13 +28,13 @@ struct netcheck {
 };
 
 
-static struct netcheck d;
+static struct netroam d;
 
 
 static bool laddr_obsolete(enum sip_transp tp, const struct sa *laddr,
 			   void *arg)
 {
-	struct netcheck *n = arg;
+	struct netroam *n = arg;
 	char ifname[256] = "???";
 	int err;
 	(void) tp;
@@ -56,7 +58,7 @@ static bool laddr_find(enum sip_transp tp, const struct sa *laddr, void *arg)
 }
 
 
-static bool netcheck_find_obsolete(struct netcheck *n)
+static bool netroam_find_obsolete(struct netroam *n)
 {
 	sip_transp_list(uag_sip(), laddr_obsolete, n);
 	return sa_isset(&n->laddr, SA_ADDR);
@@ -66,7 +68,7 @@ static bool netcheck_find_obsolete(struct netcheck *n)
 static bool sip_transp_misses_laddr(const char *ifname, const struct sa *sa,
 			     void *arg)
 {
-	struct netcheck *n = arg;
+	struct netroam *n = arg;
 
 	if (!net_ifaddr_filter(baresip_network(), ifname, sa))
 		return false;
@@ -82,7 +84,7 @@ static bool sip_transp_misses_laddr(const char *ifname, const struct sa *sa,
 
 static void poll_changes(void *arg)
 {
-	struct netcheck *n = arg;
+	struct netroam *n = arg;
 	bool changed = false;
 	net_dns_refresh(baresip_network());
 
@@ -90,15 +92,15 @@ static void poll_changes(void *arg)
 	sa_init(&n->laddr, AF_UNSPEC);
 	net_if_apply(sip_transp_misses_laddr, n);
 	if (sa_isset(&n->laddr, SA_ADDR)) {
-		debug("netcheck: new IP address %j\n", &n->laddr);
+		debug("netroam: new IP address %j\n", &n->laddr);
 		uag_transp_add(&n->laddr);
 		changed = true;
 	}
 
 	/* was a local IP removed? */
 	sa_init(&n->laddr, AF_UNSPEC);
-	if (netcheck_find_obsolete(n)) {
-		debug("netcheck: IP address %j was removed\n", &n->laddr);
+	if (netroam_find_obsolete(n)) {
+		debug("netroam: IP address %j was removed\n", &n->laddr);
 		uag_transp_rm(&n->laddr);
 		changed = true;
 	}
@@ -127,8 +129,8 @@ static int module_close(void)
 }
 
 
-const struct mod_export DECL_EXPORTS(netcheck) = {
-	"netcheck",
+const struct mod_export DECL_EXPORTS(netroam) = {
+	"netroam",
 	"application",
 	module_init,
 	module_close
--- a/modules/netroam/netroam.c
+++ b/modules/netroam/netroam.c
@@ -114,9 +114,9 @@ static int module_init(void)
 	int err = 0;
 	d.cfg = &conf_config()->net;
 	d.net = baresip_network();
-	/* TODO: a config setting for the interval */
 	/* TODO++: Use AF_NETLINK socket to be notified! (man 7 netlink) */
-	d.interval = 2;
+	d.interval = 60;
+	conf_get_u32(conf_cur(), "netroam_interval", &d.interval);
 	tmr_start(&d.tmr, d.interval * 1000, poll_changes, &d);
 
 	return err;
--- a/modules/netroam/netroam.c
+++ b/modules/netroam/netroam.c
@@ -60,6 +60,7 @@ static bool laddr_find(const char *ifname, const struct sa *laddr, void *arg)
 
 static bool netroam_find_obsolete(struct netroam *n)
 {
+	sa_init(&n->laddr, AF_UNSPEC);
 	net_laddr_apply(n->net, laddr_obsolete, n);
 	return sa_isset(&n->laddr, SA_ADDR);
 }
@@ -92,7 +93,6 @@ static void poll_changes(void *arg)
 	sa_init(&n->laddr, AF_UNSPEC);
 	net_if_apply(net_misses_laddr, n);
 	if (sa_isset(&n->laddr, SA_ADDR)) {
-		debug("netroam: new IP address %j\n", &n->laddr);
 		net_add_address(n->net, &n->laddr);
 		changed = true;
 	}
@@ -100,7 +100,6 @@ static void poll_changes(void *arg)
 	/* was a local IP removed? */
 	sa_init(&n->laddr, AF_UNSPEC);
 	if (netroam_find_obsolete(n)) {
-		debug("netroam: IP address %j was removed\n", &n->laddr);
 		net_rm_address(n->net, &n->laddr);
 		changed = true;
 	}
--- a/modules/netroam/netroam.c
+++ b/modules/netroam/netroam.c
@@ -104,22 +104,40 @@ static void poll_changes(void *arg)
 		changed = true;
 	}
 
-	tmr_start(&n->tmr, changed ? 1000 : n->interval * 1000,
-		  poll_changes, n);
+	if (n->interval)
+		tmr_start(&n->tmr, changed ? 1000 : n->interval * 1000,
+			  poll_changes, n);
 }
+
+
+static int cmd_netchange(struct re_printf *pf, void *unused)
+{
+	(void)unused;
+
+	re_hprintf(pf, "netroam: network change\n");
+	poll_changes(&d);
+	return 0;
+}
+
+
+static const struct cmd cmdv[] = {
+
+{"netchange",     0, 0, "Inform netroam about a network change",
+				cmd_netchange                                },
+};
 
 
 static int module_init(void)
 {
-	int err = 0;
 	d.cfg = &conf_config()->net;
 	d.net = baresip_network();
 	/* TODO++: Use AF_NETLINK socket to be notified! (man 7 netlink) */
 	d.interval = 60;
 	conf_get_u32(conf_cur(), "netroam_interval", &d.interval);
-	tmr_start(&d.tmr, d.interval * 1000, poll_changes, &d);
+	if (d.interval)
+		tmr_start(&d.tmr, d.interval * 1000, poll_changes, &d);
 
-	return err;
+	return cmd_register(baresip_commands(), cmdv, ARRAY_SIZE(cmdv));
 }
 
 
--- a/modules/netroam/netroam.c
+++ b/modules/netroam/netroam.c
@@ -144,6 +144,7 @@ static int module_init(void)
 static int module_close(void)
 {
 	tmr_cancel(&d.tmr);
+	cmd_unregister(baresip_commands(), cmdv);
 	return 0;
 }
 
--- a/modules/netroam/netroam.c
+++ b/modules/netroam/netroam.c
@@ -31,15 +31,15 @@ struct netroam {
 static struct netroam d;
 
 
-static bool laddr_obsolete(enum sip_transp tp, const struct sa *laddr,
+static bool laddr_obsolete(const char *ifname, const struct sa *laddr,
 			   void *arg)
 {
 	struct netroam *n = arg;
-	char ifname[256] = "???";
+	char ifn[2] = "?";
 	int err;
-	(void) tp;
+	(void) ifname;
 
-	err = net_if_getname(ifname, sizeof(ifname), sa_af(laddr), laddr);
+	err = net_if_getname(ifn, sizeof(ifn), sa_af(laddr), laddr);
 	if (err == ENODEV) {
 		sa_cpy(&n->laddr, laddr);
 		return true;
@@ -49,10 +49,10 @@ static bool laddr_obsolete(enum sip_transp tp, const struct sa *laddr,
 }
 
 
-static bool laddr_find(enum sip_transp tp, const struct sa *laddr, void *arg)
+static bool laddr_find(const char *ifname, const struct sa *laddr, void *arg)
 {
 	const struct sa *sa = arg;
-	(void) tp;
+	(void) ifname;
 
 	return sa_cmp(sa, laddr, SA_ADDR);
 }
@@ -60,12 +60,12 @@ static bool laddr_find(enum sip_transp tp, const struct sa *laddr, void *arg)
 
 static bool netroam_find_obsolete(struct netroam *n)
 {
-	sip_transp_list(uag_sip(), laddr_obsolete, n);
+	net_laddr_apply(n->net, laddr_obsolete, n);
 	return sa_isset(&n->laddr, SA_ADDR);
 }
 
 
-static bool sip_transp_misses_laddr(const char *ifname, const struct sa *sa,
+static bool net_misses_laddr(const char *ifname, const struct sa *sa,
 			     void *arg)
 {
 	struct netroam *n = arg;
@@ -73,7 +73,7 @@ static bool sip_transp_misses_laddr(const char *ifname, const struct sa *sa,
 	if (!net_ifaddr_filter(baresip_network(), ifname, sa))
 		return false;
 
-	if (!sip_transp_list(uag_sip(), laddr_find, (void*) sa)) {
+	if (!net_laddr_apply(n->net, laddr_find, (void *) sa)) {
 		sa_cpy(&n->laddr, sa);
 		return true;
 	}
@@ -90,10 +90,10 @@ static void poll_changes(void *arg)
 
 	/* was a local IP added? */
 	sa_init(&n->laddr, AF_UNSPEC);
-	net_if_apply(sip_transp_misses_laddr, n);
+	net_if_apply(net_misses_laddr, n);
 	if (sa_isset(&n->laddr, SA_ADDR)) {
 		debug("netroam: new IP address %j\n", &n->laddr);
-		uag_transp_add(&n->laddr);
+		net_add_address(n->net, &n->laddr);
 		changed = true;
 	}
 
@@ -101,7 +101,7 @@ static void poll_changes(void *arg)
 	sa_init(&n->laddr, AF_UNSPEC);
 	if (netroam_find_obsolete(n)) {
 		debug("netroam: IP address %j was removed\n", &n->laddr);
-		uag_transp_rm(&n->laddr);
+		net_rm_address(n->net, &n->laddr);
 		changed = true;
 	}
 
@@ -115,6 +115,8 @@ static int module_init(void)
 	int err = 0;
 	d.cfg = &conf_config()->net;
 	d.net = baresip_network();
+	/* TODO: a config setting for the interval */
+	/* TODO++: Use AF_NETLINK socket to be notified! (man 7 netlink) */
 	d.interval = 2;
 	tmr_start(&d.tmr, d.interval * 1000, poll_changes, &d);
 
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -861,6 +861,7 @@ int  uag_reset_transp(bool reg, bool reinvite);
 void uag_set_sub_handler(sip_msg_h *subh);
 int  uag_set_extra_params(const char *eprm);
 int  uag_enable_transport(enum sip_transp tp, bool en);
+int  uag_transp_add(const struct sa *laddr);
 struct ua   *uag_find(const struct pl *cuser);
 struct ua   *uag_find_msg(const struct sip_msg *msg);
 struct ua   *uag_find_aor(const char *aor);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -862,6 +862,7 @@ void uag_set_sub_handler(sip_msg_h *subh);
 int  uag_set_extra_params(const char *eprm);
 int  uag_enable_transport(enum sip_transp tp, bool en);
 int  uag_transp_add(const struct sa *laddr);
+int  uag_transp_rm(const struct sa *laddr);
 struct ua   *uag_find(const struct pl *cuser);
 struct ua   *uag_find_msg(const struct sip_msg *msg);
 struct ua   *uag_find_aor(const char *aor);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -703,12 +703,17 @@ int  net_alloc(struct network **netp, const struct config_net *cfg);
 int  net_use_nameserver(struct network *net,
 			const struct sa *srvv, size_t srvc);
 int  net_set_address(struct network *net, const struct sa *ip);
+int  net_add_address(struct network *net, const struct sa *ip);
+int  net_flush_addresses(struct network *net);
+int  net_rm_address(struct network *net, const struct sa *ip);
 bool net_af_enabled(const struct network *net, int af);
 int  net_set_af(struct network *net, int af);
 void net_dns_refresh(struct network *net);
 int  net_dns_debug(struct re_printf *pf, const struct network *net);
 int  net_debug(struct re_printf *pf, const struct network *net);
-void net_laddr_apply(const struct network *net, net_ifaddr_h *ifh, void *arg);
+bool net_laddr_apply(const struct network *net, net_ifaddr_h *ifh, void *arg);
+void net_set_add_handler(struct network *net, net_ifaddr_h *ifh, void *arg);
+void net_set_rm_handler(struct network *net, net_ifaddr_h *ifh, void *arg);
 bool net_ifaddr_filter(const struct network *net, const char *ifname,
 		       const struct sa *sa);
 const struct sa *net_laddr_af(const struct network *net, int af);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -709,6 +709,8 @@ void net_dns_refresh(struct network *net);
 int  net_dns_debug(struct re_printf *pf, const struct network *net);
 int  net_debug(struct re_printf *pf, const struct network *net);
 void net_laddr_apply(const struct network *net, net_ifaddr_h *ifh, void *arg);
+bool net_ifaddr_filter(const struct network *net, const char *ifname,
+		       const struct sa *sa);
 const struct sa *net_laddr_af(const struct network *net, int af);
 struct dnsc     *net_dnsc(const struct network *net);
 
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -703,10 +703,6 @@ int  net_alloc(struct network **netp, const struct config_net *cfg);
 int  net_use_nameserver(struct network *net,
 			const struct sa *srvv, size_t srvc);
 int  net_set_address(struct network *net, const struct sa *ip);
-void net_change(struct network *net, uint32_t interval,
-		net_change_h *ch, void *arg);
-void net_force_change(struct network *net);
-bool net_check(struct network *net);
 bool net_af_enabled(const struct network *net, int af);
 int  net_set_af(struct network *net, int af);
 void net_dns_refresh(struct network *net);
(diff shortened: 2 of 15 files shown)


alfredh:
I will go a head and merge the patch soon.

could you please merge the re patch first ?

juha-h:
Alfred E. Heggestad writes:

> It would also be great if someone else could test the PR.

I could try it in my Android app, but I'm wondering what needs to be
changed if anything to make the app working again after the PR.

cspiel1:
> I will go a head and merge the patch soon.
 
Should I add a config setting for the check interval in module netroam? In mainline we have 60 seconds. In this PR currently it is 2 seconds.

> I could try it in my Android app, but I'm wondering what needs to be changed if anything to make the app working again after the PR.

A good test would be to change between Wifi and LTE. You will have to add the netroam module in your baresip config. It is built automatically. If someone decides to omit this module, the alternative is that the application has to detect network changes and then restart baresip.

juha-h:
Christian Spielberger writes:

> Should I add a config setting for the check interval in module
> netroam? In mainline we have 60 seconds. In this PR currently it is 2
> seconds.

If active network of mobile device changes from WiFi to Mobile Network
or vice versa, it should not take 60 seconds before the change is
detected and re-INVITE is sent.

> A good test would be to change between Wifi and LTE. You will have to
> add the netroam module in your baresip config. It is built
> automatically. If someone decides to omit this module, the alternative
> is that the application has to detect network changes and then restart
> baresip.

Restarting baresip would be too destructive action.  The app should
somehow (e.g. via an event) be able to tell netroam module module that
there has been a change.  That way the change could be detected
immediately.

juha-h:
Does this patch also add support for multiple simultaneous interfaces so
that, for example, some UAs could be using Mobile Network and some WiFi?

cspiel1:
> Does this patch also add support for multiple simultaneous interfaces so that, for example, some UAs could be using Mobile Network and some WiFi?

Yes.

cspiel1:
About immediate network change detection:
- (A) I would like to try to add an AF_NETLINK socket to the re-main epoll loop. See `man 7 netlink`! Would this also be a solution for Android?
- (B) Also I would implement your idea with an event from the App, maybe in addition to (A). This simply will be a command added by netroam, that triggers the check for changes.

juha-h:
Christian Spielberger writes:

> About immediate network change detection:
> - (A) I would like to try to add an AF_NETLINK socket to the re-main
> epoll loop. See `man 7 netlink`! Would this also be a solution for
> Android?

Can't say.

> - (B) Also I would implement your idea with an event from the App,
> - maybe in addition to (A). This simply will be a command added by
> - netroam, that triggers the check for changes.

Sounds good.  Then in Android app, netroam polling interval could be
configured to be very long or polling could be disabled altogether in
order to reduce CPU usage and battery consumption.

cspiel1:
Added:
```
commit 05ad03cc01bebd7f1a9f0d7a23097c49499184eb (HEAD -> net_remove_check, origin/net_remove_check)
Author: Christian Spielberger <c.spielberger@commend.com>
Date:   Mon Aug 30 15:40:27 2021 +0200

    netroam: add command netchange
    
    Also, if setting netroam_interval is zero, do not start the network polling.
```

Today also tested (and fixed)
- adding/removing of IP addresses,
- plugging/unplugging of USB interface

It's ready for merge from my viewpoint.

Following PRs:
- automatic selection of laddr for incoming/outgoing calls

juha-h:
Currently in my Android app, I use these two API functions to handle network changes:

- `net_set_address` (to set and unset ipv4 or ipv6 address)
- `uag_reset_transp`

Now with `netroam`, is it so that executing command `netchange` is all that is needed or should I replace `net_set_address` calls with `net_add_address` and `net_rm_address` calla?

cspiel1:
`net_set_address()` is not really needed any longer. We have the options:
- to remove it,
- or to force to use the given IP address.

For now I decided the second.

`net_add_address()` and `net_rm_address()`: At startup net.c collects all interfaces and local IP addresses. It adds for each local IP address SIP transport instances. If you don't use the module `netroam` and the command `netchange` to signal network changes, then you should use `net_add_address()` for each IP address that is added in the system and `net_rm_address()` if one is removed.

juha-h:
OK, I'll first try to get my Android app working w/o `netroam` module by replacing `net_set_address()` calls with `net_add_address()` and `net_rm_address()` calls followed by `uag_reset_transp()` calls.

Is there any other changes needed?

cspiel1:
> OK, I'll first try to get my Android app working w/o `netroam` module by replacing `net_set_address()` calls with `net_add_address()` and `net_rm_address()` calls followed by `uag_reset_transp()` calls.
> 
> Is there any other changes needed?

Sorry, I edited may comment above again. You don't need to call `uag_reset_transp()`.

Right, you don't need `netroam` module.

juha-h:
Currently this code is executed whenever network addresses changed:
```
if (updated) {
            BaresipService.linkAddresses = linkAddresses
            Api.uag_reset_transp(register = true, reinvite = true)
            Api.net_debug()
        }
```
So without `uag_reset_transp()` will re-invites and re-registers now sent automatically when `net_add|rm_address()` call is made?

cspiel1:
> So without `uag_reset_transp()` will re-invites and re-registers now sent automatically when `net_add|rm_address()` call is made?

Yes.

juha-h:
If there is, say, one network address and it changes, the app needs to make `net_add_address()` call and `net_rm_address()` call in some order.  Will re-register and re-invite requests then be sent twice, i.e. once after both of these calls?

It would make sense to send them only once after the change has been completed.

cspiel1:
First you should call `net_add_address()`. Then `net_rm_address()`. Only the latter calls `call_reset_transp()` which sends the re-invite and `ua_register()` for the re-register, but only if the related laddr was removed.

juha-h:
OK, I got my Android app working using `net_add|rm_address()` calls on a device that only has WiFi network connection.  I'll try next to set up test environment where the device would have both WiFi and mobile data network connections.

cspiel1:
But note: The patches for the correct selection of laddr for outgoing/incoming calls are still missing. --> next PR.

modules/netroam/netroam.c
```diff
@@ -0,0 +1,156 @@
...
+	if (d.interval)
+		tmr_start(&d.tmr, d.interval * 1000, poll_changes, &d);
+
+	return cmd_register(baresip_commands(), cmdv, ARRAY_SIZE(cmdv));
+}
+
+
+static int module_close(void)
+{
+	tmr_cancel(&d.tmr);
```
alfredh, line 146: please note, a call to `cmd_unregister` is missing here

cspiel1:
Fixed the re-invite with last commit. But further:
- I think the re-register has to be done before the re-invite, right?

juha-h:
Christian Spielberger writes:

> Fixed the re-invite with last commit. But further:
> - I think the re-register has to be done before the re-invite, right?

Re-REGISTER allows new incoming requests to reach the UA, but re-INVITE
from the UA should work also before re-REGISTER as long as its Contact
URI has been updated.

But should also be OK to do re-REGISTER first.

juha-h:
juha-h writes:

> But should also be OK to do re-REGISTER first.

This is what the current uag_reset_transp does so better stick to it.

cspiel1:
The current code of `uag_transp_rm()` for re-REGISTER and re-INVITE does the same like `uag_reset_transp()`. So if the tests of @juha-h are okay we could merge this.

Then I will proceed with the laddr selection patches.

alfredh:
I think we should merge this soon (great work!)

juha-h:
Christian Spielberger writes:

> The current code of `uag_transp_rm()` for re-REGISTER and re-INVITE
> does the same like `uag_reset_transp()`. So if the tests of @juha-h
> are okay we could merge this.

My tests with net_add|rm_address() worked fine.  I have not tested any
other new API functions of the PR.

Please still keep net_set_address and uag_reset_transp for backwards
compatibility.

cspiel1:
Thanks for the tests! Regarding other new API functions: Just realized that `uag_transp_add()` and `uag_transp_rm()` are not needed in the API. Instead the App or a module (netroam) should use `net_add_address()` and `net_rm_address()`.

juha-h:
Just started to do Android tests when I got a second device with mobile and wifi connectivity.

Question:  can I call net_add|rm_address() without checking if the address has already been added or removed?

juha-h:
It turned out that in Android WiFI and Cellular data network cannot be both active at the same time.  When I have only cellular network connected, I get:
```
Active network 126 is available with caps: [ Transports: CELLULAR Capabilities: SUPL&INTERNET&NOT_RESTRICTED&TRUSTED&NOT_VPN&VALIDATED&NOT_ROAMING&FOREGROUND&NOT_CONGESTED&NOT_SUSPENDED LinkUpBandwidth>=51200Kbps LinkDnBandwidth>=102400Kbps Specifier: <3>], props: {InterfaceName: ccmni0 LinkAddresses: [ 100.66.96.68/32 ] DnsAddresses: [ /192.89.123.230,/192.89.123.231 ] Domains: null MTU: 1500 TcpBufferSizes: 2097152,4194304,8388608,262144,524288,1048576 Routes: [ 0.0.0.0/0 -> 100.66.96.68 ccmni0,100.66.96.68/32 -> 0.0.0.0 ccmni0 ]}
```
and then when I also connect WiFI, I get:
```
09-14 10:37:08.595 16525 16587 I Baresip : Network 126 is lost
09-14 10:37:08.612 16525 16587 I Baresip : Active network 128 is available with caps: [ Transports: WIFI Capabilities: INTERNET&NOT_RESTRICTED&TRUSTED&NOT_VPN&VALIDATED&NOT_ROAMING&FOREGROUND&NOT_CONGESTED&NOT_SUSPENDED LinkUpBandwidth>=1048576Kbps LinkDnBandwidth>=1048576Kbps SignalStrength: -34], props: {InterfaceName: wlan0 LinkAddresses: [ fe80::4950:4d1e:ba8e:b4d2/64,192.168.214.225/24 ] DnsAddresses: [ /192.168.214.149 ] Domains: null MTU: 0 TcpBufferSizes: 1730560,3461120,6922240,524288,1048576,4525824 Routes: [ fe80::/64 -> :: wlan0,192.168.214.0/24 -> 0.0.0.0 wlan0,0.0.0.0/0 -> 192.168.214.149 wlan0 ]}
```
That is, Cellular network is lost when WiFI becomes available.  Android is thus at any time using only one network: the one that it thinks gives the best Internet connectivity.

I still need to test what happens when VPN network becomes available.

juha-h:
I did some VPN tests and executed network debug whenever there was change in network connectivity.  For example, when VPN was active, I got:
```
09-14 18:50:43.973 26714 26762 D Baresip Lib: --- Network debug ---
09-14 18:50:43.973 26714 26762 D Baresip Lib: enabled interfaces:
09-14 18:50:43.973 26714 26762 D Baresip Lib:        tun0:  10.50.8.6
09-14 18:50:43.973 26714 26762 D Baresip Lib: network interfaces:
09-14 18:50:43.973 26714 26762 D Baresip Lib:          lo:  127.0.0.1
09-14 18:50:43.973 26714 26762 D Baresip Lib:      ccmni0:  100.73.234.15
09-14 18:50:43.973 26714 26762 D Baresip Lib:       wlan0:  192.168.214.225
09-14 18:50:43.973 26714 26762 D Baresip Lib:        tun0:  10.50.8.6
```
and when I paused VPN, I got:
```
09-14 18:51:39.401 26714 26761 D Baresip Lib: enabled interfaces:
09-14 18:51:39.401 26714 26761 D Baresip Lib:       wlan0:  192.168.214.225
09-14 18:51:39.401 26714 26761 D Baresip Lib: network interfaces:
09-14 18:51:39.401 26714 26761 D Baresip Lib:          lo:  127.0.0.1
09-14 18:51:39.401 26714 26761 D Baresip Lib:      ccmni0:  100.73.234.15
09-14 18:51:39.401 26714 26761 D Baresip Lib:       wlan0:  192.168.214.225
```
Stupid question:  why can't baresip automatically use all (non local) network interfaces?  Why would I need to make net_add|rm_address calls?

juha-h:
There is `net_dns_refresh` API function.  How about corresponding `net_address_refresh`?

juha-h:
If `net_address_refresh` would exist, netroam `poll_changes` could use it the same way as it i now using `net_dns_refresh`.  And since in Android baresip app gets automatically notified about network changes, it could also use that function when change occurs.

So please include `net_address_refresh` in net.c.

juha-h:
I looked at netroam code and started to wonder if it handles both ipv4 and ipv6 address additions/removals.  For example, when a new network becomes available, it could have both new IPv4 local address and new IPv6 local address.  Function `poll_changes` seems to be able to add or remove only one address.

juha-h:
Never mind.  I managed to implement network updates using net_add|rm_address functions.

cspiel1:
Will have a look at your comments on Monday. I am currently on vacation.

juha-h:
@cspiel1 replied:

> > OK, I'll first try to get my Android app working w/o `netroam` module by replacing `net_set_address()` calls with `net_add_address()` and `net_rm_address()` calls followed by `uag_reset_transp()` calls.
> > Is there any other changes needed?
> 
> Sorry, I edited may comment above again. You don't need to call `uag_reset_transp()`.

After tests on Android, I need to reopen this discussion.  Based on your reply and my tests `uag_reset_transp()` functionality seems to be embedded in `net_add_address()` and/or `net_rm_address()` call (which ones I'm not sure yet).

I want to have control of that myself.  For example, if user turns on flight mode, it may happen that three networks (vpn, wifi, mobile) will get disconnected at the same time.  So I make three `net_rm_address()` calls and after those I want to make a single `uag_reset_transp()` call.  Is that currently possible?

Also, when a network is added, it may not change the network that is currently active, in which case the current transport does not need to be reset and UAs don't need to re-register and calls don't need to get re-invited.

I started to examine this when I got crash in basesip lib when I removed several networks. It may have something to do with back-to-back `net_rm_address()` calls:
```
********** Crash dump: **********
Build fingerprint: 'Nokia/Essential2_00WW/ES2N_sprout:10/QP1A.190711.020/00WW_4_200:user/release-keys'
#00 0x000000000007e270 /apex/com.android.runtime/lib64/bionic/libc.so (__memcpy+96) (BuildId: 2a5abdc9c768b33656f7aa8d9ce5cf54)
#01 0x000000000015edac /data/app/com.tutpro.baresip-cklVitEszjewE0uSNTa71w==/lib/arm64/libbaresip.so (mbuf_write_str+152) (BuildId: 7e2c958edcd8eeffd5cab9f210ffdb7bc7f44a6b)
                                                                                                      mbuf_write_mem
                                                                                                      /usr/src/libbaresip-android/re/src/mbuf/mbuf.c:226:2
                                                                                                      mbuf_write_str
                                                                                                      /usr/src/libbaresip-android/re/src/mbuf/mbuf.c:304:9
#02 0x0000000000133598 /data/app/com.tutpro.baresip-cklVitEszjewE0uSNTa71w==/lib/arm64/libbaresip.so (sip_auth_encode+360) (BuildId: 7e2c958edcd8eeffd5cab9f210ffdb7bc7f44a6b)
                                                                                                      sip_auth_encode
                                                                                                      /usr/src/libbaresip-android/re/src/sip/auth.c:0:3
#03 0x00000000001364f4 /data/app/com.tutpro.baresip-cklVitEszjewE0uSNTa71w==/lib/arm64/libbaresip.so (sip_drequestf+164) (BuildId: 7e2c958edcd8eeffd5cab9f210ffdb7bc7f44a6b)
                                                                                                      sip_drequestf
                                                                                                      /usr/src/libbaresip-android/re/src/sip/request.c:879:10
#04 0x000000000013d708 /data/app/com.tutpro.baresip-cklVitEszjewE0uSNTa71w==/lib/arm64/libbaresip.so (BuildId: 7e2c958edcd8eeffd5cab9f210ffdb7bc7f44a6b)
                                                                                                      request
                                                                                                      /usr/src/libbaresip-android/re/src/sipreg/reg.c:302:9
                                                                                                      destructor
                                                                                                      /usr/src/libbaresip-android/re/src/sipreg/reg.c:81:27
#05 0x00000000001655b8 /data/app/com.tutpro.baresip-cklVitEszjewE0uSNTa71w==/lib/arm64/libbaresip.so (mem_deref+68) (BuildId: 7e2c958edcd8eeffd5cab9f210ffdb7bc7f44a6b)
                                                                                                      mem_deref
                                                                                                      /usr/src/libbaresip-android/re/src/mem/mem.c:318:3
#06 0x000000000012db74 /data/app/com.tutpro.baresip-cklVitEszjewE0uSNTa71w==/lib/arm64/libbaresip.so (reg_register+116) (BuildId: 7e2c958edcd8eeffd5cab9f210ffdb7bc7f44a6b)
                                                                                                      reg_register
                                                                                                      /usr/src/libbaresip-android/baresip/src/reg.c:211:16
#07 0x000000000011b7f4 /data/app/com.tutpro.baresip-cklVitEszjewE0uSNTa71w==/lib/arm64/libbaresip.so (BuildId: 7e2c958edcd8eeffd5cab9f210ffdb7bc7f44a6b)
                                                                                                      start_register
                                                                                                      /usr/src/libbaresip-android/baresip/src/ua.c:143:9
#08 0x000000000011e6dc /data/app/com.tutpro.baresip-cklVitEszjewE0uSNTa71w==/lib/arm64/libbaresip.so (BuildId: 7e2c958edcd8eeffd5cab9f210ffdb7bc7f44a6b)
                                                                                                      uag_transp_rm
                                                                                                      /usr/src/libbaresip-android/baresip/src/uag.c:496:11
                                                                                                      transp_rm_laddr
                                                                                                      /usr/src/libbaresip-android/baresip/src/uag.c:529:8
#09 0x00000000001170fc /data/app/com.tutpro.baresip-cklVitEszjewE0uSNTa71w==/lib/arm64/libbaresip.so (net_rm_address+116) (BuildId: 7e2c958edcd8eeffd5cab9f210ffdb7bc7f44a6b)
                                                                                                      net_rm_address
                                                                                                      /usr/src/libbaresip-android/baresip/src/net.c:523:5
#10 0x0000000000106644 /data/app/com.tutpro.baresip-cklVitEszjewE0uSNTa71w==/lib/arm64/libbaresip.so (Java_com_tutpro_baresip_Api_net_1rm_1address+248) (BuildId: 7e2c958edcd8eeffd5cab9f210ffdb7bc7f44a6b)
                                                                                                      Java_com_tutpro_baresip_Api_net_1rm_1address
                                                                                                      /usr/src/baresip-studio/app/src/main/cpp/baresip.c:1733:9
```

cspiel1:
> Just started to do Android tests when I got a second device with mobile and wifi connectivity.
> 
> Question: can I call net_add|rm_address() without checking if the address has already been added or removed?

- `net_add_address()`
   - Checks if given IP address is a local IP address.
   - Does not check if given IP address was already added. Should we add this?
- `net_rm_address()` needs no check. If the given IP address is not known, it does nothing.

juha-h:
It would simplify the app and prevent the same address from being added twice, if net_`add_address()` would check that the address does not already exist.  Would need same kind of loop over laddrs as in `net_rm_address()`.

cspiel1:
> There is `net_dns_refresh` API function. How about corresponding `net_address_refresh`?

The functionality of automatically adding/removing of single interface IP addresses now is in module `netroam`. Whereas `src/net.c` simply adds all local network interface IP addresses. A refresh in the sense of flush all, and add all again would be possible.

cspiel1:
> If `net_address_refresh` would exist, netroam `poll_changes` could use it the same way as it i now using `net_dns_refresh`. And since in Android baresip app gets automatically notified about network changes, it could also use that function when change occurs.
> 
> So please include `net_address_refresh` in net.c.

Ok. I will add this with flush all, and add all again. This may have some drawbacks, we will have to check/test this. But the behavior before the network changes was like that.

juha-h:
As I mentioned, I have managed to get my Android app working using `net_add|rm_address` functions, so would not anymore need `net_address_refresh` function.

As I explained, I still would like to control if `net_add|rm_address` functions reset transports and do re-registrations/re-invites.

cspiel1:
> I looked at netroam code and started to wonder if it handles both ipv4 and ipv6 address additions/removals. For example, when a new network becomes available, it could have both new IPv4 local address and new IPv6 local address. Function `poll_changes` seems to be able to add or remove only one address.

Thanks, I will try to reproduce/fix this. If IPv4 and a IPv6 address is added at once, then `netroam` should add both during one `poll_changes()`.

cspiel1:
> As I mentioned, I have managed to get my Android app working using `net_add|rm_address` functions, so would not anymore need `net_address_refresh` function.

Ok. Thanks!

> As I explained, I still would like to control if `net_add|rm_address` functions reset transports and do re-registrations/re-invites.

Only `net_rm_address()` will initiate re-registration if the given IP address is the local IP address of a registration. Baresip knows which UA is a registered UA and has the IP address as local IP address. Thus it should decide on its own if a re-registration is needed. Same for a re-invite of an active call. The responsible code is in `uag_transp_rm()`.

cspiel1:
- Manual re-INVITE, re-REGISTER still are possible with `uag_reset_transp()`. But this flushes and adds all transports also. You should know  when you need to call this in addition to `net_add_address()` and `net_rm_address()`.
- Will try to reproduce the crash for successive `net_rm_address()` calls.

juha-h:
Does also `net_add_address()` cause re-registrations/re-invites?  If it does, then I would not need to make `uag_reset_transp()` calls.

 For example, when device has mobile network connected and then wifi is connected and becomes the default network, calling  `net_add_address()` on the new wifi ip address should cause re-registrations/re-invites.

cspiel1:
> Does also `net_add_address()` cause re-registrations/re-invites? If it does, then I would not need to make `uag_reset_transp()` calls.
> 
> For example, when device has mobile network connected and then wifi is connected and becomes the default network, calling `net_add_address()` on the new wifi ip address should cause re-registrations/re-invites.

I see. This should be possible to decide by baresip. Needs another PR. Idea for the algorithm:
- for each registered UA,
  - get peer address P of registration
  - get local address L with route to P (system order of routes is used)
  - compare L with local address of registration
  - if there is a difference, do the re-registration
- Similar for active call(s).

juha-h:
The algorithm looks good.  Now I do this:
``` 
if (!addrsRemoved && addrsAdded)
            Api.uag_reset_transp(register = true, reinvite = true)
```
but the problem is that when, for example, VPN is connected that affects only routing to one UA, all UAs are re-registered/re-invited.

juha-h:
In order to test adding and removing of local network addresses with `net_add|rm_address` calls, I created this kind of test environment:

- Android device where baresip is running can have three IP network interfaces: mobile, WiFi, and VPN.  Any one of them can be enabled or disabled at any time.  Internet is reachable if mobile or WiFi interface is enabled.  VPN interface can only be used to reach lab.tutpro.com.  In addition to Internet, WiFi interface can also be used to reach test.tutpro.com that  has local 192.168.x.x address.
- baresip has three accounts: one in each domain lab.tutpro.com, test.tutpro.com, and sip2sip.info.

Examples of adding/removing networks starting from no networks:

- Mobile interface is enabled and `net_add_address` is used to add its address: lab.tutpro.com and sip2sip.info UAs should register successfully and registration of test.tutpro.com UA should fail.
- WiFI is enabled and `net_add_address` is used to add its address: lab.tutpro.com and sip2sip.info UAs should re-register (since default route has now changed from mobile to WiFi) and test.tutpro.com UA should register successfully
- VPN is enabled and  `net_add_address` is used to add its address: lab.tutpro.com UA should re-register successfully, since best route to lab.tutpro.com is now over VPN interface. 
- WiFi is disabled and `net_rm_address` is used to remove its address:  sip2sip.info UAs should re-register successfully (since default route has now been switched from WiFi to mobile) and re-registration of test.tutpro.com UA should fail.  There should be no effect on lab.tutpro.com UA, since it can still use the VPN interface (only underlying implementation of the tunnel has changed).
-  VPN is disabled and `net_rm_address` is used to remove its address: lab.tutpro.com UA should re-register successfully.

cspiel1:
Will prepare a PR that adds checks in `net_add_address()` if re-invite,re-register is needed.

juha-h:
Would it be OK to add API function `struct list *net_address_list(struct network *net)` that returns current laddr list of the network? It would help me to figure out if an address needs to be added or removed.

juha-h:
When I start baresip without having made any `net_add_address calls`, it still has found the local address(es):
```
09-22 09:21:52.054 16720 16765 D Baresip Lib: starting baresip
09-22 09:21:52.080 16720 16766 D Baresip Lib: Local network addresses:
09-22 09:21:52.081 16720 16766 D Baresip Lib:      ccmni0:  100.83.216.34
09-22 09:21:52.081 16720 16766 D Baresip Lib:       wlan0:  192.168.214.225
09-22 09:21:52.082 16720 16765 D Baresip Lib: --- Network debug ---
09-22 09:21:52.082 16720 16765 D Baresip Lib: enabled interfaces:
09-22 09:21:52.082 16720 16765 D Baresip Lib:      ccmni0:  100.83.216.34
09-22 09:21:52.082 16720 16765 D Baresip Lib:       wlan0:  192.168.214.225
09-22 09:21:52.082 16720 16765 D Baresip Lib: network interfaces:
09-22 09:21:52.082 16720 16765 D Baresip Lib:          lo:  127.0.0.1
09-22 09:21:52.082 16720 16765 D Baresip Lib:      ccmni0:  100.83.216.34
09-22 09:21:52.082 16720 16765 D Baresip Lib:       wlan0:  192.168.214.225
09-22 09:21:52.082 16720 16765 D Baresip Lib:  DNS Servers from Config: (3)
09-22 09:21:52.082 16720 16765 D Baresip Lib:    0: 192.168.214.149:53
09-22 09:21:52.082 16720 16765 D Baresip Lib:    1: 192.89.123.230:53
09-22 09:21:52.082 16720 16765 D Baresip Lib:    2: 192.89.123.231:53
```
How does baresip do it? 

Since baresip is able to discover the local addresses at start, would it also be able to do again after start if triggered by an API call?

If so, could this trigger call replace net_add|rm_address calls and make life much simpler?

juha-h:
Earlier I asked about adding `net_address_list()` API function.  Here is another (higher level) proposal:  Add API function
```
int net_update_addresses(struct network *net, struct list *addrs);
```
Parameter `addrs` would contain current `struct sa *` list of local addresses . The function would call `net_add_address` on those in `addrs` list that are missing from net's `laddrs` and would call `net_rm_address` on those in net's `laddrs` that are missing from `addrs`.

cspiel1:
Seems we have a moving target. The result of previous discussions and PRs is:
- The baresip core src/net.c adds all available interfaces and IP addresses.
- The  module netroam, handles changes of the network interfaces automatically. It handles changes by adding/removing single IP addresses, without having to flush and all again.
- If the application doesn't want the automatic behavior, then the manual functions `net_add_address()` and `net_rm_address()` can be used instead of module netroam.

Maybe you prefer to use the module netroam, and call command "netchange" when Android detects a network change.

juha-h:
It might be possible to use `netchange` command in Android app, if would not call `net_dns_refresh` since Android does not allow access to dns info.  May be config var could be added to choose if `net_dns_refresh` is called.

Regarding `net_add|rm_address()`, see https://github.com/baresip/baresip/issues/1611.  I have serious problems with them.

cspiel1:
> It might be possible to use `netchange` command in Android app, if would not call `net_dns_refresh` since Android does not allow access to dns info. May be config var could be added to choose if `net_dns_refresh` is called.

In `re/src/dns/ns.c` function `dns_srv_get()` has a path for Android. Does `get_android_dns()` not work? If so, are you able to fix this?

The function `net_dns_refresh()` should check if DNS settings have changed. This should be called whenever a network change was detected. So a config would not solve the problem.
> 
> Regarding `net_add|rm_address()`, see #1611. I have serious problems with them.

Ok, I will have a look on this as soon there is time. For me it turns out, that for Android module `netroam` with command `netchange` is favorable. We have it now in the default configuration enabled.

juha-h:
Christian Spielberger writes:

> In `re/src/dns/ns.c` function `dns_srv_get()` has a path for
> Android. Does `get_android_dns()` not work? If so, are you able to fix
> this?

It used to work until Google tightened the belt a few years ago.

> The function `net_dns_refresh()` should check if DNS settings have
> changed. This should be called whenever a network change was
> detected. So a config would not solve the problem.

When my Android app starts baresip, there exists dns_server lines in
config that I put there after asking Android about the interfaces and
dns servers.  So baresip always reports that it gets DNS servers from
config, e.g.

09-22 18:01:18.500  5054  5093 D Baresip Lib:  DNS Servers from Config: (4)
09-22 18:01:18.500  5054  5093 D Baresip Lib:    0: 62.241.198.245:53
09-22 18:01:18.500  5054  5093 D Baresip Lib:    1: 62.241.198.246:53
09-22 18:01:18.500  5054  5093 D Baresip Lib:    2: [2001:14b8:1000::1]:53
09-22 18:01:18.500  5054  5093 D Baresip Lib:    3: [2001:14b8:1000::2]:53

So if DNS servers come from config, netroam should not make
net_dns_refresh() call, since it is useless.  So could you add a check
to netroam module that skips net_dns_refresh() call is dns servers are
from config?

> > Regarding `net_add|rm_address()`, see #1611. I have serious problems with them.
> 
> Ok, I will have a look on this as soon there is time. For me it turns
> out, that for Android module `netroam` with command `netchange` is
> favorable. We have it now in the default configuration enabled.

I suspect that the problem has something fundamental to do with ipv6
address handling that happened when netroam module was introduced.  Now
at start, my baresip Android does not automatically detect ipv6
addresses and also setting them does not work.  This seems like a major
issue that should be fixed urgently.

juha-h:
How about calling  `net_dns_refresh()` only when `net->nsn == 0`, i.e., when there is no configured name servers?

cspiel1:
> How about calling `net_dns_refresh()` only when `net->nsn == 0`, i.e., when there is no configured name servers?

Ok, this might work. I will have a closer look.

juha-h:
Christian Spielberger writes:

> > How about calling `net_dns_refresh()` only when `net->nsn == 0`, i.e., when there is no configured name servers?
> 
> Ok, this might work. I will have a closer look.

Yes, but since net struct is not in baresip.h, I had to introduce new
netroam config param for the purpose in https://github.com/baresip/baresip/pull/1614.

cspiel1:
But `cfg->nsc` should be available.

cspiel1:
> > Does also `net_add_address()` cause re-registrations/re-invites? If it does, then I would not need to make `uag_reset_transp()` calls.
> > For example, when device has mobile network connected and then wifi is connected and becomes the default network, calling `net_add_address()` on the new wifi ip address should cause re-registrations/re-invites.
> 
> I see. This should be possible to decide by baresip. Needs another PR. Idea for the algorithm:
> 
> * for each registered UA,
>   
>   * get peer address P of registration
>   * get local address L with route to P (system order of routes is used)
>   * compare L with local address of registration
>   * if there is a difference, do the re-registration
> * Similar for active call(s).

Here the problem is, that the peer address `dst` of any request is not stored. Thus the `sipreg` object also does not store the `dst`. Also for the timer based re-REGISTERs a fresh DNS query is started. So such a check in `net_add_address()` is not possible. Instead we could
- rely on the timer based re-REGISTER,
- or always do the re-REGISTER for all UAs if the network changes somehow.

juha-h:
Actually, mobile network really disappears 60 secs after WiFi has become the active network. My conclusion is that since I have all necessary info available in Android app, I can handle net_add|rm_address calls from there and also make uag_reset_transp(register = true, reinvite = true) calls after each change.  So net_add|rm_address should not do any re-registers/re-invites on its own.  They should be done in netroam module, which I don't need to enable.

PR #1590, opened 2021-09-01
ua: select laddr with route to SDP offer address

cspiel1:
Re-opens: https://github.com/baresip/baresip/pull/1556
   
For incoming calls (INVITE) we get an SDP offer and can use the contained IP
address `dst` to find a local address which has a route to `dst`.
    
The function `sdp_connection()` replaces the `sdp_af_hint()` and still
retrieves the address family from the SDP offer. As fallback use address family
of SIP message source.
    
Checking if the address family is supported can be done also in
`ua_call_alloc()`. This avoids similar code in `sipsess_conn_handler()`.

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -752,8 +752,8 @@ int ua_call_alloc(struct call **callp, struct ua *ua,
 	memset(&cprm, 0, sizeof(cprm));
 
 	if (sa_isset(&dst, SA_ADDR)) {
-		sa_set_port(&dst, 53);
 		err = net_dst_source_addr_get(&dst, &cprm.laddr);
+		info("ua: laddr=%j\n", &cprm.laddr);
 		if (err) {
 			warning("ua: no laddr for %j (%m)\n", &dst, err);
 			return err;
--- a/src/ua.c
+++ b/src/ua.c
@@ -752,6 +752,7 @@ int ua_call_alloc(struct call **callp, struct ua *ua,
 	memset(&cprm, 0, sizeof(cprm));
 
 	if (sa_isset(&dst, SA_ADDR)) {
+		sa_set_port(&dst, 53);
 		err = net_dst_source_addr_get(&dst, &cprm.laddr);
 		if (err) {
 			warning("ua: no laddr for %j (%m)\n", &dst, err);
--- a/src/ua.c
+++ b/src/ua.c
@@ -753,7 +753,6 @@ int ua_call_alloc(struct call **callp, struct ua *ua,
 
 	if (sa_isset(&dst, SA_ADDR)) {
 		err = net_dst_source_addr_get(&dst, &cprm.laddr);
-		info("ua: laddr=%j\n", &cprm.laddr);
 		if (err) {
 			warning("ua: no laddr for %j (%m)\n", &dst, err);
 			return err;
@@ -763,7 +762,6 @@ int ua_call_alloc(struct call **callp, struct ua *ua,
 		sa_cpy(&cprm.laddr, net_laddr_af(net, af));
 	}
 
-	sa_cpy(&cprm.laddr, net_laddr_af(net, af));
 	cprm.vidmode = vmode;
 	cprm.af      = af;
 	cprm.use_rtp = use_rtp;
--- a/src/ua.c
+++ b/src/ua.c
@@ -526,23 +526,40 @@ static bool require_handler(const struct sip_hdr *hdr,
 }
 
 
-static int sdp_af_hint(struct mbuf *mb)
+static int sdp_connection(struct mbuf *mb, int *af, struct sa *sa)
 {
-	struct pl af;
+	struct pl pl1, pl2;
+	char *addr = NULL;
 	int err;
 
+	*af = AF_UNSPEC;
 	err = re_regex((char *)mbuf_buf(mb), mbuf_get_left(mb),
-		       "IN IP[46]+", &af);
+		       "IN IP[46]+ [^ \r\n]+", &pl1, &pl2);
 	if (err)
-		return AF_UNSPEC;
+		return EINVAL;
 
-	switch (af.p[0]) {
+	pl_strdup(&addr, &pl2);
+	switch (pl1.p[0]) {
 
-	case '4': return AF_INET;
-	case '6': return AF_INET6;
+	case '4': *af = AF_INET;
+		  break;
+	case '6': *af = AF_INET6;
+		  break;
 	}
 
-	return AF_UNSPEC;
+	/* OSX/iOS needs a port number for udp_connect() */
+	err = re_regex((char *)mbuf_buf(mb), mbuf_get_left(mb),
+		       "m=audio [0-9]+ ", &pl1);
+	if (err)
+		err = re_regex((char *)mbuf_buf(mb), mbuf_get_left(mb),
+				"m=video [0-9]+ ", &pl1);
+
+	if (err)
+		return EINVAL;
+
+	err = sa_set_str(sa, addr, pl_u32(&pl1));
+	mem_deref(addr);
+	return err;
 }
 
 
@@ -550,9 +567,7 @@ static int sdp_af_hint(struct mbuf *mb)
 void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
 {
 	struct config *config = conf_config();
-	const struct network *net = baresip_network();
 	const struct sip_hdr *hdr;
-	int af_sdp;
 	struct ua *ua;
 	struct call *call = NULL;
 	char to_uri[256];
@@ -602,26 +617,6 @@ void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
 				  "Content-Length: 0\r\n\r\n",
 				  &hdr->val);
 		return;
-	}
-
-	/* Check if offered media AF is supported and available */
-	af_sdp = sdp_af_hint(msg->mb);
-	if (af_sdp) {
-		if (!net_af_enabled(net, af_sdp)) {
-			warning("ua: SDP offer AF not supported (%s)\n",
-				net_af2name(af_sdp));
-			af_sdp = 0;
-		}
-		else if (!sa_isset(net_laddr_af(net, af_sdp), SA_ADDR)) {
-			warning("ua: SDP offer AF not available (%s)\n",
-				net_af2name(af_sdp));
-			af_sdp = 0;
-		}
-		if (!af_sdp) {
-			(void)sip_treply(NULL, uag_sip(), msg, 488,
-					 "Not Acceptable Here");
-			return;
-		}
 	}
 
 	(void)pl_strcpy(&msg->to.auri, to_uri, sizeof(to_uri));
@@ -717,16 +712,17 @@ int ua_call_alloc(struct call **callp, struct ua *ua,
 	const struct network *net = baresip_network();
 	struct call_prm cprm;
 	int af;
-	int af_sdp;
+	struct sa dst;
 	int err;
 
 	if (!callp || !ua)
 		return EINVAL;
 
-	if (msg && (af_sdp = sdp_af_hint(msg->mb))) {
-		info("ua: using AF from sdp offer: af=%s\n",
-		     net_af2name(af_sdp));
-		af = af_sdp;
+	if (msg && !sdp_connection(msg->mb, &af, &dst)) {
+		info("ua: using origin address %j of SDP offer\n", &dst);
+	}
+	else if (msg) {
+		af = sa_af(&msg->src);
 	}
 	else if (ua->acc->maf &&
 		   sa_isset(net_laddr_af(net, ua->acc->maf), SA_ADDR)) {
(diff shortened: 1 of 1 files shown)


juha-h:
I gave the PR a try and it didn't work for me:
```
TCP 192.26.134.1:47208 -> 192.26.134.12:5060
INVITE sip:192.26.134.12:5060;transport=tcp SIP/2.0
Via: SIP/2.0/TCP 192.26.134.1:5066;branch=z9hG4bK63313d078719932b;rport
Contact: <sip:test-0x564f1c5864e0@192.26.134.1:5066;transport=tcp>
Max-Forwards: 70
To: <sip:192.26.134.12:5060;transport=tcp>
From: "Test Account" <sip:test@test.tutpro.com>;tag=767c5d266aa26ea6
Call-ID: b56d013e3afec9d3
CSeq: 64291 INVITE
User-Agent: baresip v1.1.0 (x86_64/linux)
Allow: INVITE,ACK,BYE,CANCEL,OPTIONS,NOTIFY,SUBSCRIBE,INFO,MESSAGE,REFER
Supported: gruu
Content-Type: application/sdp
Content-Length: 882

v=0
o=- 1787544111 351707542 IN IP4 192.168.61.45
```

cspiel1:
This PR is only for incoming calls.

We had the all-together PR already. And it was to big for review.

juha-h:
I tried with incoming call and addresses in SDP were not correct in 200 OK:
```
$ baresip -s
baresip v1.1.0 Copyright (C) 2010 - 2021 Alfred E. Heggestad et al.
Local network addresses:
     wlp2s0:  192.168.133.45
   vboxnet0:  192.26.134.1
     wlp2s0:  fe80::6b31:374f:b910:ecfd
...
#
TCP 192.26.134.12:41010 -> 192.26.134.1:5068
INVITE sip:test@test.tutpro.com SIP/2.0
Via: SIP/2.0/TCP 192.26.134.12:5060;branch=z9hG4bK837b1e5fb9bf190b;rport
Contact: <sip:test-0x561725619d80@192.26.134.12:5060;transport=tcp>
Max-Forwards: 70
Route: <sip:192.26.134.1:5068;transport=tcp;lr>
To: <sip:test@test.tutpro.com>
From: <sip:test@sua.tutpro.com>;tag=af1e8db74f7efba9
Call-ID: 1fdeaef4dfaa6548
CSeq: 226 INVITE
User-Agent: baresip v1.1.0 (x86_64/linux)
Allow: INVITE,ACK,BYE,CANCEL,OPTIONS,NOTIFY,SUBSCRIBE,INFO,MESSAGE,REFER
Supported: gruu
Content-Type: application/sdp
Content-Length: 327

v=0
o=- 3478729827 515717806 IN IP4 192.26.134.12
s=-
c=IN IP4 192.26.134.12
...
```
#
TCP 192.26.134.1:5068 -> 192.26.134.12:5060
SIP/2.0 200 Answering
Via: SIP/2.0/TCP 192.26.134.12:5060;branch=z9hG4bK837b1e5fb9bf190b;rport=41010;received=192.26.134.12
To: <sip:test@test.tutpro.com>;tag=f4fc1c85fef702fc
From: <sip:test@sua.tutpro.com>;tag=af1e8db74f7efba9
Call-ID: 1fdeaef4dfaa6548
CSeq: 226 INVITE
Server: baresip v1.1.0 (x86_64/linux)
Contact: <sip:test@test.tutpro.com;gr=urn:uuid:9b21c1a8-dc58-a3d6-83d5-7be67cbe70ce>
Allow: INVITE,ACK,BYE,CANCEL,OPTIONS,NOTIFY,SUBSCRIBE,INFO,MESSAGE,REFER
Content-Type: application/sdp
Content-Length: 317

v=0
o=- 641252692 1363066048 IN IP4 192.168.133.45
s=-
c=IN IP4 192.168.133.45
...
```
If I have understood correctly, addresses in SDP should have been 192.26.134.1.

**Comment by cspiel1** [comment]
This PR changes ua.c to:
- look in SDP of INVITE for the o-line --> Address family IPv4 and remote address 192.26.134.12
- then selects with `udp_connect()` a local address  that has a route to 192.26.134.12

In master the selection of laddr is not related to the remote address.

I also don't understand why for your example 192.168.133.45 is the source address of the UDP socket. So the UDP socket is: 192.168.133.45 --> 192.26.134.12
Does audio work in both directions?

Do you see this line in your log?
```
ua: using origin address 192.26.134.12 of SDP offer
```

**Comment by juha-h** [comment]
This is how my routing table looks like:
```
# route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.133.118 0.0.0.0         UG    600    0        0 wlp2s0
169.254.0.0     0.0.0.0         255.255.0.0     U     1000   0        0 wlp2s0
192.26.134.0    0.0.0.0         255.255.255.0   U     0      0        0 vboxnet0
192.168.133.0   0.0.0.0         255.255.255.0   U     600    0        0 wlp2s0
```
and ip address of vboxnet0 is:
```
4: vboxnet0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 0a:00:27:00:00:00 brd ff:ff:ff:ff:ff:ff
    inet 192.26.134.1/24 brd 192.26.134.255 scope global vboxnet0
       valid_lft forever preferred_lft forever
```
So why is address 192.26.134.1 not used in SDP of 200 OK?

Yes, there is message:
```
ua: using origin address 192.26.134.12 of SDP offer
```

**Comment by juha-h** [comment]
Default route points to 192.168.133.118 and local address of interface wlp2s0 is 192.168.133.45. But "best" route to 192.26.134.12 is not the default route.

Host  with 192.26.134.12 may very well have no default route at all, but just route to 192.26.134.0/255.255.255.0, in which case it is not capable to sending media to 192.168.133.45.

**Comment by cspiel1** [comment]
Good question. But I am not a network expert. But I always use the new commands: `ip route`, `ip addr`

Your default route seems to be 192.168.133.118. Maybe you find more information with
```
ip route
# or
ip route show table local
ip route show table main
```
See: https://unix.stackexchange.com/questions/188584/which-order-is-the-route-table-analyzed-in

This PR uses the libre function `net_dst_source_addr_get()` of @sreimers.

**Comment by sreimers** [comment]
Can you try this test code:

```c
#include <getopt.h>
#include <re/re.h>

static void usage(void)
{
        re_fprintf(stderr, "usage: ./lookup host\n");
}


int main(int argc, char *argv[])
{
        int err;
        const char *host;
        struct sa dst_ip;
        struct sa src_ip;

        if (argc != 2) {
                usage();
                return -EINVAL;
        }

        host = argv[1];

        err = sa_set_str(&dst_ip, host, 53);
        IF_ERR_GOTO_OUT(err)

        err = net_dst_source_addr_get(&dst_ip, &src_ip);
        IF_ERR_GOTO_OUT(err);

        err = net_dst_source_addr_get(&dst_ip, &src_ip);
        IF_ERR_GOTO_OUT(err);

        re_fprintf(stdout, "src: %H > dst: %H\n", sa_print_addr, &src_ip,
                   sa_print_addr, &dst_ip);
out:
        if (err)
                re_fprintf(stderr, "error: %m\n", err);
        return err;
}
```

`gcc main.c -lre -DHAVE_INTTYPES_H -o lookup`

and then:

`./lookup 192.26.134.12`

juha-h:
In the stack discussion there was this reply:
```
The routing table is usually applied in order from "most specific" to "least specific". A destination of 0.0.0.0 with a netmask of 0.0.0.0, i.e. your default route, is the least specific possible and so will always be applied last.
```
This is that baresip must achieve, but I don't know how.

juha-h:
Here you go:
```
./lookup 192.26.134.12
src: 192.26.134.1 > dst: 192.26.134.12
```

cspiel1:
Why this works like expected?
The call to `net_dst_source_addr_get()` is twice. Mistake, or not?

sreimers:
> The call to `net_dst_source_addr_get()` is twice. Mistake, or not?

Mistake, edit the example code. Maybe the DST Port makes a difference?

cspiel1:
@juha-h Then let's have a try with the destination port 53. Please could try again if now the wanted local address is selected?

juha-h:
With this main.c:
```
include <getopt.h>
#include <re/re.h>

static void usage(void)
{
        re_fprintf(stderr, "usage: ./lookup host\n");
}


int main(int argc, char *argv[])
{
        int err;
        const char *host;
        struct sa dst_ip;
        struct sa src_ip;

        if (argc != 2) {
                usage();
                return -EINVAL;
        }

        host = argv[1];

        err = sa_set_str(&dst_ip, host, 0);
        IF_ERR_GOTO_OUT(err)

        err = net_dst_source_addr_get(&dst_ip, &src_ip);
        IF_ERR_GOTO_OUT(err);

        re_fprintf(stdout, "src: %H > dst: %H\n", sa_print_addr, &src_ip,
                   sa_print_addr, &dst_ip);
out:
        if (err)
                re_fprintf(stderr, "error: %m\n", err);
        return err;
}
```
I get:
```
$ ./lookup 192.26.134.12
src: 192.26.134.1 > dst: 192.26.134.12
```

cspiel1:
Please could you try the branch again. You should see the output:
```
ua: using origin address 192.26.134.12 of SDP offer
ua: laddr=192.26.134.1
```
And then `192.26.134.1` should be in SDP of 200 Answering.

cspiel1:
I will also install virtual box and try to reproduce your setup. Did you add a "Host-only Adapter" in vbox?

juha-h:
Yes, I see
```
ua: using origin address 192.26.134.12 of SDP offer
ua: laddr=192.26.134.1
```
but still 200 OK is not using it:
```
TCP 192.26.134.1:5068 -> 192.26.134.12:5060
SIP/2.0 200 Answering
Via: SIP/2.0/TCP 192.26.134.12:5060;branch=z9hG4bKdad5db46ae82f5b5;rport=50438;received=192.26.134.12
To: <sip:test@test.tutpro.com>;tag=20e4949a5a792c17
From: <sip:test@sua.tutpro.com>;tag=cf2edb3333e10602
Call-ID: 905f391a34a8afe6
CSeq: 22098 INVITE
Server: baresip v1.1.0 (x86_64/linux)
Contact: <sip:test@test.tutpro.com;gr=urn:uuid:9b21c1a8-dc58-a3d6-83d5-7be67cbe70ce>
Allow: INVITE,ACK,BYE,CANCEL,OPTIONS,NOTIFY,SUBSCRIBE,INFO,MESSAGE,REFER
Content-Type: application/sdp
Content-Length: 318

v=0
o=- 3116334757 1370225006 IN IP4 192.168.133.45
```

cspiel1:
Found the mistake. Sorry for the troubles! And thanks for your test with the VM!

juha-h:
Now SDP addresses were correct!

juha-h:
Christian Spielberger writes:

> I will also install virtual box and try to reproduce your setup. Did
> you add a "Host-only Adapter" in vbox?

I have two Network Adapters:

Adapter 1: Attached to NAT
Adapter 2: Host-only Adapter

The latter is configured with IP address 192.26.134.12:
```
$ more /etc/network/interfaces
source /etc/network/interfaces.d/*

# The loopback network interface
auto lo
iface lo inet loopback

# The primary network interface
allow-hotplug enp0s3
iface enp0s3 inet dhcp

auto enp0s8
iface enp0s8 inet static
  address 192.26.134.12
  netmask 255.255.255.0
```

And when I did the test I had done this:
```
$ more route.sh
ifdown enp0s3
route add default gw 192.26.134.1 metric 1
```

cspiel1:
@sreimers If it works also on MAC, or/and the demo code works on MAC with destination port 0, then this PR is ready for merge from my point of view.

cspiel1:
@juha-h Thanks for the vbox setup! I will install one for future tests.

cspiel1:
Now I have a similar setup with two network interfaces. It needs also this PR https://github.com/baresip/baresip/pull/1591 to make UA selection work at callee. Otherwise there is the error "UA not found" at callee.

```
$ ip addr
...
2: eth0: ...
    ...
    inet 10.1.0.2/24 brd 10.1.0.255 scope global eth0

3: enx503eaa4964f9
    ...
    inet 10.3.0.2/24 scope global enx503eaa4964f9
    ...
$ ip route
default via 10.1.0.1 dev eth0 onlink 
10.1.0.0/24 dev eth0 proto kernel scope link src 10.1.0.2 
10.2.0.0/24 via 10.1.0.181 dev eth0 
10.3.0.0/24 dev enx503eaa4964f9 proto kernel scope link src 10.3.0.2 
...
```
It works:
```
ua: sipsess connect via UDP 10.3.0.3:5060 --> 10.3.0.2:5060
A@jackson: selected
ua: using origin address 10.3.0.3 of SDP offer
call: alloc with params laddr=10.3.0.2, af=AF_INET, use_rtp=1
```
And `o=- 1535679361 1136262660 IN IP4 10.3.0.2.` in SDP of 200 Answering.

sreimers:
> @sreimers If it works also on MAC, or/and the demo code works on MAC with destination port 0, then this PR is ready for merge from my point of view.

Port 0 does not work on macos (can`t assign address error), but you set one for dst here?
https://github.com/baresip/baresip/pull/1590/files#diff-6e7a04f8ec629a75484be1ed2e359322851f5315d27022b968da3da51784af8bR550

cspiel1:
> > @sreimers If it works also on MAC, or/and the demo code works on MAC with destination port 0, then this PR is ready for merge from my point of view.
> 
> Port 0 does not work on macos (can`t assign address error), but you set one for dst here?
> https://github.com/baresip/baresip/pull/1590/files#diff-6e7a04f8ec629a75484be1ed2e359322851f5315d27022b968da3da51784af8bR550

Your absolutely right. Port 0 was no useful test case. The PR does it already correct, also for MAC OS.

cspiel1:
Is still somebody testing/reviewing?

juha-h:
Christian Spielberger writes:

> Is still somebody testing/reviewing?

I'm done with Linux.  I'll try on Android after the PR has been
committed.

alfredh:
thanks!

cspiel1:
Thanks for review and merge!
Coming soon: PR for selecting laddr for outgoing calls if the destination is an IP address.

PR #1603, opened 2021-09-18
Fixed last remaining ffmeg related compiler warning

juha-h:


alfredh:
I think the problem can be solved by adding a casting somewhere

juha-h:
Didn't work on Debian.

PR #1605, opened 2021-09-20
Net add laddr check 

juha-h:
Check that local address to be added has not been added already


juha-h:
Will try again.  ffmpeg stuff was not supposed to be there.

PR #1606, opened 2021-09-20
When adding local address, check that it has not been added already

juha-h:


src/net.c
```diff
@@ -205,12 +205,20 @@ static bool check_ipv6(void)
 static int net_add_laddr(struct network *net, const char *ifname,
 			 const struct sa *sa)
 {
+
```
cspiel1, line 208: empty line too much?

cspiel1:
Tested and was okay.

PR #1607, opened 2021-09-21
start without network

cspiel1:
Allows baresip to start without any network interface/local IP.

Instead only prints a warning.

The network addresses can be added later.


juha-h:
Christian Spielberger writes:

> Allows baresip to start without any network interface/local IP.

Thanks.  Tested on Android and now start w/o network(s) works fine.

PR #1614, opened 2021-09-23
Added netroam_dns config variable

juha-h:
Will disable net_dns_refresh() if set to 'no'.

### DIFF
--- a/modules/netroam/netroam.c
+++ b/modules/netroam/netroam.c
@@ -23,6 +23,7 @@ struct netroam {
 	const struct config_net *cfg;
 	struct network *net;
 	uint32_t interval;
+	bool roam_dns;
 	struct tmr tmr;
 	struct sa laddr;
 };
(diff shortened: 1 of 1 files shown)


cspiel1:
Why does `net_dns_refresh()` not work for Anrdoid?

juha-h:
Christian Spielberger writes:

> Why does `net_dns_refresh()` not work for Android?

https://developer.android.com/about/versions/oreo/android-8.0-changes.html#o-pri

But they are available to the app via connection manager callbacks as I
have shown in some examples.

So my app gets that info from the callbacks, writes corresponding
dns_server lines in config file and then starts baresip service.  And
when changes occur, my app learns them from the callbacks and then calls
net_use_nameserver() function for the update.

cspiel1:
The PR https://github.com/baresip/baresip/pull/1615 replaces this one.

PR #1615, opened 2021-09-23
net dns srv get

cspiel1:
Only query local DNS settings if necessary.

In Android function `dns_srv_get()` does not work according to @juha-h. If DNS was configured with `dns_server`, then this function needs not be called.

Another check for `config_net->nsc` in module `netroam` is not really need. Should we add it anyway?
```c
static void poll_changes(void *arg)
{
...
if (!cfg->nsc)
    net_dns_refresh(baresip_network());
...
```

### DIFF
--- a/modules/netroam/netroam.c
+++ b/modules/netroam/netroam.c
@@ -87,7 +87,9 @@ static void poll_changes(void *arg)
 {
 	struct netroam *n = arg;
 	bool changed = false;
-	net_dns_refresh(baresip_network());
+
+	if (!n->cfg->nsc)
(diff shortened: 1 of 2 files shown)


juha-h:
But since the check is not done in netroam module, there will unnecessary call of net_dns_refresh() function.  I would prefer to skip it altogether by testing cfg->nsc in netroam.c.

juha-h:
So I would like to add this test in netroam.c poll_changes:
```
	if (n->cfg->nsc == 0)
		net_dns_refresh(baresip_network());
```
I tested and it works.

Issue #1611, opened 2021-09-22
issues with adding of ipv6 addresses


juha-h:
When I start baresip in environment where cellular interface as both ipv4 and ipv6 address, ipv6 address does not show up in net_debug:
```
09-22 16:26:27.483 30616 30652 I Baresip : Network 136 link properties: {InterfaceName: rmnet_data1 LinkAddresses: [ 10.152.224.205/30,2001:14bb:ac:5547:67a0:c509:a625:cc4c/64 ] DnsAddresses: [ /62.241.198.245,/62.241.198.246,/2001:14b8:1000::1,/2001:14b8:1000::2 ] Domains: null MTU: 1464 TcpBufferSizes: 2097152,4194304,8388608,262144,524288,1048576 Routes: [ 0.0.0.0/0 -> 10.152.224.206 rmnet_data1 mtu 1464,::/0 -> fe80::bd89:360f:9d29:a217 rmnet_data1 mtu 1464,10.152.224.204/30 -> 0.0.0.0 rmnet_data1 mtu 0,2001:14bb:ac:5547::/64 -> :: rmnet_data1 mtu 0 ]}
...
09-22 16:26:27.637 30616 30654 D Baresip Lib: starting baresip
09-22 16:26:27.647 30616 30655 D Baresip Lib: Local network addresses:
09-22 16:26:27.647 30616 30655 D Baresip Lib:  rmnet_data1:  10.152.224.205
09-22 16:26:27.647 30616 30655 D Baresip Lib:       wlan0:  192.168.214.149
09-22 16:26:27.649 30616 30654 D Baresip Lib: --- Network debug ---
09-22 16:26:27.649 30616 30654 D Baresip Lib: enabled interfaces:
09-22 16:26:27.649 30616 30654 D Baresip Lib:  rmnet_data1:  10.152.224.205
09-22 16:26:27.649 30616 30654 D Baresip Lib:       wlan0:  192.168.214.149
09-22 16:26:27.649 30616 30654 D Baresip Lib: network interfaces:
09-22 16:26:27.649 30616 30654 D Baresip Lib:          lo:  127.0.0.1
09-22 16:26:27.649 30616 30654 D Baresip Lib:  rmnet_data1:  10.152.224.205
09-22 16:26:27.649 30616 30654 D Baresip Lib:       wlan0:  192.168.214.149
```
Also, when I try to add ipv6 address using `net_add_address`, nothing seems to be happening:
```
[...]

PR #1616, opened 2021-09-24
remove uag transp rm (#1611)

cspiel1:
net,uag: remove differential uag transport changes (#1611)
    
- removes the `uag_transp_rm()` and re-registration/re-invite from `net_rm_address()`
- adds function `net_add_address_ifname()`
- `net_add_address()`, `net_rm_address()` will not call `uag_transp_add()`, `uag_transp_rm()`
    
Instead `uag_reset_transp()` has to be called by the application/netroam module after one or several calls to `net_rm_address()`, `net_add_address()` or `net_add_address_ifname()`.

Previously we wanted to send only re-register for UAs for which we know that the local IP address changed. But what we didn't think about, that also the default route may change if an interface is added. Thus it is best to send re-register/re-invite for all UAs when the network changes.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -705,6 +705,8 @@ int  net_use_nameserver(struct network *net,
 			const struct sa *srvv, size_t srvc);
 int  net_set_address(struct network *net, const struct sa *ip);
 int  net_add_address(struct network *net, const struct sa *ip);
+int  net_add_address_ifname(struct network *net, const struct sa *sa,
+			    const char *ifname);
 int  net_flush_addresses(struct network *net);
 int  net_rm_address(struct network *net, const struct sa *ip);
 bool net_af_enabled(const struct network *net, int af);
@@ -713,8 +715,6 @@ void net_dns_refresh(struct network *net);
 int  net_dns_debug(struct re_printf *pf, const struct network *net);
 int  net_debug(struct re_printf *pf, const struct network *net);
 bool net_laddr_apply(const struct network *net, net_ifaddr_h *ifh, void *arg);
-void net_set_add_handler(struct network *net, net_ifaddr_h *ifh, void *arg);
-void net_set_rm_handler(struct network *net, net_ifaddr_h *ifh, void *arg);
 bool net_ifaddr_filter(const struct network *net, const char *ifname,
 		       const struct sa *sa);
 const struct sa *net_laddr_af(const struct network *net, int af);
--- a/modules/netroam/netroam.c
+++ b/modules/netroam/netroam.c
@@ -28,6 +28,7 @@ struct netroam {
 	uint32_t interval;
 	struct tmr tmr;
 	struct sa laddr;
+	bool reset;
 };
 
 
@@ -123,12 +124,19 @@ static void poll_changes(void *arg)
 		changed = true;
 	}
 
-	if (n->interval)
-		tmr_start(&n->tmr, changed ? 1000 : n->interval * 1000,
-			  poll_changes, n);
-
-	if (changed)
+	if (n->reset) {
+		uag_reset_transp(true, true);
+		n->reset = false;
 		print_changes(n);
+	}
+
+	if (changed) {
+		n->reset = true;
+		tmr_start(&n->tmr, 1000, poll_changes, n);
+	}
+	else if (n->interval) {
+		tmr_start(&n->tmr, n->interval * 1000, poll_changes, n);
+	}
 }
 
 
--- a/src/net.c
+++ b/src/net.c
@@ -407,7 +407,9 @@ int net_alloc(struct network **netp, const struct config_net *cfg)
 		goto out;
 	}
 
+#if !defined(ANDROID)
 	net_if_apply(add_laddr_filter, net);
+#endif
 	info("Local network addresses:\n");
 	if (!list_count(&net->laddrs))
 		warning("  None for net_interface: %s\n",
@@ -735,8 +737,10 @@ int net_debug(struct re_printf *pf, const struct network *net)
 	err |= re_hprintf(pf, "enabled interfaces:\n");
 	net_laddr_apply(net, if_debug_handler, argv);
 
+#if !defined(ANDROID)
 	err |= re_hprintf(pf, "network interfaces:\n");
 	err |= net_if_apply( if_debug_handler, argv);
+#endif
 
 	err |= net_dns_debug(pf, net);
 
--- a/src/net.c
+++ b/src/net.c
@@ -412,7 +412,7 @@ int net_alloc(struct network **netp, const struct config_net *cfg)
 #endif
 	info("Local network addresses:\n");
 	if (!list_count(&net->laddrs))
-		warning("  None for net_interface: %s\n",
+		info("  None for net_interface: %s\n",
 				str_isset(cfg->ifname) ? cfg->ifname : "-");
 	else
 		net_laddr_apply(net, print_addr, NULL);
--- a/src/net.c
+++ b/src/net.c
@@ -412,7 +412,7 @@ int net_alloc(struct network **netp, const struct config_net *cfg)
 #endif
 	info("Local network addresses:\n");
 	if (!list_count(&net->laddrs))
-		info("  None for net_interface: %s\n",
+		info("  None available for net_interface: %s\n",
 				str_isset(cfg->ifname) ? cfg->ifname : "-");
 	else
 		net_laddr_apply(net, print_addr, NULL);
--- a/src/net.c
+++ b/src/net.c
@@ -737,11 +737,6 @@ int net_debug(struct re_printf *pf, const struct network *net)
 	err |= re_hprintf(pf, "enabled interfaces:\n");
 	net_laddr_apply(net, if_debug_handler, argv);
 
-#if !defined(ANDROID)
-	err |= re_hprintf(pf, "network interfaces:\n");
-	err |= net_if_apply( if_debug_handler, argv);
-#endif
-
 	err |= net_dns_debug(pf, net);
 
 	return err;
--- a/src/net.c
+++ b/src/net.c
@@ -407,9 +407,7 @@ int net_alloc(struct network **netp, const struct config_net *cfg)
 		goto out;
 	}
 
-#if !defined(ANDROID)
 	net_if_apply(add_laddr_filter, net);
-#endif
 	info("Local network addresses:\n");
 	if (!list_count(&net->laddrs))
 		info("  None available for net_interface: %s\n",
--- a/src/net.c
+++ b/src/net.c
@@ -17,11 +17,6 @@ struct network {
 	uint32_t nsn;        /**< Number of configured name servers      */
 	struct sa nsvf[NET_MAX_NS];/**< Configured fallback name servers */
 	uint32_t nsnf;       /**< Number of configured fallback name servers */
-
-	net_ifaddr_h *addh;  /**< Local address added handler            */
-	void *addh_arg;
-	net_ifaddr_h *rmh;   /**< Local address removed handler          */
-	void *rmh_arg;
 };
 
 
@@ -202,8 +197,16 @@ static bool check_ipv6(void)
 }
 
 
-static int net_add_laddr(struct network *net, const char *ifname,
-			 const struct sa *sa)
+/**
+ * Add a local IP address with given interface name
+ *
+ * @param net  Network instance
+ * @param ip   IP address
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int  net_add_address_ifname(struct network *net, const struct sa *sa,
+			    const char *ifname)
 {
 	struct le *le;
 	struct laddr *laddr;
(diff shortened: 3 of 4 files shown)


cspiel1:
@juha-h Could you please test this branch on Android?
Module netroam will also not work (with IPv6) on Android because it needs `getifaddr`. So you are on the right way if you plan to use
- `net_add_address_ifname()`
- `net_rm_address()`
- `uag_reset_transp()`

juha-h:
I built 1616 based libbaresip and I'm currently testing it.  So far, I have only tried with ipv4 only device that has cellular and wifi networks. 

At start, baresip seems to learn automatically both wifi and cellular networks, although cellular network is not active.  I then use `net_add_address_ifname` to add wifi address:
```
09-24 11:12:35.038 18979 19189 D Baresip Lib: starting baresip
09-24 11:12:35.052 18979 19190 D Baresip Lib: Local network addresses:
09-24 11:12:35.052 18979 19190 D Baresip Lib:      ccmni0:  100.71.229.59
09-24 11:12:35.052 18979 19190 D Baresip Lib:       wlan0:  192.168.214.225
09-24 11:12:35.052 18979 19189 D Baresip Lib: adding address/ifname '192.168.214.225/wlan0'
09-24 11:12:35.053 18979 19189 D Baresip Lib: --- Network debug ---
09-24 11:12:35.053 18979 19189 D Baresip Lib: enabled interfaces:
09-24 11:12:35.053 18979 19189 D Baresip Lib:       wlan0:  192.168.214.225
09-24 11:12:35.053 18979 19189 D Baresip Lib: network interfaces:
09-24 11:12:35.053 18979 19189 D Baresip Lib:          lo:  127.0.0.1
09-24 11:12:35.053 18979 19189 D Baresip Lib:      ccmni0:  100.71.229.59
09-24 11:12:35.053 18979 19189 D Baresip Lib:       wlan0:  192.168.214.225
09-24 11:12:35.053 18979 19189 D Baresip Lib:  DNS Servers from Config: (3)
09-24 11:12:35.053 18979 19189 D Baresip Lib:    0: 192.168.214.149:53
09-24 11:12:35.053 18979 19189 D Baresip Lib:    1: 192.89.123.230:53
09-24 11:12:35.053 18979 19189 D Baresip Lib:    2: 192.89.123.231:53
```
The question is, should even more control be given to user by making automatic learning of local network addresses in `net_alloc()` made optional (perhaps via`const struct config_net` option)? Now it learns also cellular network, which is not usable (it does not have FOREGROUND capability):
```
NET_CAPABILITY_FOREGROUND
Indicates that this network is available for use by apps, and not a network that is being kept up in the background to facilitate fast network switching
```
Or does it not matter, because only the wifi interface is enabled?

II'll do more tests later today.

juha-h:
I tried also with device that has both ipv4 and ipv6 address.  When baresip is starting, I got to log:
```
09-24 12:22:39.565 25336 25384 D Baresip Lib: starting baresip
09-24 12:22:39.578 25336 25385 D Baresip Lib: Local network addresses:
09-24 12:22:39.578 25336 25385 D Baresip Lib:  rmnet_data2:  10.137.143.84
09-24 12:22:39.578 25336 25385 D Baresip Lib:       wlan0:  192.168.214.149
```
Then I add both addresses using `net_add_address_ifname`:
```
09-24 12:22:39.578 25336 25385 D Baresip Lib: uag: add local address 10.137.143.84
09-24 12:22:39.579 25336 25385 D Baresip Lib: uag: add local address 2001:14bb:699:f53e:e9b1:8a11:aa6b:2e47
```
Then when baresip is fully started, I call net_debug():
```
09-24 12:22:39.845 25336 25384 D Baresip Lib: --- Network debug ---
09-24 12:22:39.845 25336 25384 D Baresip Lib: enabled interfaces:
09-24 12:22:39.845 25336 25384 D Baresip Lib:  rmnet_data2:  10.137.143.84
09-24 12:22:39.845 25336 25384 D Baresip Lib:  rmnet_data2:  2001:14bb:699:f53e:e9b1:8a11:aa6b:2e47
09-24 12:22:39.845 25336 25384 D Baresip Lib: network interfaces:
09-24 12:22:39.845 25336 25384 D Baresip Lib:          lo:  127.0.0.1
09-24 12:22:39.845 25336 25384 D Baresip Lib:  rmnet_data2:  10.137.143.84
09-24 12:22:39.845 25336 25384 D Baresip Lib:       wlan0:  192.168.214.149
09-24 12:22:39.845 25336 25384 D Baresip Lib:  DNS Servers from Config: (4)
09-24 12:22:39.845 25336 25384 D Baresip Lib:    0: 62.241.198.245:53
09-24 12:22:39.845 25336 25384 D Baresip Lib:    1: 62.241.198.246:53
09-24 12:22:39.845 25336 25384 D Baresip Lib:    2: [2001:14b8:1000::1]:53
09-24 12:22:39.845 25336 25384 D Baresip Lib:    3: [2001:14b8:1000::2]:53
```
Everything seems to be OK, but why `net_degug()` under `network interfaces:` does not show the ipv6 address?

cspiel1:
There is already the config setting `net_interface`. You could set it to "none".

It seems that the Android version uses some deprecated interface for collecting IP addresses which works only for IPv4: `net_if_list()`

For now you could add the IPv6 manually after startup. When network changes, you should add/rm IPv4 and IPv6.

> Everything seems to be OK, but why `net_degug()` under `network interfaces:` does not show the ipv6 address?

Everything is Ok. The "network interfaces:" list also uses `net_if_apply()`. Should we remove this list. It is anyway not very informative and confusing. The relevant list is the "enabled interfaces:".

juha-h:
Thanks for the explanation.  If  network collection does not work in Android due to `net_if_list()`, perhaps that could then be turned off with `if defined (ANDROID)`.  Or if I set `net_interface` to "none" is the collection then not done?

I add the addresses using `net_add_address_ifname` between `baresip_init()` and `ua_init()` and that seems to be working fine.

juha-h:
I tried with `net_interface none` and got:
```
9-24 12:51:15.058 26385 26431 D Baresip Lib: Local network addresses:
09-24 12:51:15.058 26385 26431 D Baresip Lib:   None for net_interface: none
09-24 12:51:15.058 26385 26431 D Baresip Lib: Local network addresses:
09-24 12:51:15.058 26385 26431 D Baresip Lib:   None for net_interface: none
```
Does "none" have any special meaning or could the value as well have been "foobar"?

If it has special meaning, the above messages could be less confusing, e.g.:
```
Local network addresses:
 No network interface configured
```
or something like that?

cspiel1:
Ah, I see. Then `if defined (ANDROID)` seems to be the best option.

juha-h:
Is selecting outbound interface specific SDP address still worked on or it it already included in this PR?

cspiel1:
> Is selecting outbound interface specific SDP address still worked on or it it already included in this PR?

No. We should first finish the rest.

I add the ANDROID check here.

juha-h:
Looks good now:
```
09-24 14:03:17.126 27523 27696 D Baresip Lib: --- Network debug ---
09-24 14:03:17.126 27523 27696 D Baresip Lib: enabled interfaces:
09-24 14:03:17.126 27523 27696 D Baresip Lib:  rmnet_data1:  10.140.85.190
09-24 14:03:17.126 27523 27696 D Baresip Lib:  rmnet_data1:  2001:14bb:a0:5545:818d:5109:4166:b991
09-24 14:03:17.126 27523 27696 D Baresip Lib:  DNS Servers from Config: (4)
09-24 14:03:17.126 27523 27696 D Baresip Lib:    0: 62.241.198.245:53
09-24 14:03:17.126 27523 27696 D Baresip Lib:    1: 62.241.198.246:53
09-24 14:03:17.126 27523 27696 D Baresip Lib:    2: [2001:14b8:1000::1]:53
09-24 14:03:17.126 27523 27696 D Baresip Lib:    3: [2001:14b8:1000::2]:53
```
Is it so that there is no need for net_debug() to show local addresses, since they are those of enabled interfaces?

juha-h:
At start I get:
```
09-24 14:12:08.996 28219 28256 D Baresip Lib: starting baresip
09-24 14:12:09.011 28219 28257 D Baresip Lib: Local network addresses:
09-24 14:12:09.011 28219 28257 D Baresip Lib:   None for net_interface: -
```
Could the #endif moved below the if statement, since if net_if_apply() is not called, there cannot be any local network interfaces:
```
#if !defined(ANDROID)
	net_if_apply(add_laddr_filter, net);
#endif
	info("Local network addresses:\n");
	if (!list_count(&net->laddrs))
		warning("  None for net_interface: %s\n",
				str_isset(cfg->ifname) ? cfg->ifname : "-");
	else
		net_laddr_apply(net, print_addr, NULL);
```

cspiel1:
> Is it so that there is no need for net_debug() to show local addresses, since they are those of enabled interfaces?

No. The first list shows the enabled interfaces. The second shows all interfaces. This can be confusing. Should we remove the second list generally (not only for Android) ?

cspiel1:
> 09-24 14:12:09.011 28219 28257 D Baresip Lib:   None for net_interface: -
> ```
> 
> It is read and thus doesn't look good. Downgrading warning to info would help.
> 
> Or could the #endif moved below the if statement, since if net_if_apply() is not called, there cannot be any local network interfaces?
> 
> Instead of ANDROID, the define could use HAVE_GETIFADDRS, but that one is currently used only in re.

- `info` instead of `warning` is the best option.
- I think `#ifdef ANRDOID` is no good SW design. Currently this is only used in src/config.c. All other system differences are hidden in `re`. Shouldn't we solve this in `re`?

@sreimers 
- Isn't it correct that `net_getifaddrs()` provides also IPv6 on linux and MAC?
- Does the alternative `net_if_list()` provide IPv4 and IPv6 on Windows?
- Can we change the behavior so that the function gives an empty list if IPv6 can't be queried? This would mean that we remove the body of `net_if_list()` in `src/net/posix/pif.c`.

juha-h:
Christian Spielberger writes:

> > Is it so that there is no need for net_debug() to show local
> > addresses, since they are those of enabled interfaces? 
> 
> No. The first list shows the enabled interfaces. The second shows all
> interfaces. This can be confusing. Should we remove the second list
> generally (not only for Android) ?

It is a bit confusing.  Removing second list would be fine with me.

juha-h:
I agree that `#ifdef ANRDOID` is no good design.  Best if could be solved in re, since ifaddrs is available also in Android when API level >= 24. The INFO text could be `None available for net_interface`.

sreimers:
> * Isn't it correct that `net_getifaddrs()` provides also IPv6 on linux and MAC?

Yes most platforms should support getifaddrs and IPv6, only android (API level < 24 as mentioned by juha-h) has this problem,
maybe older embedded systems too?

> * Does the alternative `net_if_list()` provide IPv4 and IPv6 on Windows?

Supported by: https://github.com/baresip/re/blob/master/src/net/win32/wif.c#L25

> * Can we change the behavior so that the function gives an empty list if IPv6 can't be queried? This would mean that we remove the body of `net_if_list()` in `src/net/posix/pif.c`. It is already marked as deprecated.

From a fault tolerant perspective a IPv4 only fallback is better than no detection I think. Only IPv6 only is a problem.

BTW: The ICE Module uses `net_if_apply` for candidate detection, too. 

@juha-h can you release separate builds for android users? the `/proc/net/if_inet6` solution should work for users lower than Android 10 and for newer versions you can use API Level >=24?

cspiel1:
Thanks! I suggest its best to remove the ANDROID switch for now.

juha-h:
Sebastian Reimers writes:

> @juha-h can you release separate builds for android users? the
> `/proc/net/if_inet6` solution should work for users lower than Android
> 10 and for newer versions you can use API Level >=24?

This is getting too complicated.  I'll drop support for API levels below
24 altogether.  Already now baresip+ (the video version) requires API
level 24 or higher.  I'll make the same requirement for baresip (the
non-video version).

On the other hand, I want to keep on using net_add_address_ifname in
order to have control on which addresses are used.  All this works fine
now.  Only waiting for the code that chooses correct SDP addresses for
outgoing calls.

juha-h:
It is OK to me merge this PR, but does it break outgoing INVITEs in that IP addresses in SDP may anymore be correct?

If so, we would have broken baresip master, which should not happen.

cspiel1:
> It is OK to me merge this PR, but does it break outgoing INVITEs in that IP addresses in SDP may anymore be correct?
> 
> If so, we would have broken baresip master, which should not happen.

No, this PR "breaks" nothing, neither previous PRs. We have to improve this in future PR.

cspiel1:
This PR is ready for the merge in my opinion. Here a summary about the network patches:
- added support for multiple network addresses
- choose proper IP address for incoming calls
- This PR: cleanup and fix re-INVITE, re-REGISTER for some cases
- TODO: choose proper IP address for outgoing calls

Before the network patches there was one IPv4 and one IPv6 address. And the RTP stream for outgoing calls was with IPv4 by default.

alfredh (approved):

PR #1618, opened 2021-09-28
ua: select proper af and laddr for outgoing IP calls

cspiel1:
- For registrar-less accounts: If the dial URI of an outgoing call is an IP address, we want to select a local address with route to the destination. Note: Nothing changes for cases where the dial URI is a domain name.
- For registrar accounts: Use laddr of registration for SDP.

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -30,6 +30,7 @@ struct ua {
 	struct list hdr_filter;      /**< Filter for incoming headers        */
 	struct list custom_hdrs;     /**< List of outgoing headers           */
 	char *ansval;                /**< SIP auto answer value              */
+	struct sa dst;               /**< Current destination address        */
 };
 
 struct ua_xhdr_filter {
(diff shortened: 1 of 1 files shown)


juha-h:
Regarding registrar-less accounts, where outbound proxy has been defined with IP address, it should be possible to select local address with route to outbound proxy IP address even when dial URI hostpart is domain name.

cspiel1:
> Regarding registrar-less accounts, where (edit _single_) outbound proxy has been defined with IP address, it should be possible to select local address with route to outbound proxy IP address even when dial URI hostpart is domain name.

Okay. Will check this.

> 
> Edit: Regarding registrar accounts, registration tells to registrar, where the UA can be reached. I don't think registration has anything to do with outbound calls.

But the local IP of the registration most likely will be an IP address which is reachable also for the RTP stream. It makes sense that the RTP stream uses the same interface like the SIP registration.

> 
> Summary: everything should work in case where each UA has a single outbound proxy URI with IP address.

Okay.

juha-h:
I tested outbound calls in the case where accounts have single outbound proxy defined using IP address and correct IP address was used in SDP.

juha-h:
Thanks for the feature.  Is there still more to come, i.e., selecting af and laddr when destination (R-URI or outbound URI host) is domain name?   Just asking if it makes sense to publish new version of my Android app now or wait for the domain name support.

cspiel1:
Domains as dial URI is not so simple. This would mean some API changes and refactoring.
Reason: The DNS request is done at very low level in `re/src/sip/request.c`. And the `laddr` has to be selected in `src/ua.c`.

But I have still some patches for IPv6 link local. I hope that tomorrow there is some time to make a PR.

juha-h:
Christian Spielberger writes:

> Domains as dial URI is not so simple. This would mean some API changes
> and refactoring.
> Reason: The DNS request is done at very low level in
> `re/src/sip/request.c`. And the `laddr` has to be selected in
> `src/ua.c`.

Understand.  Hope there is some way to work it out, because support for
multiple simultaneous networks has been a major issue for a long time.

PR #1624, opened 2021-10-05
Support for IPv6 link local address for streams

cspiel1:
sets scope ID for
- RTP peer address
- RTCP peer address
- and choosing the IPv6 link local address for `laddr`


src/net.c
```diff
@@ -678,6 +694,30 @@ bool net_is_laddr(const struct network *net, struct sa *sa)
...
+
+	LIST_FOREACH(&net->laddrs, le) {
+		struct laddr *laddr = le->data;
+		struct sa *sa = &laddr->sa;
+		if (sa_af(sa) != AF_INET6 || !sa_is_linklocal(sa))
+			continue;
+
+		sa_set_scopeid(dst, sa_scopeid(&laddr->sa));
+		if (!net_dst_is_source_addr(dst, &laddr->sa))
+			return 0;
```
alfredh, line 714: perhaps the dst scope-id should only be set in the success case ?

src/ua.c
```diff
@@ -559,6 +560,7 @@ static int sdp_connection(struct mbuf *mb, int *af, struct sa *sa)
 		return EINVAL;
 
 	err = sa_set_str(sa, addr, pl_u32(&pl1));
+	err |= net_set_dst_scopeid(net, sa);
```
alfredh, line 563: add explicit check; ipv6 and link-local

alfredh:
is it possible to add a test-case for this?

since most modern machines have an ipv6 link-local address, it should work on most machines.
a simple call between A and B, with no sip proxy, and send media via IPv6 link local.

alfredh:
in the new ipv6ll test, is it possible to check that the RTP traffic is actually sent via link local ?

we can fetch the remote rtp addr used, and check that it is v6 and link-local ?

this addition would make the test slightly more robust :)

PR #1626, opened 2021-10-05
New module providing webrtc aec mobile mode filter

juha-h:
In my tests, aec mobile mode filtering is used in calls and baresip does not crash, but I don't know if there is any difference in call quality as compared to webrtc_aec module.

WebRtcAecm API calls don't take same arguments  as WebRtcAec calls,  The main difference is that WebRtcAecm API calls take sound bytes as int16_t when WebRtcAecm API them as float.

Review and other comments are appreciated.

### DIFF
--- a/modules/webrtc_aecm/aec.cpp
+++ b/modules/webrtc_aecm/aec.cpp
@@ -1,0 +1,153 @@
+/**
+ * @file aec.cpp  WebRTC Acoustic Echo Cancellation (AEC) Mobile Mode
+ *
+ * Copyright (C) 2010 Alfred E. Heggestad
+ */
+
+#include <re.h>
+#include <rem.h>
+#include <baresip.h>
+#ifdef HAVE_PTHREAD
+#include <pthread.h>
+#endif
+#include "aec.h"
+
+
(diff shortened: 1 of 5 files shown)


juha-h:
Alfred E. Heggestad writes:

> in general we should try to upgrade to the latest version of libwebrtc

I agree.  Current webrtc source that baresip webrtc_aec module is using
has two aec modes: aec and aecm.  In the latest webrtc code, aecm mode
is still there, but aec mode has been replaced with aec3 mode and
differences between the old aec mode and the new aec3 mode are big.

Today I was able to build libwebrtc.a for Android from current webrtc
source, but the lib only includes aecm mode. I haven't even tried to
include aec3 mode in the lib since looks like it has dependencies
(e.g. abseil) that are not available in NDK.

> did you test that the module is working on your hardware, e.g. a Linux
> laptop or Android phone ?

> how well does it suppress echo from the remote side ?

I tested that webrtc_aecm module was working both on my Debian laptop
and on my Android devices, but I don't have any objective measurements
on its echo suppression effectiveness.

Issue #1627, opened 2021-10-11
"Address already in use" error


juha-h:
I have defined in config:
```
sip_listen		192.168.146.45:5020
```
and when I try to start baresip, I get "Address already in use" error:
```
$ baresip -v
baresip v1.1.0 Copyright (C) 2010 - 2021 Alfred E. Heggestad et al.
Local network addresses:
     wlp2s0:  192.168.146.45
   vboxnet0:  192.26.134.1
     wlp2s0:  fe80::6b31:374f:b910:ecfd
   vboxnet0:  2002:db8:0:1::1
   vboxnet0:  fe80::800:27ff:fe00:0
uag: add local address 192.168.146.45
SIP Certificate: /home/jh/.baresip/cert.pem
ua: adding SIP CA file: /home/jh/.baresip/ca_list.pem
ua: adding SIP CA path: 
uag: add local address 192.26.134.1
tcp: sock_bind: bind: Address already in use (af=2, 192.168.146.45:5020)
ua: SIP Transport failed: Address already in use
ua: init failed (Address already in use)
ua: stop all (forced=1)
main: unloading modules..
```

juha-h:
Does net_interface allow port specification?  If not, then it is not an alternative to sip_listen.  I use sip_listen in situations where I have several SIP UAs or proxies on the same host and they all use different listening ports.

PR #1628, opened 2021-10-12
uag: respect setting sip_listen (#1627)

cspiel1:


juha-h:
I gave the PR a try and it worked OK.  Should there be some warning or error message if none of the local network addresses match the sip_listen address?  Now there isn't and registrations just mysteriously fail.

juha-h:
I added 
```
net_interface 192.168.146.45
```
and even that didn't help.  At start I correctly get only one local address:
```
Local network addresses:
     wlp2s0:  192.168.146.45
```
but in the test call, address `192.26.134.1` still appears in the SDP.

juha-h:
I agree "no" for `sip_listen`, since that is only for listening of sip requests.  But for `net_interface` I would say "yes".  My understanding has been that if `net_interface` has been specified, then that interface will be used for everything.  For example, a VPN user may want for security reasons enforce the use of VPN interface and only it.

juha-h:
It should not matter if `net_interface` is IP address or name of the network.  In both cases, baresip should use only that interface for everything.

juha-h:
I have one more wish regarding `sip_listen`.  Would it be a big deal to allow also there name of the interface instead of its IP address?  For example:
```
sip_listen		wlp2s0:5020
```
The rational is that, for example, in case of WLAN interface, the address may change each time when WLAN is turned on, but name of the interface stays the same.

PR #1630, opened 2021-10-13
select laddr for SDP with respect to net_interface

cspiel1:
- net: add function net_laddr_for()
- ua: select laddr only from network object


cspiel1:
In the Anrdoid app `net_interface` can be set to something invalid e.g. "none" in order to avoid the collection of any addresses. Then function `net_add_address()` can be used to add IPv6 addresses and `sip_listen` to set the listen port.

```
sip_listen    0.0.0.0:5060
# or
sip_listen    [0::]:5060
```
has the same result.

PR #1638, opened 2021-10-19
Use source address of request as SDP session and media address

juha-h:
There is comment in the code asking if sipsess_connect() call desc argument is really needed anymore or could it be replaced by NULL.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -2065,20 +2065,22 @@ static void redirect_handler(const struct sip_msg *msg, const char *uri,
 static int send_invite(struct call *call)
 {
 	const char *routev[1];
-	struct mbuf *desc;
 	int err;
 
 	routev[0] = account_outbound(call->acc, 0);
 
+#if 0
+	struct mbuf *desc;
 	err = call_sdp_get(call, &desc, true);
 	if (err)
 		return err;
 
-#if 0
 	info("- - - - - S D P - O f f e r - - - - -\n"
 	     "%b"
 	     "- - - - - - - - - - - - - - - - - - -\n",
 	     desc->buf, desc->end);
+
+	mem_deref(desc);
 #endif
 
 	err = sipsess_connect(&call->sess, uag_sipsess_sock(),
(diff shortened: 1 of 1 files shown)


juha-h:
> I think that it is not wanted to pass a `sdp_session` to `re/src/sipsess`. This are two independent SW layers.

I don't understand why it is not wanted.  sipsess_connect() has also call->acc argument.  Why they are not `independent SW layers`?

Actually, I though first to replace `struct sipsess` `struct mbuf *desc;` with `void *desc;` and then it could hold either `mbuf *` (text) or `struct sdp_session *`.  In case of INVITE, it would be the latter.

juha-h:
baresip call.c/send_invite() assumes that invite has sdp body and not some other body content type:
```
	err = call_sdp_get(call, &desc, true);
	if (err)
		return err;
```
Then is passes the body text to sipsess_connect().  My suggestion is to pass struct sdp_session * instead. It makes implementation simple and efficient. re code does not need to make any assumptions, because it knows if sdp argument was NULL or not.

juha-h:
But I agree that connect.c/invite() call should check if sess->sdp is not set and then include the body from desc as it was before.  I'll make this fix if we otherwise agree.

juha-h:
Christian Spielberger writes:

> The title of the PRs have to match.

OK, thanks.  The titles are now the same.  `retest` tests still fail.
I'll try to fix those if we make progress with this PR.

cspiel1:
Another solution could be that baresip makes a DNS request first. The result maybe could be buffered.

juha-h:
Christian Spielberger writes:

> The DNS request done in request.c could be made earlier in baresip to
> find the destination address, select laddr adequately.

There can be more than one destination address returned from DNS with
different priorities/weights.  In my tests this PR tries all of them in
sequence once request is successfully sent.

But I don't mind some other solution as long as it has benefits over
this one and someone else implements it.

If implementing a better solution (if one exists) takes time then I
suggest that this PR (unless bugs are found) is merged first and then
later replaced by a better one.

PR #1645, opened 2021-10-22
set sdp laddr to SIP src address

cspiel1:
replace pre-encoded SDP by sipsess_desc_handler() for INVITE

The sipsess_desc_handler() updates the laddr of the SDP session and encodes
the SDP just before the SIP INVITE is sent.


juha-h:
If setting of mediaaf does not mix well with this PR, I would not mind to give up setting it.

cspiel1:
I have to check this in the code.

cspiel1:
> Probability that audio works both ways would be higher if media ipv6 laddr would be chosen from the same interface as SIP laddr (if such address exists), so in this test `2001:14bb:a6:6c40::bb`.

This makes sense. And this needs extra handling/code.

juha-h:
Christian Spielberger writes:

> > Probability that audio works both ways would be higher if media ipv6 laddr would be chosen from the same interface as SIP laddr (if such address exists), so in this test `2001:14bb:a6:6c40::bb`.
> 
> This makes sense. And this needs extra handling/code.

If too much, it is OK with me to remove `account_set_mediaaf()` API
function.  I think there is no corresponding parameter in account config
nor a menu command.

cspiel1:
`account_set_mediaaf()` corresponds to 
`mediaaf={ipv4,ipv6,auto}`

Should we remove it and simplify the configuration? It was only added as kind of a workaround as alternative to a fix like in this PR.

juha-h:
OK with me to remove `account_set_mediaaf()`.

cspiel1:
The newest commit removes `mediaaf` account parameter. This seems to be obsolete.

More interesting would be a "media interface" `mediaif` parameter. But this would be another story.

This PR should make RTP simply work in any situation where SIP works.

juha-h:
I made my tests once more after https://github.com/baresip/baresip/pull/1645/commits/ac1889a6cdcfa1eaa23f10f024ad22b2349ac086 and had no problems.  So from my point of view, this PR could be merged.

juha-h:
Forgot to mention that of course /* and */ should be removed from here is call.c:
```
/*#if 0*/
	info("- - - - - S D P - O f f e r - - - - -\n"
	     "%b"
	     "- - - - - - - - - - - - - - - - - - -\n",
	     (*descp)->buf, (*descp)->end);
/*#endif*/
```

juha-h:
Does someone still want to review this PR or could be merged?

cspiel1:
@alfredh and/or @sreimers please have a look on it!

It contains also changes for `re`. https://github.com/baresip/re/pull/172
As a reference I used the already existing callback function: `sipsess_offer_h`

juha-h:
No comments so far. I consider this PR a major improvement to baresip and would like to get it merged.

Unless no one objects and it is not merger earlier, I'll merge it tomorrow.

juha-h:
Sebastian Reimers writes:

> Merged #1645 into master.

Thanks.

PR #1654, opened 2021-11-08
stream: split up sender/receiver

alfredh:
propsals for naming the sender/receiver structs:

1. struct sender
2. struct rtp_sender
3. struct rtpsender

same with "receiver".

Please chime in with your preference...


cspiel1:
With sender/receiver inside stream.c everything is clear. Obviously, RTP is meant.

PR #1658, opened 2021-11-16
call: new transfer call state to handle transfered calls correctly

cHuberCoffee:
* Whenever a call gets transfered via REFER message, the call state
  switches to TRANSFER. While in this state, the call should not be
  get on-hold by establishing the new one.
* It removes one problematic re-Invite (put call on-Hold) while
  the call gets closed by the peer. (normal transfer behavior)


juha-h:
Does the PR help with this:

https://groups.google.com/g/baresip/c/NUHJeMqAyTw/m/O6wxqRVuAQAJ

cHuberCoffee:
Short answer: No it does not help with your topic.

In case that a transferred call is rejected, the originator of the transfer will be notified via SIP NOTIFY (sipfrag 486 Rejected). This is only an example if the transfer-to rejects the call. In the case a SIP NOTIFY sipfrag => 300 is received (see `call.c:2360` `sipsub_notify_handler`), the call is handled via CALL_EVENT_TRANSFER_FAILED. Following down the line of the event shows me that in `ua.c` the call gets dereferenced. Also the UA_EVENT_CALL_TRANSFER_FAILED does in fact nothing but inform the user via a `info`.

juha-h:
Looks like baresip does not support nosub (rfc7614) REFER requests at all, which would make transfers simpler.

PR #1678, opened 2021-12-06
Transfer like rfc5589

cHuberCoffee:
Current behavior: In case the transfer-target refuses the call, all calls get closed.

[RFC5589 Section 6.3.1](https://datatracker.ietf.org/doc/html/rfc5589#section-6.3.1) shows a different behavior.
If the transfer-target refuses the call, the original call between transferor and transferee should be reestablished. This means the original call should be unhold.

@juha-h opened a question in the mailing list ["call transfer question"](https://groups.google.com/g/baresip/c/NUHJeMqAyTw/m/O6wxqRVuAQAJ) earlier this year. This PR changes the behavior such that a failed transfer will resume the original call.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -33,6 +33,7 @@ struct call {
 	struct sdp_session *sdp;  /**< SDP Session                          */
 	struct sipsub *sub;       /**< Call transfer REFER subscription     */
 	struct sipnot *not;       /**< REFER/NOTIFY client                  */
+	struct call *xcall;       /**< Cross ref Transfer call              */
 	struct list streaml;      /**< List of mediastreams (struct stream) */
 	struct audio *audio;      /**< Audio stream                         */
 	struct video *video;      /**< Video stream                         */
@@ -920,6 +921,7 @@ int call_alloc(struct call **callp, const struct config *cfg, struct list *lst,
 	/* inherit certain properties from original call */
 	if (xcall) {
 		call->not = mem_ref(xcall->not);
+		call->xcall = xcall;
 	}
 
 	if (call->af != AF_UNSPEC)
@@ -941,10 +943,14 @@ int call_alloc(struct call **callp, const struct config *cfg, struct list *lst,
 	list_append(lst, &call->le, call);
 
  out:
-	if (err)
+	if (err) {
 		mem_deref(call);
-	else if (callp)
+	}
+	else if (callp) {
 		*callp = call;
+		if (xcall)
+			xcall->xcall = call;
+	}
 
 	return err;
 }
@@ -1812,6 +1818,18 @@ static void sipsess_refer_handler(struct sip *sip, const struct sip_msg *msg,
 }
 
 
+static void xfer_cleanup(struct call *call, char *reason)
+{
+	if (call->xcall->state == CALL_STATE_TRANSFER) {
+		set_state(call->xcall, CALL_STATE_ESTABLISHED);
+		call_event_handler(call->xcall, CALL_EVENT_TRANSFER_FAILED,
+			reason);
+	}
+
+	call->xcall->xcall = NULL;
+}
+
+
 static void sipsess_close_handler(int err, const struct sip_msg *msg,
 				  void *arg)
 {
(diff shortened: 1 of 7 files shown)


src/call.c
```diff
@@ -862,6 +863,7 @@ int call_alloc(struct call **callp, const struct config *cfg, struct list *lst,
 	call->ansadir = SDP_SENDRECV;
 	call->ansvdir = SDP_SENDRECV;
 	call->use_rtp = prm->use_rtp;
+
```
cspiel1, line 866: unrelated change?

juha-h:
Thanks for the PR.  I tested it with my Android app and transfer worked fine when transfer target accepted the call.  But when transfer target rejected the call, calling `call_hold(call, false)` in event
```
case UA_EVENT_CALL_TRANSFER_FAILED:

```
didn't do anything, i.e., re-INVITE was not sent.  Instead the app hanged.

I was able to "fix" this by introducing 500 ms delay before calling `call_hold(call, false)`.  Then resume was done AFTER `presence: ua=0x773d9c36d0 got event 18 (TRANSFER_FAILED)` log message:
```
12-07 17:01:19.200  6757  6809 D Baresip Lib: ua event (TRANSFER_FAILED) 486 Rejected
12-07 17:01:19.200  6757  6809 D Baresip : got uaEvent transfer failed,486 Rejected/sip:test@test.tutpro.com/512183454992
12-07 17:01:19.202  6757  6810 D Baresip Lib: call: transfer failed: 486 Rejected
12-07 17:01:19.232  6757  6810 D Baresip Lib: presence: ua=0x773d9c36d0 got event 18 (TRANSFER_FAILED)
... 500ms delay is here ...
12-07 17:01:19.702  6757  6886 D Baresip : Resuming call 512183454992
12-07 17:01:19.702  6757  6810 D Baresip Lib: call: resume sip:test@tutpro.com
... everything works OK ...

```
If I resume the call immediately in UA_EVENT_CALL_TRANSFER_FAILED event, then this comes to log:
```
12-07 17:26:57.774  7038  7091 D Baresip Lib: ua event (TRANSFER_FAILED) 486 Rejected
12-07 17:26:57.774  7038  7092 D Baresip Lib: call: transfer failed: 486 Rejected
12-07 17:26:57.775  7038  7091 D Baresip : got uaEvent transfer failed,486 Rejected/sip:test@test.tutpro.com/512186282256
12-07 17:26:57.793  7038  7091 D Baresip : Resuming call 512186282256
... nothing after this and app hangs ...
```
I don't know it this is Android specific issue.  If so, I can live with the above describe delay hack.

juha-h:
Resume worked when I called `call_hold(call, false)` in native code immediately when UA_EVENT_CALL_TRANSFER_FAILED was received.  The hangup happened only when I passed UA_EVENT_CALL_TRANSFER_FAILED event to Kotlin and from there called `call_hold(call, false)` API function (which is executed within re thread).  So this problem had nothing to do with the PR itself.

juha-h:
OK with me to merge this PR.

juha-h:
One more comment: why is `menu_stop_play();` called in `case UA_EVENT_CALL_TRANSFER_FAILED`, since I don't see any play started in `call_xfer()`?

cHuberCoffee:
The code in `UA_EVENT_CALL_TRANSFER_FAILED` combines both sides, the transfereor and the transferee. If the call gets transfered, the transferee will have a signaling tone for the call transferee <-> transfer-target. If that call failed the menu has to stop any existing menu players. For the transferor, this statement has no effect since no menu audio was played in the first place.

juha-h:
Christoph Huber writes:

> The code in `UA_EVENT_CALL_TRANSFER_FAILED` combines both sides, the
> transfereor and the transferee. If the call gets transfered, the
> transferee will have a signaling tone for the call transferee <->
> transfer-target. If that call failed the menu has to stop any existing
> menu players. For the transferor, this statement has no effect since
> no menu audio was played in the first place.

OK, got it.  Two other menu related thoughts:

- currently there is no nohold music
- permission should be asked before call is made to transfer target

PR #1690, opened 2022-01-04
Support >20ms ptime values for AMR codec

adevenport:
Modified the AMR RTP depayloader to inspect the TOC entries for multiple AMR frames instead of assuming there is only a single 20ms frame in each packet.
Modified the payloader to encode multiple frames when handed more than 20ms worth of samples.

I've validated both the bandwidth-efficient and octet-aligned payload types with AMR-NB since that's all my client supports. AMRWB should probably be validated before merging this into the mainline, but I don't have a reference client to support that.

### DIFF
--- a/modules/amr/amr.h
+++ b/modules/amr/amr.h
@@ -7,7 +7,9 @@
 struct amr_aucodec {
 	struct aucodec ac;
 	bool aligned;
-	uint8_t *be_dec_arr;
+	uint8_t *dec_arr;
+	uint8_t *enc_arr;
+
 };
 
 bool amr_octet_align(const char *fmtp);
--- a/src/audio.c
+++ b/src/audio.c
@@ -55,7 +55,7 @@
 enum {
 	MAX_SRATE       = 48000,  /* Maximum sample rate in [Hz] */
 	MAX_CHANNELS    =     2,  /* Maximum number of channels  */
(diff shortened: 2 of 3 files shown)


alfredh:
thanks! do you have a list of endpoints that support AMR ?

modules/amr/amr.h
```diff
@@ -1,15 +1,17 @@
...
-
-struct amr_aucodec {
-	struct aucodec ac;
-	bool aligned;
-	uint8_t *be_dec_arr;
-};
-
-bool amr_octet_align(const char *fmtp);
-int  amr_fmtp_enc(struct mbuf *mb, const struct sdp_format *fmt,
-		  bool offer, void *arg);
```
alfredh, line 15: can you please check this file ?

it looks like it has been converted to Windows format.

src/audio.c
```diff
@@ -1368,6 +1368,8 @@ int audio_alloc(struct audio **ap, struct list *streaml,
 				   "minptime", "%u", minptime);
 	err |= sdp_media_set_lattr(stream_sdpmedia(a->strm), true,
 				   "ptime", "%u", ptime);
+	err |= sdp_media_set_lattr(stream_sdpmedia(a->strm), true,
+				   "maxptime", "%u", MAX_PTIME);
```
alfredh, line 1372: please consider to make a separate PR for the changes in audio.c

PR #1695, opened 2022-01-12
sip: add optional TCP source port

cspiel1:
related to: https://github.com/baresip/re/pull/198

Adds a new account related parameter for the SIP TCP source port.

Some Session Border Controller (SBC) or other SIP components require the
ability to set a static source port at a UAC for SIP TCP connections. If the
new setting `sip_tcpsrcport` is not used, the source port still is
randomly chosen from the ephemeral ports range.

UDP source port can already be set by `sip_listen` setting.

SIP devices that have such a configuration setting: Konftel 800, Mitel 6800/6900

### DIFF
--- a/docs/examples/accounts
+++ b/docs/examples/accounts
@@ -36,6 +36,7 @@
 #    ;stunuser=STUN/TURN/ICE-username
 #    ;stunpass=STUN/TURN/ICE-password
 #    ;stunserver=stun:[user:pass]@host[:port]
+#    ;tcpsrcport=6060
 #    ;video_codecs=h264,h263,...
 #
 # Examples:
--- a/src/account.c
+++ b/src/account.c
@@ -409,7 +409,7 @@ static int sip_params_decode(struct account *acc, const struct sip_addr *aor)
 	u32  = 0;
 	err |= param_u32(&u32, &aor->params, "tcpsrcport");
 	if (u32) {
-		if (u32 <= 0xffff)
+		if (u32 <= 65535)
 			acc->tcpsrcport = u32;
 		else
 			warning("account: invalid tcpsrcport\n");
--- a/src/account.c
+++ b/src/account.c
@@ -386,6 +386,7 @@ static int sip_params_decode(struct account *acc, const struct sip_addr *aor)
 {
 	struct pl auth_user, tmp;
 	size_t i;
+	uint32_t u32;
 	int err = 0;
 	char *value;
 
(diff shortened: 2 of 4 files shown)


cspiel1:
Convert to Draft: Because have to check if this setting makes sense to be set per account.

alfredh:
adding more config options will increase the complexity, just for the record.

is it an option to connect FROM the listening port instead ?

for example, if the TCP socket is listening on port 5050, all outgoing TCP connections
will be made FROM port 5050. it remains to be seen that it works, of course, on the
supported platforms.

cspiel1:
In the meanwhile we came to the point that we need it per account (only for registered accounts). That's why it is still a draft. The TCP Listen socket is unique. So, not really an option.

I pushed today the newest version. Since the settings is used in `src/reg.c`, I think that the "extra" parameter should not be used for this.

docs/examples/accounts
```diff
@@ -36,6 +36,7 @@
 #    ;stunuser=STUN/TURN/ICE-username
 #    ;stunpass=STUN/TURN/ICE-password
 #    ;stunserver=stun:[user:pass]@host[:port]
+#    ;tcpsrcport=6060
```
alfredh, line 39: does this config apply to TCP transport only ?

or TCP, TLS, WS, WSS ?

src/account.c
```diff
@@ -405,6 +406,14 @@ static int sip_params_decode(struct account *acc, const struct sip_addr *aor)
 	err |= param_u32(&acc->fbregint, &aor->params, "fbregint");
 	acc->pubint = 0;
 	err |= param_u32(&acc->pubint, &aor->params, "pubint");
+	u32  = 0;
+	err |= param_u32(&u32, &aor->params, "tcpsrcport");
+	if (u32) {
+		if (u32 <= 0xffff)
```
alfredh, line 412: probably better to use decimal port here (65535).

src/reg.c
```diff
@@ -229,10 +229,18 @@ int reg_register(struct reg *reg, const char *reg_uri, const char *params,
 	if (acc && acc->fbregint)
 		err = sipreg_set_fbregint(reg->sipreg, acc->fbregint);
 
+	if (acc && acc->tcpsrcport)
+		err = sipreg_set_srcport(reg->sipreg, acc->tcpsrcport);
```
alfredh, line 233: can this function fail ?

docs/examples/accounts
cspiel1: TCP and TLS. Because TLS uses also the TCP code.

WS, WSS would need other changes.

src/reg.c
cspiel1: Not really. I will change the function to return void.

alfredh:
looks good now, please go ahead and merge to master

Issue #1699, opened 2022-01-20
Mickey Mouse Voice after audio source replacement


cspiel1:
Mentioned here: https://github.com/baresip/baresip/issues/1697#issuecomment-1016155771

During a G722 call, change audio source to aufile and specify a WAV file with 8000Hz samplerate.

```
/ausrc aufile,/usr/local/share/baresip/ringback.wav
```
--> The file is streamed to the peer, but with ugly quality. It seems that the resampler is missing or working wrong.

Then change the ausrc back to pulse/alsa.

```
/ausrc pulse,xxx
```

The voice sounds distorted on the peer (Mickey Mouse voice).

The solution should go with these plans: https://github.com/baresip/re/discussions/184

@sreimers Should I start with a first PR? One of these:
- Move resample processing to a filter module.
- Remove jbuf from audio.c and stream.c and replace by adaptive aubuf.

sreimers:
> @sreimers Should I start with a first PR? One of these:

Go for it :)

alfredh:
please have a look at this code:

```c
/*
 * Read samples from Audio Source
 *
 * @note This function has REAL-TIME properties
 *
 * @note This function may be called from any thread
 */
static void ausrc_read_handler(struct auframe *af, void *arg)
{
	struct audio *a = arg;
	struct autx *tx = &a->tx;
	size_t num_bytes = auframe_size(af);

	if (tx->src_fmt != af->fmt) {
		warning("audio: ausrc format mismatch:"
			" expected=%d(%s), actual=%d(%s)\n",
			tx->src_fmt, aufmt_name(tx->src_fmt),
			af->fmt, aufmt_name(af->fmt));
		return;
	}

	if (tx->muted)
		auframe_mute(af);

	if (aubuf_cur_size(tx->aubuf) >= tx->aubuf_maxsz) {

		++tx->stats.aubuf_overrun;

		debug("audio: tx aubuf overrun (total %llu)\n",
		      tx->stats.aubuf_overrun);
	}

	(void)aubuf_write(tx->aubuf, af->sampv, num_bytes);
```

please check the auframe here, does it have samplerate 8000 Hz,
while the audio sender is sending 16000 Hz ?

I am not sure if the baresip audio core can handle switching of samplerate dynamically.
this should be possible to fix, just save the samplerate and check auframe
if samplerate is the same or different.

cspiel1:
[...]

PR #1443, opened 2021-05-11
audio: add pre filter (baresip_au_pre_filtl)

sreimers:
Needed for #1411

 Processing encoder pipeline:
```
 .   .-------.  .--------.  .-------.  .--------.  .--------.  .--------.
 |   |       |  |        |  |       |  |        |  |        |  |        |
 |O->| ausrc |->|pre_filt|->| aubuf |->| resamp |->| aufilt |->| encode |->RTP
 |   |       |  |        |  |       |  |        |  |        |  |        |
 '   '-------'  '--------'  '-------'  '--------'  '--------'  '--------'
```

Processing decoder pipeline:
```
     .--------.  .--------.  .-------.  .--------.  .--------.  .--------.
|\   |        |  |        |  |       |  |        |  |        |  |        |
| |<-| auplay |<-|pre_filt|<-| aubuf |<-| resamp |<-| aufilt |<-| decode |<-RTP
|/   |        |  |        |  |       |  |        |  |        |  |        |
     '--------'  '--------'  '-------'  '--------'  '--------'  '--------'
```

PR #1705, opened 2022-01-25
module auresamp

cspiel1:
Moves resample code from audio.c to a new module `auresamp`.

Proposed by @alfredh here: https://github.com/baresip/baresip/pull/1443

More discussions in this issue: https://github.com/baresip/baresip/issues/1699

TODOs:
- Not sure if the `struct auresamp_st` should have an "optional" inheritance by means of union.

### DIFF
--- a/modules/auresamp/auresamp.c
+++ b/modules/auresamp/auresamp.c
@@ -1,0 +1,234 @@
+/**
+ * @file auresamp.c  A filter module that inserts a resampler into the audio
+ *                   pipeline if needed
+ *
+ * Copyright (C) 2022 Commend.com - c.spielberger@commend.com
+ */
+
+#include <re.h>
+#include <rem.h>
+#include <baresip.h>
+
+enum {
+	MAX_PTIME       =    60,  /* Maximum packet time in [ms] */
+};
+
+/**
+ *  The auresamp module is one of the audio filters. The order of the filters
+ *  is specified by the order in the config file.
+ *
+ *  .    .-------.   .----------.   .-------.   .---------.
+ *  |    |       |   |          |   |       |   |         |
+ *  |O-->| ausrc |-->| auresamp |-->| aubuf |-->| encoder |--> RTP
+ *  |    |       |   |          |   |       |   |         |
+ *  '    '-------'   '----------'   '-------'   '---------'
+ *
+ *       .--------.   .-------.   .----------.   .--------.
+ * |\    |        |   |       |   |          |   |        |
+ * | |<--| auplay |<--| aubuf |<--| auresamp |<--| decode |<-- RTP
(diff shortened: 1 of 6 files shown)


modules/auresamp/auresamp.c
```diff
@@ -0,0 +1,234 @@
...
+	struct auresamp resamp;  /* resampler                */
+	struct aufilt_prm oprm;  /* filter output parameters */
+};
+
+
+static void destructor(void *arg)
+{
+	struct auresamp_st *st = arg;
+
+	mem_deref(st->rsampv);
```
alfredh, line 50: is there a `list_unlink` missing here ?

alfredh:
some more comments:

- we should check that the samplerate of the audio frame sent to the encoder
matches the samplerate of the audio codec (same with channels).

this can happen, for example if the user forgets to load the auresamp.so module

modules/auresamp/auresamp.c
cspiel1: Indeed! Because it is a filter.

modules/auresamp/auresamp.c
```diff
@@ -0,0 +1,234 @@
...
+	if (af->fmt != AUFMT_S16LE) {
+		if (!st->sampv)
+			err = sampv_alloc(st, af);
+
+		if (err)
+			return err;
+
+		auconv_to_s16(st->sampv, af->fmt, af->sampv, af->sampc);
+		sampv = st->sampv;
+	}
```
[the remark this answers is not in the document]
cspiel1: I guess not. Because module auresamp will use `auresamp` of rem, like audio.c does so far. And this resampler only supports s16le. The audio conversion that we need here is from input format to s16le and after resampling to the output format.

Module `auconv` converts the samples to the configured `config_audio.play_fmt` and `config_audio.enc_fmt`. BTW I think we should also change the initialization of `auconv` module to a similar behavior like we plan for auresamp:
- use `struct auframe` for the input format
- use `struct aufilt_prm` for the output format

But the conversion to/from s16le in `auresamp` is only internally necessary.

cspiel1:
Had to rebase. Saw that you had already the idea to use `auframe` for `encode_rtp_send()`.

cspiel1:
Next PR would be to remove the `auconv_from_s16` from `aurx_stream_decode()` and directly call `aubuf_write_auframe()`. --> - 28 lines in audio.c

Then in `rem` the function `aubuf_write_auframe()` will be made to insert audio frames sorted by the timestamp.

sreimers:
Maybe we should add `auresamp` as default filter?

https://github.com/baresip/baresip/blob/master/src/config.c#L939

cspiel1:
Yes, this makes sense.

Issue #1676, opened 2021-12-05
AV1 issue with libaom 3.2.0 on Debian 11 against Google Chrome


juha-h:
When I make a video call using AV1 codec from Debian 11 where baresip is build with libaom 3.2.0 to Google Chrome 96.0, video window does not become visible and I baresip produces to log these kind of messages:
```
foo@test.tutpro.com: Call answered: sip:test@test.tutpro.com
stream: update 'audio'
stream: receive: setting SSRC: e75f2df4
stream: update 'video'
stream: receive: setting SSRC: 179e8320
audio: Set audio decoder: opus 16000Hz 1ch
audio: player started with sample format S16LE
audio: Set audio encoder: opus 16000Hz 1ch
webrtc_aec: creating shared state: [16000 Hz, 1 channels, blocksize 160 samples]
audio: source started with sample format S16LE
audio tx pipeline:        alsa ---> webrtc_aec ---> opus
audio rx pipeline:        alsa <--- webrtc_aec <--- opus
Set video encoder: AV1  (500000 bit/s, 15.00 fps)
Set video decoder: AV1 
v4l2: /dev/video0: found valid V4L2 device (640 x 480) pixfmt=YUYV
v4l2: Current input: "Camera 1"
v4l2: current framerate is 30.00 fps
video tx pipeline:       v4l2 ---> selfview_pip ---> AV1
video rx pipeline:     (disp) <--- selfview_pip <--- AV1
foo@test.tutpro.com: Call established: sip:test@test.tutpro.com
stream: incoming rtp for 'audio' established, receiving from 192.26.134.10:60020
stream: incoming rtp for 'video' established, receiving from 192.26.134.10:60062
av1: decode error: Unspecified internal error0/0.0    
     
av1: decode error: Unspecified internal error
     
av1: decode error: Unspecified internal error0/0.0    
     
av1: decode error: Unspecified internal error0/0.0    
     
av1: decode error: Unspecified internal error
     
av1: decode error: Unspecified internal error0/0.0    
     
av1: decode error: Corrupt frame detected
     No sequence header
av1: decode error: Corrupt frame detecteds=0.0/0.0    
     No sequence header
av1: decode error: Corrupt frame detected
     No sequence header
av1: decode error: Corrupt frame detecteds=0.0/0.0    
     No sequence header
av1: decode error: Corrupt frame detecteds=0.0/0.0    
     No sequence header
av1: decode error: Corrupt frame detecteds=0.0/0.0    
[...]

PR #1710, opened 2022-01-27
stream: only call rtp_clear for audio

alfredh:
Ref: https://github.com/baresip/baresip/issues/1676

the reason for this is that the first ~50 video packets are dropped during call setup,
causing the decoder to request a new keyframe and the video start to be delayed.

@cspiel1 could you please have a look why it was added ?


alfredh:
I merged it to master.

I dont think there is a need to flush packets on the UDP socket.
This should be handled in the RTP stack, by checking sequence numbers and timestamp
of incoming packets.

PR #1714, opened 2022-01-30
ice: remove ice_mode

sreimers:
depends on https://github.com/baresip/re/pull/147


alfredh:
thanks, this looks good

PR #1718, opened 2022-02-02
Attended call transfer

cHuberCoffee:


juha-h:
I could not find in the Wiki any example command flow on how attended
call transfer is done.

Commit summary has:

* menu: new command to execute a attended call transfer

but there is three new commands in the code:

{"atransferstart", 'T', CMD_PRM, "Start attended transfer", attended_xfer  },
{"atransferexec",   0,    0, "Execute attended transfer",   exec_att_xfer  },
{"atransferabort",  0,    0, "Abort attended transfer",
abort_att_xfer },

juha-h:
Sorry that I didn't have time to do any testing of this before now.

I made a test with three UAs: test@test.tutpro.com, jh@test.tutpro.com, and sua@test.tutpro.com.  First `test` calls `jh` and and `jh` answers.  Then test gives command `T sua`.  As result, `test` puts `jh` on hold and then tries to call `sua` without hostpart:
```
call: connecting to 'sua'..
#
TCP 192.168.59.95:51432 -> 192.168.59.95:5060
INVITE sua SIP/2.0
```
It should have used hostpart of the caller (test.tutpro.com) in the R-URI, but it didn't and of course the call failed.

The other weird thing was that after the above INVITE this came to console:
```
reg: sip:test@test.tutpro.com (prio 0): Register: No such file or directory
```

PR #1742, opened 2022-02-23
audio use module auconv

cspiel1:
- audio: use module auconv for optional tx format conversion
- audio: use module auconv for optional rx format conversion
- config: add auconv as default module

TODO:
The print pipeline shows that the filters are processed in different order.
```
audio tx pipeline:       pulse ---> auconv ---> auresamp ---> opus
audio rx pipeline:       pulse <--- auconv <--- auresamp <--- opus
```
We want auconv to be processed always first. Otherwise auresamp has to do extra conversion internally if rx decoder works with float. Should we change the filter processing order to be head to top for tx and rx?

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -108,6 +108,7 @@ module			g711.so
 #module			ilbc.so
 
 # Audio filter Modules (in encoding order)
+module			auconv.so
 module			auresamp.so
 #module			vumeter.so
 #module			sndfile.so
--- a/src/audio.c
+++ b/src/audio.c
@@ -104,7 +104,6 @@ struct autx {
 	int cur_key;                  /**< Currently transmitted event     */
 	enum aufmt src_fmt;           /**< Sample format for audio source  */
 	enum aufmt enc_fmt;           /**< Sample format for encoder       */
-	bool need_conv;               /**< Sample format conversion needed */
 
 	struct {
 		uint64_t aubuf_overrun;
@@ -561,7 +560,6 @@ static void poll_aubuf_tx(struct audio *a)
 	int16_t *sampv = tx->sampv;
 	size_t sampc;
 	size_t sz;
-	size_t num_bytes;
 	struct le *le;
 	uint32_t srate;
 	uint8_t ch;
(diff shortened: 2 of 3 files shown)


cspiel1:
About my tests:
- tested with opus sample format float
```
ausrc_format		float		# s16, float, ..
audec_format 		float
```

cspiel1:
@alfredh I am not sure if your emoji reaction refer to the question above.

Should we change the filter processing order to be head to top for tx and rx? I am not sure if this could break some applications.

alfredh:
the filter system was designed to be processed in the same order as the modules are loaded.
if this is tx or rx is a good question.

if there is a mistake in the printing of the pipeline, it is probably best to fix
the printing and keep the processing order as is.

cspiel1:
I double checked again:
TX: Processing is from HEAD to TAIL. Like it is configured. Printing also correct.
RX: Processing is from TAIL to HEAD. So reverse to the configuration. Printing is done from HEAD to TAIL, but with reverse arrows `<---`. Thus it prints how it is processed.

Why RX should be processed in reverse order?
Printing of reverse arrows or forward for RX is not important for me. Only that it fits the processing order.

cspiel1:
Only added a tiny improvement in the print functions. Looks now like this:
```
audio tx pipeline:  pulse_async ---> aubuf ---> auconv ---> auresamp ---> opus
audio rx pipeline:  pulse_async <--- aubuf <--- auconv <--- auresamp <--- opus
```
But above question still persists.

alfredh:
```
-->[Filt]-->[Filt]-->[AEC]-->[Filt]-->[...]--->
                       |
                       o
                       |
<--[Filt]<--[Filt]<--[AEC]<--[Filt]<--[...]<---
```

the processing order is done like it is, because some filters might have a shared state (e.g. AEC).
so the decoding order is reverse of the encoding order.

this approach has been working fine for a long time, is there any reason to change it ?

cspiel1:
The module auresamp is not able to resample float. It has to convert the samples internally to `s16le`.

If the codec works with float, then this would be a good configuration.
```
audio tx pipeline:  pulse_async ---> aubuf ---> auresamp ---> auconv ---> opus
audio rx pipeline:  pulse_async <--- aubuf <--- auresamp <--- auconv <--- opus
```
AEC is a good example that we should not change the order for the decoding path.

PR #1747, opened 2022-02-28
Allow hanging up call that has not been ACKed yet

juha-h:
Needs https://github.com/baresip/re/pull/252

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -1121,7 +1121,8 @@ void call_hangup(struct call *call, uint16_t scode, const char *reason)
 			     sip_dialog_callid(sipsess_dialog(call->sess)),
 			     call->peer_uri);
 			sipsess_abort(call->sess);
-		} else {
+		}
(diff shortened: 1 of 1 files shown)


sreimers:
I have changed the re PR title so it matches with this one and restarted the checks.

The ccheck issues remains:
```
make ccheck
./src/call.c:112: else: ending if bracket should be on previous line
./src/call.c:1133: else: ending if bracket should be on previous line
Statistics:
~~~~~~~~~~~
Number of files processed:     c: 247  cpp: 12  h: [5](https://github.com/baresip/baresip/runs/5356794501?check_suite_focus=true#step:3:5)4  mk: [9](https://github.com/baresip/baresip/runs/5356794501?check_suite_focus=true#step:3:9)5  m4: 0  py: 1  m: 1  s: 0  java: 0  php: 0 
Number of lines with errors:   2
```

Issue #1746, opened 2022-02-27
Supported header with replaces tag missing from 200 OK response


juha-h:
While testing the new attended call transfer feature, I noticed that unlike the INVITE request, 200 OK response to it does not contain `Supported` header with `replaces` tag.  Without that information, UAC does not know if it can initiate attended call transfer.

RFC 3261 [13.3.1.4](https://www.rfc-editor.org/rfc/rfc3261.html#section-13.3.1.4) The INVITE is Accepted

   A 2xx response to an INVITE SHOULD contain the Allow header field and
   the Supported header field, and MAY contain the Accept header field.
   Including these header fields allows the UAC to determine the
   features and extensions supported by the UAS for the duration of the
   call, without probing.

juha-h:
Implemented by https://github.com/baresip/baresip/pull/1752.

PR #1752, opened 2022-03-01
call: send supported header for 200 answering/ok

cHuberCoffee:
* Add supported header into 200 Answering/OK sip package: Fix for #1746
* Traced with wireshark. Now Supported header is included in 200 package.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -1259,10 +1259,16 @@ int call_answer(struct call *call, uint16_t scode, enum vidmode vmode)
 	if (err)
 		return err;
 
-	err = sipsess_answer(call->sess, scode, "Answering", desc,
-			     "Allow: %H\r\n"
-			     "%H", ua_print_allowed, call->ua,
-			     ua_print_supported, call->ua);
+	if (scode >= 200 && scode < 300) {
+		err = sipsess_answer(call->sess, scode, "Answering", desc,
+				"Allow: %H\r\n"
+				"%H", ua_print_allowed, call->ua,
+				ua_print_supported, call->ua);
+	}
+	else {
+		err = sipsess_answer(call->sess, scode, "Answering", desc,
+				"Allow: %H\r\n", ua_print_allowed, call->ua);
+	}
 
 	call->answered = true;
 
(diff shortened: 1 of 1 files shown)


juha-h:
Christoph Huber writes:

> * Add supported header into 200 Answering/OK sip package: Fix for #1746
> * Traced with wireshark. Now Supported header is included in 200
> * package.

There are several places in call.c where Allow header is added.  Should
Supported header also be added in all of them?

cHuberCoffee:
I convert this PR into a draft until we have a solution for all the places a Supported header can appear. So we have one PR for the hole topic.

The [RFC 3261 13.3.1.4](https://www.rfc-editor.org/rfc/rfc3261.html#section-13.3.1.4) says every 2xx answer of an invite should have the supported header, which is now done. Since the Supported header should only be included for 2xx responses and the `scode` is a parameter for the function `call_answer`, we should include a `scode` check for 2xx and include the Supported header only in this case.
Other places for the supported header are:
*`sipsess_refer_handler`: 202 Accepted is an answer of an Refer message. As described in [RFC 3515 2.2](https://datatracker.ietf.org/doc/html/rfc3515#section-2.2) the supported header is optional for a 2xx response.
*`call_accept & call_progress`: 180 Ringing & 183 Session Progress: [RFC 3261 20](https://datatracker.ietf.org/doc/html/rfc3261#section-20) These two packages does not have the Supported header field.

@juha-h what do you think about the 202 Accepted? Should we include the Supported header too?

juha-h:
I don't see need for Supported header in 202 Accepted to REFER, since usually the dialog already exists when REFER is sent.  If REFER is used as initial request, it does not establish a dialog.  No need for Supported in 18x responses either.

cHuberCoffee:
I added the check for 2xx scodes in `call_answer` since it may be called with other scodes.

juha-h:
> I don't see need for Supported header in 202 Accepted to REFER, since usually the dialog already exists when REFER is sent. If REFER is used as initial request, it does not establish a dialog. No need for Supported in 18x responses either.

I see it the same way for the REFER and the 18x codes

juha-h:
I tested the PR and Supported was now included also in 200 Answering.

Issue #1733, opened 2022-02-18
audio_alert does not stop alerting when the call ends


negbie:
I have two baresip (from current main branch) users registered who are part of a group, so that both get an INVITE. When a caller calls this group and ends the call, audio_alert still plays the alerting sound and the only way to stop it is to close baresip.

juha-h:
Eugen Biegler writes:

> I have two baresip users registered who are part of a group, so that
> both get an INVITE. When a caller calls this group and ends the call,
> audio_alert still plays the alerting sound and the only way to stop it
> is to close baresip.

Doesn't sound like baresip issue.  I often make calls where my SIP Proxy
parallel forks the call to two baresip UASs.  When one of them answers,
my proxy cancels the INVITE to the other baresip.

negbie:
I haven't explained it well enough. I use one baresip client for two users who are in a parallel ringing group. When the call comes in both are ringing. None of them answers the call, the caller just hangs up (both users get the CANCEL) but audio_alert still plays the alerting sound.

juha-h:
Eugen Biegler writes:

> I haven't explained it well enough. I use one baresip client for two
> users who are in a parallel ringing group. When the call comes in both
> are ringing. None of them answers the call, the caller just hangs up
> (both users get the CANCEL) but audio_alert still plays the alerting
> sound.

I just tried with two baresips registering the same account AoR.  Then I
called that AoR using a third SIP UA having different account AoR.  Both
baresips started to ring.  Then I canceled the call at the calling SIP
UA and both baresips stopped ringing.

negbie:
I have only one baresip instance with two different account AoR registered. In the config I use
call_hold_other_calls   no

Do you think the problem is that I use only one baresip instance for that?

juha-h:
Start baresip with -s option and check if it receive two CANCEL request matching the two INVITEs when the caller hangs up.

negbie:
Already checked that. Both get a CANCEL and for both I can see
closed: Connection reset by peer
audio: destroyed (started=0)

negbie:
From SIP side everything looks ok and if both users get a call sequentially it works like expected too. Only when both users get an INVITE at the same time ring.wav keeps playing.

negbie:
This is the verbose output without SIP

ua: sipsess connect via TLS 1.1.1.1:5061 --> 2.2.2.2:33916
xxx@xxx.pbx.de: selected for 1-0x5602a29af450
[...]

PR #1760, opened 2022-03-06
Stop audio_alert on UA_EVENT_CALL_CLOSED

negbie:
Fix for https://github.com/baresip/baresip/issues/1733

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -650,6 +650,7 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 			play_resume(call);
 		}
 
+		menu_stop_play();
 		hash_apply(menu.ovaufile->ht, ovaufile_del, call);
 		break;
 


negbie:
Keep in mind that I did not test this for redial or xfer. Mby for these cases menu_stop_play() shouldn't be called?

negbie:
What I mean is that mby it's better to use this?

		if (!(menu.redial_attempts) && !(menu.xfer_call == call || menu.xfer_targ == call)) {
			menu_stop_play();
		}

alfredh:
thanks for the PR!

does the fix solve your problem ?

cspiel1:
Everything is there already. What exactly does not work as you expect?

```c
		if (!str_cmp(call_id(call), menu.callid)) {
			if (count==1)
				menu_play_closed(call);

			menu_selcall(NULL);
			play_resume(call);
		}
```

negbie:
> thanks for the PR!
> 
> does the fix solve your problem ?

Yes it does but I would like to get feedback from someone who is more familiar with this codebase if this change has no side effects.

> Everything is there already. What exactly does not work as you expect?
I've explained it here https://github.com/baresip/baresip/issues/1733 in more detail but in short:
if one baresip instance has two registered users who are part of a parallel ringing group and none of them accepts an incoming call, ring.wav will be played even if the caller ends the call. You have to close baresip to make it stop.

negbie:
@cspiel1 yes menu_play_closed will call menu_stop_play() too but for this particular case you won't get into that condition.

cspiel1:
What the mentioned code segment should do is:
- `menu.callid` is the call ID of the "current call". This is the call for which the ring.wav is played.
- If the call that terminates is the current call, then the ring playback should be stopped.
- Then another call is selected as current call. In your case the other incoming call. The ring-tone is started for the second call.
- If the second call also terminates, you should get another CALL_CLOSED event and again the callid should match and the ringtone is stopped.

For short: You should always here the signal tone for the current call. This could also be ringback.wav for an outgoing call.

The setup is not so easy. I will check the code first.

cspiel1:
> @cspiel1 yes menu_play_closed will call menu_stop_play() too but for this particular case you won't get into that condition.

I guess the call_id does not match. Right? If so, why?

cspiel1:
Tried to reproduce with two accounts on two different SIP servers. With no luck.

How are your accounts `aor` looking?

cspiel1:
Now tried with two accounts on one SIP registrar. Everything works correctly.
```
/listcalls

User-Agent: alice@jackson
--- Active calls (0) ---


User-Agent: 101@192.168.110.56
--- Active calls (1) ---
> [line 1, id def36c91-55f1-4a91-99d9-1718b3452ff5]  0:00:00   INCOMING             sip:102@192.168.110.56


User-Agent: 103@192.168.110.56
--- Active calls (1) ---
> [line 1, id a925563b-ec7c-4e0c-8c67-fbb189c992f8]  0:00:00   INCOMING             sip:102@192.168.110.56

/hangupall
call: terminate call 'f3875ce47e20b4c3' with sip:101@192.168.110.56:16060;transport=tcp
call: terminate call 'd7dee295bf276dfc' with sip:103@192.168.110.56:16060;transport=tcp
```
The ring tone stops.

What exactly is meant with ring-group? I guess two INVITEs nearly simultaneously.

negbie:
@cspiel1 thank you very much for looking into this.
AOR's are:

<sip:11@192.168.1.22;transport=tls>;auth_pass=XXX;mediaenc=srtp-mand;answermode=manual
<sip:22@192.168.1.22;transport=tls>;auth_pass=XXX;mediaenc=srtp-mand;answermode=manual

Both are in a parallel ringing group so both get an INVITE at the simultaneously. I've added some additional debug log lines:

Right after CANCEL, 487 for fist AOR

menu: call_id(call): c2855d48dd668942-Acc3338-B2b2386
menu: menu.callid: c2855d48dd668942-Acc3338-B2b2386
menu: count: 2 

Right after CANCEL, 487 for second AOR

menu: call_id(call): c2855d48dd668942-Acc3338-B2b2386
menu: menu.callid: 
menu: count: 1 

So as you can see when the first call is closed the condition (!str_cmp(call_id(call), menu.callid)) is true but
count is 2 so the condition (count==1) is false and menu_play_closed(call) won't be called.

For the second call menu.callid is empty so (!str_cmp(call_id(call), menu.callid)) is false and menu_play_closed(call) is not called too.

negbie:
So do you thing it would be ok to change
```

		if (!str_cmp(call_id(call), menu.callid)) {
			if (count==1)
				menu_play_closed(call);

			menu_selcall(NULL);
			play_resume(call);
		}

```
to
```

		if (!str_cmp(call_id(call), menu.callid)) {
			if (count>=1)
				menu_play_closed(call);

			menu_selcall(NULL);
			play_resume(call);
		}
```

cspiel1:
`count>=1` is not okay. Because it is the opposite what is wanted:
- The error-tone should only be played if the last call terminates.

The error is that the `menu.callid` is empty after the first call terminates. 

Anyway, it would be okay if we add an `else` path.
```c
			if (count==1)
				menu_play_closed(call);
			else
				menu_stop_play();
```

But the problem still persists that `menu_selcall(NULL);` does not select the second incoming call `c2` and then the `play_resume(call)` will not start again for `c2`. What would be expected if the second CANCEL does not come immediately.

negbie:
Ah ok makes sense. For me the else condition would be enough but I can understand if you want to resolve the root cause. To help with that it would take me more time to get involved.

cspiel1:
Could you please add some debug logging to `menu_selcall()` and `filter_call()`?

The function `filter_call()` should be called two times after the first call terminates.

cspiel1:
I'll further trying to reproduce this. I guess that not the "simultaneous" INVITEs are the problem. More that the order of call terminations is relevant.

negbie:
menu filter_call: call_state(call) != fa->state
menu filter_call: call_state(call) != fa->state
menu filter_call: call_state(call) != fa->state
menu filter_call: call_state(call) != fa->state
menu filter_call: call_state(call) != fa->state
menu filter_call: call_state(call) != fa->state
menu filter_call: call_state(call) != fa->state
menu filter_call: call_state(call) != fa->state
menu filter_call: !str_cmp(call_id(call), fa->callid_old)
menu filter_call: !str_cmp(call_id(call), fa->callid_old)

I added some debug logs in filter_call to see how on which conditions it returns. It never returns true.

juha-h:
I was able to reproduce the original issue.  I can try again with your proposed PR when it is ready.

cspiel1:
Thanks! I got the setup from @negbie and can reproduce this now. The SIP flow is like this:
- INVITE a <--
- 180 Ringing -->
- INVITE b <--
- 180 Ringing -->
- ...
- CANCEL <--
- CANCEL <--

The CANCEL are arriving very fast (with same timestamp in wireshark). Most likely they are processed in one re-main loop iteration.

So, most likely I am able to fix this now and open a PR soon.

cspiel1:
The problem is that the call-id of both calls is identical. This is a situation we didn't consider when we chose to track the current call by its call-id: https://github.com/baresip/baresip/pull/1207 

I have to think about this.

cspiel1:
@alfredh Found your comment. https://github.com/baresip/baresip/pull/1199#issuecomment-772390766 

Wouldn't it possible to replace `menu.callid` by a pointer `menu.current_call`?
The pointer can be set to NULL (not deref) when the last call terminates.

juha-h:
My wireshark pacpng is here:

https://tutpro.com/tmp/baresip.pcapng

Also in it CANCELs arrive at the same ms.

cspiel1:
> My wireshark pacpng is here:
> 
> https://tutpro.com/tmp/baresip.pcapng
> 
> Also in it CANCELs arrive at the same ms.

Thanks! You have the same situation. The call-ids are identical for both calls `532531d8a31bbeae`.

juha-h:
Christian Spielberger writes:

> Thanks! You have the same situation. The call-ids are identical for
> both calls `532531d8a31bbeae`.

That is because my SIP proxy parallel forks the one call.

cspiel1:
I'll open a PR where we can proceed the discussion about tracking the "current call".

juha-h:
Christian Spielberger writes:
 
 > Thanks! You have the same situation. The call-ids are identical for
 > both calls `532531d8a31bbeae`.

But Via branch values differ as they should.

cspiel1:
Here is the PR: https://github.com/baresip/baresip/pull/1763

juha-h:
Christian Spielberger writes:

> Here is the PR: https://github.com/baresip/baresip/pull/1763

Worked fine in my test.  In my Android app, I have always identified
calls by pointers.

PR #1761, opened 2022-03-07
Added new API function call_supported() and used it in menu module

juha-h:
Added new API function `call_supported()` and used it in menu module `atransferstart` command


juha-h:
Any comments on this PR?  The new API command is not changing current behavior and thus should be OK to merge.

PR #1763, opened 2022-03-08
menu select other call on hangup

cspiel1:
- menu: fix current call selection on `CALL_STATE_CLOSED` (#1733)
- menu: add call pointer match to `filter_arg` (#1733)
- menu: replace current call-id by pointer (#1733)

In some setups it is possible to receive multiple incoming calls with equal
call-id. This PR solves tracking of the current call if there are multiple
calls with identical call-id.

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -28,10 +28,12 @@ enum {
 
 struct filter_arg {
 	enum call_state state;
-	const char *callid_old;
+	const struct call *exclude;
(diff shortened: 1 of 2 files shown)


cspiel1:
Simpler use case for reproducing the issue:
Call to local UA (loop back call).
--> Result: There is an outgoing and an incoming call with same call-id.
On hangup of both calls the ringback tone is not stopped.

This PR fixes this also.

PR #1769, opened 2022-03-12
audio: fix rx_thread (adaptive jitter buffer)

### LINKED ISSUE #1768: Dropped a frame to reduce latency
I just created new version of my Android app using latest baresip/re/rem libs and noticed that audio quality is very bad.  I get lots of these to log during call:
```
Dropped a frame to reduce latency
```
I don't believe that there is any latency, since both UAC and UAS run on my laptop.  Audio works fine with earlier version of my Android app from Feb 24.

When I look the log messages, in Feb 24 version I get this when call starts:
```
03-12 19:05:24.784 18904 18940 D Baresip+ Lib: audio: source started with sample format S16LE
03-12 19:05:24.784 18904 18940 D Baresip+ Lib: audio tx pipeline:    opensles ---> webrtc_aecm ---> opus
03-12 19:05:24.784 18904 18940 D Baresip+ Lib: audio rx pipeline:    opensles <--- webrtc_aecm <--- opus
```
With today's version I get:
```
03-12 19:16:35.474 19151 19549 D Baresip Lib: audio tx pipeline:    opensles ---> aubuf ---> webrtc_aecm ---> opus
03-12 19:16:35.474 19151 19549 D Baresip Lib: audio rx pipeline:    opensles <--- aubuf <--- webrtc_aecm <--- opus
```
That is `aubuf` is now on the pipeline and might explain why my app has become useless.

sreimers:
tmr_jiffies should only be used to compare/measure time in ms. But pthread_cond_timedwait relies on a REALTIME value.

fixes #1768

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -735,14 +735,17 @@ static void *rx_thread(void *arg)
 {
 	struct audio *a = arg;
 	struct timespec ts;
-	uint64_t ms;
+	static const uint16_t ms = 500;
 	int err = 0;
 
 	while (a->rx.thr.run) {
-
-		ms = tmr_jiffies() + 500;
-		ts.tv_sec = (time_t)(ms / 1000);
-		ts.tv_nsec = (ms % 1000) * 1000000UL;
+		clock_gettime(CLOCK_REALTIME, &ts);
+		ts.tv_sec += ms / 1000;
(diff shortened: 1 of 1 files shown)


juha-h:
Based on my tests, this PR has fixed  the `Dropped a frame to reduce latency #1768` issue.

PR #1773, opened 2022-03-17
test: refactoring of test_account_uri_complete

alfredh:
rewrite the test so that each array entry is a pair of input string and expected output string.
reduces complexity, not sure if the explicit compare function was needed ?

@cspiel1 please review

### DIFF
--- a/test/account.c
+++ b/test/account.c
@@ -77,110 +77,77 @@ int test_account(void)
 }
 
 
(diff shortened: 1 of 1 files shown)


cspiel1:
Looks good!
Since `retest` tests `sip_addr_decode()` sufficient, the compare function is not needed.

PR #1780, opened 2022-03-26
Added API function call_diverteruri

juha-h:
Added API function `call_diverteruri` taken from History-Info or Diversion header.  Needed in order to be able to show to user who (if any) has diverted the incoming call.  Even though Diversion header has officially been replaced by History-Info, Diversion is still widely used.  Both have worked fine in my tests.


alfredh:
thanks, this patch look fine.

PR #1784, opened 2022-03-30
aubuf adaptive jitter buffer

cspiel1:
Relates to https://github.com/baresip/rem/pull/33
Discussion: https://github.com/baresip/re/discussions/184

- audio: adapt to adaptive aubuf, adaptive jbuf becomes obsolete
- audio: do not drop reordered packets
- config,stream: support for new jbuf type "minimize"
- stream: also for JBUF_ADAPTIVE decode immediately 

This makes JBUF_ADAPTIVE completely obsolete what is the final goal
of the current PR. This is necessary because otherwise no decoding
would be done if `jitter_buffer_type   adaptive` is configured.

An alternative would be to keep all the JBUF_ADAPTIVE functions in
stream.c and audio.c for a deprecation period.

@sreimers, @alfredh, @juha-h 
I suppose that nobody uses JBUF_ADAPTIVE except my company, thus I would prefer to directly drop all the obsolete code in baresip and also in `re/src/jbuf`. Let me know what you would prefer!
- A jbuf adaptive deprecation period, or
- immediate cleanup? This would/should not change the JBUF_FIXED behavior.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -50,6 +50,7 @@ auplay_format		s16		# s16, float, ..
 auenc_format		s16		# s16, float, ..
 audec_format		s16		# s16, float, ..
 audio_buffer		20-160		# ms
+audio_buffer_mode	fixed		# fixed, adaptive
 audio_telev_pt		101		# payload type for telephone-event
 
 # Video
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -51,6 +51,7 @@ auenc_format		s16		# s16, float, ..
 audec_format		s16		# s16, float, ..
 audio_buffer		20-160		# ms
 audio_buffer_mode	fixed		# fixed, adaptive
+audio_silence		-35.0		# in [dB]
 audio_telev_pt		101		# payload type for telephone-event
 
 # Video
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -277,6 +277,7 @@ int  conf_get_vidsz(const struct conf *conf, const char *name,
 		    struct vidsz *sz);
 int  conf_get_sa(const struct conf *conf, const char *name, struct sa *sa);
 enum jbuf_type conf_get_jbuf_type(const struct pl *pl);
+bool conf_aubuf_adaptive(const struct pl *pl);
 void conf_close(void);
 struct conf *conf_cur(void);
 int conf_loadfile(struct mbuf **mbp, const char *filename);
@@ -338,6 +339,7 @@ struct config_audio {
 	int enc_fmt;            /**< Audio encoder sample format    */
 	int dec_fmt;            /**< Audio decoder sample format    */
 	struct range buffer;    /**< Audio receive buffer in [ms]   */
+	bool adaptive;          /**< Enable adaptive audio buffer   */
 	uint32_t telev_pt;      /**< Payload type for tel.-event    */
 };
 
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -363,7 +363,6 @@ struct config_avt {
 	bool rtcp_mux;          /**< RTP/RTCP multiplexing          */
 	enum jbuf_type jbtype;  /**< Jitter buffer type             */
 	struct range jbuf_del;  /**< Delay, number of frames        */
-	uint32_t jbuf_wish;     /**< Startup wish delay of frames   */
 	bool rtp_stats;         /**< Enable RTP statistics          */
 	uint32_t rtp_timeout;   /**< RTP Timeout in seconds (0=off) */
 	bool bundle;            /**< Media Multiplexing (BUNDLE)    */
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -340,6 +340,7 @@ struct config_audio {
 	int dec_fmt;            /**< Audio decoder sample format    */
 	struct range buffer;    /**< Audio receive buffer in [ms]   */
 	bool adaptive;          /**< Enable adaptive audio buffer   */
+	double silence;         /**< Silence volume in [dB]         */
 	uint32_t telev_pt;      /**< Payload type for tel.-event    */
 };
 
--- a/modules/multicast/player.c
+++ b/modules/multicast/player.c
@@ -537,7 +537,8 @@ int mcplayer_start(struct jbuf *jbuf, const struct aucodec *ac)
 			goto out;
 		}
 
-		aubuf_set_mode(player->aubuf, AUBUF_ADAPTIVE);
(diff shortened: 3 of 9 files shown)


sreimers:
Maybe we can cleanup and `jitter_buffer_type adaptive` is a deprecated fallback to JBUF_MINIMIZE?

cspiel1:
> Maybe we can cleanup and `jitter_buffer_type adaptive` is a deprecated fallback to JBUF_MINIMIZE? Or we keep the `adaptive` term instead of `minimize`

I would prefer to drop all the jbuf adaptive code and keep the term `adaptive` instead of `minimize`. Should I do the cleanup right now?

cspiel1:
Another question is:
Should we activate `AUBUF_ADAPTIVE` via baresip config to minimize the risk for the PR?
If everything works, `AUBUF_ADAPTIVE` should be used by default for optimal audio quality without having to adjust the buffers in the config on the network quality.

Currently, it is activated for `rx` and not for `tx`.

The `AUBUF_ADAPTIVE` in theory works also for `tx` in order to handle SW jitter of the audio source. But I still have to test this.

sreimers:
> I would prefer to drop all the jbuf adaptive code and keep the term `adaptive` instead of `minimize`. Should I do the cleanup right now?

I would say yes, introducing a new term/config option is more confusing I think. And its easier to maintain/improve one adaptive jitter buffer solution.

> The `AUBUF_ADAPTIVE` in theory works also for `tx` in order to handle SW jitter of the audio source. But I still have to test this.

I think `rx` only is fine. the remote site has to deal with jitter anyway and for `tx` we have a thread mode for decoupling from the audio thread. Something we can consider for `rx` after merging this PR.

> Should we activate `AUBUF_ADAPTIVE` via baresip config to minimize the risk for the PR?
> If everything works, `AUBUF_ADAPTIVE` should be used by default for optimal audio quality without having to adjust the buffers in the config on the network quality.

Keeping `jitter_buffer_delay fixed` as default and migrate after some time sounds good.

cspiel1:
Thus all combinations are possible and should work:
```
enum jbuf_type {
	JBUF_OFF,                   /* No jitter buffer --> best option for G.711 with AUBUF_ADAPTIVE */
	JBUF_FIXED,               /* The old behavior for packet re-ordering. */
	JBUF_ADAPTIVE         /* Packet re-ordering with minimum latency. --> Perfect for stateful codecs. */
};
```
```
enum aubuf_mode {
	AUBUF_FIXED,              /* The old behavior to simply decouple the threads. */
	AUBUF_ADAPTIVE        /* Compute jitter and optimize latency. */
};
```

`JBUF_FIXED + AUBUF_FIXED` will be default for some period.

In order to simplify the configuration for the user, finally these settings should be default:
- `AUBUF_ADAPTIVE`
- `JBUF_OFF` for state-less codecs and `JBUF_ADAPTIVE` for others. Maybe we find a solution to select this also autmatically.

cspiel1:
Currently struggling around with the config for `enum aubuf_mode`, which is defined in `rem`.

If we add to `struct config_audio {`
```
enum aubuf_mode bufmod; /**< Audio buffer mode              */
```

Then `baresip.h` has it's first `rem.h` dependency. We need this setting in `audio.c` and in `modules/multicast/receiver.c` for now. And gcc tells me that "ISO C forbids forward references to ‘enum’ types".
 Should we
- read config setting `audio_buffer_mode` directly and compare the strings,
- instead of the `enum` use an `int` in `struct config_audio`,
- add the `#include <rem.h>` in all the c-files before `#include <baresip.h>`,
- split `baresip.h` into `baresip_config.h` and `baresip.h`

cspiel1:
What is this? This has nothing to do with this PR.
```
src/udp/udp.c: In function ‘udp_flush’:
111
src/udp/udp.c:1004:27: error: pointer targets in passing argument 2 of ‘recvfrom’ differ in signedness [-Werror=pointer-sign]
112
 1004 |   while (recvfrom(us->fd, buf, sizeof(buf),
113
      |                           ^~~
114
      |                           |
115
      |                           uint8_t * {aka unsigned char *}
```

alfredh:
this looks good.

please feel free to merge to main.

cspiel1:
We have to complete first the re/rem PRs.

cspiel1:
There is still missing the config for the aubuf mode. Should I implement one? Or is it okay that we switch to ABUF_ADAPTIVE for the rx stream?

Note: I had problems with the header file. https://github.com/baresip/baresip/pull/1784#issuecomment-1084420339

sreimers:
A difficult decision, I am not sure, but would perhaps prefer this variant:
> * instead of the `enum` use an `int` in `struct config_audio`,

cspiel1:
> A difficult decision, I am not sure, but would perhaps prefer this variant:
> 
> > * instead of the `enum` use an `int` in `struct config_audio`,

Thanks! For now a bool is enough.

cspiel1:
I want to update a commit message "aubuf mode config draft". Then it's ready.

cspiel1:
Now it should be okay for the merge. Thanks for reviewing!

juha-h:
Currently I have in my Android app
```
jitter_buffer_type adaptive
jitter_buffer_delay 5-10
jitter_buffer_wish 6
```
After this commit, is that still valid or should it be changes somehow?  Is there a migration guide somewhere?

juha-h:
juha-h writes:

> Currently I have in my Android app
> ```
> jitter_buffer_type adaptive
> jitter_buffer_delay 5-10
> jitter_buffer_wish 6
> ```
> After this commit, is that still valid or should it be changed
> somehow?  Is there a migration guide somewhere?

After checking core_config_template() function, the above still looks
valid.  Is there some relationship between jitter_buffer_type and the
new audio_buffer_mode?  Is it fine if audio_buffer_mode is "fixed" while 
jitter_buffer_type is "adaptive"?

cspiel1:
`jitter_buffer_wish` is obsolete

I updated the wiki.

Your settings could be transferred to:
```
jitter_buffer_type adaptive
jitter_buffer_delay 5-10
audio_buffer_mode adaptive
audio_buffer 40-200
```

juha-h:
Christian Spielberger writes:

> `jitter_buffer_wish` is obsolete
> 
> I updated the wiki.
> 
> Your settings could be transferred to:
> ```
> jitter_buffer_type adaptive
> jitter_buffer_delay 5-10
> audio_buffer_mode adaptive
> audio_buffer 40-200
> ```

Thank you very much.

PR #1797, opened 2022-04-07
Multicast jbuf and aubuf changes

cHuberCoffee:


alfredh:
here are some comments:

1. you may consider to move this module to baresip-apps. It is fine to keep it here also.

2. this module has a copy of audio.c , we should try to make the audio.c api generic enough
so that it can also be used here.

alfredh:
here are some comments:

1. you may consider to move this module to baresip-apps. It is fine to keep it here also.

2. this module has a copy of audio.c , we should try to make the audio.c api generic enough
so that it can also be used here.

cHuberCoffee:
Im still working on a few points which are already in progress. I'm open to move the module to baresip-apps but not immediately.

Yes you are right, a good solution would be to dedup a lot of code from the multicast/player.c and multicast/source.c
I will have an eye on this topic and try a bit around on whats possible.

PR #1800, opened 2022-04-08
cmake: preliminary support

alfredh:


alfredh:
this version supports only baresip core, we also need to add the modules.

does anyone have a good idea of how to handle modules in CMake ?
anyone with CMake experience here, please feel free to help out :)

sreimers:
Maybe we can re-use some parts of this old PR: https://github.com/baresip/baresip/pull/1039

PR #1801, opened 2022-04-08
stream: remove mbuf_get_left check in rtp_handler

GGGO:


alfredh:
thanks @GGGO -- lets try that now

if anyone complains, please be prepared to revert the change or fix it in a different way.

NOTE: the error in the selftest is a bug in selftest

PR #1817, opened 2022-04-25
Updated debian version

juha-h:
I noticed that debian changelog version was still 1.0.0, but that latest tag was 2.0.2.  So I went and updated debian changelog accordingly.

In the future, should this be done each time a new tag is created?

How come macos tests take MUCH longer time than all other tests altogether?

Looks like macos check was not able to download something.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+baresip (2.0.2) unstable; urgency=medium
+
+  * version 2.0.2
+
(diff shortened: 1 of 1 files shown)


alfredh:
yes, this should also be updated during a release.

we are planning release 2.3.0 around 1st may

juha-h:
OK, since I use Debian, I can try to keep watch when new tags are created.

PR #1822, opened 2022-04-28
audio: flush aubuf if ssrc changes

cspiel1:


src/audio.c
```diff
@@ -146,6 +146,7 @@ struct aurx {
 	const struct aucodec *ac;     /**< Current audio decoder           */
 	struct audec_state *dec;      /**< Audio decoder state (optional)  */
 	struct aubuf *aubuf;          /**< Incoming audio buffer           */
+	uint32_t ssrc;                /**< Incoming syncronizing source     */
```
sreimers, line 149: syncronizing -> synchronization

PR #1832, opened 2022-05-05
multicast: support error code EAGAIN of jbuf_get()

cspiel1:
Reducing jitter buffer for low latency did not work.


modules/multicast/receiver.c
```diff
@@ -411,16 +411,19 @@ static int player_decode(struct mcreceiver *mcreceiver)
 {
 	void *mb = NULL;
 	struct rtp_header hdr;
+	int err1;
```
sreimers, line 414: maybe something like `res` or `jerr` is more meaningful.

PR #1839, opened 2022-05-10
h264: move from rem to re

sreimers:


cspiel1:
```
  CC [M]  build-x86_64/modules/avcodec/decode.o
In file included from ../rem/include/rem_video.h:12,
                 from ../rem/include/rem.h:16,
                 from modules/avcodec/decode.c:9:
../rem/include/rem_h264.h:9:6: error: redeclaration of ‘enum h264_nalu’
    9 | enum h264_nalu {
      |      ^~~~~~~~~
```

sreimers:
Can`t reproduce. Can check you if rem headers are up to date?

cspiel1:
Yes, everything is up to date. I checked again by rm /usr/local/include/re and rem. Then make && make install in re and rem.
 But now the gcc error is "redeclaration".
 
 And I have:
 ```
$ find -name "*_h264.h"
./re/re_h264.h
./rem/rem_h264.h
```

sreimers:
rem_h264.h should not exist anymore:

https://github.com/baresip/rem/tree/main/include

And is not referenced: https://github.com/baresip/rem/blob/main/include/rem_video.h

cspiel1:
Sorry, the remote was wrong for my local `main`!

Issue #852, opened 2019-11-15
audio: Marker bit handling


premultiply:
I think patch https://github.com/alfredh/baresip/pull/724 makes decoding of RFC 3640 AAC mpeg4-generic streams impossible as they have set the marker bit on each RTP packet.
https://tools.ietf.org/html/rfc3640#section-3.1

Maybe the current RX (and TX) marker bit handling has to be completly redesigned.
It would be better if the audio encoding and decoding functions have direct access to the tx/rx marker bits.

premultiply:
After a little more investigation I think there is a general problem with the current marker bit handling when they are set on all packets as requiered for RFC 3640.
But wasn't able to get it exactly why no data is passed continuously to the decoder in this case. It looks to me like just the first RX packet is passed through and all further packets are discarded.

alfredh:
this looks like video decoding, where the marker bit indicates the
last packet of a complete frame (fragmented).

so far we did not need to use the marker bit for audio codecs.

do you have a copy of a Wireshark trace from an endpoint that supports this ?

premultiply:
Yes. I sent you an E-mail with a Wireshark trace.

alfredh:
one solution is to extend the aucodec API with an optional marker bit flag:

old:

```c
typedef int (auenc_encode_h)(struct auenc_state *aes,
			     uint8_t *buf, size_t *len,
			     int fmt, const void *sampv, size_t sampc);
```

new:

```c
typedef int (auenc_encode_h)(struct auenc_state *aes,
			     bool *marker, uint8_t *buf, size_t *len,
			     int fmt, const void *sampv, size_t sampc);
```


the question is if there are other fields of the RTP header that also
should be passed to/from aucodec ?

alfredh:
at the moment I would like to suggest the following solution:

1. add marker bit to encode API (`bool *marker`)
2. add RTP header to decode API (`struct rtp_header`)

premultiply:
Sounds good to me.

alfredh:
patch related to aac mpeg4-generic work:

https://gist.github.com/alfredh/90dd76b912ef86a38dee61e7252c5e71

I did some testing with Linphone for iOS which supports AAC-ELD (etc)
using mpeg4-generic

alfredh:
here is a patch for review/testing:

https://gist.github.com/alfredh/4a81beef98479302b504b201dc2ae347

after some more thinking I think its enough to add the marker bit
to the encode/decode functions.

[...]

PR #1841, opened 2022-05-12
stream,audio: flush RX buffers if marker bit is set

cspiel1:
We had audio artefacts in playback of the RX stream after periods of silence, in which no RTP packets are transmitted by the sender.

I am pretty sure that the RX buffers have to be flushed, if the RTP marker bit is set. Due to RFC-3551 and to an already removed code snippet of @alfredh the marker bit indicates the beginning of a new talkspurt. 

When searching the git log I found this commit:
```
commit fdb22bc3725483a9500323df85627e223435b38e
Author: Alfred E. Heggestad <alfred.heggestad@gmail.com>
Date:   Sun Dec 1 11:21:44 2019 +0100

    aucodec: add marker bit -- ref #852

diff --git a/src/stream.c b/src/stream.c
index 9e434feec..6e6a75904 100644
--- a/src/stream.c
+++ b/src/stream.c
@@ -260,9 +260,11 @@ static void rtp_handler(const struct sa *src, const struct rtp_header *hdr,
        if (!(sdp_media_ldir(s->sdp) & SDP_RECVONLY))
                return;
 
+#if 0
        /* The marker bit indicates the beginning of a talkspurt. */
        if (hdr->m && s->type == MEDIA_AUDIO)
                flush = true;
+#endif
```
And then this issue with this comment:
https://github.com/baresip/baresip/issues/852#issuecomment-560089704

@alfredh At least for G.722 this PR seems to be the right solution. How should we proceed here?

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -844,6 +844,9 @@ static int aurx_stream_decode(struct aurx *rx, const struct rtp_header *hdr,
 
 		rx->ssrc = hdr->ssrc;
 	}
+	else if (hdr->m) {
+		aubuf_flush(rx->aubuf);
+	}
 
 	err = aubuf_write_auframe(rx->aubuf, &af);
 	if (err)
--- a/src/stream.c
+++ b/src/stream.c
@@ -388,6 +388,10 @@ static void rtp_handler(const struct sa *src, const struct rtp_header *hdr,
 		s->rx.pseq = hdr->seq - 1;
 		flush = true;
 	}
+	else if (hdr->m && s->type == MEDIA_AUDIO) {
+		/* The marker bit indicates the beginning of a talkspurt. */
+		flush = true;
+	}
 
 	/* payload-type changed? */
 	err = s->pth(hdr->pt, mb, s->arg);


sreimers:
Another solution could be a separate RX poll thread (like we did for TX) this way the decoding is independent from the main thread workload and jbuf automatically runs empty if no packets are received (no artefacts). If necessary we can adapt/reset the jitter buffer calculation, aubuf itself should read already silence if empty. I can prepare a PR in the next days (since I need this also for better scaling and less software jitter).

cspiel1:
`jbuf`: Does not run below `jb->wish` in order to be able to reorder dis-ordered RTP packets.
`aubuf`: Yes, it runs empty, but:
- `ab->ts` has to be set to zero, and
-  in adaptive mode the `ajb` struct has to be reset if the marker bit was set. See function `ajb_reset()`

An extra thread should be configurable, in order to make low latency solution possible.

Currently I am working on a configurable aubuf directly after decoding:

```
       If an AEC is in the filter list aufilt, then the aubuf should be
       constant and small. A second audio buffer aubufdec should be added via
       config if network jitter is expected.

       .--------.   .-------.   .--------.   .----------.   .--------.
 |\    |        |   |       |   |        |   |  opt.    |   |        |
 | |<--| auplay |<--| aubuf |<--| aufilt |<--| aubufdec |<--| decode |<--- RTP
 |/    |        |   |       |   |        |   |          |   |        |
       '--------'   '-------'   '--------'   '----------'   '--------'
```
The `aubufdec` makes most sense, if it runs in adaptive mode to handle changing network jitter best. If running in adaptive mode, an extra thread might not be required.

sreimers:
As mentioned in https://github.com/baresip/baresip/issues/852 the marker bit is only a hint for a talkspurt and can be reused for other definitions. So I think its not very reliable and flushing buffers with active streams should be avoided in other cases.

Can we reuse/extend the RTP timeout handling to detect this case (no RTP packets are transmitted for silence) and only flush if timeout is detected (maybe together with the marker bit)?

https://github.com/baresip/baresip/blob/main/src/stream.c#L239-L245

```
The marker bit is a hint; the beginning of a talkspurt can also be computed by comparing the difference in timestamps and sequence numbers between two packets, assuming the timestamp clock rate is known.
```
https://www.cs.columbia.edu/~hgs/rtp/faq.html#marker

cspiel1:
Yes, maybe a combination of marker bit and jump in timestamp.

alfredh:
it has been discussed before, the marker bit is just a "hint".

you can rely on it if you want, but please keep in mind that other SIP clients may not set it.
so then the logic will not work with them.

if there is a long pause in the stream with no packets received, you might
have when the stream resumes:

- A long jump in sequence number
- A long jump in timestamp
- A change in SSRC

if the codec is the same as before the pause, there should be no need to flush buffers.

If you are going from e.g. PCMU to G722, then the sample rate will change, and flushing
of audio buffers are needed.

that was my 2 cents, feel free to take it with a pinch of salt :)

src/audio.c
```diff
@@ -844,6 +844,9 @@ static int aurx_stream_decode(struct aurx *rx, const struct rtp_header *hdr,
 
 		rx->ssrc = hdr->ssrc;
 	}
+	else if (hdr->m) {
+		aubuf_flush(rx->aubuf);
+	}
```
cspiel1, line 849: This is not needed. The aubuf runs empty --> "underrun", then the `ajb` object can be reset. A `rem` PR will follow.

cspiel1: - In case of an underrun there is a fix in `rem`: https://github.com/baresip/rem/pull/51
- In case of SSRC change, there is already a `aubuf_flush`
- In case of timestamp jump: Needs a fix here in `audio.c`, maybe in combination with marker bit.

src/stream.c
```diff
@@ -388,6 +388,10 @@ static void rtp_handler(const struct sa *src, const struct rtp_header *hdr,
 		s->rx.pseq = hdr->seq - 1;
 		flush = true;
 	}
+	else if (hdr->m && s->type == MEDIA_AUDIO) {
+		/* The marker bit indicates the beginning of a talkspurt. */
+		flush = true;
+	}
```
cspiel1, line 394: For the jbuf:
- in case of SSRC change, there is already a `jbuf_flush()`.
- in case of timestamp/seq number jump --> needs a fix in `jbuf.c`

cspiel1: jbuf will not run empty because `jbuf_put()` and `jbuf_get()` is called in same thread. The timestamp/seq number jump can not be defined clearly.
Thus added a timeout check in this `re` PR: https://github.com/baresip/re/pull/370

cspiel1:
> Another solution could be a separate RX poll thread (like we did for TX) this way the decoding is independent from the main thread workload and jbuf automatically runs empty if no packets are received (no artefacts). If necessary we can adapt/reset the jitter buffer calculation, aubuf itself should read already silence if empty. I can prepare a PR in the next days (since I need this also for better scaling and less software jitter).

Hi Sebastian, do you have already a branch? Today I realized that we thread the RX poll thread for our old platform. Maybe we should add a "real-time" epoll loop to `re`.

PR #1842, opened 2022-05-12
audio: add optional decoding buffer

cspiel1:
This is only for discussion so far.
```
       If an AEC is in the filter list aufilt, then the aubuf should be
       constant and small. A second audio buffer aubufdec should be added via
       config if network jitter is expected.

       .--------.   .-------.   .--------.   .----------.   .--------.
 |\    |        |   |       |   |        |   |  opt.    |   |        |
 | |<--| auplay |<--| aubuf |<--| aufilt |<--| aubufdec |<--| decode |<--- RTP
 |/    |        |   |       |   |        |   |          |   |        |
       '--------'   '-------'   '--------'   '----------'   '--------'

```

The aubufdec makes mostly sense, if it runs in adaptive mode to handle changing network jitter best. If running in adaptive mode, an extra thread might not be required.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -1449,7 +1449,7 @@ static int rx_filter_thread(void *arg)
 
 		if (af.sampc && sampc != af.sampc) {
 			sampc = af.sampc;
-			ptime = sampc * 1000 / (ac->srate * ac->ch);
+			ptime = (uint32_t) (sampc*1000 / (ac->srate * ac->ch));
 		}
 
 		(void)rx_push_aubuf(rx, &af);
--- a/src/audio.c
+++ b/src/audio.c
@@ -1320,27 +1320,30 @@ static int autx_print_pipeline(struct re_printf *pf, const struct autx *autx)
 }
 
 
-static int aurx_print_pipeline(struct re_printf *pf, const struct aurx *aurx)
+static int aurx_print_pipeline(struct re_printf *pf, const struct aurx *rx)
 {
 	struct le *le;
 	int err;
 
-	if (!aurx)
+	if (!rx)
 		return 0;
 
 	err = re_hprintf(pf, "audio rx pipeline:  %10s",
-			 aurx->ap ? aurx->ap->name : "(play)");
+			 rx->ap ? rx->ap->name : "(play)");
 
 	err |= re_hprintf(pf, " <--- aubuf");
-	for (le = list_head(&aurx->filtl); le; le = le->next) {
+	for (le = list_head(&rx->filtl); le; le = le->next) {
 		struct aufilt_dec_st *st = le->data;
 
 		if (st->af->dech)
 			err |= re_hprintf(pf, " <--- %s", st->af->name);
 	}
 
+	if (rx->aubufdec)
+		err |= re_hprintf(pf, " <--- aubufdec");
+
 	err |= re_hprintf(pf, " <--- %s\n",
-			  aurx->ac ? aurx->ac->name : "(decoder)");
+			  rx->ac ? rx->ac->name : "(decoder)");
 
 	return err;
 }
--- a/src/audio.c
+++ b/src/audio.c
@@ -137,6 +137,16 @@ struct autx {
  |/    |        |   |       |   |        |   |        |
        '--------'   '-------'   '--------'   '--------'
 
+       If an AEC is in the filter list aufilt, then the aubuf should be
+       constant and small. A second audio buffer aubufdec should be added via
+       config if network jitter is expected.
+
+       .--------.   .-------.   .--------.   .----------.   .--------.
+ |\    |        |   |       |   |        |   |  opt.    |   |        |
+ | |<--| auplay |<--| aubuf |<--| aufilt |<--| aubufdec |<--| decode |<--- RTP
+ |/    |        |   |       |   |        |   |          |   |        |
+       '--------'   '-------'   '--------'   '----------'   '--------'
+
  \endverbatim
  */
 struct aurx {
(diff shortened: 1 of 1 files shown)


sreimers:
> If an AEC is in the filter list aufilt, then the aubuf should be constant and small.

With aubuf you mean the `auplay aubuf` should constant and small for AEC or the `aubufdec`?

cspiel1:
The `auplay` `aubuf` should be small and constant. Where "constant" is most important for an AEC.

sreimers:
I see, can we put the AEC in a "auplay aufilt" stage?:
 ```
auplay <-- auplay aufilt <-- aubuf <-- aufilt <-- decode <-- RTP
```
If the AEC calculation is fast enough this should archive the best results.


> But the thread you meant is for the RTP and decode, right?

Right.

**Comment by cspiel1** [comment]
> I see, can we put the AEC in a "auplay aufilt" stage?:
> 
> ```
> auplay <-- auplay aufilt <-- aubuf <-- aufilt <-- decode <-- RTP
> ```

This was my first draft. When discussing this with our audio experts, we found that indeed AEC has to be very fast because the auplay thread has the strictest real time constraints in the whole decode pipeline. The AEC (e.g. our audio-core) can not guaranty this real time constraints in general.

> 
> If the AEC calculation is fast enough this should archive the best results.
> 
> > But the thread you meant is for the RTP and decode, right?
> 
> Right.

The RTP+decode thread is a good idea. It does not overlap with this PR.

**Comment by sreimers** [comment]
The decode part for AEC is normally not heavy (since the most work is done in encode) but maybe depends on the implementation.

Some quick `webrtc_aec` aufilt measurements (20ms ptime):

20-85 microseconds (decode)
100-350 microseconds (encode)

So putting webrtc_aec decode in auplay thread should be fast enough for realtime.

By `The AEC (e.g. our audio-core) ` you mean a custom/hardware AEC?
Can we put the `aubufdec` into a aufilter module?

**Comment by cspiel1** [comment]
> 20-85 microseconds (decode) 100-350 microseconds (encode)
I will discuss this with our audio experts. Maybe they could make similar measurements with our audio-core. 

> By `The AEC (e.g. our audio-core) ` you mean a custom/hardware AEC?
Yes a custom audio-core containing AEC, noise suppression, ...

> Can we put the `aubufdec` into a aufilter module?
Maybe. Then there would not be any change in the baresip core. I'll give it a try.

**Comment by cspiel1** [comment]
And I just realized that if an decoding aubuf is added another thread is needed for the filter list, otherwise the auplay aubuf will have the underruns. This thread can be put also in the module. I will call this module `auajb` for "Audio - Adaptive Jitter Buffer" with one config setting `decode_buffer   20-160`.

**Comment by juha-h** [comment]
Christian Spielberger writes:

> And I just realized that if an decoding aubuf is added another thread
> is needed for the filter list, otherwise the auplay aubuf will have
> the underruns. This thread can be put also in the module. I will call
> this module `auajb` for "Audio - Adaptive Jitter Buffer" with one
> config setting `decode_buffer   20-160`.

After you have done all these changes, please let non-audio experts like
me to know what changes are needed in current config in order to make it
optional in common use cases.  For example, should the new module be
loaded.

File: src/audio.c [hunk 1]
```diff
@@ -173,6 +184,14 @@ struct aurx {
 	enum jbuf_type jbtype;       /**< Jitter buffer type               */
 	volatile int32_t wcnt;       /**< Write handler call count         */
 
+#ifdef HAVE_PTHREAD
```
alfredh, line 187: when using the new C11 thrd_t api, I think the ifdef is no longer needed ?

src/audio.c
```diff
@@ -173,6 +184,14 @@ struct aurx {
...
+	struct {
+		thrd_t thrd;         /**< Audio RX filter thread           */
+		mtx_t mtx;           /**< Mutex for stopping thread        */
```
alfredh, line 190: is it possible to use Atomic bool here, instead of mutex ?

src/audio.c
```diff
@@ -751,20 +779,84 @@ static int stream_pt_handler(uint8_t pt, struct mbuf *mb, void *arg)
 }
 
 
+static int process_decfilt(struct aurx *rx, struct auframe *af)
+{
+	struct le *le;
```
alfredh, line 784: le can now be declared inside the for loop (C99 style)

src/audio.c
```diff
@@ -751,20 +779,84 @@ static int stream_pt_handler(uint8_t pt, struct mbuf *mb, void *arg)
...
+
+		if (err)
+			return err;
+	}
+
+	return 0;
+}
+
+
+static int rx_push_aubuf(struct aurx *rx, struct auframe *af)
```
alfredh, line 802: auframe can be made const ?

src/audio.c
```diff
@@ -751,20 +779,84 @@ static int stream_pt_handler(uint8_t pt, struct mbuf *mb, void *arg)
...
 static int aurx_stream_decode(struct aurx *rx, const struct rtp_header *hdr,
 			      struct mbuf *mb, unsigned lostc, bool drop)
 {
 	struct auframe af;
 	size_t sampc = AUDIO_SAMPSZ;
 	bool marker = hdr->m;
-	struct le *le;
 	int err = 0;
 	const struct aucodec *ac = rx->ac;
+	bool flush = rx->ssrc != hdr->ssrc && rx->ssrc;
```
alfredh, line 852: ssrc value 0 is also a valid SSRC. I think there should be a variable ssrc_set that can be used ?

src/audio.c
cspiel1: I think so. This mtx solution is maybe overcautious. I saw this in module ausine.

src/audio.c
cspiel1: okay, we can simplify this:
```c
bool flush = rx->ssrc != hdr->ssrc;
```
because it doesn't make a difference if we flush at the beginning. The first frame is pushed to `aubuf` anyway.

PR #1848, opened 2022-05-21
dns: add query cache

sreimers:
The dns test checks for rotating entries and so these cannot be cached.

depends on: https://github.com/baresip/re/pull/369

### DIFF
--- a/test/ua.c
+++ b/test/ua.c
@@ -524,7 +524,7 @@ static int reg_auth_dns(enum sip_transp tp)
 
 	memset(&t, 0, sizeof(t));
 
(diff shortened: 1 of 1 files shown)


juha-h:
I have noticed in my Android app that even with re_thread_async, dns queries sometimes fail when dns query cache is in use.  In those cases, I get to log:
```
reg: sip:foo@test.tutpro.com (prio 0): Register: Destination address required [89]
```
When I turn caching off with `dnsc_cache_max(net_dnsc(baresip_network()), 0)` I do not get these registration failures.  So I'll keep it that way for now.

PR #1850, opened 2022-05-22
avfilter: migrate to C11 mutex

alfredh:
I noticed that destroy is missing from the wrapper api ?

it is probably a good idea to migrate lock usage from "struct lock" to C11 mutex.
after that we can delete the "lock" module

### DIFF
--- a/modules/avfilter/avfilter.c
+++ b/modules/avfilter/avfilter.c
@@ -152,7 +152,7 @@ static int module_init(void)
 
 static int module_close(void)
 {
-	/* todo: mutex destroy ? */
+	mtx_destroy(&lock);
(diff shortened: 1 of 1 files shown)


sreimers:
> I noticed that destroy is missing from the wrapper api ?

I added a PR https://github.com/baresip/re/pull/371

> it is probably a good idea to migrate lock usage from "struct lock" to C11 mutex. after that we can delete the "lock" module
Yes I think most `lock` usage can be replaced. The only missing feature is read/write lock (is not defined by C11 mutex).

alfredh:
thanks!

I updated the PR with mtx_destroy.

PR #1853, opened 2022-05-22
audio: C11 mutex

alfredh:


alfredh:
one test fails with:

```
/usr/bin/ld: build-x86_64/src/audio.o: undefined reference to symbol 'mtx_unlock@@GLIBC_2.28'
[621](https://github.com/baresip/baresip/runs/6544008690?check_suite_focus=true#step:9:622)
/usr/bin/ld: /lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
[622](https://github.com/baresip/baresip/runs/6544008690?check_suite_focus=true#step:9:623)
collect2: error: ld returned 1 exit status
```

sreimers:
Looks like we have a `HAVE_PTHREAD=` check here: https://github.com/baresip/baresip/blob/main/.github/workflows/extra.yml#L53

I think we can remove this. I would like to deprecate HAVE_PTHREAD since its the default fallback for threads now.

alfredh:
great,

yes I agree. It would be nice to get rid of the HAVE_PTHREAD check in baresip or applications.
this info should instead be configured in the library.

I will also make an attempt to convert pthread_create to thrd_create

PR #1865, opened 2022-05-28
av1: add packetize handler

alfredh:
this enables support for avformat pass through of AV1

1. avformat module can enable pass_through mode
2. the mediafile read can contain AV1 bitstream
3. the bitstream will not be decoded by avformat, but passed raw to av1.so codec module
4. the av1 packetize handler will convert the bitstream to RTP packets

### DIFF
--- a/modules/av1/encode.c
+++ b/modules/av1/encode.c
@@ -290,3 +290,36 @@ int av1_encode_packet(struct videnc_state *ves, bool update,
 
 	return err;
 }
+
+
+int av1_encode_packetize(struct videnc_state *ves,
+			 const struct vidpacket *packet)
+{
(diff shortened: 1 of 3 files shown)


modules/av1/encode.c
```diff
@@ -290,3 +290,36 @@ int av1_encode_packet(struct videnc_state *ves, bool update,
 
 	return err;
 }
+
+
+int av1_encode_packetize(struct videnc_state *ves,
+			 const struct vidpacket *packet)
+{
+	struct mbuf *mb2;
```
sreimers, line 298: `mb2` is called `mb_pkt`  on other places, maybe more meaningful here too?

PR #1886, opened 2022-06-03
A "real-time" thread for RTP RX

cspiel1:
- stream: RX real-time - add RX thread
- stream: RX real-time - make some fields atomic, add TODOs
- stream,metric: RX real-time - make metric thread-safe


sreimers:
Looks like `cnd_destroy` is missing will make a PR for posix and win32.

cspiel1:
Thanks to the multi-threading https://github.com/baresip/re/wiki/Development-(re-usage)#multithreading support of `re` the changes are only in `baresip`.

Now it turns out that the main part of this PR is to make everything thread-safe that is written/read from the global re_main thread and the "real-time" re_main thread.

- Commit 2 makes some fields atomic.
- Commit 3 makes `metric` thread-safe.
- Added also some TODOs.

Any hints to make this clean and simple would be great.

src/core.h
```diff
@@ -144,33 +144,19 @@ int conf_get_csv(const struct conf *conf, const char *name,
...
-	/* counters: */
-	uint32_t n_packets;
-	uint32_t n_bytes;
-	uint32_t n_err;
-
-	/* bitrate calculation */
-	uint32_t cur_bitrate;
-	uint64_t ts_last;
-	uint32_t n_bytes_last;
-};
```
alfredh, line 167: can you please make this a separate patch, moving "struct metric" to metric.c ?

src/core.h
```diff
@@ -144,33 +144,19 @@ int conf_get_csv(const struct conf *conf, const char *name,
...
+struct metric;
 
 int      metric_init(struct metric *metric);
 void     metric_reset(struct metric *metric);
 void     metric_add_packet(struct metric *metric, size_t packetsize);
 double   metric_avg_bitrate(const struct metric *metric);
+uint32_t metric_n_packets(struct metric *metric);
+uint32_t metric_n_bytes(struct metric *metric);
+uint32_t metric_n_err(struct metric *metric);
+uint32_t metric_bitrate(struct metric *metric);
```
alfredh, line 156: read-only accessors can use "const"

src/core.h
cspiel1: It is a separate commit. You mean a separate PR? --> https://github.com/baresip/baresip/pull/1895

cspiel1:
Example for missing thread safety in this PR:
`stream.rtcp_stats`
- is **written** in new "real-time" thread function `rtcp_handler()`,
- is **read** in `print_rtp_stats()` from global main thread.
- And `rtcp_handler()` calls callback functions `rtcph` and `sessrtcph`. It seems not be okay, if another thread calls these functions.

cspiel1:
> Example for missing thread safety in this PR: `stream.rtcp_stats`
> 
> * is **written** in new "real-time" thread function `rtcp_handler()`,
> * is **read** in `print_rtp_stats()` from global main thread.
> * And `rtcp_handler()` calls callback functions `rtcph` and `sessrtcph`. It might be problematic if another thread calls these functions.

Answer myself here:
- `print_rtp_stats()` is called after the `rx_thread` finished --> Ok
- Another read of `rtcp_stats` is in call.c function `stream_rtcp_handler()`. But this runs also in the rx thread. --> Ok
- and a read in module `rtcpsummary.c`, after audio stopped but the stream still exists --> needs a mutex. We need a `stream_rx_lock()` function.
- also module `vidinfo` --> `stream_rx_lock()`

For simpler API: Instead of a `stream_rx_lock()` we should make `stream_rtcp_stats()` thread safe. It shouldn't return a pointer.

cspiel1:
- Made `stream_rtcp_stats()` thread-safe. We loose the `const` qualifiers for struct stream for some functions because the mtx has to be locked inside.
- Call  to `s->rtcph`. The `rtcp_handler` in video.c is thread safe and audio.c doesn't have one.
- TODO: `s->sessrtcph` has to be thread-safe.

cspiel1:
- `s->rtph`: audio.c - `stream_recv_handler()` and `aurx_stream_decode()` seems thread-safe to me. --> should be verified with helgrind.
- `s->rtpestabh`: TODO
- `s->pth`: stream.c - `stream_pt_handler()` -> `handle_telev()`, `audio_decoder_set()` have to be made thread-safe.

modules/rtcpsummary/rtcpsummary.c
```diff
@@ -9,12 +9,15 @@
 
 
 static void print_rtcp_summary_line(const struct call *call,
-				    const struct stream *s)
+				    struct stream *s)
```
alfredh, line 12: is this module used ?

Perhaps we should move the xrtpstat code to a separate module, and allow it
to print the stats to a SIP header or stdout ?

src/audio.c
```diff
@@ -113,7 +113,7 @@ struct autx {
 		RE_ATOMIC bool run;   /**< Audio transmit thread running   */
 	} thr;
 
-	mtx_t lock;
+	mtx_t mtx;
```
alfredh, line 116: this diff is already quite large ...

please make the renaming lock->mtx a separate PR

alfredh, line 116: this diff is already quite large ...

please make the renaming lock->mtx a separate PR

src/audio.c
```diff
@@ -716,17 +722,47 @@ static void ausrc_error_handler(int err, const char *str, void *arg)
...
+	mtx_unlock(&a->rx.mtx);
+	if (err)
 		return;
 
 	digit = telev_code2digit(event);
-	if (digit >= 0 && a->eventh)
-		a->eventh(digit, end, a->arg);
+	if (digit >= 0) {
+		struct dtmf_event *e;
+		e = mem_zalloc(sizeof(*e), NULL);
```
alfredh, line 759: check for OOM ?

alfredh, line 759: check for OOM ?

src/audio.c
```diff
@@ -1189,8 +1231,8 @@ int audio_alloc(struct audio **ap, struct list *streaml,
 	rx->pt     = -1;
 	rx->ptime  = ptime;
 
-	err  = mtx_init(&tx->lock, mtx_plain);
-	err |= mtx_init(&rx->lock, mtx_plain);
+	err  = mtx_init(&tx->mtx, mtx_plain);
+	err |= mtx_init(&rx->mtx, mtx_recursive);
```
alfredh, line 1235: what is the reason for using Recursive lock here ?

alfredh, line 1235: what is the reason for using Recursive lock here ?

src/audio.c
```diff
@@ -1304,7 +1346,7 @@ static int autx_print_pipeline(struct re_printf *pf, const struct autx *autx)
 }
 
 
-static int aurx_print_pipeline(struct re_printf *pf, const struct aurx *rx)
+static int aurx_print_pipeline(struct re_printf *pf, struct aurx *rx)
```
alfredh, line 1349: this change should not be needed, if you use the new `mtx_alloc` function

alfredh, line 1349: this change should not be needed, if you use the new `mtx_alloc` function

src/stream.c
```diff
@@ -659,6 +728,8 @@ static int receiver_init(struct receiver *rx)
 	tmr_init(&rx->tmr_rtp);
 
 	rx->pseq = -1;
+	mtx_init(&rx->mtx, mtx_plain);
```
alfredh, line 731: error checking is missing here ...

alfredh, line 731: error checking is missing here ...

cspiel1:
Needs another rebase after merge of: https://github.com/baresip/baresip/pull/1906

src/audio.c
```diff
@@ -1413,44 +1463,78 @@ static int aufilt_setup(struct audio *a, struct list *aufiltl)
...
+{
+	const struct aucodec *ac = rx->ac;
+	size_t size = rx->last_sampc * aufmt_sample_size(rx->dec_fmt);
+
+	if (!size)
+		return EINVAL;
+
+	if (size != rx->fsize) {
+		rx->sampvf = mem_deref(rx->sampvf);
+		rx->sampvf = mem_zalloc(size, NULL);
```
alfredh, line 1476: you can also use `mem_realloc` or `mem_reallocarray` here ...

alfredh:
moving the processing of incoming audio packets from main thread to new thread
will most likely have many unknown side effects.

I suggest that this patch is tested really well, before merging

also please ask @juha-h for one extra round of testing on Android :)

modules/rtcpsummary/rtcpsummary.c
cspiel1: I have no idea if `rtcpsummary` is used.

Also have no idea how to make `call->sess` thread safe.
- `stream_rtcp_handler()` is called from RX thread,
-  while e.g. `call_hangup()` does a `mem_deref(call->sess)` in the main thread.

Another mutex in `struct call`?

@sreimers, @alfredh : The changes for thread-safety are already very large and complicated. Is there enough benefit for an RX real-time thread?

Or should we close this?

Is there a chance for a redesign of the RTP processing in order to have it thread-safe out of the box?

cspiel1:
A decode thread would not improve the audio quality.

The idea was to handle the incoming RTP file descriptors in a separate thread to avoid underruns if e.g. a long blocking command of any module is invoked during a call.

There is a "long async blocking" warning in re `main.c` that could help to detect such "ugly" commands.
@sreimers : Should we close this PR?

sreimers:
> @sreimers : Should we close this PR?

Yes we can close this for now.

alfredh:
the design around re main is based on the assumption that all the event loop code
executes very fast. If a function blocks for a long time, this breaks the design.
You should find which function is blocking and try to improve it.


it is not a bad idea to have dedicated worker threads for Audio and Video,
but the implications has to be analyzed properly

juha-h:
Alfred E. Heggestad writes:

> moving the processing of incoming audio packets from main thread to new thread
> will most likely have many unknown side effects.
> 
> I suggest that this patch is tested really well, before merging
> 
> also please ask @juha-h for one extra round of testing on Android :)

I hope that entering and leaving this new thread is done automatically
in baresip lib so that I don't need to add any thread enter/leave calls in
the app.

cspiel1:
> the design around re main is based on the assumption that all the event loop code executes very fast. If a function blocks for a long time, this breaks the design. You should find which function is blocking and try to improve it.

I agree. Such long blocking calls should be improved.
On the other side our opinion is that this is not enough for a reliable audio processing. Especially for video calls on embedded devices the SW jitter in the polling main loop is significantly higher than for audio only calls. Further any user command might block the main loop for a few milli seconds.

We approximately have these SW jitter values:
- during call start: 1 - 80 ms
- typical: 17 - 23 ms
- worst case:  5 - 35 ms

> it is not a bad idea to have dedicated worker threads for Audio and Video, but the implications has to be analyzed properly

Right! We plan to work on this in the comming year. Maybe I'll start with defining some constrains that help to make everything thread-safe. Like: "struct RX is only accessed from audio thread."

PR #1897, opened 2022-06-09
audio: make rx.filtl thread-safe

cspiel1:
Involved threads are:
- global re main thread
- rx_filter_thread (only used if config contains: `decode_buffer_mode`)

commits
- audio: rename lock to mtx
- audio: make rx.filtl thread-safe

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -303,7 +303,9 @@ static void stop_rx(struct aurx *rx)
 	rx->auplay = mem_deref(rx->auplay);
 	rx->aubuf  = mem_deref(rx->aubuf);
 	rx->aubufdec  = mem_deref(rx->aubufdec);
+	mtx_lock(rx->mtx);
 	list_flush(&rx->filtl);
+	mtx_unlock(rx->mtx);
 }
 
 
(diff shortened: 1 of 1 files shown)


src/audio.c
```diff
@@ -1304,7 +1311,7 @@ static int autx_print_pipeline(struct re_printf *pf, const struct autx *autx)
 }
 
 
-static int aurx_print_pipeline(struct re_printf *pf, const struct aurx *rx)
+static int aurx_print_pipeline(struct re_printf *pf, struct aurx *rx)
```
alfredh, line 1314: we should try to avoid this

with this change, we are loosing the "const-ness".
removing const means that the function is modifying the state.

this was working with the old "struct lock", why is it not working with "mtx_t" ?

alfredh, line 1314: we should try to avoid this

with this change, we are loosing the "const-ness".
removing const means that the function is modifying the state.

this was working with the old "struct lock", why is it not working with "mtx_t" ?

alfredh:
proposal to split the PR into 2 PRs, one for renaming the mtx and one for the rest
(this is mainly for readability and future debugging)

src/audio.c
sreimers: `mtx_alloc` is now available: https://github.com/baresip/re/commit/79a6168f144125da99b0173f96cfda20b7e925b8

alfredh: this is great !

we should try to keep the const-ness, it gives a hint to the compiler about what
can be optimized, and it also improves readability.

I think there are 1-2 places in re+rem that also has the same issue with const and mtx_t

PR #1907, opened 2022-06-15
Pulseaudio Async Interface Module

cHuberCoffee:
A new pulseaudio module taking advantage of the asynchronous interface of pulseaudio.

### DIFF
--- a/modules/pulse_async/player.c
+++ b/modules/pulse_async/player.c
@@ -1,0 +1,49 @@
+/**
+ * @file player.c  Pulseaudio sound driver - player (asynchronous API)
+ *
+ * Copyright (C) 2021 Commend.com - h.ramoser@commend.com,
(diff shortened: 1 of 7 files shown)


modules/pulse_async/player.c
```diff
@@ -0,0 +1,168 @@
...
+	size_t sz = len;
+	static uint64_t t0 = 0;
+	uint64_t t;
+
+	if (st->b->shutdown)
+		goto out;
+
+	t = tmr_jiffies_usec();
+	if (!t0)
+		t0 = t;
```
sreimers, line 146: Looks like `t0` and `t` are unused?

alfredh:
many thanks!

can you also add a CMake snippet cmake/FindPulse.cmake ?

we are migrating from GNU Make to CMake...

PR #1913, opened 2022-06-19
In start_registering, create register clients if reg list is empty

juha-h:
Also, had to change location of `create_register_clients` and `add_extension` functions.

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -193,11 +193,11 @@ static int start_register(struct ua *ua, bool fallback)
 		}
 	}
 
-	if (!fallback)
-		ua_event(ua, UA_EVENT_REGISTERING, NULL, NULL);
-
 	if (list_isempty(&ua->regl))
 		create_register_clients(ua);
+
+	if (!fallback && !list_isempty(&ua->regl))
+		ua_event(ua, UA_EVENT_REGISTERING, NULL, NULL);
 
 	for (le = ua->regl.head, i=0; le; le = le->next, i++) {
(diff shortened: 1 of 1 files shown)


src/ua.c
```diff
@@ -138,6 +196,9 @@ static int start_register(struct ua *ua, bool fallback)
 	if (!fallback)
 		ua_event(ua, UA_EVENT_REGISTERING, NULL, NULL);
 
+	if (list_isempty(&ua->regl))
+		create_register_clients(ua);
+
```
cspiel1, line 198: Does this change anything?

juha-h:
Christian Spielberger writes:

> > @@ -138,6 +196,9 @@ static int start_register(struct ua *ua, bool fallback)
>  	if (!fallback)
>  		ua_event(ua, UA_EVENT_REGISTERING, NULL, NULL);
>  
> +	if (list_isempty(&ua->regl))
> +		create_register_clients(ua);
> +
> 
> Does this change anything?

Yes, it creates register clients if baresip was initially started with
regint=0.

juha-h:
Just to add to my previous comment, register clients are not added by `create_register_clients` if regint=0 like it is when baresip is started with ;regint=0.  When /uareg is called with regint>0, subsequent call of `create_register_clients` does add them.

cspiel1:
Okay, now I understand. So this should be okay.

juha-h:
Christian Spielberger writes:

> Okay, now I understand. So this should be okay.

Thanks, I'll merge the PR.

PR #1914, opened 2022-06-21
uag: add timestamps to SIP trace

cspiel1:


src/uag.c
```diff
@@ -3,6 +3,8 @@
  *
  * Copyright (C) 2010 Alfred E. Heggestad
  */
+#include <time.h>
+#include <sys/time.h>
```
sreimers, line 7: `sys/time.h` needs a ifdef:

```c
#ifdef HAVE_SYS_TIME_H
#include <sys/time.h>
#endif
```

src/uag.c
```diff
@@ -499,15 +501,26 @@ static void sip_trace_handler(bool tx, enum sip_transp tp,
...
+	struct tm *local;
+	struct timeval curTime;
+	int msecs;
 	(void)tx;
 	(void)arg;
 
+	time(&now);
+	local = localtime(&now);
+
+	gettimeofday(&curTime, NULL);
```
sreimers, line 514: For WIN32 we need something like this:

```c
#ifdef WIN32
SYSTEMTIME st;
GetSystemTime(&st);
msecs = st.wMilliseconds;
#else
struct timeval curTime;
gettimeofday(&curTime, NULL);
msecs = (int) (curTime.tv_usec / 1000);
#endif
```

src/uag.c
```diff
@@ -499,15 +504,38 @@ static void sip_trace_handler(bool tx, enum sip_transp tp,
...
+	struct tm* local;
+
+	(void)clock_gettime(CLOCK_REALTIME, &ts);
+	ms = ts.tv_nsec / 1000000;
+	local = localtime(&ts.tv_sec);
+	h = local->tm_hour;
+	m = local->tm_min;
+	s = local->tm_sec;
+#endif
+
```
alfredh, line 528: it would be nice to make this code portable and put it in re.

we have some similar functions in re/src/fmt/time.c

did you have a look at `fmt_gmtime`

PR #1915, opened 2022-06-21
event: make UA events thread safe

cspiel1:
Ensures that the event handlers are called in the same thread from which
uag_event_register() was called. This is normally the baresip main thread.

This also fixes the thread check warning for the module event in module
ctrl_dbus.

### DIFF
--- a/src/event.c
+++ b/src/event.c
@@ -389,16 +389,16 @@ void ua_event(struct ua *ua, enum ua_event ev, struct call *call,
 		struct ua_event_data *ed;
 		le = le->next;
 
-		if (thrd_current() != eh->thrd) {
+		if (thrd_equal(thrd_current(), eh->thrd)) {
+			eh->h(ua, ev, call, buf, eh->arg);
+		}
+		else {
 			if (!eh->mqueue)
 				continue;
 
 			ed = ua_event_data_alloc(eh, ua, ev, call, buf);
 			if (ed)
 				mqueue_push(eh->mqueue, 0, ed);
-		}
-		else {
-			eh->h(ua, ev, call, buf, eh->arg);
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


src/event.c
```diff
@@ -376,9 +444,21 @@ void module_event(const char *module, const char *event, struct ua *ua,
 	le = ehl.head;
 	while (le) {
 		struct ua_eh *eh = le->data;
+		struct ua_event_data *ed;
+		enum ua_event ev = UA_EVENT_MODULE;
 		le = le->next;
 
-		eh->h(ua, UA_EVENT_MODULE, call, buf, eh->arg);
+		if (thrd_current() != eh->thrd) {
```
sreimers, line 451: Comparing threads should be done with thrd_equal().

alfredh:
not sure about this ...

perhaps it is better to require the sender to send from the right thread ?

juha-h:
In my Android app I have one `ua_event_handler`, that attaches to Android applications thread and then calls applications Java function:
```
static void ua_event_handler(struct ua *ua, enum ua_event ev,
			     struct call *call, const char *prm, void *arg)
{
    const char *event;
    char event_buf[256];
    int len;
    ...
    switch (ev) {
    ....
    case UA_EVENT_CALL_OUTGOING:
            len = re_snprintf(event_buf, sizeof event_buf, "call outgoing");
            break;
    ...
    }

    event = event_buf;

    JavaVM *javaVM = g_ctx.javaVM;
    JNIEnv *env;
    jint res = (*javaVM)->GetEnv(javaVM, (void**)&env, JNI_VERSION_1_6);
    if (res != JNI_OK) {
        res = (*javaVM)->AttachCurrentThread(javaVM, &env, NULL);
        if (JNI_OK != res) {
            LOGE("failed to AttachCurrentThread, ErrorCode = %d\n", res);
            return;
        }
    }

    jmethodID methodId = (*env)->GetMethodID(env, g_ctx.mainActivityClz, "uaEvent", "(Ljava/lang/String;JJ)V");
    jstring jEvent = (*env)->NewStringUTF(env, event)
    (*env)->CallVoidMethod(env, g_ctx.mainActivityObj, methodId, jEvent, (jlong)ua, (jlong)call);
    (*env)->DeleteLocalRef(env, jEvent);
}
```
Why would that handler need to be called from the same thread as from which
uag_event_register() was called?

cspiel1:
This are good questions. We should restrict the call to `ua_event()` and `module_event()` to the main thread.
And module which run additional threads and implement an UA event handler have to be made thread-safe. They can use:
- `mqueue` or
- `mtx_t`

Thus I'll replace this PR by a fix in module `ctrl_dbus`. It calls `module_event()` from a "wrong" thread.

cspiel1:
https://github.com/baresip/baresip/pull/1921 replaces this one.

PR #1916, opened 2022-06-21
ctrl_dbus: use new thread and mtx interface

cspiel1:


modules/ctrl_dbus/ctrl_dbus.c
```diff
@@ -61,10 +61,12 @@
  * Copyright (C) 2020 commend.com - Christian Spielberger
  */
 
+#include <pthread.h>
```
sreimers, line 64: `pthread.h` should not be necessary anymore.

PR #1918, opened 2022-06-22
Un-subscribe mwi at un-register

juha-h:
When UNREGISTERING event is received, un-SUBSCRIBE possible MWI subscription, because at least when outbound is used, SIP proxy cannot anymore route NOTIFY requests to baresip based on gruu.


alfredh:
@juha-h can you volunteer as a maintainer for the MWI module ?

juha-h:
Yes, I can try.

PR #1940, opened 2022-07-06
Add CMake target to install baresip executable, library and modules

widgetii:
The same approach used for re and rem libraries


alfredh:
many thanks for your contributions!

For Cmake we are planning two things:

1. add support for cmake in CI (github workflows)
2. add support for cmake to all supported modules


If you have time to help out with some of this, it would be really great!
(it seems like you are an CMake expert :)

PR #1944, opened 2022-07-09
ci: test cmake

alfredh:


.github/workflows/cmake.yml
```diff
@@ -0,0 +1,37 @@
...
+        secret: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: make re/rem
+      run: |
+        for p in re rem; do cd $p && sudo PATH="$PATH" make install && cd - && sudo rm -Rf $p; done
+        sudo ldconfig
+
+    - name: baresip
+      run: |
+        cmake -DCMAKE_C_FLAGS="-Werror" . && make
```
sreimers, line 37: Maybe we should use a build directory like so:

`cmake -B build -DCMAKE_C_FLAGS="-Werror" && cmake --build build --parallel`

alfredh:
thanks, I started with Ubuntu only.

Later we can add macOS and Windows.

PR #1960, opened 2022-07-15
Stun uri cred

viordash:
This PR allows to use the user credentials **[user:password]** from the stunserver URI, like so:
`# ;stunserver=stun:[user:password]@host[:port]`

When writing the parsing part, I relied on your [stun.patch](https://gist.github.com/alfredh/ab96b75b16a3c313770d9a85836388c9)


viordash:
[629e0c0](https://github.com/baresip/baresip/pull/1960/commits/629e0c0bb6bfc2ea5193c800a40514f29b7a3a1c)
Added a case to the tests when the URI of the stunserver parameter itself does not contain user/passw and there are also no stunuser/stunpass parameters

src/account.c
```diff
@@ -81,29 +81,40 @@ static int param_u32(uint32_t *v, const struct pl *params, const char *name)
...
 		err |= param_dstr(&acc->stun_user, &aor->params, "stunuser");
+	else if (pl_isset(&uri.user))
+		err |= re_sdprintf(&acc->stun_user, "%H",
+					uri_user_unescape, &uri.user);
 
 	if (0 == msg_param_exists(&aor->params, "stunpass", &tmp))
 		err |= param_dstr(&acc->stun_pass, &aor->params, "stunpass");
+	else if (pl_isset(&uri.password))
+		err |= re_sdprintf(&acc->stun_pass, "%H",
+					uri_password_unescape, &uri.password);
```
alfredh, line 117: the uri "password" field is deprecated

better to use a parameter e.g. "password=XXXX" or the credential field
in the STUN uri spec

viordash: You are completely correct that the "password" uri field is deprecated. But I don't see how to prevent stunserver from being used with this field. Moreover, in [;stunserver=stun:[user:pass]@host[:port]](https://github.com/baresip/baresip/blob/main/docs/examples/accounts) this field is indicated as an example for the stunserver parameter.
Might be worth generating a warning message if this option is used?
`
	else if (pl_isset(&uri.password)) {
		err |= re_sdprintf(&acc->stun_pass, "%H",
					uri_password_unescape, &uri.password);
		warning("account: use of the format "user:pass" in the stunserver field is deprecated\n",
				&srv, err);			
	}
`

alfredh:
hi,

the first step is to remove usage of "password" field in any uri. If the password field is set,
we should print a warning about deprecation.

the STUN username and password can already be set using "stunuser" and "stunpass"

for additional fields, we could look to this for inspiration:

https://developer.mozilla.org/en-US/docs/Web/API/RTCIceServer

viordash:
Hi
I didn't quite understand you, you are suggesting to print a warning if any uri has a password. But will that uri's password **continue to be used** for authentication?
If I understand this point correctly, then what do you think is the best place to generate a warning message? Directly in the body of the `uri_decode(struct uri *uri, const struct pl *pl)` method if there is a password parameter in the uri? To do this, I do not know how to generate a message in the "Re" library, the warning method is presented in log.c in the BareSIP project

If you think that the password parameter in the uri should also be ignored for authentication, then that would be a bit of a problem for me. The reason is that several instances of the program have a password in the stun uri, I don't want (and probably can't even) migrate from them to "stunuser" and "stunpass".

alfredh:
okay, then let us merge this patch if it helps you

but in the long run we should phase out usage of "password" in uri

(in first stage print a warning, then remove the field from struct uri)

src/account.c
```diff
@@ -81,29 +81,40 @@ static int param_u32(uint32_t *v, const struct pl *params, const char *name)
...
-		err = stunuri_decode(&acc->stun_host, &srv);
+		err = uri_decode(&uri, &srv);
 		if (err) {
-			warning("account: decode '%r' failed: %m\n",
+			warning("account: decode '%r' failed (%m)\n",
 				&srv, err);
+			return err;
 		}
+
+		stunuri_decode_uri(&acc->stun_host, &uri);
```
alfredh, line 102: check return value here

src/account.c
```diff
@@ -81,29 +81,40 @@ static int param_u32(uint32_t *v, const struct pl *params, const char *name)
...
 
 	if (0 == msg_param_exists(&aor->params, "stunuser", &tmp))
 		err |= param_dstr(&acc->stun_user, &aor->params, "stunuser");
+	else if (pl_isset(&uri.user))
+		err |= re_sdprintf(&acc->stun_user, "%H",
+					uri_user_unescape, &uri.user);
 
 	if (0 == msg_param_exists(&aor->params, "stunpass", &tmp))
 		err |= param_dstr(&acc->stun_pass, &aor->params, "stunpass");
+	else if (pl_isset(&uri.password))
```
alfredh, line 115: we can print a message here saying "deprecated field used"

src/stunuri.c
```diff
@@ -71,21 +67,44 @@ int stunuri_decode(struct stun_uri **sup, const struct pl *pl)
...
+
+/**
+ * Decode a STUN uri from URI object
+ *
+ * @param sup Pointer to allocated STUN uri
+ * @param uri URI object
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int stunuri_decode_uri(struct stun_uri **sup, const struct uri *uri) {
```
alfredh, line 83: curly brackets for functions should be on the line after function declaration, like this:

```
function
{
}
```

test/account.c
```diff
@@ -151,3 +151,103 @@ int test_account_uri_complete(void)
...
+int test_account_stun_uri_complete(void)
+{
+	struct account *acc = NULL;
+	int err = 0;
+	struct sip_addr *addr;
+
+	err = account_alloc(&acc, "\"A\" <sip:A@proxy.com>"
+		";stunuser=bob@bob.com"
+		";stunpass=taj:aa"
+		";stunserver=\"stun:42:19@stunserver.org\"");
```
alfredh, line 165: maybe remove password from testcode

src/account.c
viordash: [5681fba](https://github.com/baresip/baresip/pull/1960/commits/5681fbafdfab838962deab8dc12ca5e9fd90f5af)

src/account.c
viordash: [d7b8d0d](https://github.com/baresip/baresip/pull/1960/commits/d7b8d0d951f7800f5cfcfee67bf026b922d48149)
It seems to me that the most correct place to generate a message is inside the stunuri_decode_uri method. In the future, if the password in uri is completely disabled, it is possible to return an error

src/stunuri.c
viordash: [9f7cac4](https://github.com/baresip/baresip/pull/1960/commits/9f7cac47f4edc12a50dd95faf60441eb601e5509)

test/account.c
viordash: These are not real passwords, or do you mean something else?

viordash:
It seems to me that the documentation, docs/examples/accounts, should have a note about the password in the uri.

alfredh:
thanks for the update


in general we dont want any warnings when running selftest


after applying your patch, a new warning is printed with selftest:


```
[ RUN      ] test_account_stun_uri_complete
The "user:password" format in the stunserver userinfo field is deprecated.
The "user:password" format in the stunserver userinfo field is deprecated.
The "user:password" format in the stunserver userinfo field is deprecated.
[       OK ]
```

I suggest to keep the warning, but remove usage of the uri "password" field in the testcode

(The testcode also serves as documentation for people to see how to use the API)

viordash:
Thank you.
'test_account' already checks parsing stun uri parameters as they are recommended to be used and so the newly added 'test_account_stun_uri_complete' can be removed

viordash:
Thanks

PR #1961, opened 2022-07-16
ci: add cmake/macos

alfredh:


alfredh:
Linking modules is failing on MacOS:

```
[ 77%] Linking C shared module account.so
Undefined symbols for architecture x86_64:
  "_auconv_from_s16", referenced from:
      _common_resample in auresamp.c.o
  "_auconv_to_s16", referenced from:
      _common_resample in auresamp.c.o
```

I am wondering if we have to add some linker flags for modules/MacOS ?

alfredh:
there are two ways to do this, not sure which method is best:

1. make all public symbols in baresip visible, and link each module to -lbaresip
2. for each module, use runtime symbol look undefined symbols

we have something like this in re.mk:


```
ifeq ($(OS),darwin)
	MOD_LFLAGS	+= -undefined dynamic_lookup
endif
```

adding this seems to help:


```
   if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
      set_target_properties(${mod} PROPERTIES
                            LINK_FLAGS "-undefined dynamic_lookup")
    endif()
```

PR #1964, opened 2022-07-16
Revert PR #1922

juha-h:
Revert PR #1922 due to serious issues with early media and perhaps something else too.


juha-h:
Thanks for the merge.  If there was some bug in https://github.com/baresip/baresip/pull/1922 that can now be fixed and tested before new merge.

PR #1973, opened 2022-07-18
Improve RFC 3262

maximilianfridrich:
This PR improves early media confirmed handling by making use of the `prack_handler` in re.

Now the early media session is considered confirmed after a valid PRACK or successful response to a PRACK is received which aligns better with the RFC.

Related:
- https://github.com/baresip/re/pull/450


cspiel1:
We try to reproduce this failure from the workflows:
```
./selftest
free(): invalid next size (fast)
make: *** [Makefile:335: test] Aborted (core dumped)
Error: Process completed with exit code 2.
```

maximilianfridrich:
There is no reason this should fail. All checks passed succesfully and then I [added some whitespace](https://github.com/baresip/baresip/compare/8df0844469dc9a89bf55fe1abe55213b233b8f12..0dba6ee7414cf41abc12101ba195dabc407bedb0). Not sure what happened there and I can't reproduce it. Can you trigger a re-run of the tests? @cspiel1

cspiel1:
So, this is ready for merge. First the `re` PR.

PR #1974, opened 2022-07-18
audio: remove unused aubuf for decoding

cspiel1:
The aubuf for decoding and the rx_filter_thread was introduced in order to
combine the benefits of an adaptive audio buffer with an AEC as audio filter.
If a sound daemon like pulseaudio is used, the AEC should be integrated into
this daemon. But if baresip directly connects to ALSA the aubuf between the
filters and the audio player has to be fixed and as short as possible.

We think that the rx_filter_thread with the decoding aubuf has not enough
benefits for the lots of extra code that was necessary. Instead one could use
a modified ALSA module which integrates an AEC solution, like we did. Then the
AEC runs in the auplay thread.

So this commit is a cleanup for unused code parts.

---
We don't think that anybody uses the `rx_filter_thread`. But this should only be for discussion for now.

Notes for further cleanup / different solution:
- @sreimers You proposed ones that the filters may ran also in the auplay thread. We could make configurable if the dec-filters are running in the main thread (default) or in the auplay thread. But so far I don't think that there is any need for this, except there is somebody who wants to run an AEC as baresip filter and to have the adaptive aubuf.
- If we let the dec-filters run in the main-thread on all cases, we could also remove the lots of `mtx_lock` around `rx.filtl`. Presumed, that we don't plan another try for the RX RTP real-time thread.

Should we remove also the mutex locks?

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -130,16 +130,6 @@ struct autx {
  |/    |        |   |       |   |        |   |        |
        '--------'   '-------'   '--------'   '--------'
 
-       If an AEC is in the filter list aufilt, then the aubuf should be
-       constant and small. A second audio buffer aubufdec should be added via
-       config if network jitter is expected.
-
-       .--------.   .-------.   .--------.   .----------.   .--------.
- |\    |        |   |       |   |        |   |  opt.    |   |        |
- | |<--| auplay |<--| aubuf |<--| aufilt |<--| aubufdec |<--| decode |<--- RTP
- |/    |        |   |       |   |        |   |          |   |        |
-       '--------'   '-------'   '--------'   '----------'   '--------'
-
  \endverbatim
  */
 struct aurx {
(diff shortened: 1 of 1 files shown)


sreimers:
> * @sreimers You proposed ones that the filters may ran also in the auplay thread. We could make configurable if the dec-filters are running in the main thread (default) or in the auplay thread. But so far I don't think that there is any need for this, except there is somebody who wants to run an AEC as baresip filter and to have the adaptive aubuf.

I currently have also no need for this anymore, we can implement this when we see a use case.

> * If we let the dec-filters run in the main-thread on all cases, we could also remove the lots of `mtx_lock` around `rx.filtl`. Presumed, that we don't plan another try for the RX RTP real-time thread.

I'm unsure, maybe we can leave these for a while.

cspiel1:
Okay, in this case and if no one else needs the `rx_filter_thread` we could merge this.

PR #1975, opened 2022-07-18
Modules cmake

viordash:
for #1971 
small contribution :)
I'll add more as soon as I have free time.


alfredh:
Many thanks, please keep'em coming

PR #1977, opened 2022-07-20
pulse_async: reduce number of reconnect attempts

RobertMi21:


modules/pulse_async/pulse.c
```diff
@@ -72,8 +73,8 @@ static void qh(int id, void *data, void *arg)
 	if (pa.paconn)
 		pa.paconn = mem_deref(pa.paconn);
 
-	if (paconn_start(&pa.paconn))
-		tmr_start(&pa.rc, 1000, reconnth, NULL);
+	pa.retry = 0;
+	tmr_start(&pa.rc, 1000, reconnth, NULL);
```
cspiel1, line 77: Should'nt we try for ever, but increase the delay?
See also: `reconnth()`

RobertMi21: Add delay increase between 800ms and 2000ms

cspiel1: Increasing from 800ms to 2000ms has not much benefit. Tomorrow, we try a good delay and maybe we will retry for ever.

Maybe we should add a `module_event()` that informs the application that there had been `n` retries to connect to pulseaudio.

modules/pulse_async/pulse.c
```diff
@@ -205,7 +216,9 @@ static int paconn_start(struct paconn_st **ppaconn)
  */
 static int pa_start(void)
 {
-	int err = 0;
+	int err;
+
+	pa.retry = 0;
```
cspiel1, line 222: remove the new line

modules/pulse_async/pulse.c
```diff
@@ -24,6 +24,7 @@
  * the pulseaudio async interface.
  */
 
+#define RECONN_DELAY 1500
```
cspiel1, line 27: baresip uses enum instead of define for integer constants

PR #1979, opened 2022-07-21
audio: always start reading in TX thread

cspiel1:
This fixes a major bug with setting:
```
audio_txmode		thread
```
If microphone frames are smaller than `tx->psize`, the aubuf never returned frames.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -516,6 +516,7 @@ static void poll_aubuf_tx(struct audio *a)
 	struct le *le;
 	uint32_t srate;
 	uint8_t ch;
+	bool underrun = false;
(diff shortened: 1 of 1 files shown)


cspiel1:
Thanks, with POLL mode there seems to be the same problem when looking in the code:
```c
		if (aubuf_cur_size(tx->aubuf) < tx->psize)
			break;

		poll_aubuf_tx(a);
```

cspiel1:
It issue occurred in combination with `pulse_async`. Because the psize was not reached exactly during start of the TX stream.

Here is the PR for the POLL mode. https://github.com/baresip/baresip/pull/1980

PR #1982, opened 2022-07-21
Improve re_atomic handling

sreimers:
Since we can not rely on C11 `_Atomic` (`RE_ATOMIC`) keyword only (no C99 and MSVC support). We have to be explicit about atomic usage. This is also needed for C11 too, since _Atomic has implicit sequentially-consistent memory ordering by default. On weak memory ordering architectures like ARM these needs extra memory barriers and can lead to performance drops if relaxed ordering is enough.

We have to be careful about `re_atomic_weak` (relaxed memory) usage, but for `boolean run` this should be fine and is lock and data-race free. Also with explicit definitions its much fewer code than mutex locking.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -262,8 +262,9 @@ static void stop_tx(struct autx *tx, struct audio *a)
 	if (!tx || !a)
 		return;
 
-	if (a->cfg.txmode == AUDIO_MODE_THREAD && tx->thr.run) {
-		tx->thr.run = false;
+	if (a->cfg.txmode == AUDIO_MODE_THREAD &&
+	    re_atomic_rlx(&tx->thr.run)) {
+		re_atomic_rlx_set(&tx->thr.run, false);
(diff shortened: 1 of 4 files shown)


src/audio.c
```diff
@@ -1203,7 +1204,7 @@ static int tx_thread(void *arg)
 	uint64_t ts = 0;
 
 	mtx_lock(tx->mtx);
-	while (tx->thr.run) {
+	while (re_atomic_weak(&tx->thr.run)) {
```
sreimers, line 1207: Maybe we can replace this one with stronger memory ordering and can remove some mutex locks.

alfredh:
there is a build error in the CI:

```
src/audio.c:266:6: error: implicit declaration of function ‘re_atomic_weak’; did you mean ‘re_atomic_seq’? [-Werror=implicit-function-declaration]
```

Issue #1963, opened 2022-07-16
no early audio with latest main


juha-h:
An Android user complained about audio problems at the start of a call to a PSTN number and I started to wonder if it might be early audio related.

I have an app that, when called, first responds with 183 Session Progress and then plays an early audio message.  After the message has been played, it responds with 408 Request Timeout.

Call to this app works fine with baresip/re/rem from May 14 (first log below).  But when I make the same call with baresip/re/rem from today, I don't hear the early audio message (second log below) and at the end of the call baresip tells that it didn't receive any audio messages.

Could this be related to https://github.com/baresip/baresip/issues/1957?

------------------ first log
```
call: connecting to 'sip:foo@ea.test.tutpro.com'..
call id: 0106864b7180731c

#
TCP 192.168.183.215:36966 -> 192.168.183.215:5060
INVITE sip:foo@ea.test.tutpro.com SIP/2.0
Via: SIP/2.0/TCP 192.168.183.215:45559;branch=z9hG4bKed4b88a2b6b58275;rport
Contact: <sip:test@test.tutpro.com;gr=urn:uuid:9b21c1a8-dc58-a3d6-83d5-7be67cbe70ce>
Max-Forwards: 70
Proxy-Authorization: Digest username="test", realm="test.tutpro.com", nonce="YtKXfWLSllGH1GMtGv8kou6r5K4z5U8wsLYThSV1gUV/ecdK9s8MQGqs9q+A", uri="sip:foo@ea.test.tutpro.com", response="a93607027687adc9f2886b88dbebf586", cnonce="5c93b6f0aa9a6c3d", qop=auth, nc=00000001, algorithm=MD5
Route: <sip:siika.tutpro.com:5060;transport=tcp;lr>
To: <sip:foo@ea.test.tutpro.com>
From: "Test Account" <sip:test@test.tutpro.com>;tag=e31e5c2863510360
Call-ID: 0106864b7180731c
CSeq: 60735 INVITE
User-Agent: baresip v2.3.0 (x86_64/linux)
Allow: INVITE,ACK,BYE,CANCEL,OPTIONS,NOTIFY,SUBSCRIBE,INFO,MESSAGE,REFER
Supported: gruu,path,outbound,replaces
Content-Type: application/sdp
Content-Length: 754

v=0
o=- 1858384871 1231713726 IN IP4 192.168.183.215
s=-
c=IN IP4 192.168.183.215
t=0 0
[...]

Issue #1957, opened 2022-07-14
No free-line signal since baresip v2.5.0


robert-scheck:
When calling somebody (outside of the own PBX), there is no free-line signal since baresip v2.5.0 anymore (baresip v2.5.0, re v2.5.0, rem v2.5.0). With "free-line signal", I mean "call-connected signal", so the classic POTS sound sequence that you normally hear as a caller while the phone is ringing at the callee. The last version known to work properly is baresip v2.4.0 (baresip v2.4.0, re v2.4.0, rem v2.4.0).

Any pointers?

sreimers:
On a quick guess, maybe this PR is related: https://github.com/baresip/baresip/pull/1922

robert-scheck:
Yes, good catch! Reverting commit 28fa0176148381f2e79ec0d310371436a57cdea1 makes it working again.

sreimers:
@RobertMi21 Can you check this?

RobertMi21:
Sorry for the troubles with my changes. I did not notice that early audio and the free-line signal did not work after my changes. I'll try to find a working fix for #1922 .

robert-scheck:
With "free-line signal", I mean "call-connected signal", so the classic POTS sound sequence that you normally hear as a caller while the phone is ringing at the callee. Or in German "Freiton" aka "Freizeichen"; Wikipedia says this is also called "ring-back tone" in English. And yes, #1963 is a duplicate of this one.

PR #1986, opened 2022-07-22
call: do not stop streams on session progress

cspiel1:
This should be the working replacement for PR https://github.com/baresip/baresip/pull/1922 which caused these issues (which are almost the same):
- https://github.com/baresip/baresip/issues/1957
- https://github.com/baresip/baresip/issues/1963

If a UAS sends multiple 183 Session Progress with unchanged SDP, then the streams where stopped and restarted always. This PR fixes this behavior.

The problem with the PR https://github.com/baresip/baresip/pull/1922 was that `call_stream_start()` was missing, or more concretely `stream_enable()` which is needed for processing the RX stream. E.g. for the "free-line signal", which is early-audio from callee (or UAS) to the caller.

@robert-scheck, @juha-h  : please could you test this version?

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -2260,18 +2260,15 @@ static void sipsess_progr_handler(const struct sip_msg *msg, void *arg)
 		break;
 	}
 
-	call_stream_stop(call);
-
-	if (media)
-		call_stream_start(call, false);
-
-	if (media)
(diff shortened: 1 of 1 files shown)


juha-h:
I tested using the same early audio app as in https://github.com/baresip/baresip/issues/1963 and now the call worked fine.

PR #1989, opened 2022-07-23
event: add on hold info

copiltembel:
This PR:

Adds the call.on_hold information to the event dictionary

### DIFF
--- a/src/event.c
+++ b/src/event.c
@@ -215,10 +215,9 @@ int event_encode_dict(struct odict *od, struct ua *ua, enum ua_event ev,
 		if (user_data) {
 			err |= odict_entry_add(od, "userdata", ODICT_STRING,
 				user_data);
+			if (err)
+				goto out;
 		}
-
-		if (err)
-			goto out;
 	}
 
 	if (str_isset(prm)) {
--- a/src/event.c
+++ b/src/event.c
@@ -208,7 +208,7 @@ int event_encode_dict(struct odict *od, struct ua *ua, enum ua_event ev,
 			err |= odict_entry_add(od, "diverteruri", ODICT_STRING,
 					       call_diverteruri(call));
 
-		err |= odict_entry_add(od, "onhold", ODICT_BOOL, 
+		err |= odict_entry_add(od, "onhold", ODICT_BOOL,
 				call_is_onhold(call));
 
 		const char *user_data = call_user_data(call);
--- a/src/event.c
+++ b/src/event.c
@@ -208,13 +208,17 @@ int event_encode_dict(struct odict *od, struct ua *ua, enum ua_event ev,
 			err |= odict_entry_add(od, "diverteruri", ODICT_STRING,
 					       call_diverteruri(call));
 
+		err |= odict_entry_add(od, "onhold", ODICT_BOOL, 
+				call_is_onhold(call));
+
 		const char *user_data = call_user_data(call);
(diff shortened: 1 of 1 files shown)


src/event.c
```diff
@@ -208,13 +208,17 @@ int event_encode_dict(struct odict *od, struct ua *ua, enum ua_event ev,
 			err |= odict_entry_add(od, "diverteruri", ODICT_STRING,
 					       call_diverteruri(call));
 
+		err |= odict_entry_add(od, "onhold", ODICT_BOOL,
+				call_is_onhold(call));
+
```
cspiel1, line 213: This is only the local on-hold flag. The application should know that it put the call on hold locally. Is this worth to be put to the event?
Should we add it only if `call_is_onhold()`?

src/event.c
```diff
@@ -208,13 +208,17 @@ int event_encode_dict(struct odict *od, struct ua *ua, enum ua_event ev,
...
 		const char *user_data = call_user_data(call);
 		if (user_data) {
 			err |= odict_entry_add(od, "userdata", ODICT_STRING,
 				user_data);
-			if (err)
-				goto out;
 		}
+
+		if (err)
+			goto out;
```
cspiel1, line 221: I think this should be at least a separate commit. Normally we make separate PRs, for different fixes/features.

juha-h:
Christian Spielberger writes:

> This is only the local on-hold flag. The application should know that it put the call on hold locally. Is this worth to be put to the event?
> Should we add it only if `call_is_onhold()`?

In my Android app, call has two properties: call.onhold and call.held.
So both are needed.

src/event.c
copiltembel: OK, I've moved this change to a separate PR: https://github.com/baresip/baresip/pull/1992

src/event.c
copiltembel: For me it is important to know if a call has been put on hold, regardless if it was triggered by the local client or the remote party. I am using the mqtt module to get updates about the call states and HOLD is something I am interested in.

If the remote party sets the call to HOLD, the call_is_onhold is not updated?

sreimers:
You can rebase/merge main to fix clang ci issues.

juha-h:
copiltembel writes:

> For me it is important to know if a call has been put on hold, regardless if it was triggered by the local client or the remote party. I am using the mqtt module to get updates about the call states and HOLD is something I am interested in.
> 
> If the remote party sets the call to HOLD, the call_is_onhold is not
> updated?

Haven't checked, but a call is on hold when local party has put the call
on hold and the call is held when remote party has put the call on hold.

When call is put on hold by one party of the call, re-INVITE has 

a=sendonly

and 200 OK has

a=recvonly

If also the other party puts the call on hold, 200 OK has

a=inactive

So two separate checks would be needed: is call on hold (i.e.. put on
hold by the local party) and is call held (i.e., put on hold by the
remote party).

copiltembel:
> So two separate checks would be needed: is call on hold (i.e.. put on
> hold by the local party) and is call held (i.e., put on hold by the
> remote party).

That makes sense! So we need a second field on the call that keeps track of the "is call held" status.

cspiel1:
If a re-INVITE has `a=sendonly`, then "the remote put us on-hold" is only an interpretation. This could be done by the application. The audio directions are already encoded to the event data.

PR #1988, opened 2022-07-23
call: fix heap-buffer-overflow in prack_handler

sreimers:
We had some strange selftest CI errors sometimes, looks like these are related to this heap-buffer-overflow:

SUMMARY: AddressSanitizer: heap-buffer-overflow src/call.c:2024:25 in prack_handler

Found by:
`make CC=clang EXTRA_CFLAGS="-fsanitize=address" EXTRA_LFLAGS="-fsanitize=address"`

Related: https://github.com/baresip/baresip/pull/1973#issuecomment-1190151908

CC: @cspiel1 @maximilianfridrich

PR #1995, opened 2022-07-24
Added mwi module cmake build

juha-h:


juha-h:
I have no clue why MinGW-w64 Test failed.

sreimers:
Yes nothing todo with this PR, we have still a heap-buffer-overflow in prack_handler. #1988

alfredh:
Perhaps run selftest with verbose logging so we can see which test failed:

```
./selftest -v
```

juha-h:
Can't find selftest program, but looks like the problem is now solved, since all checks have passed.

PR #1997, opened 2022-07-24
ci: test re/rem with cmake

alfredh:


.github/workflows/cmake.yml
```diff
@@ -48,7 +53,13 @@ jobs:
...
     - name: make re/rem
       run: |
-        for p in re rem; do cd $p && sudo PATH="$PATH" make install && cd - && sudo rm -Rf $p; done
+        for p in re rem; do
+          cmake -S $p -B $p/build
+          cmake --build $p/build
+        done
+        for p in re rem; do
+          mv $p ../.
+        done
```
alfredh, line 62: this one changes building of re/rem from make to cmake

the last step is a clever trick to move re/rem one level up, so that baresip can find them

please review if this is a good idea (please review this hunk)

CMakeLists.txt
```diff
@@ -80,6 +80,7 @@ if(USE_OPENSSL)
     -DUSE_OPENSSL_SRTP
     -DUSE_TLS
   )
+  include_directories( ${OPENSSL_INCLUDE_DIR} )
```
alfredh, line 83: looks like this one is needed for MacOS, other building fails because it cannot
find openssl headers ...

Issue #1996, opened 2022-07-24
heap-buffer-overflow, in call prack_handler


sreimers:
Looks like `call->early_confirmed` writes to a dereferenced object (dangling `call` pointer).
```
[ RUN      ] test_call_answer_hangup_a
==175531==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x615000004b5c at pc 0x55eb612f853f bp 0x7ffcacefa580 sp 0x7ffcacefa578
WRITE of size 1 at 0x615000004b5c thread T0
    #0 0x55eb612f853e in prack_handler baresip/src/call.c:2027:25
    #1 0x55eb6146467a in resp_handler re/src/sipsess/prack.c:84:3
    #2 0x55eb6143b10f in terminate re/src/sip/request.c:107:3
    #3 0x55eb6143a698 in response_handler re/src/sip/request.c:156:2
    #4 0x55eb6152b04c in response_handler re/src/sip/ctrans.c:292:4
    #5 0x55eb614465c0 in sip_recv re/src/sip/transp.c:357:7
    #6 0x55eb614407ef in udp_recv_handler re/src/sip/transp.c:431:2
    #7 0x55eb614d7507 in udp_read re/src/udp/udp.c:244:2
    #8 0x55eb614d5d5e in udp_read_handler re/src/udp/udp.c:257:2
    #9 0x55eb6150ca61 in fd_poll re/src/main/main.c:916:4
    #10 0x55eb6150abda in re_main re/src/main/main.c:1096:9
    #11 0x55eb612cc576 in re_main_timeout baresip/test/test.c:39:2
    #12 0x55eb6128c2e6 in test_call_answer_hangup_a baresip/test/call.c:518:8
    #13 0x55eb612cd969 in run_tests baresip/test/main.c:100:9
    #14 0x55eb612cd310 in main baresip/test/main.c:276:9
    #15 0x7f850803228f  (/usr/lib/libc.so.6+0x2928f) (BuildId: 60df1df31f02a7b23da83e8ef923359885b81492)
    #16 0x7f8508032349 in __libc_start_main (/usr/lib/libc.so.6+0x29349) (BuildId: 60df1df31f02a7b23da83e8ef923359885b81492)
    #17 0x55eb6118b4a4 in _start /build/glibc/src/glibc/csu/../sysdeps/x86_64/start.S:115

0x615000004b5c is located 36 bytes to the left of 488-byte region [0x615000004b80,0x615000004d68)
[...]

PR #1999, opened 2022-07-25
Fix prack handling

maximilianfridrich:
Re-enable prack handler after the source of issue #1996 has been fixed in https://github.com/baresip/re/pull/457.


sreimers:
Thanks!

PR #2000, opened 2022-07-25
cmake: add ctrl_dbus module

cspiel1:


cspiel1:
`cmakelint` is fine. But what about `ccheck` for the cmake build?

cspiel1:
Next problem: `gdbus-codegen` is missing on Mac OS.

modules/ctrl_dbus/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
+project(ctrl_dbus)
+
+set(SRCS ctrl_dbus.c baresipbus.c)
+
+find_package(PkgConfig REQUIRED)
+
+pkg_check_modules(GLIB REQUIRED glib-2.0>=0.29.1)
+pkg_check_modules(GIO  REQUIRED gio-unix-2.0>=0.29.1)
```
sreimers, line 8: These checks should be moved to a separat FindGLIB.cmake module in the `cmake` folder, so the main CMakeLists.txt includes `ctrl_dbus` only if requirements are met (like AAC).

cspiel1: okay, thanks!

cspiel1: I added a minimalist `FindGLIB.cmake`.

Do we need the other find methods like in `FindAAC.cmake` and the very first `FindOPUS.cmake` ? This seems to be a manual method. Modul `PkgConfig` should do all this alreaydy.

cspiel1:
What about the `if(NOT WIN32)` in `FindOPUS.cmake` around the PkgConfig? Other `Find...` don't have this.

cmake/modules.cmake
```diff
@@ -67,6 +68,9 @@ if(FFMPEG_FOUND)
   list(APPEND MODULES avformat)
   list(APPEND MODULES swscale)
 endif()
+if(GIO_FOUND)
+  add_subdirectory(modules/ctrl_dbus)
```
sreimers, line 72: This should be `list(APPEND MODULES ctrl_dbus)` now, sorry for the inconvenience, but this way we can build modules optional.

modules/ctrl_dbus/CMakeLists.txt
cspiel1: Answer myself:
The cmake manual suggest to add more checks. I can't understand why this is needed.
https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html

Anyway, added these checks like in the other cmake/Find... modules.

cspiel1:
Next steps:
- Fix in cmake: ctrl_dbus should be built only if the tool `gdbus-codegen` is available.
- Fix the workflow: Add a `brew install glib-utils` anywhere? Currently struggling with this.

cspiel1:
With the newest commit I think the build will succeed, but the `ctrl_dbus` will not be built. After the cmake.yml was extended with the `install packages` job, we should check if ctrl_dbus is built. Then I will mark this PR as "ready".

sreimers:
Ok merged: e4658723645ee3384e1c6f8730504924d7c41b0d

cspiel1:
`ctrl_dbus` is built now for normal and static cmake build on macOS.

From my side it is ready for merge. This PR has many commits, so pls don't forget to use "Squash and merge". I suggest also to simplify the commit message. ;-)

modules/ctrl_dbus/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+    --generate-c-code baresipbus --c-namespace DBus
+    --interface-prefix com.github.
+    ${CMAKE_SOURCE_DIR}/modules/${PROJECT_NAME}/com.github.Baresip.xml
+  DEPENDS com.github.Baresip.xml
+  BYPRODUCTS baresipbus.c baresipbus.h)
+
+set_target_properties(${PROJECT_NAME} PROPERTIES PREFIX "")
+
+list(APPEND MODULES ${PROJECT_NAME})
+set(MODULES ${MODULES} PARENT_SCOPE)
```
sreimers, line 39: These lines can be removed now (append is done in modules.cmake):

```
list(APPEND MODULES ${PROJECT_NAME})
set(MODULES ${MODULES} PARENT_SCOPE)
```

cspiel1: Should I remove also `set_target_properties(${PROJECT_NAME} PROPERTIES PREFIX "")`?

sreimers: Prefix setting is still needed (cmake puts a `lib` prefix by default)

sreimers: Wait I will add a global one in `CMakeLists.txt`. Will create a PR. Nothing todo for you.

PR #2001, opened 2022-07-25
cmake: add generated CMakeFiles to .gitignore

cspiel1:


cspiel1:
cmake should not be called in the source directory. Instead in any sub-directory. Then the CMakeFiles will be created also in this sub-directory and it is not necessary to have it in .gitignore.

PR #2006, opened 2022-07-26
Added cmake of zrtp module

juha-h:
Don't know why this PS cannot be automatically merged.  Solved merge conflict.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,7 @@ find_package(FFMPEG COMPONENTS avcodec avfilter avformat swscale)
 find_package(OPUS)
 find_package(PULSE)
 find_package(X11)
+find_package(ZRTP)
 
 
 ##############################################################################
@@ -205,6 +206,9 @@ if(PULSE_FOUND)
 endif()
 if(X11_FOUND)
   add_subdirectory(modules/x11)
+endif()
+if(ZRTP_FOUND)
+  add_subdirectory(modules/zrtp)
 endif()
 
 if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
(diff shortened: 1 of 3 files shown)


alfredh:
many thanks,

Currently we have two ZRTP modules:

- zrtp.so
- gzrtp.so

Perhaps it is enough to have one zrtp module ?

juha-h:
Alfred E. Heggestad writes:

> Currently we have two ZRTP modules:
> 
> - zrtp.so
> - gzrtp.so
> 
> Perhaps it is enough to have one zrtp module ?

gzrtp may have some benefits over zrtp, but is currently lacking handling
of several events.

alfredh:
okay, I think we should try to compare them and select one of them to maintain.
the other one will then be deleted.

the most important thing is who is using it, how many are using it,
and the license and support from the underlying library.

juha-h:
Alfred E. Heggestad writes:

> okay, I think we should try to compare them and select one of them to
> maintain. the other one will then be deleted.

zrtp module is based on FreeSwitch libzrtp that I have forked and made
some minor fixes.  There has been no bug fixes in the FreeSwitch lib
after 2016.

gzrtp module is based on zrtp lib by Werner Dittmann and is more
actively maintained.  So most likely, it would be the best choice for
baresip zrtp module.

However for some reason the developer of gzrtp module has not included
calls to event event handlers that are needed in order to make zrtp
module useful in non-cli baresip apps, like in my Android apps.

> the most important thing is who is using it, how many are using it,
> and the license and support from the underlying library.

gzrtp module is written in c++ that I don't know anything about.  For
that reason it would be hard for me to add the missing stuff to gzrtp
module.

PR #2015, opened 2022-07-27
Added cmake of g722 module

juha-h:


CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ find_package(FFMPEG COMPONENTS
     avutil avcodec avfilter avformat swscale swresample avdevice)
 find_package(OPUS)
 find_package(PULSE)
+find_package(SPANDSP COMPONENTS g722)
```
sreimers, line 39: Looks like here is also no component support in FindSPANDSP.

PR #2016, opened 2022-07-27
Added cmake of vp8 and vp9 modules

juha-h:


CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ find_package(FFMPEG COMPONENTS
     avutil avcodec avfilter avformat swscale swresample avdevice)
 find_package(OPUS)
 find_package(PULSE)
+find_package(VPX COMPONENTS vp8 vp9)
```
sreimers, line 39: Looks like FindVPX.cmake has no component support I think. So `find_package(VPX)` should be enough.

sreimers (approved):

PR #2029, opened 2022-07-29
[WIP] ci: test windows and cmake

alfredh:


alfredh:
it currently fails with this error:

```
Could NOT find RE (missing: RE_LIBRARY)
```

sreimers:
I think i had the same error with `retest`, can you try this:

https://github.com/baresip/retest/commit/f47d505f8ffafa9b2202eb38219ea844e3a8dc32

alfredh:
thanks, that helped

it got a bit further, now it is failing here:

```
D:\a\baresip\baresip\modules\ausine\ausine.c(8,10): fatal error C1083: Cannot open include file: 'pthread.h': No such file or directory [D:\a\baresip\baresip\build\modules\ausine\ausine.vcxproj]
```

there are also some other minor changes needed, for example the "stdio" module
should not be enabled for Windows platform.

alfredh:
next error:


```
dshow.cpp
D:\a\baresip\rem\include\rem_aubuf.h(31,3): error C7555: use of designated initializers requires at least '/std:c++20' 
D:\a\baresip\rem\include\rem_aubuf.h(53,3): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\baresip\baresip\build\modules\dshow\dshow.vcxproj]
D:\a\baresip\rem\include\rem_aubuf.h(68,3): error C7555: use of designated initializers requires at least '/std:c++20' [D:\a\baresip\baresip\build\modules\dshow\dshow.vcxproj]
```


here is the code:

```c
static inline int aubuf_write(struct aubuf *ab, const uint8_t *p, size_t sz)
{
        struct auframe af = {
                .fmt = AUFMT_RAW,
                .srate = 0,
                .sampv = (uint8_t *)p,
                .sampc = sz,
                .timestamp = 0,
                .level = AULEVEL_UNDEF
        };

        return aubuf_write_auframe(ab, &af);
}
```


this can be solved by changing the function from inline to non-inline

sreimers:
Since `aubuf` is not used by dshow, we can remove the `#include <rem.h>` and include only needed/explicit rem headers?

```c
#include <re.h>
#include <rem_au.h>
#include <baresip.h>
```

alfredh:
some new warnings from the Windows compiler:


```
D:\a\baresip\baresip\src\audio.c(1437,46): warning C4267: 'function': conversion from 'size_t' to 'uint16_t', possible loss of data [D:\a\baresip\baresip\build\baresip.vcxproj]
D:\a\baresip\baresip\src\audio.c(1438,46): warning C4267: 'function': conversion from 'size_t' to 'uint16_t', possible loss of data [D:\a\baresip\baresip\build\baresip.vcxproj]
```

alfredh:
Current error:

```
D:\a\baresip\baresip\src\config.c(7,10): fatal error C1083: Cannot open include file: 'dirent.h': No such file or directory [D:\a\baresip\baresip\build\baresip.vcxproj]
```


in config.c we use `opendir` to count the number of modules.
it is possible to port this function to use native Win32 API.

PR #2034, opened 2022-07-30
gsm: remove deprecated module

alfredh:
proposal to remove gsm.so

The GSM codec is now quite old and there exist better alternatives such as AMR and AMR-WB.


ivorget:
A little late to this but imho it's a little early to be have removed GSM codec support. I accept AMR is a better codec but AFAIK it isn't yet supported in default config in Freeswitch and Asterisk (& likely others) leading to little operator support for it. Also most popular android softphones do not seem to support it.

In my case, my SIP provider using Freeswitch only provides G.729 and GSM as their high compression codecs so not having GSM makes baresip a non-runner for me when making sip calls to landline over mobile data.

That said, the pre-android 12 default SIP dialer supports both AMR and GSM so if no one else cares about it's removal from baresip then I guess it's no biggie.

juha-h:
My baresip Android app supports both AMR and G.729 codecs.

ivorget:
Sorry should have explained, G729 sounded kinda nasty when I tested it - my provider may have been sending me the lower quality G729A like many do. Besides unlike GSM, G729 doesn't handle background noise nor packet loss as well.

When I check around popular providers like voip.ms, GSM seems to be the next most commonly supported compression codec after G729/G729A. That's probably my biggest gripe with it's removal.

It's your prerogative though, I can focus more on linphone if you guys prefer to keep it out.

juha-h:
Latest version (48.1.0) of my Android baresip app now includes gsm codec.  Takes a few days to appear in F-Droid/Play Store.

ivorget:
Great! Thanks for reconsidering.

However maybe one of the reverts was missed because 48.1.0 is already in the play store but GSM doesn't show in it's audio settings codec list.

juha-h:
ivorget writes:

> However maybe one of the reverts was missed because 48.1.0 is already
> in the play store but GSM doesn't show in it's audio settings codec
> list.

I created 48.1.0 today.  Earlier version already in the store should be
48.0.1.

ivorget:
I double-checked, it's definitely 48.1.0 from today.

juha-h:
ivorget writes:

> I double-checked, it's definitely 48.1.0 from today.

That was fast.  I just installed 48.1.0 from Play Store and it does have
GSM codec.  You need to activate it in Account's Audio Codecs list.

ivorget:
Ah OK, I'm afraid I was completely misled by the main audio settings module list which I thought was the complete list of built in codecs so tbh I had barely even tested baresip after that. You maybe should consider adding some clarifying text there in case others are similarly misled.

The other main feature I'd been looking for is AMR-WB and I'd seen somewhere you support that so I was thinking about recompiling to add it in - but now I see you have AMR-WB included by default which is great.

ivorget:
I misread it that is, probably not fair to say I was misled unless I find a lot of misreaders.

Issue #2042, opened 2022-07-31
audio call is automatically modified to audio-video call after 200 OK is received


juha-h:
Using latest main, I noticed the following with my baresip+ app:

- baresip+ sends audio only INVITE to Linux baresip
- after receiving audio only 200 OK, I get to log:
```
07-31 10:58:47.924 21275 21347 D Baresip+ Lib: call: got SDP answer (472 bytes)
07-31 10:58:47.945 21275 21346 D Baresip+ Lib: ua event (CALL_REMOTE_SDP) answer
07-31 10:58:47.946 21275 21346 D Baresip+ Lib: video RTP/AVP
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:   local formats:
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:      96 VP9/90000/1 () *
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:      97 VP8/90000/1 () *
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:      98 H264/90000/1 () *
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:      99 H264/90000/1 () *
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:      34 H263/90000/1 () *
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:     100 H265/90000/1 (profile-id=1) *
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:     101 AV1/90000/1 () *
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:   remote formats:
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:   local attributes:
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:     rtcp-rsize
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:     ssrc='132011099 cname:sip:bar@test.tutpro.com'
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:     mid='1'
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:     framerate='15.00'
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:     rtcp-fb='* nack pli'
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:     content='main'
07-31 10:58:47.946 21275 21346 D Baresip+ Lib:     label='2'
[...]

PR #2046, opened 2022-08-03
call: remember media dir for established state

cspiel1:
Fixes issue #2042
- The function call_set_media_direction() should remember the audio/video dir
  for the established state
- Rename call_set_media_ansdir() to call_set_media_estdir(). This function sets
  the audio/video dir for the established state, before the call is established


cspiel1:
Tested:
- @juha-h with the Android App
- Me with early-video

Renames an API function.

PR #2050, opened 2022-08-03
play: ring tone fixes if file_ausrc is set

cspiel1:
- aufile: use aubuf_read_auframe()
- play: use new aubuf API and don't start auplay too early


cspiel1:
There is another problem with same setup. The ringing of an incoming call stops unexpectedly with:
```
main: thread check: called from a NON-RE thread without thread_enter()!
```
So converted to draft. Maybe the fix fits to the above somehow. I have to analyze.

PR #2061, opened 2022-08-08
play: tmr_polling has to check if ausrc is used

cspiel1:


sreimers:
Should we add a sanity check in `start_ausrc` too?

sreimers:
Thanks!

PR #2066, opened 2022-08-11
FindAMR.cmake fixes/improvements

juha-h:


juha-h:
I have now tested this on Debian 11 with each of the packages `libopencore-amrnb-dev`, `libopencore-amrwb-dev`, and `libvo-amrwbenc-dev` installed/removed.

juha-h:
Forgot to mention that there are some debug message() calls that could be removed.  Can be done later when the build has been tested on other platforms.

PR #2067, opened 2022-08-12
hook up webrtc to main cmake file

alfredh:


webrtc/src/main.c
```diff
@@ -17,9 +17,9 @@
 #include <re_dbg.h>
 
 
-static const char *modpath = "/usr/local/lib/baresip/modules";
+static const char *modpath = "/usr/local/share/baresip";
```
alfredh, line 20: I noticed that the modules path has changed for cmake install. is this correct ?

sreimers: Yes I think we should fix this: https://github.com/baresip/baresip/pull/2068

PR #2080, opened 2022-08-22
test/call: Add test_call_change_videodir

maximilianfridrich:
In response to #2073.

### DIFF
--- a/test/call.c
+++ b/test/call.c
@@ -70,6 +70,7 @@ struct fixture {
 	bool stop_on_rtp;
 	bool stop_on_rtcp;
 	bool stop_on_audio_video;
+	bool accept_session_updates;
 };
 
 
@@ -365,6 +366,17 @@ static void event_handler(struct ua *ua, enum ua_event ev,
 
 		break;
 
+	case UA_EVENT_CALL_REMOTE_SDP:
+		if (f->accept_session_updates &&
+		    call_state(call) == CALL_STATE_ESTABLISHED &&
+		    !call_is_outgoing(call) &&
+		    sdp_media_ldir(stream_sdpmedia(video_strm(call_video(
+				call)))) == SDP_SENDRECV &&
+		    sdp_media_rdir(stream_sdpmedia(video_strm(call_video(
+				call)))) == SDP_INACTIVE) {
+			re_cancel();
+		}
+		break;
 	case UA_EVENT_CALL_MENC:
 		++ag->n_mediaenc;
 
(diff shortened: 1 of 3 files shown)


test/call.c
```diff
@@ -884,6 +898,104 @@ int test_call_video(void)
...
+	ASSERT_TRUE(a_video_rdir == SDP_SENDRECV);
+	ASSERT_TRUE(b_video_ldir == SDP_SENDRECV);
+	ASSERT_TRUE(b_video_rdir == SDP_INACTIVE);
+
+	/* Set video sendrecv */
+	err = call_set_video_dir(ua_call(f->a.ua), SDP_SENDRECV);
+	TEST_ERR(err);
+	err = re_main_timeout(10000);
+	err |= re_main_timeout(10000);
+	err |= re_main_timeout(10000);
```
cspiel1, line 969: The `if` in the `event_handler()` should be extended that these multiple `re_main_timeout()` become one.

test/call.c
```diff
@@ -365,6 +366,17 @@ static void event_handler(struct ua *ua, enum ua_event ev,
...
 		break;
 
+	case UA_EVENT_CALL_REMOTE_SDP:
+		if (f->accept_session_updates &&
+		    call_state(call) == CALL_STATE_ESTABLISHED &&
+		    !call_is_outgoing(call) &&
+		    sdp_media_ldir(stream_sdpmedia(video_strm(call_video(
+				call)))) == SDP_SENDRECV &&
+		    sdp_media_rdir(stream_sdpmedia(video_strm(call_video(
+				call)))) == SDP_INACTIVE) {
```
cspiel1, line 376: Maybe we could use a "state" variable set in the test-routine instead of the long if condition.

test/call.c
```diff
@@ -412,6 +424,20 @@ static void event_handler(struct ua *ua, enum ua_event ev,
...
+		    !call_is_outgoing(call) &&
+		    0 == str_casecmp(prm, "video")&&
+		    sdp_media_ldir(stream_sdpmedia(video_strm(call_video(
+				ua_call(f->a.ua))))) == SDP_SENDRECV &&
+		    sdp_media_rdir(stream_sdpmedia(video_strm(call_video(
+				ua_call(f->a.ua))))) == SDP_SENDRECV &&
+		    sdp_media_ldir(stream_sdpmedia(video_strm(call_video(
+				ua_call(f->b.ua))))) == SDP_SENDRECV &&
+		    sdp_media_rdir(stream_sdpmedia(video_strm(call_video(
+				ua_call(f->b.ua))))) == SDP_SENDRECV) {
```
cspiel1, line 438: Same here. A state variable.

test/call.c
```diff
@@ -884,6 +912,102 @@ int test_call_video(void)
...
+	TEST_ERR(fix.err);
+
+	/* verify that video was enabled for this call */
+	ASSERT_EQ(1, fix.a.n_established);
+	ASSERT_EQ(1, fix.b.n_established);
+
+	ASSERT_TRUE(call_has_video(ua_call(f->a.ua)));
+	ASSERT_TRUE(call_has_video(ua_call(f->b.ua)));
+
+	/* Set video inactive */
```
cspiel1, line 956: Set "state" here to "inactive requested".

test/call.c
```diff
@@ -884,6 +912,102 @@ int test_call_video(void)
...
+			video_strm(call_video(ua_call(f->b.ua)))));
+	b_video_rdir = sdp_media_rdir(stream_sdpmedia(
+			video_strm(call_video(ua_call(f->b.ua)))));
+
+	ASSERT_TRUE(a_video_ldir == SDP_INACTIVE);
+	ASSERT_TRUE(a_video_rdir == SDP_SENDRECV);
+	ASSERT_TRUE(b_video_ldir == SDP_SENDRECV);
+	ASSERT_TRUE(b_video_rdir == SDP_INACTIVE);
+
+	/* Set video sendrecv */
```
cspiel1, line 978: Set "state" here to "sendrecv requested"!

test/call.c
maximilianfridrich: I tried and couldn't get it to work. These events are triggered too often and not when I would expect it.

test/call.c
maximilianfridrich: I tried and couldn't get it to work. These events are triggered too often and not when I would expect it.

test/call.c
maximilianfridrich: See other comment regarding state.

test/call.c
maximilianfridrich: See other comment regarding state.

PR #2092, opened 2022-08-27
Add both Threads and ZLIB to find_package

widgetii:
This fixes Threads package links on my system (cross-compiler arm-linux-musleabi):

```
CMake Error at CMakeLists.txt:196 (target_link_libraries):
  Target "baresip" links to:

    Threads::Threads

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

Because currently it uses `Threads::Threads` in `target_link_libraries` but it never found in `find_package`.
The same idea is for ZLIB, because it seems that clear to use the same way

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,8 @@ include(GNUInstallDirs)
 include(CheckIncludeFile)
 find_package(RE REQUIRED)
 find_package(REM REQUIRED)
+find_package(Threads REQUIRED)
+find_package(ZLIB REQUIRED)
 
 ##############################################################################
 #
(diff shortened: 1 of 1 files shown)


widgetii:
I found it failed on Windows, doesn't it support libz, or it's an optional dependency?

sreimers:
Is your re main up to date? `find_package(Threads REQUIRED)` is imported from `re-config.cmake` now.

widgetii:
> Is your re main up to date? `find_package(Threads REQUIRED)` is imported from `re-config.cmake` now.

Just updated, didn't find `re-config.cmake` in the repo

sreimers:
See: https://github.com/baresip/baresip/pull/2089

widgetii:
Got it, you're using upper directory with re as dependency
`find_package(re CONFIG REQUIRED HINTS ../re/cmake)`

but it seems is not a good idea to hardcode paths used on target machine. Moreover, it breaks flexible build systems like Buildroot because it builds `re`, `rem` and `baresip` components as dependency graph and there is no good way to layout them all together

sreimers:
Its only a hint path, if installed it follows the cmake [Config Mode Search Procedure](https://cmake.org/cmake/help/latest/command/find_package.html?highlight=find_package#id7)

PR #2097, opened 2022-08-28
Add audio driver for HiSilicon IP camera

widgetii:
This adds support for HiSilicon SoC based IP-cameras using their platform SDK to act baresip as intercom


cspiel1:
Isn't there an ALSA driver?
https://patchwork.kernel.org/project/alsa-devel/patch/E1dAAxe-0002Ps-B4@finisterre/

widgetii:
Maybe it's another HiSilicon SoC type, but HiSilicon IPC's don't have ALSA drivers at all (as all IPC manufactures used their own custom kernel and userspace drivers)

alfredh:
the text is a bit misleading, because it says "IP camera" and the patch is actually an audio driver.

this sounds like a module for specially interested people.
perhaps we should start by adding it to baresip-apps first ?

widgetii:
Every modern IP camera has an audio chip (or built-in block inside SoC) which provides two-way audio communication, so IP camera can be used as an intercom directly. 

This PR implements this for HiSilicon-based chips, from my perspective there is no difference between say ALSA audio source and custom hardware vendor APIs

If PR will be accepted (now I'm not sure), I have plans to continue development and eventually add video support

modules/hisilicon/hisi_play.c
```diff
@@ -0,0 +1,157 @@
...
+#include <baresip.h>
+#include <pthread.h>
+
+#include "hisi.h"
+
+#include <mpi_audio.h>
+#include <mpi_sys.h>
+
+struct auplay_st {
+	pthread_t thread;
```
alfredh, line 18: please use C11 threads here

modules/hisilicon/module.mk
```diff
@@ -0,0 +1,13 @@
...
+# Copyright (C) 2010 Alfred E. Heggestad
+#
+
+MOD		:= hisilicon
+$(MOD)_SRCS	+= hisi.c hisi_src.c hisi_play.c
+$(MOD)_LFLAGS	+= -L$(HISI_SDK_DIR)/mpp/lib
+$(MOD)_LFLAGS	+= -lmpi -lVoiceEngine -lsecurec -lupvqe -ldnvqe
+$(MOD)_CFLAGS	+= -I-L$(HISI_SDK_DIR)/mpp/include
+
+include mk/mod.mk
```
alfredh, line 13: given that the old Makefile system will be phased out, it is probably not necessary to add this file

alfredh:
if we want to add this feature/module it should be generic enough that many people are interested in it.
If there is only 1 person using a module, it is probably better to maintain it elsewhere.

We have this repo for baresip apps that are not officially supported:

We have added modules in the past, and the people who submitted them promised to help
with maintainance. But when the work was needed the people was nowhere to be found.

so if we add this module, will you step up as a maintainer for the module, see:

widgetii:
> https://github.com/baresip/baresip/wiki/Modules

No problem for that, this is just it was planned. I'm a core developer of [OpenIPC](http://github.com/openipc/) project, and we decided to bring baresip as main solution for IP cameras to use them as cheap intercoms by everyone who wants to replace original vendor firmware by open sourced one.

modules/hisilicon/hisi_play.c
```diff
@@ -133,14 +133,13 @@ int hisi_play_alloc(struct auplay_st **stp, const struct auplay *ap,
...
+	if (HI_SUCCESS != ret) {
 		printf("error %d\n", ret);
 		return EINVAL;
 	}
 
 	st->run = true;
-	err = thrd_create(&st->thread, write_thread, st);
-	if (err) {
+	if (thrd_success != thrd_create(&st->thread, write_thread, st)) {
 		st->run = false;
```
sreimers, line 143: `err = EAGAIN`, should be set here.

modules/hisilicon/hisi.c
```diff
@@ -0,0 +1,74 @@
...
+
+#include "hisi.h"
+
+#include <mpi_sys.h>
+#include <mpi_vb.h>
+
+static struct ausrc *ausrc;
+static struct auplay *auplay;
+
+static int init_hw(void) {
```
alfredh, line 18: curly bracket on next line:

```c
int foo()
{
}
```

modules/hisilicon/hisi.c
```diff
@@ -0,0 +1,74 @@
...
+	int ret = HI_MPI_SYS_Init();
+
+	if (HI_SUCCESS != ret)
+	{
```
alfredh, line 22: please move curly bracket to previous line:

```c
if (cond) {
}
```

modules/hisilicon/hisi.c
```diff
@@ -0,0 +1,74 @@
...
+		printf("HI_MPI_SYS_Init failed with %d\n", ret);
```
alfredh, line 23: please use `warning()` instead of printf

modules/hisilicon/hisi.c
```diff
@@ -0,0 +1,74 @@
...
+unsigned audio_frame_size(unsigned srate) {
+	switch (srate) {
+		case 8000: return 8000 / 50;
+		case 12000: return 12000 / 50;
+		case 16000: return 16000 / 50;
+		case 24000: return 24000 / 50;
+		case 48000: return 48000 / 50;
+		default:
+			    return 320;
+	}
```
alfredh, line 38: can be replaced with a simpler formula:

```c
return srate / 50;
```

can you also describe what is 50 and how is it calculated ?

modules/hisilicon/hisi_play.c
```diff
@@ -0,0 +1,157 @@
...
+
+#include "hisi.h"
+
+#include <mpi_audio.h>
+#include <mpi_sys.h>
+
+struct auplay_st {
+	thrd_t thread;
+	volatile bool run;
+	void *sampv;
```
alfredh, line 20: which sample format is used here ?

if 16-bit signed then perhaps change to `int16_t *sampv` ?

modules/hisilicon/hisi_play.c
```diff
@@ -0,0 +1,157 @@
...
+	struct auplay_st *st = arg;
+
+	/* Wait for termination of other thread */
+	if (st->run) {
+		debug("hisi: stopping playback thread\n");
+		st->run = false;
+		thrd_join(st->thread, NULL);
+	}
+
+	int ret = HI_MPI_AO_DisableChn(0, 0);
```
alfredh, line 39: does the driver handle concurrency or multiple sessions ?

modules/hisilicon/hisi_play.c
```diff
@@ -0,0 +1,157 @@
...
+	if (HI_SUCCESS != ret) {
+		printf("error %d\n", ret);
```
alfredh, line 41: consider ignoring errors on close, or use warning() instead

modules/hisilicon/hisi_play.c
```diff
@@ -0,0 +1,157 @@
...
+	auframe_init(&af, st->prm.fmt, st->sampv, st->sampc, st->prm.srate,
+			st->prm.ch);
+
+	while (st->run) {
+		st->wh(&af, st->arg);
+
+		AUDIO_FRAME_S stData = {
+			.enBitwidth = AUDIO_BIT_WIDTH_16,
+			.enSoundmode = AUDIO_SOUND_MODE_MONO,
+			.u32Len = st->sampc * 2,
```
alfredh, line 67: replace 2 with sizeof(int16_t)

modules/hisilicon/hisi_play.c
```diff
@@ -0,0 +1,157 @@
...
+			.u64VirAddr[0] = st->sampv,
+		};
+		int ret = HI_MPI_AO_SendFrame(0, 0, &stData, -1);
+		if (ret != HI_SUCCESS) {
+			printf("error %d\n", ret);
```
alfredh, line 72: use `warning()` instead of printf

modules/hisilicon/hisi_play.c
```diff
@@ -0,0 +1,157 @@
...
+		.enWorkmode = AIO_MODE_I2S_MASTER,
+		.enSoundmode = AUDIO_SOUND_MODE_MONO,
+		.u32EXFlag = 0,
+		.u32FrmNum = 2, /* keep it small for low latency */
+		.u32PtNumPerFrm = st->sampc,
+		.u32ChnCnt = 1,
+		.u32ClkSel = 0,
+		.enI2sType = AIO_I2STYPE_INNERCODEC,
+	};
+	int ret = HI_MPI_AO_SetPubAttr(AoDevId, &stAioAttr);
```
alfredh, line 121: value set in `ret` is not checked here

modules/hisilicon/hisi_src.c
```diff
@@ -0,0 +1,261 @@
...
+		ret = HI_MPI_AI_GetFrame(dev, chn, &stFrame, &stAecFrm, -1);
+		if (HI_SUCCESS != ret) {
+			printf("HI_MPI_AI_GetFrame(%d, %d),"
+				" failed with %#x!\n", dev, chn, ret);
+			continue;
+		}
+
+		struct auframe af;
+
+		memcpy(st->sampv, stFrame.u64VirAddr[0], stFrame.u32Len);
```
alfredh, line 75: perhaps you can use the buffer directly here, avoiding memcpy ?

use auframe and set buffer to point to stFrame.u64VirAddr[0]

alfredh:
please also update the PR title to indicate that it is an audio driver

alfredh:
I am not sure about this PR.

1. Since the hardware is a IP camera it would be logical to start with a vidsrc driver, not audio driver
2. The hardware supported is used by a small minority.
3. You did not address my comments about multiple sessions.


I think it would be better to gather some support for this from multiple people.
If there is a really strong interest for this, we can consider it.

Otherwise please consider:

widgetii:
Then why did you ask me to do more and more work on it (changing thread, makefile stuff and so on)? Next time, make it decision early, please.

cspiel1:
This should go to baresip-apps. It is like a playground for baresip add-ons that might be helpful for anybody.
Maybe `baresip-apps/modules/auhisilicon`? In order to state that it is a audio driver.

widgetii:
It makes no sense to open source the code and place it to the repo `that are not officially supported`.  I expect that open sourced code will live its own life (one will use it, another one find bugs and send patches).

We will resolve this issue with our team internally.

Thanks everybody for your time

cspiel1:
Okay! By the way: baresip-apps is officially supported. E.g. we once discussed to add a github action in order to build baresip-apps when ever `re`, `rem` or `baresip` changes. So lets say, somebody changes the API and this breaks the build of your module then this author has also to make a PR for baresip-apps and fix the build.

PR #2101, opened 2022-08-30
instance getters of the video structure

viordash:
When attempting to apply your WebRTC implementation, the WebRTC session and video encoder could not be associated. In my case, this was necessary, since the video source was determined from the pool at the stage of creating a session for a new connection.
This PR adds methods to get a instance of struct video from the `(struct vtx *)arg` argument of the `videnc_update_h` callback, which is equal to the instance in [webrtc/src/sess.c:103](https://github.com/baresip/baresip/blob/4756f56b704fc6377c1cbde10828f75d9c6086d0/webrtc/src/sess.c#L103) `structure media_track * media->u.vid`
The `vrx_video` method added for complementarity.

In the case of audio, it is possible to associate a session and an audio track, the audio instance in [webrtc/src/sess.c:95](https://github.com/baresip/baresip/blob/4756f56b704fc6377c1cbde10828f75d9c6086d0/webrtc/src/sess.c#L95) `struct media_track * media->u.au` is equal to `void *arg` of the [ausrc_alloc_h](https://github.com/baresip/baresip/blob/4756f56b704fc6377c1cbde10828f75d9c6086d0/include/baresip.h#L512) callback


include/baresip.h
```diff
@@ -1316,6 +1316,8 @@ struct config_audio *audio_config(struct audio *au);
  */
 
 struct video;
+struct vtx;
+struct vrx;
```
alfredh, line 1320: these types are not supposed to be exposed outside video.c

it is supposed to be a hidden implementation specific detail.

alfredh: you are not supposed to call this method directly.

It is an implementation of the "vidcodec" interface.

alfredh:
I dont think this PR is needed.

Please share your complete code so that we can understand your usecase.

include/baresip.h
viordash: `av1_encode_update` was given as an example. Based on his code, I made a proprietary video source

viordash:
Sorry, now I don't have time to prepare an example for you that would not go beyond privacy. I can close this issue, and when I prepare an example, I will restore it. 
Thank you for your time

PR #2104, opened 2022-08-31
Added cmake of webrtc_aecm module

juha-h:
`make` generates lots of warnings like in gzrtp module.


juha-h:
WEBRTC_AECM_LIBRARY has PATHS `../webrtc/obj/local/x86_64`.  Any idea with what `x86_64` should be replaced so that this would work with different architectures such as `arm64-v8a` and `armeabi-v7a`?

sreimers:
This depends on the Cross compilation toolchain `-DCMAKE_TOOLCHAIN_FILE=path/to/file` I think, see:

https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-android-with-the-ndk

You can check if `CMAKE_ANDROID_ARCH_ABI` is set.

sreimers:
> `make` generates lots of warnings like in gzrtp module.

You can try:

```cmake
target_include_directories(${PROJECT_NAME} SYSTEM PRIVATE ${WEBRTC_AECM_INCLUDE_DIRS})
```

This should prevent warnings from included headers. Like we did in `module.mk`

juha-h:
> This depends on the Cross compilation toolchain `-DCMAKE_TOOLCHAIN_FILE=path/to/file` I think, see:
> 
> https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-android-with-the-ndk
> 
> You can check if `CMAKE_ANDROID_ARCH_ABI` is set.

Yes, but how should the line
```
PATHS ../webrtc/obj/local/x86_64
```
be changed so that it takes the architecture from CMAKE_ANDROID_ARCH_ABI?  Now x86_64 is hardcoded that is not good.

sreimers:
You can use `../webrtc/obj/local/${CMAKE_ANDROID_ARCH_ABI}` as additional HINTS path (tried before PATHS) .

juha-h:
I tried
```
target_include_directories(${PROJECT_NAME} SYSTEM PRIVATE ${WEBRTC_AECM_INCLUDE_DIRS})
```
and the warnings didn't disappear (webrtc_aecm and gzrtp modules).

Many warnings are due to a struct having the same name as function.

Looks like the warnings are compiler dependent, since I don't get any warning when I use clang compilers included in Android toolchains.

PR #2106, opened 2022-09-01
call: do not set call state to answered, if session progress (PRACK)

RobertMi21:


maximilianfridrich:
Looks good to me.
This is a bug I introduced with baresip/re#430 - the answer handler is called in re/sipsess/connect.c:94 when SDP answers are received through reliable 183 responses. The reasoning for this is that if a SDP answer is received through a reliable provisional response, the answer counts as "final" and the media session is established (see RFC 3262 and 6337 for details). I think checking the scode in baresip is a good fix for this.

PR #2113, opened 2022-09-04
Update video in menu when UA_EVENT_CALL_REMOTE_SDP is recieved

juha-h:


cspiel1:
In call.c `UA_EVENT_CALL_REMOTE_SDP` is sent in function `update_media()`.
There is already:
```c
	if (stream_is_ready(video_strm(call->video))) {
		err |= video_update(call->video, call->peer_uri);
	}
```

Is'nt it called twice now?

juha-h:
Christian Spielberger writes:

> In call.c `UA_EVENT_CALL_REMOTE_SDP` is sent in function `update_media()`.
> There is already:
> ```c
> 	if (stream_is_ready(video_strm(call->video))) {
> 		err |= video_update(call->video, call->peer_uri);
> 	}
> ```
> 
> Is'nt it called twice now?

If it is only the one done in menu has the desired effect.  In the above,
the call is made when stream is ready.  Is the stream ready when video
has been disabled?

If not, perhaps the same test could be added to menu that I proposed
earlier:

if (vrdir == SDP_INACTIVE)
    video_update(call->video, call->peer_uri);

cspiel1:
The stream is not ready in case of `SDP_INACTIVE`. What exactly was missing before this PR?

juha-h:
X11 window was not closed when video was made inactive.

cspiel1:
I understand. I will have a look. I guess the solution should be in `call.c`.

juha-h:
Thanks, I didn't dare to touch the core.

cspiel1:
Your solution works only if `menu` is loaded.

juha-h:
Yes, but in most cases when X11 display is used, menu is also used. But any improvement is of course welcome.

PR #2115, opened 2022-09-06
refer out of dialog

cspiel1:
- ua: add ua_handle_refer
- ua: add UA_EVENT_REFER
- ua: add Refer-Sub to REFER response
- ua,menu: add API function and command for sending out-of-dialog REFER
- ua: add mandatory headers to REFER
- ua,menu: do URI complete for refer-to URI
- account: add UAS authentication
- ua: add authorization check for incoming out-of-dialog REFER
- ua: user@host for digest realm
- ua: handle REFER set Refer-Sub defaul to true
- ua: SIP request handler should not print warning and return false
- test: add test for out-of-dialog REFER

### DIFF
--- a/docs/examples/accounts
+++ b/docs/examples/accounts
@@ -38,6 +38,8 @@
 #    ;stunpass=STUN/TURN/ICE-password
 #    ;stunserver=stun:[user:pass]@host[:port]  # see RFC 3986: Use of the format "user:password" in the userinfo field is deprecated.
 #    ;tcpsrcport=6060
+#    ;uas_user=username
+#    ;uas_pass=password
 #    ;video_codecs=h264,h263,...
 #
 # Examples:
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -138,7 +138,10 @@ enum sipansbeep account_sipansbeep(const struct account *acc);
 void account_set_sipansbeep(struct account *acc, enum sipansbeep beep);
 void account_set_autelev_pt(struct account *acc, uint32_t pt);
 uint32_t account_autelev_pt(struct account *acc);
-
+bool account_uasmethod_deny(const struct account *acc, const struct pl *met);
+bool account_uasmethod_allow(const struct account *acc, const struct pl *met);
+const char* account_uas_user(const struct account *acc);
+const char* account_uas_pass(const struct account *acc);
 
 /*
  * Call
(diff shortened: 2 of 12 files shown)


juha-h:
Would it be appropriate at least in some use cases to ask confirmation from the user if he/she would like to accept the refer?

oej:
> Would it be appropriate at least in some use cases to ask confirmation from the user if he/she would like to accept the refer?

I think that the RFC requires user approval, as it would be many security issues otherwise. Refer out of dialog is often used in call control scenarios, where there are ACL filters and auth to prevent abuse, but allow automated call setups.

cspiel1:
For security reason we decided to introduce UAS authorization. Thus, for incoming requests we have a user/password to be validated. `;uas_user=<user>;uas_pass=<pass>`. For now it is only applied to incoming REFER.

A further PR could introduce also a `;refer=<auto,user>` account setting. With `auto` the INVITE is done by module menu, like now. With `user`, the user/application has to invoke command `dial` "manually".

juha-h:
Christian Spielberger writes:

> For security reason we decided to introduce UAS authorization. Thus,
> for incoming requests we have a user/password to be
> validated. `;uas_user=<user>;uas_pass=<pass>`. For now it is only
> applied to incoming REFER.

That verifies who UAC is, but does not restrict which URIs and when the
UAS can be referred to.

> A further PR could introduce also a `;refer=<auto,user>` account
> setting. With `auto` the INVITE is done by module menu, like now. With
> `user`, the user/application has to invoke command `dial` "manually".

Yes, 'user' option would give refer control to the UAS.

maximilianfridrich:
> I think that the RFC requires user approval

[Section 2.4.2](https://datatracker.ietf.org/doc/html/rfc3515#section-2.4.2) of RFC 3515 states

```
A UA accepting a well-formed REFER request SHOULD request approval
from the user to proceed (this request could be satisfied with an
interactive query or through accessing configured policy).
```

We agree that for security reasons, out-of-dialog REFERs need to be authorized by some mechanism which is why we chose to implement [standard SIP request authentication](https://www.rfc-editor.org/rfc/rfc3261#section-22). Further, the security consideration of RFC 3515 [state](https://datatracker.ietf.org/doc/html/rfc3515#section-5.2):

```
In some cases, the user can provide authorization for some REFER
requests ahead of time by providing policy to the user agent.  This
is appropriate, for instance, for call transfer as discussed in [4].
Here, a properly authenticated REFER request within an existing SIP
dialog to a sip:, sips:, or tel: URI may be accepted through policy
without interactively obtaining the user's authorization.
```

So we are clearly RFC compliant and acting as suggested by the security negotiations. Of course, adding a BareSIP option to ask for user confirmation could provide another alternative to non-interactive authorization.

cspiel1:
ACL is not supported currently. If `uas_user` is not specified, then SIP 403 Forbidden will be returned.

PR #2121, opened 2022-09-12
account: do not append IP address to unregistered accounts.

RobertMi21:


juha-h:
Could you explain what is the motivation of this PR?  I'm also wondering, if this kind of function really belongs to baresip core.

RobertMi21:
> Could you explain what is the motivation of this PR? I'm also wondering, if this kind of function really belongs to baresip core.

After the tests failed I've thought about it and I think we need a solution outside of the core implementation.

The actual behavior I was trying to change was:

When I trigger a call to foobar (/dial foobar) on a device with no SIP Server configured, I receive a video stream showing the video from the device’s own camera.

I going to close the PR, because this is no resolution for this behavior.

cspiel1:
baresip calls itself, if `dial foo` with accounts like this:
`x@127.0.0.1` or `x@<local-ip>`

Issue #2123, opened 2022-09-14
Quit delay with latest main


juha-h:
I had an older baresip/re/rem main build (most likely done on Sept 9) and Quit worked fine.  With today's build, Quit takes long time.  The difference between the older build and now is that now when mwi is unsubscribed, baresip returns `404 not found` to the final NOTIFY:
```
Quit
ua: stop all (forced=0)
mwi: shutdown of sip:test@test.tutpro.com
18:12:26.081#
TCP 192.168.238.215:56384 -> 192.168.238.215:5060
SUBSCRIBE sip:127.0.0.1:5080;alias=127.0.0.1~33817~2;transport=tcp SIP/2.0
Via: SIP/2.0/TCP 192.168.238.215:5020;branch=z9hG4bK070f5feb0cfce3ae;rport
Contact: <sip:test@test.tutpro.com;gr=urn:uuid:9b21c1a8-dc58-a3d6-83d5-7be67cbe70ce>
Max-Forwards: 70
Proxy-Authorization: Digest username="test", realm="test.tutpro.com", nonce="YyHwfGMh71DNg7CNNXmugcWYrTd2ol1eG7Ig1v3mfKkYLZhcCSa563BlKbKA", uri="sip:127.0.0.1:5080;alias=127.0.0.1~33817~2;transport=tcp", response="945493183286702c71171f40d9c525a5", cnonce="a70f0ce3c74bd591", qop=auth, nc=00000002, algorithm=MD5
Route: <sip:192.168.238.215;transport=tcp;r2=on;lr>
Route: <sip:127.0.0.1:5070;transport=tcp;r2=on;lr>
To: <sip:test@test.tutpro.com>;tag=48b5b8952e8d0bca6256941910840e90-cd73d5e4
From: <sip:test@test.tutpro.com>;tag=406d5b374dc71c2a
Call-ID: 1f6c07b0d1ca0941
CSeq: 41017 SUBSCRIBE
User-Agent: baresip v2.7.0 (x86_64/linux)
Event: message-summary
Expires: 0
Accept: application/simple-message-summary
Content-Length: 0

presence: ua=0x55adee1aaec0 got event 8 (SHUTDOWN)
presence: subscriber: closing 0 subs
ua: in use (1) by app module
18:12:26.084#
TCP 192.168.238.215:5060 -> 192.168.238.215:56384
SIP/2.0 200 OK
[...]

PR #2124, opened 2022-09-14
Restored original working behavior in uag request_handler

juha-h:


cspiel1:
I understand. Yes, this fix for https://github.com/baresip/baresip/issues/2123 looks good.

PR #2134, opened 2022-09-20
call, event, audio: send DTMF via hidden call

cspiel1:
UA event handlers are processed in configuration order from h_1 to h_n. Any UA
event handler h_x  may set the calls event stop flag true. Following UA event
handlers are not processed. The next UA event is processed again by the UA
event handlers h_1 to h_x and not by h_{x+1} to h_n.

- call, event: add a UA event stop processing flag
- audio: add audio_txtelev_empty()

Note: If the flag is not set explicitely to `true` by the application, nothing will change for applications that rely on UA events.

Note2: A well known more clean solution would be to add a `return bool` to the UA event handler. But this would mean a break in the API and a lot of changes in different modules.

refers to:
- https://github.com/baresip/re/pull/537
- https://github.com/baresip/baresip-apps/pull/19

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -595,9 +595,13 @@ static void check_telev(struct audio *a, struct autx *tx)
 
 bool audio_txtelev_empty(const struct audio *au)
 {
-	const struct autx *tx = &au->tx;
+	const struct autx *tx;
 	bool empty;
 
+	if (!au)
+		return true;
+
+	tx = &au->tx;
 	mtx_lock(tx->mtx);
 	empty = telev_is_empty(au->telev);
 	mtx_unlock(tx->mtx);
--- a/src/audio.c
+++ b/src/audio.c
@@ -593,6 +593,18 @@ static void check_telev(struct audio *a, struct autx *tx)
 }
 
 
+bool audio_txtelev_empty(const struct audio *au)
+{
+	const struct autx *tx = &au->tx;
+	bool empty;
+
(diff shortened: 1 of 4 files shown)


juha-h:
Christian Spielberger writes:

> - call, event: add a UA event stop processing flag

I suggest that in the future it is explicitly mentioned if a new PR
includes an API change so that baresip devs can check if the change
breaks their apps.

cspiel1:
It is the APP that may set the new flag. If it is not set to `true`, then nothing changes.
E.g. The intercom module in baresip-apps uses this in
https://github.com/baresip/baresip-apps/pull/19

src/audio.c
```diff
@@ -593,6 +593,18 @@ static void check_telev(struct audio *a, struct autx *tx)
 }
 
 
+bool audio_txtelev_empty(const struct audio *au)
+{
```
sreimers, line 597: `au` NULL check is missing.

juha-h:
Christian Spielberger writes:

> It is the APP that may set the new flag. If it is not set to `true`,
> then nothing changes.

My understanding from this PR is that since flag was added, app's build
fails if the argument is missing.

cspiel1:
> My understanding from this PR is that since flag was added, app's build fails if the argument is missing.

Do you have a compiler error message? The flag is not missing. It is in `struct call`.

juha-h:
Sorry, I misread the baresip.h diff.

PR #2135, opened 2022-09-21
URI completion

maximilianfridrich:
When providing `account_uri_complete` with an incomplete dialstring and an account, it would always append the local URI of the account to the dialstring - regardless of the account config. However, when using a local account (with `regint=0`), this could cause BareSIP to try calling itself.

Now, an incomplete URI is only completed for accounts with `regint != 0`, otherwise `account_uri_complete` returns -1 because the string would not be a valid SIP URI.

Further, a few potential memory leaks were fixed in modules using this function.

### DIFF
--- a/modules/gtk/gtk_mod.c
+++ b/modules/gtk/gtk_mod.c
@@ -784,7 +784,7 @@ int gtk_mod_connect(struct gtk_mod *mod, const char *uri)
 
 	err = account_uri_complete(ua_account(mod->ua_cur), uribuf, uri);
 	if (err)
-		return EINVAL;
+		goto out;
 
 	uribuf->pos = 0;
 	err = mbuf_strdup(uribuf, &uri_copy, uribuf->end);
@@ -817,7 +817,7 @@ int gtk_mod_connect_attended(struct gtk_mod *mod, const char *uri,
 
 	err = account_uri_complete(ua_account(mod->ua_cur), uribuf, uri);
 	if (err)
-		return EINVAL;
+		goto out;
 
 	uribuf->pos = 0;
 	err = mbuf_strdup(uribuf, &uri_copy, uribuf->end);
--- a/modules/menu/static_menu.c
+++ b/modules/menu/static_menu.c
@@ -970,7 +970,7 @@ static int options_command(struct re_printf *pf, void *arg)
 	if (err) {
 		(void)re_hprintf(pf, "options_command failed to "
 				 "complete uri\n");
-		return EINVAL;
+		goto out;
 	}
 
 	mem_deref(uri);
@@ -1033,7 +1033,7 @@ static int cmd_refer(struct re_printf *pf, void *arg)
 	err = account_uri_complete(ua_account(ua), uribuf, uri);
 	if (err) {
 		(void)re_hprintf(pf, "invalid URI\n");
-		return EINVAL;
+		goto out;
 	}
 
 	mem_deref(uri);
@@ -1047,7 +1047,7 @@ static int cmd_refer(struct re_printf *pf, void *arg)
 	err = account_uri_complete(ua_account(ua), uribuf, touri);
 	if (err) {
 		(void)re_hprintf(pf, "invalid URI\n");
-		return EINVAL;
+		goto out;
 	}
 
 	mem_deref(touri);
--- a/src/account.c
+++ b/src/account.c
@@ -7,7 +7,6 @@
 #include <re.h>
 #include <baresip.h>
 #include "core.h"
-#include <ctype.h>
 
 
 enum {
@@ -1709,7 +1708,8 @@ const char *account_extra(const struct account *acc)
 
 
 /**
- * Auto complete a SIP uri, add scheme and domain if missing
+ * Auto complete a SIP uri, add scheme and domain if missing.
+ * If regint of the account is 0, then hostport is not appended to the URI.
  *
  * @param acc User-Agent account
  * @param buf Target buffer to print SIP uri
@@ -1720,85 +1720,27 @@ const char *account_extra(const struct account *acc)
 int account_uri_complete(const struct account *acc, struct mbuf *buf,
 			 const char *uri)
 {
-	struct sa sa_addr;
-	size_t len;
-	bool uri_is_ip;
-	char *uridup;
-	char *host;
-	char *c;
-	int err = 0;
+	char hostport[128];
+	char port[8];
+	bool print_port;
+	struct uri *luri;
 
 	if (!buf || !uri)
 		return EINVAL;
 
-	/* Skip initial whitespace */
-	while (isspace(*uri))
-		++uri;
+	if (!acc || acc->regint == 0)
+		return uri_complete(uri, NULL, buf);
 
-	len = str_len(uri);
+	luri = account_luri(acc);
+	print_port = luri->port != SIP_PORT && luri->port != 0;
+	re_snprintf(port, sizeof(port), "%d", luri->port);
 
-	/* Append sip: scheme if missing */
-	if (0 != re_regex(uri, len, "sip:"))
-		err |= mbuf_printf(buf, "sip:");
+	re_snprintf(hostport, sizeof(hostport), "%r%s%s",
+		    &luri->host,
+		    print_port ? ":" : "",
+		    print_port ? port : "");
 
-	err |= mbuf_write_str(buf, uri);
-
-	if (!acc || err)
-		return err;
-
-	/* Append domain if missing and uri is not IP address */
-
-	/* check if uri is valid IP address */
-	err = str_dup(&uridup, uri);
-	if (err)
-		return err;
-
-	if (!strncmp(uridup, "sip:", 4))
-		host = uridup + 4;
-	else
-		host = uridup;
-
-	c = strchr(host, ';');
-	if (c)
-		*c = 0;
-
-	uri_is_ip =
-		!sa_decode(&sa_addr, host, strlen(host)) ||
-		!sa_set_str(&sa_addr, host, 0);
-	if (!uri_is_ip && host[0] == '[' && (c = strchr(host, ']'))) {
-		*c = 0;
-		uri_is_ip = !sa_set_str(&sa_addr, host+1, 0);
-	}
-
-	mem_deref(uridup);
-
-	if (0 != re_regex(uri, len, "[^@]+@[^]+", NULL, NULL) &&
-		1 != uri_is_ip) {
-		if (acc->regint == 0)
-			return err;
-#if HAVE_INET6
-		if (AF_INET6 == acc->luri.af)
-			err |= mbuf_printf(buf, "@[%r]",
-					   &acc->luri.host);
-		else
-#endif
-			err |= mbuf_printf(buf, "@%r",
-					   &acc->luri.host);
-
-		/* Also append port if specified and not 5060 */
-		switch (acc->luri.port) {
-
-		case 0:
-		case SIP_PORT:
-			break;
-
-		default:
-			err |= mbuf_printf(buf, ":%u", acc->luri.port);
-			break;
-		}
-	}
-
-	return err;
+	return uri_complete(uri, hostport, buf);
 }
 
 
--- a/src/account.c
+++ b/src/account.c
@@ -1774,6 +1774,8 @@ int account_uri_complete(const struct account *acc, struct mbuf *buf,
 
 	if (0 != re_regex(uri, len, "[^@]+@[^]+", NULL, NULL) &&
 		1 != uri_is_ip) {
+		if (acc->regint == 0)
+			return err;
 #if HAVE_INET6
 		if (AF_INET6 == acc->luri.af)
 			err |= mbuf_printf(buf, "@[%r]",
--- a/test/account.c
+++ b/test/account.c
@@ -130,6 +130,9 @@ int test_account_uri_complete(void)
 	err = account_alloc(&acc, "\"A\" <sip:A@proxy.com>");
 	TEST_ERR(err);
 
+	err = account_set_regint(acc, 3600);
+	TEST_ERR(err);
+
 	mb = mbuf_alloc(256);
 	ASSERT_TRUE(mb != NULL);
 
--- a/test/ua.c
+++ b/test/ua.c
@@ -209,6 +209,13 @@ int test_ua_alloc(void)
 	/* verify URI complete function */
 	err = account_uri_complete(ua_account(ua), mb, "bob");
 	ASSERT_EQ(0, err);
+	TEST_STRCMP("sip:bob", 7, mb->buf, mb->end);
+
+	mbuf_reset(mb);
+	err = account_set_regint(ua_account(ua), 3600);
+	ASSERT_EQ(0, err);
+	err = account_uri_complete(ua_account(ua), mb, "bob");
+	ASSERT_EQ(0, err);
 	TEST_STRCMP("sip:bob@test.invalid", 20, mb->buf, mb->end);
 
 	mem_deref(ua);


src/account.c
```diff
@@ -1774,6 +1774,8 @@ int account_uri_complete(const struct account *acc, struct mbuf *buf,
 
 	if (0 != re_regex(uri, len, "[^@]+@[^]+", NULL, NULL) &&
 		1 != uri_is_ip) {
+		if (acc->regint == 0)
+			return -1;
```
sreimers, line 1778: A meaningful error code should be used, like `ENOTSUP`

juha-h:
As I have commented before, URI completion is application dependent business.  In some cases even if regint=0 it makes sense to complete the URI with domain of the account.  If URI completion function is in the API, it should be general purpose having enough arguments to cover all options.

cspiel1:
> As I have commented before, URI completion is application dependent business. In some cases even if regint=0 it makes sense to complete the URI with domain of the account. If URI completion function is in the API, it should be general purpose having enough arguments to cover all options.

Yes, I agree. But without this APP logic it should not append something to non-registration accounts. On the other hand calling a user at the SIP registrar without specifying the registrars host should be possible.

maximilianfridrich:
I see where you are coming from @juha-h. However, relying on this URI completion feature of accounts is probably misusing local accounts as contacts and I think contacts should be created and used for such a use case. Unless I am missing something.

src/account.c
cspiel1: Sorry, this was a miss-understanding.

We should return 0 in this case. So the call to a local hostname should be possible.

maximilianfridrich:
I have updated the PR. Please have a look at the modified ua test which shows the new behavior quite well. Since this changes old behavior we have to be careful with this. @juha-h do you think this would break things?

juha-h:
I still don't understand what regint has to do with URI completion.  Doesn't outbound proxy play any role in this?  I suggest that the usage comment in front of the function explains these things.

I personally don't use this function for Request URI completion, since I do URI completion by an utility function in the app.  For example, if user calls a phone number, there exists an additional account configuration variable ;telephony_provider that (if defined) is in such cases used as R-URI hostpart.

So, again, it doesn't make much sense to have this function in baresip API.

juha-h:
My suggestion is to make the API function generic by changing account arg to hostpart arg.  The function would then add scheme (sip:) if missing and hostpart if missing.  Apps would then call the function if needed.

cspiel1:
We don't have this logic in our application. We don't want to track the accounts also in the application. And our users should not be forced to select an account for an outgoing call.

If you don't want this logic to use in the Android APP, than it simply has to provide the complete target SIP URI for an outgoing call.

juha-h:
I didn't suggest that your user would need to select the account.  Your app would select it based on whatever criteria and then pass hostpart of the account (instead of the account itself) as argument to the API function.

cspiel1:
okay, I will see what we can do.

juha-h:
This function
```
int uri_complete( const char *hostpart, struct mbuf *buf, const char *uri)
```
would be generic and thus perhaps would belong to re instead of baresip.

cspiel1:
Just talked to my colleagues from the APP development. There are already a lot of installations that rely on this behavior:

They configure APP contacts (not baresip contacts) simply with the target user-id, e.g. `101`.
1. `uag_find_requri()` selects the first registered UA for the outgoing call.
2. `account_uri_complete()` completes the SIP URI and appends the host part.

We use module `menu` command `dial`/`dialdir` for the outgoing call.

But if I remember correctly, the android APP does not use module menu. And the host append logic is only used by module `menu` and `gtk` (which are the bottom layer applications).

The call of `account_uri_complete()` from `uag.c` only prepends `sip:` if missing.

cspiel1:
- menu uses both: `uag_find_requri()` and `account_uri_complete()` (it simply takes first registered account)
- gtk uses only `account_uri_complete()` and has a separate user selection of the UA

For both it is possible to call just a user-id. The `struct account` knows what has to be appended. That's why it is an account function.

maximilianfridrich:
Have we reached a conclusion yet? I'm fine with leaving it as is or implementing a new function (where exactly?) with a signature and logic as suggested by @juha-h. I can use the existing `account_luri()` function to get the hostpart of the account so it would be pretty straightforward to implement.

juha-h:
Maximilian Fridrich writes:

> Have we reached a conclusion yet? I'm fine with leaving it as is or
> implementing a new function (where exactly?) with a signature and
> logic as suggested by @juha-h. I can use the existing `account_luri()`
> function to get the hostpart of the account so it would be pretty
> straightforward to implement.

Do what you need to be done.  I can then add generic
`uri_complete(uri, hostpart, buf)` function to re_uri.h.

maximilianfridrich:
Please have a look at this updated version. I have implemented the `uri_complete` function in re - as suggested - which `account_uri_complete` uses and it checks regint before calling `uri_complete`.
Is this satisfactory?

re PR:
- baresip/re#539

juha-h:
Maximilian Fridrich writes:

> Please have a look at this updated version. I have implemented the
> `uri_complete` function in re - as suggested - which
> `account_uri_complete` uses and it checks regint before calling
> `uri_complete`.
> Is this satisfactory?

I was out and haven't looked your code yet, but this is what I was
planning:

- add scheme "sip:" if missing

- if string that follows "sip:" is an IP address or contains port (:xxx)
  or parameters (;param), do nothing more

- if string that follows "sip:" is valid user, add "@" and hostpart
  argument, which may contain port and params

Does this match with what you did?

maximilianfridrich:
Yes, that's what `uri_complete` does. The behavior is the exact same as `account_uri_complete` was before this PR.

juha-h:
Maximilian Fridrich writes:

> Yes, that's what `uri_complete` does. The behavior is the exact same
> as `account_uri_complete` was before this PR.

Based on comments in the PR, I'm not sure if it does what I proposed:

    /* Append sip: scheme if missing */
	if (0 != re_regex(uri, len, "sip:"))
		err |= mbuf_printf(buf, "sip:");

	err |= mbuf_write_str(buf, uri);

	if (err)
		return err;

	/* Append domain if missing and uri is not IP address */
	/* check if uri is valid IP address */

If sip: is not missing, domain should not be appended.  For example, if
uri is

  sip:foo.bar

then for.bar should be considered as domain name, not userpart.  That
is, by including scheme, the user indicates that uri should be used as
is.

juha-h:
I still don't understand what regint has to do with uri completion.  Please explain.

In my Android app, long press of an account in the spinner toggles account's registration. It has no effect whatsoever on making calls.

maximilianfridrich:
Please see the very first comment introducing the PR. We have observed an issue where BareSIP tries calling itself when using local accounts (i.e. regint=0).

cspiel1:
The Android app will not be affected by this change.

maximilianfridrich:
> That is, by including scheme, the user indicates that uri should be used as is.

I took your suggested function signature `uri_complete(uri, hostpart, buf)`. You indicate that you want the URI as is by passing `hostpart=NULL`.

cspiel1:
> > That is, by including scheme, the user indicates that uri should be used as is.
> 
> I took your suggested function signature `uri_complete(uri, hostpart, buf)`. You indicate that you want the URI as is by passing `hostpart=NULL`.

What is the benefit of such a function?

juha-h:
Maximilian Fridrich writes:

> Please see the very first comment introducing the PR. We have observed
> an issue where BareSIP tries calling itself when using local accounts
> (i.e. regint=0).

I have read the comment and don't understand it. Give a concrete
example.

According to my understanding, if account ***@***.*** calls car, completed
uri is ***@***.*** no matter if ***@***.*** is currently registered or not.

What is wrong with that?

juha-h:
Christian Spielberger writes:

> The Android app will not be affected by this change.

Yes, I know it and that was not my point.  I just try to understand what
registration status has to do with uri completion.

juha-h:
Maximilian Fridrich writes:

> > That is, by including scheme, the user indicates that uri should be
> > used as is. 
> 
> I took your suggested function signature `uri_complete(uri, hostpart,
> buf)`. You indicate that you want the URI as is by passing `hostpart=NULL`

That was before I took closer look at uri_complete function and made the
proposal yesterday.

juha-h:
Christian Spielberger writes:

> What is the benefit of such a function?

If you ask about the proposed uri_complete re function, its benefit is
each app does not need to implement it because it is included re lib.

cspiel1:
Here are a few examples for local accounts:
```
<sip:A@localhost>;regint=0
<sip:A@192.168.1.11>;regint=0  /* where this is our local IP address*/
<sip:A@office>;regint=0   /* where "office" is the hostname of our device */
```
In non of these cases it would make sense to append the host part to the dial address.

Further if there would be an account like this
```
<sip:A@192.168.1.21>;regint=0  /* where this is the IP of another device */
```
This would make no sense.

juha-h:
Based on your examples, I still don't see what regint has to do with uri completion.  If A@office calls B, the B should be completed to sip:B@office no matter if A is registered on not. See also my example in https://github.com/baresip/baresip/pull/2135#issuecomment-1255867063.

juha-h:
Christian Spielberger writes:

> Here are a few examples for local accounts:
> ```
> ***@***.***>;regint=0
> ***@***.***>;regint=0  /* where this is our local IP address*/
> ***@***.***>;regint=0   /* where "office" is the hostname of our device */
> ```
> In non of these cases it would make sense to append the host part to
> the dial address.

Just to add to my previous comment, it could make sense to append
hostpart to dial address, for example, the same host 192.168.1.11 or
office could host several UAs.

But my point is that if it makes sense or not has nothing to do with
regint.

cspiel1:
We want to append the host-part if and only if it references a SIP registrar. Because only in this case it makes sense to target the SIP INVITE to this host.

And we have a lot of customer installations where the configured contacts only have set the user-id for an outgoing call. These user-ids are accociated to UAs registered all to the same SIP Registrar. In these installations there is only one SIP Registrar. And thus it makes sense to append the host part of the account.

juha-h:
If URI completion is made to depend on regint, it breaks my existing baresip Linux usage.  I just tested like this:

- accounts test@test.tutpro.com and jh@test.tutpro.com are started with ;regint=600
- test@test.tutpro.com dials `jh`:
```
- >                               jh
test@test.tutpro.com: selected for request
call uri: sip:jh@test.tutpro.com
```
- account test@test.tutpro.com is re-started with ;regint=0
- account  test@test.tutpro.com dials `jh`:
 ```
>                               jh
test@test.tutpro.com: selected for request
call uri: sip:jh@test.tutpro.com
```
In both cases, dialed string is correctly completed as it should.

Does it mean that after this PR, the second call the dial string would not get completed with test.tutpro.com?

**Comment by maximilianfridrich** [comment]
> Does it mean that after this PR, the second call the dial string would not get completed with test.tutpro.com?

That is correct. This is not the intended use case for local accounts, contacts should be used for scenarios like that.

**Comment by cspiel1** [comment]
- host-part refers to a SIP Registrar iff (if and only if) `regint` is positive
- host-part refers to a SIP Proxy iff `;outbound` is set
- otherwise the host-part refers to the local machine

We don't want to append a host-part that refers to the local machine if the user does not specify it.

**Comment by juha-h** [comment]
> * host-part refers to a SIP Registrar iff (if and only if) `regint` is positive
> 
>     * host-part refers to a SIP Proxy iff `;outbound` is set

How can both choices by `iff`, since ;outbound can be set when ;regint > 0?

**Comment by cspiel1** [comment]
Right. The second was wrong. The host-part does not refer a SIP Proxy. The `outbound` value does.

**Comment by juha-h** [comment]
Maximilian Fridrich writes:

> > Does it mean that after this PR, the second call the dial string
> > would not get completed with test.tutpro.com? 
> 
> That is correct. This is not the intended use case for local accounts,
> contacts should be used for scenarios like that.

Sorry, but I didn't understand.  Are you saying that my example where
`jh` is called when ;regint=0, is something that should not be done or
are you going to fix the PR?

**Comment by cspiel1** [comment]
Example account:
```
<sip:1@192.168.167.13>;regint=0
```
The user dials xyz. Then previously this ended in a call to the local baresip instance.

Now we have this error message, like it is expected by a normal user.
```
>                              xyz
1@192.168.167.13: selected for request
call uri: sip:xyz
call: connecting to 'sip:xyz;transport=udp'..
call id: 70c08f19dc401652
sip:xyz;transport=udp: session closed: Destination address required [89]
```

**Comment by maximilianfridrich** [comment]
> Are you saying that my example where `jh` is called when ;regint=0, is something that should not be done or are you going to fix the PR?

It's not that it shouldn't be done, but it should not be expected that a hostname is added to a dialstring when you are using a local account. If you want those shortcuts you will have to configure contacts.

I would leave the PR as is.

**Comment by juha-h** [comment]
This PR (nor any other PR) should not break normal, common, existing behavior.  You seem to have some special case that may warrant a custom solution.

I also tested with two accounts sip:test@192.168.238.215 and sip:jh@192.168.238.215, both with ;regint=0.  Dial from test to `jh` was correctly completed and the call worked fine.

**Comment by cspiel1** [comment]
I assume that you have to baresip instances on one machine. No, you have two accounts on one baresip instance and want to have a SIP call to your-self. This is not a "normal" use case in our sense.

But there is also a `retest` that calls b@localhost from a@localhost. This is still possible if the target is specified completely.

In your case you would have to call `jh@192.168.238.215` and it will work. If you dial `jh` it would not work. There might be another host with this domain in the local subnet. This is another use case we want to support. Calling a `domain` in the local subnet.

Currently we have a work around that we prepend a dummy string and call `x@domain`. This is also we want to get rid of.

**Comment by cspiel1** [comment]
Very long discussion. I think @alfredh should make a decision what would be best. The current code has these contras:
- unexpected self calls,
- calling a domain in the local sub-net is not possible

So I suggest to:
- accept this PR, or
- remove the code that appends anything. In this case we will have to append the hostname in the APP code. The APP has to track the registration state of at least one account and append the hostname.

**Comment by juha-h** [comment]
As I have mentioned many times, this PR shouldn't break existing  behavior. Registration has nothing to do with dial string completion. See my example. If you need something else, add a new Account option for it.

**Comment by juha-h** [comment]
I'm traveling and haven't read all comments but I try once more to make my point.

If both ***@***.*** and ***@***.*** are registered, test dialing jh will get completed to ***@***.*** Then test doesn't want to be distributed and unregisters. That should not cause any chance to calling jh. String jh must still be completed as it did before unregister.

Maybe I have misunderstood the PR. If not, add for example ;no_complete account parameter.

**Comment by juha-h** [comment]
Something else has changed too with this PR.  Before the PR, when I start baresip with one registering account and make a call w/o domain, it works fine:
```
baresip is ready.
test@test.tutpro.com: (prio 0) {1/TCP/v4} 200 OK (OpenSIPg SIP Proxy (5.6.0-1 (x86_64/linux))) [1 binding]
All 1 useragent registered successfully! (1028 ms)
mwi: subscribing to messages for sip:test@test.tutpro.com
----- MWI for sip:test@test.tutpro.com -----
Messages-Waiting: no
Message-Account: sip:test@vm.test.tutpro.com
Voice-Message: 0/0 (0/0)

>                               jh
test@test.tutpro.com: selected for request
call uri: sip:jh@test.tutpro.com
```
When I applied the PR (baresip and re), this happens:
```
baresip is ready.
test@test.tutpro.com: (prio 0) {1/TCP/v4} 200 OK (OpenSIPg SIP Proxy (5.6.0-1 (x86_64/linux))) [1 binding]
All 1 useragent registered successfully! (1018 ms)
mwi: subscribing to messages for sip:test@test.tutpro.com
----- MWI for sip:test@test.tutpro.com -----
Messages-Waiting: no
Message-Account: sip:test@vm.test.tutpro.com
Voice-Message: 0/0 (0/0)

>                               jh
ua: failed to complete uri: jh
could not find UA for jh
>               jh@test.tutpro.com
ua: failed to complete uri: jh@test.tutpro.com
could not find UA for jh@test.tutpro.com
```
This even more serious issue than the one I have commented earlier.

File: modules/gtk/gtk_mod.c [hunk 2]
```diff
@@ -784,7 +784,7 @@ int gtk_mod_connect(struct gtk_mod *mod, const char *uri)
 
 	err = account_uri_complete(ua_account(mod->ua_cur), uribuf, uri);
 	if (err)
-		return EINVAL;
+		goto out;
```
alfredh, line 787: can you please make a separate PR for this ?

it makes it easier to read the main patch

also we should consider moving "gtk" to baresip-apps.

modules/menu/static_menu.c
```diff
@@ -970,7 +970,7 @@ static int options_command(struct re_printf *pf, void *arg)
 	if (err) {
 		(void)re_hprintf(pf, "options_command failed to "
 				 "complete uri\n");
-		return EINVAL;
+		goto out;
```
alfredh, line 973: please make a separate PR for this

src/account.c
```diff
@@ -1720,83 +1720,27 @@ const char *account_extra(const struct account *acc)
...
+		    print_port ? port : "");
 
-		default:
-			err |= mbuf_printf(buf, ":%u", acc->luri.port);
-			break;
-		}
-	}
-
-	return err;
+	return uri_complete(uri, hostport, buf);
```
alfredh, line 1743: I dont think it matters where the function `uri_complete` is located.

The logical behaviour would be the same, even if located in A or B or C.

test/account.c
```diff
@@ -130,6 +130,9 @@ int test_account_uri_complete(void)
 	err = account_alloc(&acc, "\"A\" <sip:A@proxy.com>");
 	TEST_ERR(err);
 
+	err = account_set_regint(acc, 3600);
+	TEST_ERR(err);
+
```
alfredh, line 135: I think the regint should be zero for testcode. Can you run the testcode and make sure
the program does not send any packets to proxy.com ?

alfredh:
The uri complete function was added to append the local domain when dialing a username.

I think we can split the function or perhaps rename it to something like:

 `append_domain();`

Which would do exactly that; append local domain.

If you want to dial directly a domain without userpart, e.g. "foo.bar" then this function
should not be invoked.

If we cannot find a solution to make it work in all cases, it would be a solution
to add a new config option to turn this off/on, e.g. "auto_append_domain=yes|no"

juha-h:
@alfredh Thanks for your comments. "auto_append_domain=yes|no" could be account specific option.

See also https://github.com/baresip/baresip/pull/2135#issuecomment-1256472627, which shows that the PR badly breaks the very normal dial usage and it is not mentioned in the PR description.

maximilianfridrich:
Thank you @alfredh.
The behavior mentioned by @juha-h of course shouldn't happen and looks serious. I will try to reproduce it.
I will take a look at the code again tomorrow, and see if we can iron out any differences. I think the "auto_append_domain=yes|no" idea could be an option for us and this way we might not have to change existing behavior. I will consult with @cspiel1.

Maybe there is even an alternative where we can fix the issue of BareSIP trying to call itself without touching URI completion, but we haven't been able to think of one yet.

cspiel1:
> ua: failed to complete uri: jh
could not find UA for jh

Can't reproduce this. This seems to be not related to this PR.
What exactly did not work?
- prepending "sip:" --> "sip:jh"
- Appending host part will be done like before if `regint` was set, like in your example. So this can't be the reason.
- "Could not find UA for jh" is not related to this PR. We are not able to reproduce this.

maximilianfridrich:
Thank you for your input. It looks like we will not be able to make URI completion dependent on the account registration interval while still keeping the behavior in other scenarios unchanged. Further, we are not particularly fond of adding another account configuration option to BareSIP.

I will close this PR and we will try to find another solution to our problem of BareSIP calling itself when local accounts are used.

Thanks again for everyone's input.

cspiel1:
We will append the hostpart in the APP.

The current code has another problem that can't be solved by anything in the APP. A call to a local domain without "@" in the dial URI is not possible. I will open an issue.

juha-h:
> If the application calls `ua_unregister()` or the new function `ua_stop_register()` the account parameter `regint` will not be touched. So this use-case still will work like you describe this here.

Yes, I noticed it.  In my Android app, when user un-registers by long touch on the account in the spinner, also regint will be changed:
```
    Api.account_set_regint(acc.accp, 0)
    Api.ua_unregister(ua.uap)
```

cspiel1:
I suggest not to change the regint.

Anyway, here is the issue that should be solved: https://github.com/baresip/baresip/issues/2150

PR #2140, opened 2022-09-23
sndfile Module - include peer_uri in filename when recording audio.  …

ninp0:
…Per discussion here: https://github.com/baresip/baresip/discussions/1990


cspiel1:
But then it is specific to a call. You said that there might other use-cases for stream/audio, stream/video. E.g. peerconn.c

Other drawback of this PR. It adds a parameter to `call_alloc()`. We talked once about moving the allocation of audio/video streams after the SDP negotiation. Maybe we should first try this steps. The peer_uri is set in `call_accept()` for incoming calls and in `call_connect()` for outgoing calls. This is only a little bit later in the call stack when incoming/outgoing INVITEs occur.

For now also setter functions for stream might help. Also the `peer_uri` might be placed better in `stream` instead of `audio`.

ninp0:
Ideally this would be configurable within ~/.baresip/config

For example:
~/.baresip/config:
```
module                 sndfile.so
snd_path                .
# options %peer_uri OR local_uri
snd_name              %peer_uri
# OR
snd_name              %local_uri
```

In some circumstances I could see where including the filename to contain the local_uri (i.e. incoming caller id) would be useful, for example a call center wanting to record incoming calls.  In other circumstances (such as my own), having the filename contain the peer_uri (i.e. outgoing destination number) would be useful (e.g. telephony-based security assessments).

stream_param initially looked great since local_uri was already being populated within: https://github.com/baresip/baresip/blob/main/src/call.c#L797

Appending strm_parm.peer_uri = call->peer_uri was returning null because it was being instantiated  a little later in the flow, which is what prompted me to add a parameter to call_alloc().

Please let me know the design you'd prefer (i.e. where to reference peer_uri & local_uri to ultimately pass into https://github.com/baresip/baresip/blob/main/modules/sndfile/sndfile.c#L82) and I can submit another PR.  Thanks.

cspiel1:
I will try to move the `call_stream_alloc()` call to a later point. For outgoing calls it is anyway in `sipsess_desc_handler()` which is called by `re` in order to serialize the SDP for the INVITE.

For incoming calls it should also be possible to alloc the streams later.

cspiel1:
After this #2154 was reviewed, we can proceed here.

PR #2154, opened 2022-09-26
call: allocate streams after peer_uri was set

cspiel1:
Related to: https://github.com/baresip/baresip/pull/2140

What changes:
- incoming calls: allocate the streams in `call_accept()`
- outgoing calls: allocate them in `sipsess_desc_handler()`, when the SDP offer is generated. Nothing changes for non-IP address calls.

What relates to https://github.com/baresip/baresip/pull/2140:
For now the peer_uri should be available when the streams are allocated. In a
next PR it could be passed to the streams.

Maybe in another PR the streams could be allocated after the SDP negotiation.
Currently this is not possible due to the "common audio or video codecs" check
which depends on the audio/video streams.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -1003,9 +1003,6 @@ int call_alloc(struct call **callp, const struct config *cfg, struct list *lst,
 		call->xcall = xcall;
 	}
 
-	if (call->af != AF_UNSPEC)
-		call_streams_alloc(call, got_offer);
-
 	if (cfg->avt.rtp_timeout) {
 		call_enable_rtp_timeout(call, cfg->avt.rtp_timeout*1000);
 	}
@@ -2146,6 +2143,10 @@ int call_accept(struct call *call, struct sipsess_sock *sess_sock,
 		if (err)
 			return err;
 	}
+
+	err = call_streams_alloc(call, got_offer);
+	if (err)
+		return err;
 
 	if (got_offer) {
 
--- a/src/call.c
+++ b/src/call.c
@@ -783,7 +783,7 @@ static void call_decode_sip_autoanswer(struct call *call,
 }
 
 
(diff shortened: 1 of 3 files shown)


cspiel1:
Needs now some tests.

src/ua.c
```diff
@@ -906,6 +907,11 @@ void ua_handle_options(struct ua *ua, const struct sip_msg *msg)
 			return;
 		}
 
+		got_offer = (mbuf_get_left(msg->mb) > 0);
```
sreimers, line 910: Can't you reuse `call->got_offer` here?

src/call.c
```diff
@@ -783,7 +783,7 @@ static void call_decode_sip_autoanswer(struct call *call,
 }
 
 
-static int call_streams_alloc(struct call *call, bool got_offer)
+int call_streams_alloc(struct call *call, bool got_offer)
```
sreimers, line 786: Is `got_offer` arg needed here? Is there a difference with `call->got_offer`? At first look it looks like the same check.

cspiel1: A cleanup might be possible.
`got_offer`:
- It is set `true`, if an SDP offer is received
- It is set to `false`, if an SDP answer is received

In `call_accept()` we simply could set the `call->got_offer` before the call to `call_streams_alloc()`.
In `sipsess_desc_handler()` we passed false. But even more correct it would be to simply leave `call->got_offer` untouched. See next commit!

cspiel1: There is also the incoming OPTIONS. I think we may also set it in `call_alloc()`. Then we have the correct `call->got_offer` also in `ua_handle_options()`.

sreimers: The got_offer` refactoring looks great, less confusing now. Thanks.

PR #2162, opened 2022-09-28
Optimization of Opensles for old Android phones, and some Chinese AOSP

KORuL:


juha-h:
KORuL writes:

>   * Optimization of Opensles for old Android phones, and some Chinese
>   * AOSP

Lots on changes.  Are you sure that there is no negative effects on new
Android devices?

src/audio.c
```diff
@@ -472,7 +472,7 @@ static void encode_rtp_send(struct audio *a, struct autx *tx,
 
 		uint32_t rtp_ts = tx->ts_ext & 0xffffffff;
 
-		if (len) {
+		if (len && !tx->muted) {
```
sreimers, line 475: Not sending packets if muted can lead to problems like NAT Timeouts etc. Can you explain why this is needed?

KORuL:
Yes, I'm sure your baresip is used in my android application. This improved the sound quality.
When it's mute, it's better not to send anything, bad old androids have a side effect from the microphone to the speaker and vice versa.
There are also optimizations in opensles as in ndk-samples from google and telegram

KORuL:
I was wrong, I had to do two pullrequest
in one opensles optimization
in the second do not send empty packets

KORuL:
can you aprove only the first commit?

src/audio.c
KORuL: I was wrong, I had to do two pullrequest
in one opensles optimization
in the second do not send empty packets

sreimers: Yes two PRs are needed, but sending empty packets is a feature not a bug. As said many things rely on sending RTP packets and bad things like NAT/TURN timeouts can happen. On a conference call its normal that one person is muted for minutes.

> When it's mute, it's better not to send anything, bad old androids have a side effect from the microphone to the speaker and vice versa.

Can you explain this further. Do you mean audible noise introduced by network traffic? Mute means all source frames are silenced (overridden with zero).

KORuL: My case is an old Chinese tablet and a bad connection to the intercom.
I have a call going through and if not mute, then the tablet has an echo from its microphone to its speaker. Therefore, it was decided not to send packets when they are not talking on the tablet, but only to receive. Packets are only sent when speech is received on the tablet. Since the connection is bad, the traffic is very important.
I also can’t overcome that even with a good video connection, the picture often crumbles and ripples

alfredh:
this reminds me that we should add Android build to the CI.


can you please split up the patch into logical chunks ?

modules/opensles/player.c
```diff
@@ -76,6 +77,9 @@ static int createOutput(struct auplay_st *st)
 	if (SL_RESULT_SUCCESS != r)
 		return ENODEV;
 
+	//SLEnvironmentalReverbItf reverb;
+	//r = (*st->outputMixObject)->GetInterface(st->outputMixObject, SL_IID_ENVITONMENTALREVERB, &reverb);
+
```
alfredh, line 82: please remove commented out lines

modules/opensles/player.c
```diff
@@ -90,22 +94,20 @@ static int createPlayer(struct auplay_st *st, struct auplay_prm *prm)
 	SLDataLocator_AndroidSimpleBufferQueue loc_bufq = {
 		SL_DATALOCATOR_ANDROIDSIMPLEBUFFERQUEUE, 2
 	};
-	uint32_t ch_mask = prm->ch == 2
-		? SL_SPEAKER_FRONT_LEFT | SL_SPEAKER_FRONT_RIGHT
-		: SL_SPEAKER_FRONT_CENTER;
+	int speakers = SL_ANDROID_SPEAKER_USE_DEFAULT;
```
alfredh, line 97: why has type changes from uint32_t to int ?

modules/opensles/player.c
```diff
@@ -117,6 +119,21 @@ static int createPlayer(struct auplay_st *st, struct auplay_prm *prm)
...
+    if (SL_RESULT_SUCCESS == r) {
+        SLint32 stream_type = SL_ANDROID_STREAM_VOICE;
+
+        (*player_config)->SetConfiguration(player_config,
+                                               SL_ANDROID_KEY_STREAM_TYPE, &stream_type, sizeof(SLint32));
+
+        SLuint32 performance = SL_ANDROID_PERFORMANCE_NONE;
+        (*player_config)->SetConfiguration(player_config,
+										   SL_ANDROID_KEY_PERFORMANCE_MODE, &performance, sizeof(SLuint32));
+    }
```
alfredh, line 135: please check the indentation

modules/opensles/player.c
```diff
@@ -139,6 +156,22 @@ static int createPlayer(struct auplay_st *st, struct auplay_prm *prm)
...
+                                            &volume);
+    if (SL_RESULT_SUCCESS == r) {
+        (*volume)->SetVolumeLevel(volume, 0);
+    }
+
+//    SLEffectSendItf sendEffect;
+//    r = (*st->bqPlayerObject)->GetInterface(st->bqPlayerObject,
+//											SL_IID_EFFECTSEND,
+//                                            &sendEffect);
+//    if (SL_RESULT_SUCCESS == r) {
```
alfredh, line 171: please remove commented out lines

modules/opensles/recorder.c
```diff
@@ -107,6 +109,49 @@ static int createAudioRecorder(struct ausrc_st *st, struct ausrc_prm *prm)
...
+
+
+    SLAndroidAcousticEchoCancellationItf player_aec;
+    r = (*st->recObject)->GetInterface(st->recObject,
+                                            SL_IID_ANDROIDACOUSTICECHOCANCELLATION,
+                                            &player_aec);
+    if (SL_RESULT_SUCCESS == r) {
+        SLboolean enable = SL_BOOLEAN_TRUE;
+        (*player_aec)->SetEnabled(player_aec, enable);
+    }
```
alfredh, line 135: should AEC always be enabled ?

There are cases where the AEC is provided by software,
or the user wants to disable AEC on purpose.

modules/opensles/recorder.c
```diff
@@ -107,6 +109,49 @@ static int createAudioRecorder(struct ausrc_st *st, struct ausrc_prm *prm)
...
+
+    SLAndroidNoiseSuppressionItf player_noise;
+    r = (*st->recObject)->GetInterface(st->recObject,
+                                            SL_IID_ANDROIDAUTOMATICGAINCONTROL,
+                                            &player_noise);
+    if (SL_RESULT_SUCCESS == r) {
+        SLboolean enable = SL_BOOLEAN_TRUE;
+        (*player_noise)->SetEnabled(player_noise, enable);
+    }
+
```
alfredh, line 154: should AGC and Noise Removal be configurable ?

Issue #2150, opened 2022-09-26
Dial to a domain without "@" is not possible through `menu`


cspiel1:
Suppose that `office` is a domain in the local network. It should be possible to initiate an outgoing call to domain `office`.

Current behavior:
Dialing e.g. `office` through module `menu` is not possible, because the host part of the selected account will be appended. The DNS resolution fails.

e.g. configured account:
```
<sip:alice@localhost>;regint=0
```

```
>                           office
A@localhost: selected for request
call uri: sip:office@localhost
```

cspiel1:
We should remove the whole append host part code. This would simplify the behavior and code by 39 lines in account.
The APP should append the host part. It can't be decided much more flexible in the APP.

E.g. the decision could be passed to the user which account should be used for the outgoing call.

cspiel1:
The `account_uri_complete()` then would be dropped, and the prepending of "sip:" could be done in `ua_connect_dir()` in ua.c.

alfredh:
is "office" an endpoint or a domain ?

Can you prepend a dummy userpart to it:

```
sip:foo@office ?
```


Perhaps consider this; if "office" can be resolved via SIP DNS mechanism,
and it returns value entry, we should consider it as a "complete" uri.

cspiel1:
` sip:office` is a complete URI. The domain can be resolved by DNS resolution. In a subnet setup with specific DNS configuration.

Pre-pending a dummy user is our work-around. And the appending of the hostname was also our "failure" in the past. It lead to so many problems and ugly code. Sorry for that! But now we want to get rid of it.

alfredh:
How about using this instead:

```
sip:office@company.com
```

sip:office is not a globally routable AOR/URI ?

cspiel1:
`office` is the user part in this SIP URI. `company.com` is the domain.

In my example `office` is the domain. No user is given.

From https://www.rfc-editor.org/rfc/rfc3261#section-25:
```
SIP-URI          =  "sip:" [ userinfo ] hostport
```

The user part is optional.

juha-h:
I sent similar comment as email, but looks it never arrived:
```
Date: Fri, 30 Sep 2022 10:15:07 +0300
To: baresip/baresip <reply+ABP3HREKF356CJI5F4MTWSGBIPEDXEVBNHHFFGGDJY@reply.github.com>
Cc: baresip/baresip <baresip@noreply.github.com>,
    Subscribed <subscribed@noreply.github.com>
[...]

PR #2163, opened 2022-09-28
simplify URI complete

cspiel1:
Solves: #2150
Relates to: https://github.com/baresip/re/pull/557

Adds new setting 
```
 append_domain false|true
 ```

edited: 
And simplifies the URI completion by using directly the pl strings we got from command parameter parsing. Allocating an mbuf is not necessary any longer. This reduces the code in the app modules.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -128,10 +128,8 @@ const char *account_medianat(const struct account *acc);
 const char *account_mwi(const struct account *acc);
 const char *account_call_transfer(const struct account *acc);
 const char *account_extra(const struct account *acc);
-int account_uri_complete(const struct account *acc, struct mbuf *buf,
-			 const char *uri);
-int account_uri_complete_strdup(const struct account *acc, char **strp,
-			 const char *uri);
+int account_uri_complete(const struct account *acc, char **strp,
+			 const struct pl *uri);
 int account_answerdelay(const struct account *acc);
 void account_set_answerdelay(struct account *acc, int adelay);
 bool account_sip_autoanswer(const struct account *acc);
@@ -896,7 +894,7 @@ struct ua   *uag_find(const struct pl *cuser);
 struct ua   *uag_find_msg(const struct sip_msg *msg);
 struct ua   *uag_find_aor(const char *aor);
 struct ua   *uag_find_param(const char *name, const char *val);
-struct ua   *uag_find_requri(const char *requri);
+struct ua   *uag_find_requri(const struct pl *requri);
 struct sip  *uag_sip(void);
 struct list *uag_list(void);
 uint32_t     uag_call_count(void);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -128,8 +128,10 @@ const char *account_medianat(const struct account *acc);
 const char *account_mwi(const struct account *acc);
 const char *account_call_transfer(const struct account *acc);
 const char *account_extra(const struct account *acc);
-int account_uri_complete(const struct account *acc, char **strp,
-			 const struct pl *uri);
+int account_uri_complete_strdup(const struct account *acc, char **strp,
+				const struct pl *uri);
+int account_uri_complete(const struct account *acc, struct mbuf *buf,
+			 const char *uri);
 int account_answerdelay(const struct account *acc);
 void account_set_answerdelay(struct account *acc, int adelay);
 bool account_sip_autoanswer(const struct account *acc);
@@ -894,7 +896,8 @@ struct ua   *uag_find(const struct pl *cuser);
 struct ua   *uag_find_msg(const struct sip_msg *msg);
 struct ua   *uag_find_aor(const char *aor);
 struct ua   *uag_find_param(const char *name, const char *val);
-struct ua   *uag_find_requri(const struct pl *requri);
+struct ua   *uag_find_requri_pl(const struct pl *requri);
+struct ua   *uag_find_requri(const char *requri);
 struct sip  *uag_sip(void);
 struct list *uag_list(void);
 uint32_t     uag_call_count(void);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -130,6 +130,8 @@ const char *account_call_transfer(const struct account *acc);
 const char *account_extra(const struct account *acc);
 int account_uri_complete(const struct account *acc, struct mbuf *buf,
 			 const char *uri);
+int account_uri_complete_strdup(const struct account *acc, char **strp,
+			 const char *uri);
 int account_answerdelay(const struct account *acc);
 void account_set_answerdelay(struct account *acc, int adelay);
 bool account_sip_autoanswer(const struct account *acc);
--- a/modules/contact/contact.c
+++ b/modules/contact/contact.c
@@ -84,6 +84,7 @@ static int cmd_dial_contact(struct re_printf *pf, void *arg)
 {
 	struct contact *cnt;
 	const char *uri;
+	struct pl pl;
 	int err = 0;
 	(void)arg;
 
@@ -93,8 +94,9 @@ static int cmd_dial_contact(struct re_printf *pf, void *arg)
 	}
 
 	uri = contact_uri(cnt);
+	pl_set_str(&pl, uri);
 
-	err = ua_connect(uag_find_requri(uri), NULL, NULL, uri, VIDMODE_ON);
+	err = ua_connect(uag_find_requri(&pl), NULL, NULL, uri, VIDMODE_ON);
 	if (err) {
 		warning("contact: ua_connect(%s) failed: %m\n",
 			uri, err);
@@ -109,6 +111,7 @@ static int cmd_message(struct re_printf *pf, void *arg)
 	const struct cmd_arg *carg = arg;
 	struct contact *cnt;
 	const char *uri;
+	struct pl pl;
 	int err = 0;
 
 	cnt = contacts_current(baresip_contacts());
@@ -117,8 +120,9 @@ static int cmd_message(struct re_printf *pf, void *arg)
 	}
 
 	uri = contact_uri(cnt);
+	pl_set_str(&pl, uri);
 
-	err = message_send(uag_find_requri(uri), uri, carg->prm,
+	err = message_send(uag_find_requri(&pl), uri, carg->prm,
 			   send_resp_handler, NULL);
 	if (err) {
 		(void)re_hprintf(pf, "contact: message_send(%s) failed (%m)\n",
--- a/modules/contact/contact.c
+++ b/modules/contact/contact.c
@@ -84,7 +84,6 @@ static int cmd_dial_contact(struct re_printf *pf, void *arg)
 {
 	struct contact *cnt;
 	const char *uri;
-	struct pl pl;
 	int err = 0;
 	(void)arg;
 
@@ -94,9 +93,8 @@ static int cmd_dial_contact(struct re_printf *pf, void *arg)
 	}
 
 	uri = contact_uri(cnt);
-	pl_set_str(&pl, uri);
 
-	err = ua_connect(uag_find_requri(&pl), NULL, NULL, uri, VIDMODE_ON);
+	err = ua_connect(uag_find_requri(uri), NULL, NULL, uri, VIDMODE_ON);
 	if (err) {
 		warning("contact: ua_connect(%s) failed: %m\n",
 			uri, err);
@@ -111,7 +109,6 @@ static int cmd_message(struct re_printf *pf, void *arg)
 	const struct cmd_arg *carg = arg;
 	struct contact *cnt;
 	const char *uri;
-	struct pl pl;
 	int err = 0;
 
 	cnt = contacts_current(baresip_contacts());
@@ -120,9 +117,8 @@ static int cmd_message(struct re_printf *pf, void *arg)
 	}
 
 	uri = contact_uri(cnt);
-	pl_set_str(&pl, uri);
 
-	err = message_send(uag_find_requri(&pl), uri, carg->prm,
+	err = message_send(uag_find_requri(uri), uri, carg->prm,
 			   send_resp_handler, NULL);
 	if (err) {
 		(void)re_hprintf(pf, "contact: message_send(%s) failed (%m)\n",
--- a/modules/gtk/gtk_mod.c
+++ b/modules/gtk/gtk_mod.c
@@ -771,22 +771,30 @@ static gboolean status_icon_on_button_press(GtkStatusIcon *status_icon,
 
 int gtk_mod_connect(struct gtk_mod *mod, const char *uri)
 {
-	char *uric = NULL;
-	struct pl pl;
+	struct mbuf *uribuf = NULL;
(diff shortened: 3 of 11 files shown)

### CODE REVIEW

**Comment by juha-h** [comment]
> Appending host part should be handled in the application. Also menu will not append any host name now. If there are multiple choice, then the user should select one and give the complete SIP URI.

So if there is only one account in accounts file, say `foo@example.com`, and user dials `bar`, it will not get completed to `bar@example.com`?  Hope I misunderstood the above.

**Comment by cspiel1** [comment]
No, we want to get rid of this. This reduces the code complexity dramatically and concerns only modules menu and gtk in baresip.
We could keep `account_uri_complete()` in the API anyway if you want to use it in the Android App.

**Comment by juha-h** [comment]
I don't need `account_uri_complete()`, but it is sad if I need to always add the domain even in the case when there there is one account (my example in the above).  I think that it is very common use case of cli baresip app. Can't you by default in such a case always select the only account?

**Comment by alfredh** [comment]
what kind of domain is "office" ? is it possible to dial foo@office ?

We should take a look at RFC 3261 and see what is a valid SIP URI.

I think we can solve this by adding a new config:

```
  uri_complete  false|true
  append_domain false|true
```

If the config item is enabled, then the curren uri complete behaviour will be triggered.

**Comment by cspiel1** [comment]
> I don't need `account_uri_complete()`, but it is sad if I need to always add the domain even in the case when there there is one account (my example in the above). I think that it is very common use case of cli baresip app. Can't you by default in such a case always select the only account?

Yes, but only if it is a registered account. But this is what led to endless discussion. So we should never append it.

Maybe we can simplify the string operations similar to the new print function `uristr_prepend_sip()` proposed in the `re` PR and reduce the ugly code in `menu` and `gtk`, add the new config and checks. But do we really need this?

We think that appending the host part of a local account is always wrong. Thus a static setting makes not much sense.

**Comment by cspiel1** [comment]
Again though about it. Using `menu` by a user and by other SW is a difference. And append host part and also port number for a dial URI could be annoying for the user. He has the option to use contacts and configure them similar to the accounts.

But I'll try it with a setting. Maybe code could be hold clean anyway.

**Comment by juha-h** [comment]
I'm not sure that I understood everything in the above.  I would like one way or another to keep the current dial behavior, i.e., account's domain (if it exists) is appended to d string if d string does not include sip: or @.  baresip cli app becomes too cumbersome to use if @domain must always be included in dial string.

**Comment by cspiel1** [comment]
Okay, currently working on the configurable solution.
But let me simplify the code by putting the calls to `account_uri_complete()` to `ua.c`. Nothing will change if the application (menu, gtk, Android App, ...) always appends the host part. This reduces the code in `menu` and `gtk`.

**Comment by cspiel1** [comment]
The behavior is now:
- If `append_domain true` then `ua.c` does the completion,
- otherwise it isn't done.

Default is `yes`.

**Comment by juha-h** [comment]
I did some testing of this PR.  Dial with just userpart worked as expected:
```
>                               jh
test@test.tutpro.com: selected for request
call uri: jh
call: connecting to 'sip:jh@test.tutpro.com'..
```
But when I added scheme, it still thought that I dialed userpart:
```
>                           sip:jh
test@test.tutpro.com: selected for request
call uri: sip:jh
call: connecting to 'sip:jh@test.tutpro.com'..
```
It should have check from /etc/hosts and dial the IP address.

**Comment by cspiel1** [comment]
What exactly is your expected behavior? Should be `sip:xxx` be a complete SIP URI and `xxx` not?
Should we append the host part only if the given dial URI does not start with "sip:"?

**Comment by juha-h** [comment]
> What exactly is your expected behavior? Should be `sip:xxx` be a complete SIP URI and `xxx` not? Should we append the host part only if the given dial URI does not start with "sip:"?

Yes, if user includes scheme in dial string, then it indicates that the user has given complete Request-URI and it should be used as is.

**Comment by cspiel1** [comment]
okay, this makes sense. I'll add a commit.

**Comment by cspiel1** [comment]
A rebase was necessary because of API check. At least on my device.

**Comment by juha-h** [comment]
Thanks, now URI completion behaved as expected in my tests.

File: include/baresip.h [hunk 1]
```diff
@@ -894,7 +894,7 @@ struct ua   *uag_find(const struct pl *cuser);
 struct ua   *uag_find_msg(const struct sip_msg *msg);
 struct ua   *uag_find_aor(const char *aor);
 struct ua   *uag_find_param(const char *name, const char *val);
-struct ua   *uag_find_requri(const char *requri);
+struct ua   *uag_find_requri(const struct pl *requri);
```
alfredh, line 897: this breaks the API. any way to avoid this ?

modules/gtk/gtk_mod.c
```diff
@@ -771,30 +771,22 @@ static gboolean status_icon_on_button_press(GtkStatusIcon *status_icon,
 
 int gtk_mod_connect(struct gtk_mod *mod, const char *uri)
 {
-	struct mbuf *uribuf = NULL;
-	char *uri_copy = NULL;
+	char *uric = NULL;
+	struct pl pl;
```
alfredh, line 775: I think we should focus on menu.so and ignore gtk.so in this patch.

alfredh:
can you fix this conflict:

```
modules/menu/static_menu.c
```

include/baresip.h
cspiel1: We have `pl` from parsing the `carg->prm` with `re_regex()`. The change in the API makes sense in order to reduce the code lines. E.g.:
```diff
-       err = pl_strdup(&uri, &pluri);
-       if (err)
-               goto out;
-
        if (!ua)
-               ua = uag_find_requri(uri);
+               ua = uag_find_requri(&pluri);
```

modules/gtk/gtk_mod.c
cspiel1: The second API change is `account_uri_complete()` which simplifies the code like this:
```patch
-       uribuf = mbuf_alloc(64);
-       if (!uribuf)
-               return ENOMEM;
-
-       err = account_uri_complete(ua_account(mod->ua_cur), uribuf, uri);
-       if (err)
-               goto out;
-
-       uribuf->pos = 0;
-       err = mbuf_strdup(uribuf, &uri_copy, uribuf->end);
+       pl_set_str(&pl, uri);
+       err = account_uri_complete(ua_account(mod->ua_cur), &uric, &pl);
```

include/baresip.h
alfredh, line 897: In order to avoid breaking the API, perhaps add a new function:

```c
struct ua   *uag_find_requri_pl(const struct pl *requri);
```

modules/contact/contact.c
```diff
@@ -84,6 +84,7 @@ static int cmd_dial_contact(struct re_printf *pf, void *arg)
 {
 	struct contact *cnt;
 	const char *uri;
+	struct pl pl;
```
alfredh, line 87: The diff in contact.c is unnecessary, just because the API was broken.

src/account.c
```diff
@@ -1711,47 +1711,61 @@ const char *account_extra(const struct account *acc)
...
+	err |= mbuf_write_pl(mb, &trimmed);
 	if (!acc || err)
-		return err;
+		goto out;
 
-	/* Append domain if missing and uri is not IP address */
+	if (complete)
+		goto out;
 
+	conf_get_bool(conf_cur(), "append_domain", &append);
```
alfredh, line 1760: this logic should be outside libbaresip core

It should be a specific config item for "menu" module.

src/config.c
```diff
@@ -495,6 +495,7 @@ int config_print(struct re_printf *pf, const struct config *cfg)
 			 "call_local_timeout\t%u\n"
 			 "call_max_calls\t\t%u\n"
 			 "call_hold_other_calls\t%s\n"
+			 "append_domain\t\t%s\n"
```
alfredh, line 498: please move to menu config

test/account.c
```diff
@@ -80,73 +80,70 @@ int test_account(void)
...
-		const char *in;
-		const char *out;
+		struct pl in;
+		struct pl out;
 	} testv[] = {
 
-		{ "192.168.1.2",
-		  "sip:192.168.1.2" },
+		{ PL("192.168.1.2"),
+		  PL("sip:192.168.1.2") },
```
alfredh, line 88: can you please rewrite this to make minimal changes to the test code ?

cspiel1: I could leave wrapper function with the current parameter list for `account_uri_complete()` and `uag_find_requri()`. Then
- let test code unchanged
- let contacts and gtk unchanged
- menu will make use of new function and become simplified

src/account.c
cspiel1: Then this setting makes no sense any longer.
We will avoid the automatic host part append by appending the host part in the application if it makes sense for us, or prepend the scheme "sip:".

This PR turns into a cleanup PR for `menu`. Following PRs could also reduce the code for `gtk`.

juha-h:
There is new config variable append_domain, but I didn't find it used in the PR.

cspiel1:
> There is new config variable append_domain, but I didn't find it used in the PR.

The config setting was discussed away: https://github.com/baresip/baresip/pull/2163#discussion_r990942702
But I missed to remove it completely.

src/uag.c
```diff
@@ -1128,6 +1128,32 @@ struct ua *uag_find_requri(const char *requri)
...
+struct ua *uag_find_requri_pl(const struct pl *requri)
+{
+	char *uric;
+	struct ua *ua;
+	int err;
+
+	if (!pl_isset(requri))
+		return NULL;
+
+	err = pl_strdup(&uric, requri);
```
alfredh, line 1147: perhaps swap the logic so that the _pl version is the base version ?

then you can avoid strdup

test/account.c
```diff
@@ -100,7 +100,7 @@ int test_account_uri_complete(void)
 		  "sip:user@domain.com:5677" },
 
 		{ "sip:user",
-		  "sip:user@proxy.com" },
+		  "sip:user" },
```
alfredh, line 103: valid change ?

cspiel1: Yes, but a bad example. Will find a better one.

src/uag.c
cspiel1: Okay, I had this already and also for `account_uri_complete()`. Your right it simplifies the code. Will revert 917038fcbe292eb73c2abafd421deba0cffb1269.

modules/menu/static_menu.c
```diff
@@ -522,43 +516,28 @@ static int dial_handler(struct re_printf *pf, void *arg)
 			clean_number(uri);
 	}
 
+	pl_set_str(&pluri, uri);
 	if (!ua)
-		ua = uag_find_requri(uri);
+		ua = uag_find_requri_pl(&pluri);
```
alfredh, line 521: is this change needed ? 

the input param is already in "uri" ?

modules/menu/static_menu.c
```diff
@@ -627,6 +606,7 @@ static int cmd_dialdir(struct re_printf *pf, void *arg)
 	}
 
 	if (err) {
+		dname = pl_null;
```
alfredh, line 609: is this change related ?

modules/menu/static_menu.c
```diff
@@ -667,34 +643,33 @@ static int cmd_dialdir(struct re_printf *pf, void *arg)
...
 
 	if (menu->adelay >= 0) {
 		ua_set_autoanswer_value(ua, menu->ansval);
 		(void)ua_enable_autoanswer(ua, menu->adelay,
-				auto_answer_method(pf));
+					   auto_answer_method(pf));
 	}
 
 	re_hprintf(pf, "call uri: %s\n", uri);
-
```
alfredh, line 697: whitespace changes ...

src/account.c
```diff
@@ -1719,39 +1719,81 @@ const char *account_extra(const struct account *acc)
...
+	if (!strp || !pl_isset(uri))
 		return EINVAL;
 
 	/* Skip initial whitespace */
-	while (isspace(*uri))
-		++uri;
+	trimmed = *uri;
+	err = pl_ltrim(&trimmed);
+	if (err)
+		return err;
```
alfredh, line 1769: this seems to mix two types of changes; refactoring and logical changes.

even if more tedious to do, it is better to keep those changes in separate PR's

src/account.c
```diff
@@ -1719,39 +1719,81 @@ const char *account_extra(const struct account *acc)
...
+	err |= mbuf_write_pl(mb, &trimmed);
 	if (!acc || err)
-		return err;
+		goto out;
 
-	/* Append domain if missing and uri is not IP address */
+	if (complete)
+		goto out;
 
+	conf_get_bool(conf_cur(), "append_domain", &append);
```
alfredh, line 1788: config item used ?

src/ua.c
```diff
@@ -1248,8 +1248,8 @@ int ua_connect_dir(struct ua *ua, struct call **callp,
 	if (err)
 		goto out;
 
-	pl.p = (char *)dialbuf->buf;
-	pl.l = dialbuf->end;
+	mbuf_set_pos(dialbuf, 0);
+	pl_set_mbuf(&pl, dialbuf);
```
alfredh, line 1252: unrelated change. please make a separate PR for this.

src/ua.c
```diff
@@ -1490,7 +1490,7 @@ int ua_refer_send(struct ua *ua, const char *uri, const char *referto,
 			   ua_print_supported, ua,
 			   referto);
 	if (err) {
-		warning("ua: send options: (%m)\n", err);
+		warning("ua: send refer: (%m)\n", err);
```
alfredh, line 1493: unrelated change

src/uag.c
```diff
@@ -1031,34 +1031,43 @@ struct ua *uag_find_param(const char *name, const char *value)
...
  * @return User-Agent (UA) if found, otherwise NULL
  */
-struct ua *uag_find_requri(const char *requri)
+struct ua   *uag_find_requri(const char *requri)
+{
+	struct pl pl;
+
+	pl_set_str(&pl, requri);
+	return uag_find_requri_pl(&pl);
+}
```
alfredh, line 1040: this looks good!

test/account.c
```diff
@@ -99,8 +99,8 @@ int test_account_uri_complete(void)
 		{ "user@domain.com:5677",
 		  "sip:user@domain.com:5677" },
 
-		{ "sip:user",
-		  "sip:user@proxy.com" },
+		{ "sip:office.local",
+		  "sip:office.local" },
```
alfredh, line 103: please keep the old testcase, and add office.local as an extra case

alfredh:
in general this PR contains a mix of many things.

The description no longer matches the content of the PR.


The PR contains many different logical changes, my advice is to split them up into smaller PRs.
Even if the patch is is just 2 lines of cleanup, it is a nice patch because it is straight to the point.

test/account.c
cspiel1: The original test case has the problem that it will fail. We decided together to change the behavior.

The URI "sip:user" is a bad example, because `user` would be in this case a domain, not a user.

PR #2165, opened 2022-09-28
sndfile Module - filename includes strm->cname (i.e. call->local_uri)…

ninp0:
… && strm->peer values (i.e. call->peer_uri) to derive source and destination of recorded call

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -64,7 +64,8 @@ struct stream {
 	struct menc_sess *mencs; /**< Media encryption session state        */
 	struct menc_media *mes;  /**< Media Encryption media state          */
 	enum media_type type;    /**< Media type, e.g. audio/video          */
-	char *cname;             /**< RTCP Canonical end-point identifier   */
+	char *cname;             /**< RTCP Canonical incoming end-point     */
+	char *peer;              /**< RTCP Canonical outgoing end-point     */
 	char *mid;               /**< Media stream identification           */
 	bool rtcp_mux;           /**< RTP/RTCP multiplex supported by peer  */
 	bool terminated;         /**< Stream is terminated flag             */
@@ -145,6 +146,7 @@ static void stream_destructor(void *arg)
 	mem_deref(s->bundle);  /* NOTE: deref before rtp */
 	mem_deref(s->rtp);
 	mem_deref(s->cname);
+	mem_deref(s->peer);
 	mem_deref(s->mid);
 }
 
@@ -709,6 +711,10 @@ int stream_alloc(struct stream **sp, struct list *streaml,
 	}
 
 	err = str_dup(&s->cname, prm->cname);
+	if (err)
+		goto out;
+
+	err = str_dup(&s->peer, prm->peer);
 	if (err)
 		goto out;
 
@@ -1515,6 +1521,38 @@ const char *stream_name(const struct stream *strm)
 		return NULL;
 
 	return media_name(strm->type);
+}
+
+
+/**
+ * Get the value of the stream->cname (i.e. call->local_uri)
+ *
+ * @param strm Stream object
+ *
+ * @return Name of stream type
+ */
+const char *stream_cname(const struct stream *strm)
+{
+	if (!strm)
+		return NULL;
+
+	return strm->cname;
+}
+
+
+/**
+ * Get the value of the stream->peer (i.e. call->peer_uri)
+ *
+ * @param strm Stream object
+ *
+ * @return Name of stream type
+ */
+const char *stream_peer(const struct stream *strm)
+{
+	if (!strm)
+		return NULL;
+
+	return strm->peer;
 }
 
 
--- a/src/stream.c
+++ b/src/stream.c
@@ -64,8 +64,8 @@ struct stream {
 	struct menc_sess *mencs; /**< Media encryption session state        */
 	struct menc_media *mes;  /**< Media Encryption media state          */
 	enum media_type type;    /**< Media type, e.g. audio/video          */
-	char *cname;             /**< RTCP Canonical incoming end-point     */
-	char *peer;              /**< RTCP Canonical outgoing end-point     */
+	char *cname;             /**< RTCP Canonical end-point identifier   */
+	char *peer;              /**< RTCP Canonical end-point identifier   */
 	char *mid;               /**< Media stream identification           */
 	bool rtcp_mux;           /**< RTP/RTCP multiplex supported by peer  */
 	bool terminated;         /**< Stream is terminated flag             */
@@ -714,9 +714,11 @@ int stream_alloc(struct stream **sp, struct list *streaml,
 	if (err)
 		goto out;
 
-	err = str_dup(&s->peer, prm->peer);
-	if (err)
-		goto out;
+	if (prm->peer) {
+		err = str_dup(&s->peer, prm->peer);
+		if (err)
+			goto out;
+	}
 
 	/* Jitter buffer */
 	if (prm->use_rtp && cfg->jbtype != JBUF_OFF && cfg->jbuf_del.max) {
@@ -1525,11 +1527,11 @@ const char *stream_name(const struct stream *strm)
 
 
 /**
- * Get the value of the stream->cname (i.e. call->local_uri)
+ * Get the value of the RTCP Canonical end-point identifier
  *
  * @param strm Stream object
  *
- * @return Name of stream type
+ * @return Canonical end-point identifier
  */
 const char *stream_cname(const struct stream *strm)
 {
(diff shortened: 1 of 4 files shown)


cspiel1:
Please do not delete the PRs and open new one each iteration!
The correct way is to push your changes by:
- `git push --force` in case of a rebase was needed
- simply adding commits and pushing without force

You don't have to change anything at the PR itself.

src/stream.c
```diff
@@ -64,7 +64,8 @@ struct stream {
 	struct menc_sess *mencs; /**< Media encryption session state        */
 	struct menc_media *mes;  /**< Media Encryption media state          */
 	enum media_type type;    /**< Media type, e.g. audio/video          */
-	char *cname;             /**< RTCP Canonical end-point identifier   */
+	char *cname;             /**< RTCP Canonical incoming end-point     */
+	char *peer;              /**< RTCP Canonical outgoing end-point     */
```
cspiel1, line 68: update the Doxygen

src/stream.c
```diff
@@ -1518,6 +1524,38 @@ const char *stream_name(const struct stream *strm)
 }
 
 
+/**
+ * Get the value of the stream->cname (i.e. call->local_uri)
```
cspiel1, line 1528: "Get the value of the RTCP Canonical end-point identifier"
Remove the hint about `call->local_uri`

src/stream.c
```diff
@@ -1518,6 +1524,38 @@ const char *stream_name(const struct stream *strm)
...
+ *
+ * @param strm Stream object
+ *
+ * @return Name of stream type
```
cspiel1, line 1532: @return Canonical end-point identifier

src/stream.c
```diff
@@ -1518,6 +1524,38 @@ const char *stream_name(const struct stream *strm)
...
+{
+	if (!strm)
+		return NULL;
+
+	return strm->cname;
+}
+
+
+/**
+ * Get the value of the stream->peer (i.e. call->peer_uri)
```
cspiel1, line 1544: "Get the value of the peer name or identifier"

cspiel1:
The code looks good now. Please update the mentioned Doxygen strings, then we come close to the merge.

alfredh:
```
[ RUN      ] test_ua_options
[81](https://github.com/baresip/baresip/actions/runs/3146850922/jobs/5120376303#step:4:82)
selftest: re_main() loop timed out -- test hung..
[82](https://github.com/baresip/baresip/actions/runs/3146850922/jobs/5120376303#step:4:83)
selftest: ASSERT_EQ: /home/runner/work/baresip/baresip/baresip-win32/baresip/test/ua.c:707: options_resp_handler(): expected=-1874543620(0x9044bbfc), actual=0(0x0)
[83](https://github.com/baresip/baresip/actions/runs/3146850922/jobs/5120376303#step:4:84)
TEST_ERR: /home/runner/work/baresip/baresip/baresip-win32/baresip/test/ua.c:816: (Unknown error [138])
[84](https://github.com/baresip/baresip/actions/runs/3146850922/jobs/5120376303#step:4:85)
test_ua_options: test failed (Unknown error [138])
[85](https://github.com/baresip/baresip/actions/runs/3146850922/jobs/5120376303#step:4:86)
test failed (Unknown error [138])
[86](https://github.com/baresip/baresip/actions/runs/3146850922/jobs/5120376303#step:4:87)
re main loop:
[87](https://github.com/baresip/baresip/actions/runs/3146850922/jobs/5120376303#step:4:88)
  maxfds:  64
[88](https://github.com/baresip/baresip/actions/runs/3146850922/jobs/5120376303#step:4:89)
  nfds:    29
[89](https://github.com/baresip/baresip/actions/runs/3146850922/jobs/5120376303#step:4:90)
  method:  2 (select)
```

cspiel1:
Yes. The peer is not set. There are two solutions.
- add a NULL pointer check before the `str_dup` in this PR
- I will create another PR that sets the `call->peer_uri` for incoming OPTIONS

src/stream.c
```diff
@@ -712,6 +714,10 @@ int stream_alloc(struct stream **sp, struct list *streaml,
 	if (err)
 		goto out;
 
+	err = str_dup(&s->peer, prm->peer);
```
cspiel1, line 717:
```c
if (prm->peer) {
	err = str_dup(&s->peer, prm->peer);
	if (err)
		goto out; 
}
```

ninp0:
@cspiel1 - all set, PR submitted...thanks for you help on this, I really appreciate your feedback and guiding me through the codebase.  
@alfredh - a thanks goes out to you as well!

src/stream.c
```diff
@@ -65,6 +65,7 @@ struct stream {
 	struct menc_media *mes;  /**< Media Encryption media state          */
 	enum media_type type;    /**< Media type, e.g. audio/video          */
 	char *cname;             /**< RTCP Canonical end-point identifier   */
+	char *peer;              /**< RTCP Canonical end-point identifier   */
```
cspiel1, line 68: pls update Doxygen!

src/stream.c
```diff
@@ -1518,6 +1526,38 @@ const char *stream_name(const struct stream *strm)
...
+{
+	if (!strm)
+		return NULL;
+
+	return strm->cname;
+}
+
+
+/**
+ * Get the value of the RTCP Canonical end-point identifier
```
cspiel1, line 1546: here also!

src/stream.c
```diff
@@ -1518,6 +1526,38 @@ const char *stream_name(const struct stream *strm)
...
+ *
+ * @param strm Stream object
+ *
+ * @return Canonical end-point identifier
```
cspiel1, line 1550: and here!

cspiel1:
> @cspiel1 - all set, PR submitted...thanks for you help on this, I really appreciate your feedback and guiding me through the code base. @alfredh - a thanks goes out to you as well!

your welcome! :-)

alfredh:
can you wait with merging this until after 2.8.0 has been released ?

The 2.8.0 release is scheduled for tomorrow (Saturday 1. Oct 2022).

I dont think we should squeeze in any last minute changes before the release.
Many times this can lead to unforseen bugs.

juha-h:
Not really related to this PR, but when I tested it I noticed that `/insmod sndfile` during the call does not add the module to the audio pipelines.  Would baresip architecture in theory allow that?

cspiel1:
> Not really related to this PR, but when I tested it I noticed that `/insmod sndfile` during the call does not add the module to the audio pipelines. Would baresip architecture in theory allow that?

Currently the filters are allocated only at startup of the streams.

juha-h:
OK, so it would not be possible to add `Rec` button to the call interface like Linphone has.

alfredh:
> Not really related to this PR, but when I tested it I noticed that `/insmod sndfile` during the call does not add the module to the audio pipelines. Would baresip architecture in theory allow that?

insmod will add the module to the current list of modules/filters.
The new filter will be used in any new calls.

It will not modify any existing calls.

We can add this feature, but it should not be done on the insmod/rmmod level.
Does all the calls want the new filter ?

it is better to explicitly add a command, such as:

```
/add_filter sndfile
```

cspiel1:
Currently not. This would need a PR. If you look in `audio.c`: The `baresip_aufiltl()` is used only in `audio_start()`. And this is called only here:
```c
	if (!tx->ausrc) {
		err |= audio_start(a);
	}
```

and here
```c
	if (!rx->auplay)
		err |= audio_start(a);
```

But then in the `aufilt_setup()` we have:
```c
	if (list_isempty(&tx->filtl))
		update_enc = true;
...
	if (list_isempty(&rx->filtl))
		update_dec = true;
```
So it is done only once. This would need a small to medium re-design.

sreimers:
Maybe its easier to load the `sndfile` filter and register a module specific record enable/disable command.

cspiel1:
Yes, and this command has to manipulate the RX/TX filter list, with locks set:
```c
		mtx_lock(tx->mtx);
		list_append(&tx->filtl, &encst->le, encst);
		mtx_unlock(tx->mtx);
```

sreimers:
Sry a bit misleading, I meant that this could a `sndfile` specific command/operation, to check if recording is wished. We can add a module config to record all calls by default (`auto`) or `manual`. In `manual` mode the recording is prepared but not written and sndfile does nothing in decode/encode.

This needs no changes to the filter loading.

cspiel1:
This is much better!

src/stream.c
ninp0: @cspiel1 , I'm not entirely clear about the request - should I change the contents of the comments to reflect something specific or run a doxygen command to generate update documentation?

src/stream.c
ninp0: @cspiel1 , same question...I'm not entirely clear about the request - should I change the contents of the comments to reflect something specific or run a doxygen command to generate update documentation?

src/stream.c
ninp0: @cspiel1 , same question...I'm not entirely clear about the request - should I change the contents of the comments to reflect something specific or run a doxygen command to generate update documentation?

cspiel1:
This was missing: https://github.com/baresip/baresip/pull/2202

PR #2169, opened 2022-09-29
log: optional timestamps

cspiel1:


alfredh:
what about also enabling timestamps for libre subsystem:

```c
/** Debug flags */
enum dbg_flags {
	DBG_NONE = 0,                 /**< No debug flags         */
	DBG_TIME = 1<<0,              /**< Print timestamp flag   */
	DBG_ANSI = 1<<1,              /**< Print ANSI color codes */
	DBG_ALL  = DBG_TIME|DBG_ANSI  /**< All flags enabled      */
};
```

cspiel1:
Didn't notice this feature. I will try to make use of it.

cspiel1:
The `re` logging is not used. I set it anyway.

Also I found that disabling the colored output might be helpful. We had this request also in my company. Can we add this to the PR?

PR #2170, opened 2022-09-29
cmake: add netroam module

robert-scheck:


modules/netroam/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+project(netroam)
+
+set(SRCS netroam.c)
+
+if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+    set(SRCS netroam.c netlink.c)
```
sreimers, line 6:
```suggestion
    list(APPEND SRCS netlink.c)
```

cmake/modules.cmake
```diff
@@ -160,6 +160,9 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   list(APPEND MODULES avcapture)
   list(APPEND MODULES coreaudio)
 endif()
+if(UNIX)
+  list(APPEND MODULES netroam)
+endif()
```
sreimers, line 165: I think `netroam` poll timer should work on all platforms.

sreimers:
Thanks!

PR #2180, opened 2022-09-29
cmake: synchronize behaviour of -DSHARE_PATH="…" with GNU Makefiles

robert-scheck:
The current implementation leads with `cmake` to `-DSHARE_PATH=\"\"` which leads to an empty string after `audio_path` in a freshly generated `.baresip/config` - which is wrong. According to the GNU Makefiles, it should be by default something like `/usr/share/baresip` (if not overriden).

Patch behaviour (backwards compliant) is to
- use override via `SHARE_PATH`, otherwise
- use distribution default via `SHARE_INSTALL_PREFIX`, otherwise
- use defaults from `cmake --install` during sound files installation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,9 +83,11 @@ include_directories(
 
 link_directories(/usr/local/lib)
 
-add_definitions(
-  -DSHARE_PATH="${SHARE_PATH}"
-)
+if(SHARE_PATH)
+  add_definitions(-DSHARE_PATH="${SHARE_PATH}")
+else()
+  add_definitions(-DSHARE_PATH="${CMAKE_INSTALL_FULL_DATADIR}/baresip")
+endif()
 
 if(DEFAULT_CAFILE)
   add_definitions(-DDEFAULT_CAFILE="${DEFAULT_CAFILE}")


CMakeLists.txt
```diff
@@ -83,9 +83,15 @@ include_directories(
...
-add_definitions(
-  -DSHARE_PATH="${SHARE_PATH}"
-)
+if(SHARE_PATH)
+  add_definitions(-DSHARE_PATH="${SHARE_PATH}")
+elseif(SHARE_INSTALL_PREFIX)
+  add_definitions(-DSHARE_PATH="${SHARE_INSTALL_PREFIX}/baresip")
+else()
+  add_definitions(
+    -DSHARE_PATH="${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR}/baresip"
```
sreimers, line 92:
```suggestion
    -DSHARE_PATH="${CMAKE_INSTALL_FULL_DATADIR}/baresip"
```

sreimers:
Looks like `SHARE_INSTALL_PREFIX` is a old variant (no references in modern CMake documentation) which is replaced by more generic GNUInstallDirs, is this needed, or is `CMAKE_INSTALL_FULL_DATADIR` enough?

CMakeLists.txt
robert-scheck: Ha, `CMAKE_INSTALL_FULL_DATADIR` was actually what I was looking for instead of `SHARE_INSTALL_PREFIX`; that simplifies the code even.

PR #2181, opened 2022-09-30
cmake: synchronize behaviour of -DMOD_PATH="…" with GNU Makefiles

robert-scheck:
The current implementation with `cmake` doesn't allow to set `MOD_PATH` nicely. Patch behaviour (backwards compliant) is to

  * use override via `MOD_PATH`, otherwise
  * use distribution default via `LIB_INSTALL_DIR`, otherwise
  * do nothing (which leads to defaults in C code, like previously)


sreimers:
> * use distribution default via `LIB_INSTALL_DIR`, otherwise

Looks like `LIB_INSTALL_DIR` is very specific? CMake provides with GNUInstallDirs loaded `CMAKE_INSTALL_LIBDIR` this looks more generic.

robert-scheck:
Unfortunately, `CMAKE_INSTALL_LIBDIR` contains only `lib64` (or `lib`) rather than `/usr/lib64` or `/usr/lib`.

robert-scheck:
Ah, `CMAKE_INSTALL_FULL_LIBDIR` seems to do the trick…thank you! :)

PR #2184, opened 2022-09-30
 Save tx traffic for weak android devices

KORuL:


sreimers:
This is a hardware specific problem and only a workaround with many negative side effects for other users. Maybe the [intercom](https://github.com/baresip/baresip-apps/tree/main/modules/intercom) module helps in your use-case scenario.

alfredh:
for example when using Kamailio/RTP-proxy it is vital that client sends audio RTP packets
all the time, even if muted. Otherwise the NAT bindings will time out.

When using Media-NAT (e.g. ICE) this is handled by the transport protocol.
Also when using RTP/RTCP mux the RTCP packets will keep the NAT bindings open.

PR #2185, opened 2022-09-30
revert the commit that breaks the call to baresip-studio android

KORuL:


sreimers:
Please open a issue with more details first, so we can analyze your problem.

PR #2187, opened 2022-09-30
cmake: add ABI (soname) versioning

### LINKED ISSUE #2178: Upstream .so name versioning?
As of writing, there doesn't seem to be any upstream .so name versioning for `libbaresip.so` when building baresip using `cmake`.

Quoting from section "SONAME handling" from https://docs.fedoraproject.org/en-US/packaging-guidelines/#_downstream_so_name_versioning:

> The SONAME field is written to the shared object by linker, using (at least in case of `ld`) the `-soname SONAME` flags. This can be passed as an option to gcc like this:
>
> ```
> $ gcc $CFLAGS -Wl,-soname,libfoo.so.0.n -o libfoo.so.0.n
> ```
> 
> If you want to check if the SONAME field is set and what value it has, use the `objdump` command (from `binutils`):
>
> ```
> $ objdump -p /path/to/libfoo.so.0.n | grep 'SONAME'
> ```

Please note that this issue is also meant as very kind try to convince you as upstream to start versioning your .so library (as our Fedora Packaging Guidelines request).

robert-scheck:
Fixes #2178

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,8 @@ cmake_minimum_required(VERSION 3.10)
 
 project(baresip VERSION 2.7.0)
 
+set(PROJECT_SOVERSION 0) # bump if ABI breaks
+
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake)
 
 ##############################################################################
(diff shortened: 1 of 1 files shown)


robert-scheck:
Given baresip ships meanwhile, at least when built via `cmake`, the `/usr/include/baresip.h` header file with public interfaces, ABI versioning makes IMHO sense.

PR #2191, opened 2022-10-01
cmake: add mpa module

### LINKED ISSUE #2177: Please add mpa module to cmake
Please add mpa module to cmake :)

I am happy to try this myself (with the goal of a pull request), however I would like to get your feedback before: The mpa module currently requires `-ltwolame -lmp3lame -lmpg123 -lspeexdsp`. Would a single `cmake/FindMPA.cmake` be the desired result? Or should this end up in 4 different cmake modules, one per library? And who is the best to answer this, maybe @sreimers?

robert-scheck:
Fixes #2177


robert-scheck:
@sreimers, it would be outstanding if this pull request could make it into (today's?) v2.8.0, because it would allow me a full switch to `cmake` for re, rem and baresip for all Fedora and EPEL builds.

PR #2193, opened 2022-10-01
release v2.8.0

cspiel1:
depends on:
- https://github.com/baresip/re/pull/551
- https://github.com/baresip/rem/pull/92


cspiel1:
> Check ABI compatibility (only re and rem for now)

CMakeLists.txt got a `PROJECT_SOVERSION` yesterday. No counterpart in Makefile.

I'll inc ABI version because some enums got new values.

PR #2194, opened 2022-10-01
Merge stable release v2.8.0

cspiel1:


cspiel1:
@sreimers may I merge the stable?

sreimers:
> @sreimers may I merge the stable?

Merge and push should be fine too.

alfredh:
great work!

cspiel1:
ok, thanks!
But see the newest baresip PR! There is a little mistake with the baresip version number in the header file.

cspiel1:
@sreimers please could you complete the stable merge? I will look in the afternoon what exactly are your steps.

cspiel1:
Ah, nothing more? So thanks for guiding me!

sreimers:
Yes if there is no conflict, merge is enough. Only if we release from stable branch we have to select the correct branch on the release page (instead main).

PR #2196, opened 2022-10-01
baresip.h: bump BARESIP_VERSION

cspiel1:


cspiel1:
Maybe obsolete. But an empty CHANGELOG would be the alternative.

PR #2203, opened 2022-10-02
cmake: add pre version handling

sreimers:


CMakeLists.txt
```diff
@@ -72,6 +82,14 @@ if(CMAKE_C_COMPILER_ID MATCHES "Clang")
 endif()
 
 find_package(re CONFIG REQUIRED HINTS ../re/cmake)
+
+list(APPEND RE_DEFINITIONS
```
cspiel1, line 86: Are `RE_DEFINITIONS` taken from `re`?
Do we overwrite `VERSION` and `VER_xxx` here?

sreimers: RE_DEFINITIONS is loaded from `re/cmake/re-config.cmake`, and each project append version specific definitions.

PR #2220, opened 2022-10-04
call, menu: support display name for outgoing calls

cspiel1:
This enables parsing the display name out of the dial URI and report it via
UA events to the application.

### DIFF
--- a/modules/menu/static_menu.c
+++ b/modules/menu/static_menu.c
@@ -606,9 +606,9 @@ static int cmd_dialdir(struct re_printf *pf, void *arg)
 			"Audio & video must not be"
 			" inactive at the same time\n";
 
-	/* with display name */
+	/* full form with display name */
 	err = re_regex(carg->prm, str_len(carg->prm),
-		"[^ \t\r\n<]*[ \t\r\n]*<[^>]+>[ \t\r\n]*"
+		"[~ \t\r\n<]*[ \t\r\n]*<[^>]+>[ \t\r\n]*"
 		"audio=[^ \t\r\n]*[ \t\r\n]*video=[^ \t\r\n]*",
 		&dname, NULL, &pluri, NULL, &argdir[0], NULL, &argdir[1]);
 	if (err) {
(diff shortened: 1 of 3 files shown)


src/call.c
```diff
@@ -1119,6 +1119,9 @@ int call_connect(struct call *call, const struct pl *paddr)
 			err = pl_strdup(&call->peer_uri, &addr.auri);
 		}
 
+		if (pl_isset(&addr.dname))
+			pl_strdup(&call->peer_name, &addr.dname);
```
maximilianfridrich, line 1123: This is problematic because the display-name and userinfo part of a SIP URI are semantically different. Especially because display-names are often quoted strings which are no allowed in the userinfo part. Also, they can be completely different strings.

Setting the `peer_name` here is probably not needed anyways.

modules/menu/static_menu.c
```diff
@@ -605,9 +606,18 @@ static int cmd_dialdir(struct re_printf *pf, void *arg)
...
 	err = re_regex(carg->prm, str_len(carg->prm),
-		"[^ ]* audio=[^ ]* video=[^ ]*",
-		&pluri, &argdir[0], &argdir[1]);
+		"[^ \t\r\n<]*[ \t\r\n]*<[^>]+>[ \t\r\n]*"
+		"audio=[^ \t\r\n]*[ \t\r\n]*video=[^ \t\r\n]*",
+		&dname, NULL, &pluri, NULL, &argdir[0], NULL, &argdir[1]);
+	if (err) {
+		dname = pl_null;
+		err = re_regex(carg->prm, str_len(carg->prm),
+			       "[~ ]+ audio=[^ ]* video=[^ ]*",
```
maximilianfridrich, line 617: Is there a reason why the first group (`[~ ]+`) of this regex is different from the previous implementation (`[^ ]*`)?

src/call.c
maximilianfridrich: E.g. `/dialdir "2201" <2201@localhost> sendrecv` will result in BareSIP trying to call `<sip:"2201"@localhost>`.
And `/dialdir "Alice" <bob@localhost> sendrecv` will result in BareSIP trying to call `<sip:"Alice"@localhost>`.

cspiel1: We need to set `call->peer_name`. See `event.c`:
```c
		peerdisplayname = call_peername(call);
		if (peerdisplayname){
				err |= odict_entry_add(od, "peerdisplayname",
						ODICT_STRING, peerdisplayname);
		}
```

This solutions is taken from incoming call:
```c
	if (pl_isset(&msg->from.dname)) {
		err = pl_strdup(&call->peer_name, &msg->from.dname);
		if (err)
			return err;
	}
```

Parsing of the display name out of the SIP URI is done by `sip_address_decode()`.

modules/menu/static_menu.c
cspiel1: '~' is with `qesc` (might  mean "quote escaping"). The quotes at beging and end are removed.

src/call.c
maximilianfridrich: You're right, this line is not the problem. Nonetheless, the behavior is very concerning:

> `/dialdir "2201" <2202@localhost> sendrecv` will result in BareSIP trying to call `<sip:"2201"@localhost>`

cspiel1: Added a commit to support also the short form with display name:
```
/dialdir "2201" <2202@localhost> sendrecv
```

maximilianfridrich: Yes, this line is not problematic at all. I ran into a parsing issue that you have fixed - substituting the display name for the userpart was never intended. Looks good to me.

maximilianfridrich:
Looks good to me.

PR #2233, opened 2022-10-07
portaudio: add mediadev_add with mediadev driver fields

sreimers:


include/baresip.h
```diff
@@ -465,6 +465,14 @@ struct contact *contacts_current(const struct contacts *contacts);
...
 	struct le   le;
 	char  *name;
+	struct {
+		int host_idx;
+		int dev_idx;
+		bool default_src;
+		bool default_play;
+		uint32_t in_channels;
+		uint32_t out_channels;
+	} info;
```
alfredh, line 475: This can be split in src/play:

```c
struct {
    uint32_t channels;
    bool is_default;
} src, play;
```

... since the property types for AUSRC and AUPLAY are the same, most of the time ...

include/baresip.h
```diff
@@ -465,6 +465,14 @@ struct contact *contacts_current(const struct contacts *contacts);
 struct mediadev {
 	struct le   le;
 	char  *name;
+	struct {
+		int host_idx;
+		int dev_idx;
```
alfredh, line 470: host and device index is Portaudio specific.

This API is supposed to be generic...

sreimers: Maybe we can use `char *index` and `char *driver`, most audio APIs have some type of identifiers. Yes Portaudio is a bit special, since it provides multiple host drivers, but without populating these, its hard to tell the difference if a device is listed multiple times.

alfredh: perhaps split the API into generic and driver specific portions:

```c
/** Defines a media device */
struct mediadev {
	struct le   le;
	char  *name;

	/* Generic portion: */
	channels, default etc;
	...

	/* Module/driver specific: */
	int host_index;
	int device_index;
	...
};
```

PR #2267, opened 2022-10-17
account: do not complete dial URI if scheme is included

cspiel1:


alfredh:
great, thanks for your work!

PR #2268, opened 2022-10-17
menu: simplify URI complete

cspiel1:
- Introduce two new functions `account_uri_complete_strdup()` and `uag_find_requri_pl()`.
- These allow to simplify the code in `modules/menu/static_menu.c`.


cspiel1:
Thanks for your review!

PR #2281, opened 2022-10-23
make: detect and add swscale module in modules.mk

agorgl:


alfredh:
CMake is now the main build system.

The old GNU make build system will become deprecated and removed.

agorgl:
Official Arch Linux packages depend on make, I suppose they should update to cmake too.
I just made it in order for swscale module to be built in the next baresip release on Arch

PR #2285, opened 2022-10-26
sdl: small improvements

sreimers:
Recommended by: https://wiki.libsdl.org/SDL_VideoInit

### DIFF
--- a/modules/sdl/sdl.c
+++ b/modules/sdl/sdl.c
@@ -30,7 +30,7 @@ struct vidisp_st {
 };
 
 
-static struct vidisp *vid;
+static struct vidisp *vid = NULL;
 
 
 static void event_handler(void *arg);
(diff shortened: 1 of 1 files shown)


modules/sdl/sdl.c
```diff
@@ -140,6 +140,7 @@ static void destructor(void *arg)
 
 	/* needed to close the window */
 	SDL_PumpEvents();
+	SDL_VideoQuit();
```
alfredh, line 143: is it correct to quit video in the destructor ?

does it work to create windows 2,3,4 etc times ?

alfredh:
I think we can get some inspiration from the FFmpeg code:

https://github.com/FFmpeg/FFmpeg/blob/master/libavdevice/sdl2.c

modules/sdl/sdl.c
sreimers: My fault, will fix this this.

PR #2286, opened 2022-10-26
cmake: add APP_MODULES symlinks

sreimers:


sreimers:
Should be fine to merge now. It fixes APP Module linking at build stage.

PR #2298, opened 2022-11-02
ua: use sdp connection data instead origin

sreimers:
WebRTC based clients use always `IP4 127.0.0.1` or `IP4 0.0.0.0` as origin `<addrtype> <unicast-address>` for privacy reasons.

https://www.rfc-editor.org/rfc/rfc4566.html#section-5.2

```
 For privacy reasons, it is sometimes desirable to obfuscate the
   username and IP address of the session originator.  If this is a
   concern, an arbitrary <username> and private <unicast-address> MAY be
   chosen to populate the "o=" field, provided that these are selected
   in a manner that does not affect the global uniqueness of the field.
```

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -622,7 +622,7 @@ static int sdp_connection(struct mbuf *mb, int *af, struct sa *sa)
 
 	*af = AF_UNSPEC;
 	err = re_regex((char *)mbuf_buf(mb), mbuf_get_left(mb),
-		       "IN IP[46]+ [^ \r\n]+", &pl1, &pl2);
+		       "c=IN IP[46]1 [^ \r\n]+", &pl1, &pl2);
 	if (err)
 		return EINVAL;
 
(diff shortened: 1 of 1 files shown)


sreimers:
I wonder if sdp origin address is reliable at all?

alfredh:
it is probably better to use the SDP Connection field instead.

A valid SDP must have one or more connection fields, either one in the session level,
or one in the media levels.

sreimers:
@cspiel1 can you check if [sdp connection-address](https://www.rfc-editor.org/rfc/rfc4566.html#section-5.7) works for you? it looks like its used only for preferred address family detection and should more reliable than origin.

WebRTC based clients use always ` IP4 127.0.0.1` or `IP4 0.0.0.0` and as origin `<addrtype> <unicast-address>` for privacy reasons.

cspiel1:
Okay, I will test it. I also think that this is an improvement.

cspiel1:
I tested with:
- IPv4
- IPv6
- IPv6 link local

- The logging `ua: using connection-address ... of SDP offer` was printed like expected.
- Audio+Video works in both directions.

sreimers:
Thanks

PR #2311, opened 2022-11-04
Stop segfaulting when no URI is passed to dial command

SimonHyde-BBC:
Without these changes a dial command without a parameter causes baresip to segfault.


alfredh:
thank you!

PR #2316, opened 2022-11-07
fixes 2315 and GTK errors on quit

### LINKED ISSUE #2315: Regression: Using GTK3 UI 'Dial…' leads to segmentation fault with v2.9.0
Using GTK3 UI `Dial…` leads to segmentation fault (input phone number, press "Call", segmentation fault) with v2.9.0. Given this worked until (including) v2.8.2, this has been quite likely introduced commit 9a9c3eabc88098d4be67f555c68363f445fcc48f – @cspiel1?

Note: I'm using `gtk_clean_number yes`, if that matters.

Ah, and the phone number I put into is not a secret: `0800-1071023` (Vodafone Germany test number for phones and DTMF, only reachable from Vodafone customers). The backtrace also should not contain anything that I consider a secret (possibly some RFC 1918 IP addresses, username/password from a test environment).

mbattista:
Closes #2315 
Small fix of errors shown when closing baresip via GTK

### DIFF
--- a/modules/gtk/gtk_mod.c
+++ b/modules/gtk/gtk_mod.c
@@ -131,9 +131,6 @@ static void menu_on_quit(GtkMenuItem *menuItem, gpointer arg)
 	struct gtk_mod *mod = arg;
 	(void)menuItem;
 
-	gtk_widget_destroy(GTK_WIDGET(mod->app_menu));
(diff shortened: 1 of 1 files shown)


sreimers:
Thanks!

PR #2320, opened 2022-11-08
acc,stream: add rtcp_mux account param

sreimers:


juha-h:
I noticed that new account param rtcp_mux param was introduced and thought that it would be useful to add it to my Android app.  But then I noticed that no API functions were introduced to get/set value of the parameter.

PR #2328, opened 2022-11-14
opus_multistream: update mimetype to ad-hoc standard

alfredh:


alfredh:
https://gitlab.freedesktop.org/gstreamer/gst-plugins-good/-/blob/master/gst/rtp/gstrtpopuspay.c

PR #2330, opened 2022-11-15
TLS server support SNI based certificate selection

cspiel1:
For incoming SIP requests a certificate is selected by SNI in the TLS client hello.

Multiple certificates can be configured by means of the account parameter `;cert=`.

relates to: https://github.com/baresip/re/pull/596

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -1066,6 +1066,7 @@ static const char *autoans_header_name(enum answer_method met)
 int ua_alloc(struct ua **uap, const char *aor)
 {
 	struct ua *ua;
+	struct uri *luri;
 	char *buf = NULL;
 	int err;
 
@@ -1126,6 +1127,15 @@ int ua_alloc(struct ua **uap, const char *aor)
 			return err;
 		}
 
+		luri = account_luri(ua->acc);
+		err = tls_add_certf(uag_tls(), ua->acc->cert,
+				    luri ? &luri->host : NULL);
+		if (err) {
+			warning("uag: SIP/TLS add server "
+				"certificate %s failed: %m\n",
+				ua->acc->cert, err);
+			return err;
+		}
 	}
 
 	err = create_register_clients(ua);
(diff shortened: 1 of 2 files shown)


alfredh:
can you describe how incoming calls are working in your setup ?

Are you using SIP outbound or are you making an incoming TCP-connection also for incoming calls ?

cspiel1:
Good question! Because for incoming calls there is always the question if the INVITE is sent through the same TCP connection that is opened for the REGISTER, or is the INVITE sent to default port 5061.

This PR is related to the second case. Another examples is an OPTIONS request that is sent by e.g. Asterisk in order to check if the UAC is still alive. These out-of-dialog requests are (by default) sent to the port that is given in the Contact header.

E.g. Asterisk has a setting `rewrite_contact` that ignores the port number in the Contact header and replaces this by the source port number of the REGISTER. We think this setting should be enabled for NAT setups. But not for LAN setups.

alfredh:
does it matter to Asterisk which certificate you use for the OPTIONS reply ?

to me it sounds like a lot of work (TLS + SNI + Certs) just to answer incoming OPTIONS from Asterisk.

alfredh:
it is fine to add this PR, I am just curious about your network setup.

I would like to suggest that we implement this properly in re first.
Please add a test using SNI in retest.


After this work is done we can make PR in baresip

cspiel1:
Should we move this to v2.12.0?

alfredh:
I think that is up to you ... no objections from my side

alfredh:
looks like your merge broke the build:

```
/home/alfredh/git/baresip/src/ua.c: In function ‘ua_alloc’:
/home/alfredh/git/baresip/src/ua.c:1132:28: warning: passing argument 3 of ‘tls_add_certf’ from incompatible pointer type [-Wincompatible-pointer-types]
 1132 |         luri ? &luri->host : NULL);
      |                            ^
      |                            |
      |                            struct pl *
```

alfredh:
I had to revert your commit, since it was breaking the build.

Please make a new PR for this patch ...

cspiel1:
Okay. Here it is: https://github.com/baresip/baresip/pull/2399

PR #2333, opened 2022-11-17
Add OPTIONS handling for webrtc demo

RenSym:
The WebRTC demo currently does not handle OPTIONS http requests. This works fine as long as you only use the demo's given frontend on port 9000.

To see how well Flutter WebRTC works with our backend using BareSIP, I made a Flutter program that connects to the BareSIP WebRTC demo backend.

Since this is CORS, and some browsers send "preflight requests", an OPTIONS request before each of certain types of requests. I had to add this commit to enable PUT/PATCH requests.

Network traffic in firefox looks like this when communicating with the demo:
grafik

Tested on Firefox
Works on Chrome (Doesn't seem to do preflight requests)

This PR fixes that.

### DIFF
--- a/webrtc/src/demo.c
+++ b/webrtc/src/demo.c
@@ -248,6 +248,13 @@ static void http_req_handler(struct http_conn *conn,
 			return;
 		}
 	}
+    else if (0 == pl_strcasecmp(&msg->met, "OPTIONS")) {
+        http_reply(conn, 204, "OK",
+                   "Content-Length: 0\r\n"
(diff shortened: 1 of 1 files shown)


sreimers:
`test/ccheck.py webrtc` reports:

```
./webrtc/src/demo.c:251: starts with 4 spaces, use tab instead
./webrtc/src/demo.c:252: starts with 8 spaces, use tab instead
./webrtc/src/demo.c:257: starts with 4 spaces, use tab instead
```

sreimers:
Thanks

PR #2336, opened 2022-11-18
mk: remove rtpext.c from srcs.mk

cspiel1:
Fixes old Makefile build


alfredh:
not sure if this is needed, this file be deleted in the next release ...?

sreimers:
I think we should test and fix Makefile build a last time for this release.

PR #2341, opened 2022-11-21
ci/build: switch to cmake build

cspiel1:
In github workflow for baresip-apps a dependency for avcodec is missing and not detected by cmake. This adds the avcodec component `hwcontext` to the cmake check.

Result in the baresip-apps workflow is now:
```
Could NOT find FFMPEG (missing: hwcontext) (found version "55.78.100")
```

I suppose also that the include for `hwcontext.h` is missing. But I am not sure.


cspiel1:
Because `avcodec.h` already contains
```c
#include "libavutil/hwcontext.h"
```
It should not be installed without `hwcontext.h`. Thus an extra check for component `hwcontext` is not required.

PR #2345, opened 2022-11-25
uag,call: do not override status code and reason

cspiel1:
Helps when someone wants to analyse unexpected call termination by the peer.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -1204,10 +1204,12 @@ void call_hangup(struct call *call, uint16_t scode, const char *reason)
 			sipsess_abort(call->sess);
 		}
 		else {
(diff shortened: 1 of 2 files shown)


alfredh:
hi @cspiel1 

can you please update the Milestone and indicate which release you want this in?

The next release is 2.10.0 due 7. Dec

alfredh:
I think the merge window for 2.10.0 is closed ?

during the 2 weeks before the release, we will only merge critical bug fixes to main.
is this a critical fix ?

cspiel1:
This is not critical. It helps when someone wants to analyse unexpected call terminations by the peer.
v2.11.0 is okay.

PR #2346, opened 2022-11-25
stream: send RTP NAT pinhole opener until RTP is received

cspiel1:
In a NAT setup SIP proxies may learn our external IP address by our sent RTP
packets. The dummy RTP packets that are sent initially when we have a receive-
only stream in order to open the NAT pinhole may be sent to early after SIP
signaling for a SIP proxy (e.g. kamailio). Or there might be multiple different
causes why NAT pinhole opening packets don't arrive at the SIP proxy.

This commit adds a timer with increasing interval that sends the pinhole opener
RTP packets until RTP data is received.


alfredh:
what if the endpoint only wants to send RTP, and not receive RTP ?

cspiel1:
Then `stream_open_natpinhole()` isn't called.

In `video.c:1197`:
```c
	if (dir & SDP_RECVONLY) {
		err |= stream_open_natpinhole(v->strm);
...
```

alfredh:
I changed the milestone to 2.11.0

please let me know if you agree

src/stream.c
```diff
@@ -1381,6 +1386,47 @@ static void update_menc_muxed(struct list *streaml, bool secure)
...
+	mb = mbuf_alloc(RTP_HEADER_SIZE);
+	if (!mb)
+		return;
+
+	tmr_start(&strm->tx.tmr_natph, phwait(&strm->tx),
+		  natpinhole_handler, strm);
+	mbuf_set_end(mb, RTP_HEADER_SIZE);
+	mbuf_advance(mb, RTP_HEADER_SIZE);
+
+	/* Send a dummy RTP packet to open NAT pinhole */
```
alfredh, line 1418: I think we should make this configurable on/off.

for deployments without NAT it is a bit annoying to see RTP ping packets in the trace.

cspiel1:
Default should be `yes` because it should not break existing installations that rely on the pinhole packets.

PR #2348, opened 2022-11-27
ctrl_tcp: fix re_net order

sreimers:


alfredh:
should we merge this before 2.10.0 ?

please add a milestone.

perhaps we should create a label for code that we want to merge during stable window ?

cspiel1:
A label for stable window would help.

alfredh:
is this change still needed if we move re_sock_t to re_types.h ?

PR #2352, opened 2022-12-07
Added rtcp_mux related API functions

juha-h:


src/account.c
```diff
@@ -1083,6 +1083,34 @@ int account_set_call_transfer(struct account *acc, const char *value)
...
 
+/**
+ * Sets rtcp_mux on (value "yes") or off (value "no")
+ *
+ * @param acc      User-Agent account
+ * @param value    "yes" or "no"
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int account_set_rtcp_mux(struct account *acc, const char *value)
```
alfredh, line 1094: perhaps it is better to use type `bool` for value ?

src/account.c
```diff
@@ -1710,6 +1738,22 @@ const char *account_call_transfer(const struct account *acc)
...
 
 
+/**
+ * Get rtcp_mux capability of an account
+ *
+ * @param acc User-Agent account
+ *
+ * @return "yes" or "no"
+ */
+const char *account_rtcp_mux(const struct account *acc)
```
alfredh, line 1748: return `bool` instead

alfredh:
NOTE: Suggestion to remove global `rtcp_mux` from config

PR #2353, opened 2022-12-10
Removed rtcp_mux config variable

juha-h:


alfredh:
https://github.com/baresip/baresip/blob/main/include/baresip.h#L379

also remove variable

juha-h:
I'm not sure if the struct config_avt variable can be removed, since, for example, it is used in webrtc/main.c to tell than rtcp_mux needs to be enabled.

alfredh:
aha, you are right

PR #2366, opened 2022-12-23
menu: debounce tone playback with a timer

cspiel1:
There are different situations where the ring-/ringback tone is started and
after a few millisecond stopped again. Fast start-stop sequences of the
playback should be avoided. These are such situations:
- First 180 Ringing, then 183 Session progress.
- Callee has automatic call answer configured.

Separates also tones.c from menu.c.

### DIFF
--- a/modules/menu/tone.c
+++ b/modules/menu/tone.c
@@ -1,0 +1,76 @@
+/**
+ * @file tone.c  Signal tones
+ *
+ * Copyright (C) 2022 Commend.com - c.spielberger@commend.com
+ */
+#include <string.h>
+#include <re.h>
+#include <baresip.h>
+#include "tone.h"
+
+enum {
+	DEBOUNCE_DELAY =   20,
+};
+
+struct {
+	struct play *play;            /**< Current audio player state     */
+	int repeat;
+	char *filename;
+	enum Device device;
+	struct tmr tmr_play;
+} tone;
+
+
+void tone_init(void)
+{
+	memset(&tone, 0, sizeof(tone));
+}
+
+
+static void do_play(void *arg)
+{
+	enum Device device = tone.device;
+	struct config *cfg = conf_config();
+	struct player *player = baresip_player();
+	const char *play_mod = cfg->audio.alert_mod;
+	const char *play_dev = cfg->audio.alert_dev;
+	(void) arg;
+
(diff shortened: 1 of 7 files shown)


modules/menu/tone.c
```diff
@@ -0,0 +1,76 @@
...
+	DEBOUNCE_DELAY =   20,
+};
+
+struct {
+	struct play *play;            /**< Current audio player state     */
+	int repeat;
+	char *filename;
+	enum Device device;
+	struct tmr tmr_play;
+} tone;
```
alfredh, line 21: can this be made static ?

modules/menu/tone.h
```diff
@@ -0,0 +1,15 @@
+/**
+ * @file tone.h module menu internal API for playing tones
+ *
+ * Copyright (c) 2022 Commend.com - c.spielberger@commend.com
+ */
+
+enum Device {
```
alfredh, line 7: please use lowercase `enum device`

modules/menu/tone.h
```diff
@@ -0,0 +1,15 @@
+/**
+ * @file tone.h module menu internal API for playing tones
```
alfredh, line 2: can we reuse the existing `menu.h` ?

alfredh:
in order to avoid conflict with other modules when STATIC=1, please make sure that all
global symbols are prefixed with the name of the module (e.g. "menu").


```
$ nm menu.so  | grep " T "
000000000000b04f T dial_menu_register
000000000000b099 T dial_menu_unregister
000000000000bf2d T dynamic_menu_register
000000000000bf77 T dynamic_menu_unregister
00000000000076a5 T menu_callcur
0000000000006043 T menu_find_call
0000000000005fdd T menu_find_call_state
0000000000007538 T menu_get
000000000000784c T menu_param_decode
00000000000075c3 T menu_selcall
0000000000007715 T menu_ua_carg
0000000000007702 T menu_uacur
0000000000005e39 T menu_update_callstatus
000000000000b015 T static_menu_register
000000000000b034 T static_menu_unregister
```

cspiel1:
Thanks for review Alfred! While I was working on other/similar issues with parallel outgoing calls, I realized that we need a more intelligent logic for starting the playback tone resp. early audio. This logic will most likely make this PR obsolete. Sorry for the circumstances!

PR #2368, opened 2022-12-27
call,menu: send UPDATE only if allowed (RFC-3311)

cspiel1:
The check if UPDATE is allowed is missing.

On the other side, a local change of the early media processing is always allowed.


cspiel1:
Maybe we replace this by a "PRACK refactoring" PR of @maximilianfridrich .

cspiel1:
replaced by https://github.com/baresip/baresip/pull/2401

maximilianfridrich:
Note that #2401 doesn't change the logic of `set_video_dir` or `call_modify`, while this PR does. If we need this logic, we can consider re-opening this PR @cspiel1.

PR #2371, opened 2022-12-27
pulse: rename to pulse_simple.so

alfredh:


robert-scheck:
Just to understand that from downstream perspective: Is the `pulse_async` module meant to be the successor of `pulse` (which was later renamed to `pulse` again)?

PR #2381, opened 2022-12-28
pulse: rename pulse_async.so to pulse.so (default)

alfredh:


cspiel1:
Good idea!

We had it in testing/production already for a while. It seems to be stable.

PR #2383, opened 2022-12-28
remove pulse_simple.so -- use pulse.so

alfredh:
it should be enough with 1 pulse module, lets use the one that is using the advanced API


robert-scheck:
@alfredh, I guess `README.md` should be updated accordingly, too?

PR #2385, opened 2022-12-29
sndfile: add pid to filename

cspiel1:


modules/sndfile/sndfile.c
```diff
@@ -94,10 +96,11 @@ static SNDFILE *openfile(const struct aufilt_prm *prm,
 	const char *peer = stream_peer(strm);
 
 	(void)re_snprintf(filename, sizeof(filename),
-			  "%s/dump-%s=>%s-%H-%s.wav",
+			  "%s/dump-%s=>%s-%H-%s-%u.wav",
 			  file_path,
 			  cname, peer,
-			  timestamp_print, tm, enc ? "enc" : "dec");
+			  timestamp_print, tm, enc ? "enc" : "dec",
+			  getpid());
```
alfredh, line 103: will this work on Windows ?

I dont think we have CI for Windows so maybe we should try to set it up first

cspiel1: I don't know. Currently, I am not sure if I can help here. I asked my colleague.

You mean like in `re` the `mingw.yml`?

martgut: I did a quick test on Windows with gcc from MinGW-W64
Following code:
```
#include <stdio.h>
#include <sys/types.h>
#include <unistd.h>

int main()
{
  pid_t pid = getpid();
  printf("pid: %u", pid);
}
```
This compiles and runs without problem:
```
C:\git\gitmart\sandbox\c\pid>gcc main.c -o main

C:\git\gitmart\sandbox\c\pid>.\main
pid: 14212

```
Does this help?

alfredh: thanks, that is very nice

We also have to support Native Windows with Visual Studio compiler.

alfredh:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/posix-getpid?view=msvc-170

alfredh:
are you using two instances of baresip on the same machine ?

please note that you can also configure the path so that the dumpfiles goes into
separate directories:


```
# sndfile
snd_path               /tmp
```

cspiel1:
We will check if we can live with the directory solution.

cspiel1:
Yes, there are multiple instances with different config already lying in specific directories. The `snd_path` should be set differently. This makes the PR unnecessary.

PR #2386, opened 2022-12-29
audiounit: use C11 mutex

alfredh:


modules/audiounit/recorder.c
```diff
@@ -234,9 +233,7 @@ int audiounit_recorder_alloc(struct ausrc_st **stp, const struct ausrc *as,
 	if (err)
 		goto out;
 
-	err = pthread_mutex_init(&st->mutex, NULL);
-	if (err)
-		goto out;
+	mtx_init(&st->mutex, mtx_plain);
```
sreimers, line 236: `mtx_init` can fail, can be handled like this:

```
	err = mtx_init(m, mtx_plain) != thrd_success;
	if (err) {
		err = ENOMEM;
		goto out;
	}
```

or replaced by `mutex_alloc()`

alfredh: thanks, updated

PR #2395, opened 2022-12-31
Ci pylint

alfredh:
proposal to group all linters into one lint.yml


alfredh:
hi @robert-scheck 


there is a new Pylint warning in your script:


```
tools/fritzbox2baresip:78:14: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
```

not sure if you know about it ...

robert-scheck:
Thanks for the pointer, will submit a PR.

PR #2396, opened 2022-12-31
fritzbox2baresip: use open with explicitly specifying an encoding

robert-scheck:
As pointed out by @alfredh in https://github.com/baresip/baresip/pull/2395#issuecomment-1368219513, there is a warning using a newer `pylint` that should be addressed:

```
tools/fritzbox2baresip:78:14: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
```


alfredh:
thanks for the quick fix!

PR #2400, opened 2023-01-03
Async Video decode

sreimers:
Requires: https://github.com/baresip/re/pull/628 (avoids out of order decoding)

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -444,7 +444,7 @@ static void rtp_handler(const struct sa *src, const struct rtp_header *hdr,
  */
 int stream_decode(struct stream *s)
 {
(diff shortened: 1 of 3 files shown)


cspiel1:
I will study soon your proposal. Is it also possible to implement a "Real-Time audio RX thread" based on this, related to this somehow https://github.com/baresip/baresip/pull/1886?

sreimers:
> I will study soon your proposal. Is it also possible to implement a "Real-Time audio RX thread" based on this, related to this somehow #1886?

Maybe I replace this PR, with a normal Video RX thread. Using re_async in this case has some drawbacks and for Video RX NACK handling it would be even more complicated.

PR #2401, opened 2023-01-04
PRACK refactoring

maximilianfridrich:
- Do not allow answer if awaiting PRACK to a 1xx response with SDP.
- Refactor call_refresh_allowed to use new re function `sipsess_refresh_allowed`.

Related re PR:
- https://github.com/baresip/re/pull/630

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -59,7 +59,6 @@ struct call {
 	bool answered;            /**< True if call has been answered       */
 	bool got_offer;           /**< Got SDP Offer from Peer              */
 	bool on_hold;             /**< True if call is on hold (local)      */
-	bool early_confirmed;     /**< Early media confirmed by PRACK       */
 	struct mnat_sess *mnats;  /**< Media NAT session                    */
 	bool mnat_wait;           /**< Waiting for MNAT to establish        */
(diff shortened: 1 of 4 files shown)


src/call.c
```diff
@@ -1358,6 +1357,12 @@ int call_answer(struct call *call, uint16_t scode, enum vidmode vmode)
 		return 0;
 	}
 
+	if (sipsess_awaiting_prack(call->sess)) {
+		info("call: answer: can not answer because we are awaiting a "
+		     "PRACK to a 1xx response with SDP\n");
+		return 0;
```
cspiel1, line 1363: Shouldn't `call_answer()` return an error code here?

maximilianfridrich: Good find, thank you. What do you think would be appropriate to return? `EPERM`? `EAGAIN`?

cspiel1: Not sure. EAGAIN sounds not bad.

The other return 0 in a separate PR.

maximilianfridrich: Done.

The other fix is in #2405.

alfredh:
this looks fine to me. Please go ahead and merge patches to main

PR #2403, opened 2023-01-05
menu: ringback/early audio handling for parallel calls

cspiel1:
This PR introduces a logic for handling ringback tones and early media for multiple outgoing calls (parallel call feature).

- call,menu: no early audio if multiple outgoing calls
- menu: delayed play for ringback and ring tone

Follow up PR:
- These changes: https://github.com/baresip/baresip/pull/2368 . Check if SIP UPDATE is allowed directly in `call_modify()` and return err if not allowed.

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -336,9 +336,9 @@ static void play_ringback(const struct call *call)
 
 static void check_ringback(struct call *call)
 {
-	enum sdp_dir ardir = sdp_media_rdir(stream_sdpmedia(
-					    audio_strm(call_audio(call))));
-	bool ring = !(ardir & SDP_RECVONLY);
+	enum sdp_dir adir = sdp_media_dir(stream_sdpmedia(
+					  audio_strm(call_audio(call))));
+	bool ring = !(adir & SDP_RECVONLY);
 
 	if (ring && !menu.ringback &&
 	    !menu_find_call(active_call_test, NULL)) {
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -23,6 +23,7 @@ static struct menu menu;
 
 enum {
 	MIN_RINGTIME = 1000,
+	TONE_DELAY   =   20,
 };
 
 
@@ -333,9 +334,27 @@ static void play_ringback(const struct call *call)
 }
 
 
-static void play_resume(const struct call *closed)
+static void check_ringback(struct call *call)
+{
+	bool ring;
+	enum sdp_dir ardir;
+
+	ardir = sdp_media_rdir(stream_sdpmedia(audio_strm(call_audio(call))));
+	ring = (ardir & SDP_RECVONLY) ? false : true;
+	if (ring && !menu.ringback &&
+	    !menu_find_call(active_call_test, NULL)) {
+		play_ringback(call);
+	}
+	else if (!ring) {
+		menu_stop_play();
+	}
+}
+
+
+static void delayed_play(void *arg)
 {
 	struct call *call = menu_callcur();
+	(void) arg;
 
 	switch (call_state(call)) {
 	case CALL_STATE_INCOMING:
@@ -343,11 +362,10 @@ static void play_resume(const struct call *closed)
 		break;
 	case CALL_STATE_RINGING:
 	case CALL_STATE_EARLY:
-		if (!menu.ringback && !menu_find_call(active_call_test,
-						      closed))
-			play_ringback(call);
+		check_ringback(call);
 		break;
 	default:
+		menu_stop_play();
 		break;
 	}
 }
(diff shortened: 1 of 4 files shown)


src/call.c
```diff
@@ -2322,12 +2337,12 @@ static void sipsess_progr_handler(const struct sip_msg *msg, void *arg)
...
 
 	if (media) {
-		call_stream_start(call, false);
 		call_event_handler(call, CALL_EVENT_PROGRESS, call->peer_uri);
+		call_stream_start(call, false);
 	}
 	else {
-		call_stream_stop(call);
 		call_event_handler(call, CALL_EVENT_RINGING, call->peer_uri);
+		call_stream_stop(call);
```
alfredh, line 2345: Not sure about this.

The handler may deref the call object.
When returning it will operate on a invalid call object.

modules/menu/menu.c
```diff
@@ -23,6 +23,7 @@ static struct menu menu;
 
 enum {
 	MIN_RINGTIME = 1000,
+	TONE_DELAY   =   20,
```
alfredh, line 26: perhaps add a comment about which unit is used

src/call.c
cspiel1: Right, I have to analyze old commits. We had this problem once. There might be a general problem if an event handler is allowed to deref the call. Especially, if there are multiple event handlers.

cspiel1: Would it be a good idea to:
```c
	mem_ref(call);
	if (media) {
		call_event_handler(call, CALL_EVENT_PROGRESS, call->peer_uri);
		call_stream_start(call, false);
		mem_deref(call);
	}
	else {
		call_event_handler(call, CALL_EVENT_RINGING, call->peer_uri);
		call_stream_stop(call);
		mem_deref(call);
	}
```

cspiel1: This https://github.com/baresip/baresip/pull/2403/commits/16a28ce26b77a4d39d1dd5caf66f145996f8cf0a ensures that the `call_stream_start()` does not operate on a destroyed call object.

modules/menu/menu.c
```diff
@@ -117,6 +118,35 @@ static char *errorcode_key_aufile(uint16_t scode)
...
+
+	if (!call_is_outgoing(call))
+		return;
+
+	ardir = sdp_media_rdir(stream_sdpmedia(audio_strm(call_audio(call))));
+	if (ardir & SDP_RECVONLY) {
+		call_set_audio_ldir(call, ardir & SDP_SENDONLY);
+		/* TODO: 100rel used and allowed */
+/*                if (call_refresh_allowed(call))*/
+/*                        call_modify(call);*/
```
alfredh, line 145: perhaps uncomment the code or remove commented out code before commit

modules/menu/menu.c
```diff
@@ -304,21 +334,38 @@ static void play_ringback(const struct call *call)
...
 
-static void play_resume(const struct call *closed)
+static void check_ringback(struct call *call)
+{
+	bool ring;
+	enum sdp_dir ardir;
+
+	ardir = sdp_media_rdir(stream_sdpmedia(audio_strm(call_audio(call))));
+	ring = (ardir & SDP_RECVONLY) ? false : true;
+	if (ring && !menu.ringback &&
```
alfredh, line 344: adding empty newline before `if` makes the code more reable

modules/menu/menu.c
```diff
@@ -304,21 +334,38 @@ static void play_ringback(const struct call *call)
...
 
 
-static void play_resume(const struct call *closed)
+static void check_ringback(struct call *call)
+{
+	bool ring;
+	enum sdp_dir ardir;
+
+	ardir = sdp_media_rdir(stream_sdpmedia(audio_strm(call_audio(call))));
+	ring = (ardir & SDP_RECVONLY) ? false : true;
```
alfredh, line 343: `ardir` and `ring` can be declared and set at the same time

also:

`ring = !(ardir & SDP_RECVONLY)`

modules/menu/menu.h
```diff
@@ -32,6 +32,7 @@ struct menu{
 	int32_t adelay;               /**< Outgoing auto answer delay     */
 	char *ansval;                 /**< Call-Info/Alert-Info value     */
 	struct odict *ovaufile;       /**< Override aufile dictionary     */
+	struct tmr tmr_play;          /**< Tones play timer               */
```
alfredh, line 35: did you forget to call `tmr_cancel` ?

cspiel1: Indeed. I add it to `menu_stop_play()`.

alfredh:
what is the next step here ? I see that the PR is marked as "Draft".

cspiel1:
I added a commit: https://github.com/baresip/baresip/pull/2403/commits/4de3d4de237c68dd33706d342acbb5c2457dc780
And rebased on current main.

I'll repreat the tests for parallel outgoing calls. Then will mark as "ready for review".

cspiel1:
In a next PR: 
Check if SIP UPDATE is allowed directly in `call_modify()` and return err if not allowed. Or a level deeper, in `sipsess_modify()`? Or even in `sipsess_update()`?

@alfredh, @sreimers Let me know what you think about this!

PR #2404, opened 2023-01-05
video: add rx thread

sreimers:


alfredh:
which codec is this needed for ?

I think FFmpeg has its own decoding threads. That means that when you call `avcodec_send_packet`
it will copy the packet and add it to a queue, and the decoding work is done by a different thread.

https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/avcodec.h#L1466

sreimers:
In my measurements `avcodec_send_packet` has every ~100ms a latency between 1-2ms (720p, H264, thread_count=1 and 2Mbit/s), if there are multiple parallel video calls this can block the main loop multiple times.

cspiel1:
access to vrx:
- r+w in `video_stream_decode()` in `vrx_thread`. --> ok
- r+w in `stream_pt_handler()` calls `video_decoder_set()` in main thread --> mutex is used, ok
  - video.c:865 set `pt_rx` should be removed, it is out of the lock. In `video_decoder_set` the mutex is locked. Also if codec can't be found, it should be set.
- `stream_recv_handler()` calls `request_picture_update()` and reads+writes `vrx->tmr_picup` and `vrx->n_picup` in main thread
- `request_picture_update()` is also called from `video_stream_decode()` in `vrx_thread` (when a decode error occurs). --> NOK, maybe an `mqueue` should trigger this
- `video_debug()` reads some fields of `vrx` that are changed in the `vrx_thread` without a lock. I think here we don't want a lock. How can this be done save?
- `video_prinf()` reads `vrx.efps` in main thread and is written also in main thread. -->ok.

Thus most fields of `vrx` are r+w in the `vrx_thread`. Some fields are r+w in main thread. And some are r+w in `vrx_thread` and read in the main thread. Some are also r+w in both.

Is it possible to separate these completely such that `vrx` fields are only r+w in the `vrx_thread`?

cspiel1:
The async API is nice. But maybe not what exactly what we need here. Because we have already the main thread that could do non-decoding and video frame processing tasks. Anyway, this is not the main question.

The main question is how to avoid concurrency problems without extra locks.

Maybe we could extract a part from `vrx` that has to be read from the main event loop. Let's call it `vrx.shared`. A duplicate of `vrx.shared` could be passed via an `mqueue` to the main thread. E.g. to function `request_picture_update()`.

Regard write access to `vrx` in main thread.  E.g. `++vrx->n_picup;` in `request_picture_update()`:
Can we avoid this completely? Second solution: Another `mqeue` that is processed in the video RX thread.

sreimers:
I updated the PR:

- mqueue can handle only one direction (thread -> main loop). I'm using `thread_async` as a shared mqueue callback (no worker callback).
- `vidqdec->vrxc` is now used for shared statistics and format data (no locking needed for `video_debug()`).

TODO:

- Check filter (`vidfilt`) and display (`vidisp`) modules are thread safe too

cspiel1:
> mqueue can handle only one direction

Right, I updated the ASCII art image. https://github.com/baresip/baresip/pull/2404#issuecomment-1380405754

Perfect, I'll try to understand your changes.

src/video.c
```diff
@@ -180,6 +185,17 @@ struct vidqent {
 };
 
 
+struct vidqdec {
+	struct le le;
+	struct rtp_header *hdr;
+	struct mbuf *mb;
+	struct vrx vrxc;
```
cspiel1, line 192: Most of the vrx fields are not used in the `vrx_thread`. Should we separate those in another struct?
```c
vrxc->n_intra
vrxc->size
vrxc->fmt
vrxc->filtl  /* This maybe better taken from vrxo. */
vrxc->stats.disp_frames
vrxc->frames
vrxc->ts_recv
```
But what do think about the following proposal?
- Start an extra re polling loop for the rtp_listen socket in the `vrx_thread`.
- Thus RTP read, decoding, filter and display all is done in the `vrx_thread`.
- The `vrx_thread` "owns" the vrx object.
- Changes of `vrx` from main thread have to be reported by means of a `vrx_update` object. Only `vrx_update` has to be locked. It is copied before it is processed in the `vrx_thread`.
- The data reported from `vrx_thread` to main thread is reported with a `vrx_info` object to the main thread. Only `vrx_info` has to be locked. It is copied before it is processed in the main thread.

sreimers: > Most of the vrx fields are not used in the `vrx_thread`. Should we separate those in another struct?

Maybe within separate PR. 

> But what do think about the following proposal?
> Start an extra re polling loop for the rtp_listen socket in the vrx_thread

For this PR I would like to keep the scope small. But in general the most problematic part are the udp helpers (DTLS/SRTP etc.) I think. These helpers have `tmr` objects and need to be transferred to the dedicated `vrx_thread` re loop then. Or everything has to be setuped from `vrx_thread` directly. Maybe not the only part that needs to be handled for full thread safety here. Another problem is maybe BUNDLE (audio/video and RTP/RTCP share one RTP connection).

cspiel1: Ah, I see. I keep this in mind when starting with RX audio thread.

So let's then proceed with this video PR. The `vrxc` and `vrxo` helps to reduce long blocking locks now, right? The `lock_rx` is locked only short in both threads. So no decoding, display is done while the lock is hold.

While the `lock_dec` is hopefully not very often locked by the main thread.
-  Switching off the video RX might block while one (last) video frame is decoded and sent to the display.

More concretely:
- video_stop_display(). Will be called if the user switches off video. This could lead to a long blocking situation and cause an audio artifact.
- video_decoder_set(). Very seldom, only if codec changes.
- set_vidisp() only called by `video_start_display()`. Maybe will not be blocked by the decoding thread.

Maybe we could improve this in a follow up PR.

> Check filter (vidfilt) and display (vidisp) modules are thread safe too

Also with the `lock_dec` mutex? Updates of the filter list, decoder and display could be handled similar.

src/video.c
```diff
@@ -876,7 +942,20 @@ static void stream_recv_handler(const struct rtp_header *hdr,
...
+		return;
+
+	q->hdr = mem_ref((void *)hdr);
+	q->mb  = mem_ref(mb);
+
+	update_rtp_timestamp(&v->vrx.ts_recv, hdr->ts);
+
+	mtx_lock(v->vrx.lock_rx);
+	q->vrxc = v->vrx; /* copy vrx */
+	list_append(&v->vrx.recvq, &q->le, q);
```
alfredh, line 956: the queue should have an upper limit for maximum number of entries.

if the decode thread is slower than the main thread, it will start to accumulate
entries in the queue. After some time this will lead to extra delay for the video stream.

alfredh:
is it an option to use a worker thread for all incoming RTP traffic ?
the thread can be shared between audio+video and use `udp_thread_attach` to listen
directly on the port. We can also use multiple threads with SO_REUSE_ADDR
and let the kernel distribute the work evenly between them.

cspiel1:
We think there should be a worker thread for incoming RTP audio. And another one for incoming RTP video.

sreimers:
So lets see what needs to be done:

- `mnat/menc` (tcp-, turn-.. timers) sessions have to be allocated properly within the new `RX_THREAD`
- `sdp` handling needs to be thread safe
- each `RX_THREAD` needs its own `dnsc` instance (i think we can ignore dns server changes, since only needed for session start)

I will start with thread safe sdp handling, since we have already some small race condition issues. I think the easiest way is to hide all public sdp structs (like `sdp_format`) and make the api itself thread safe with setters/getters etc.

cspiel1:
> * `mnat/menc` (tcp-, turn-.. timers) sessions have to be allocated properly within the new `RX_THREAD`

Why?
Do you mean `mnat_handler()` in `call.c` has to be called from RX thread?

> * `sdp` handling needs to be thread safe

Okay.


> * each `RX_THREAD` needs its own `dnsc` instance (i think we can ignore dns server changes, since only needed for session start)

Why?
DNS is not directly needed for RTP. Normally it is done before the RTP stream starts. Then maybe for changes of the RTP stream, right?

Is it possible to serialize and pass changes for RTP processing from main thread to the RX thread?

What could change? In `update_audio()`/`update_video()` there is only the `..._decoder_set()` call which changes the decoder. When the decoder needs to be changed? If the payload type changes, the decoder has to change. This can be  done in the RX thread. What else can be changed for the RX processing while the RX thread is running?

sreimers:
Take a look at the TURN Module. The RTP UDP Helpers are setuped from the main thread. These involve DNS handling and start timers. The timers belong to the main thread. Like the turnc refresh timers. Or more complicated with ICE handling. I see only two solutions, allocate everything within the right thread, so udp helper related timers also belong to `RX_THREAD` or all mnat/menc handling have to be made itself thread-safe.

cspiel1:
### mnat
1. Would it be enough to make the `mnat_handler()` thread safe?
More concretely `mnat_handler()` calls `call_modify()`, calls `update_media()`, calls `update_video()` resp `update_audio()`. Which influence has mnat on a running RTP RX stream?

2. Now I understand the general problem of UDP helpers. turn.c has no `recv_handler()`. But ice has one. Does this mean the `helper_recv_handler()` of ice/comp.c calls the `mnat_handler()`?

3. The `menc` has a `recv_handler()` that does decryption. Of course this should be done in the RX thread.

sreimers:
> turn.c has no `recv_handler()`

but libre turnc has one:

https://github.com/baresip/re/blob/main/src/turn/turnc.c#L311

And the `refresh_request` timer accesses the `turnc`.

mnat_handler can also be called from the timer, right:

https://github.com/baresip/re/blob/main/src/turn/turnc.c#L79

cspiel1:
Thus finally any call of the `mnat_handler()`
- could be done from RX and from main thread, right?
- might change `laddr` or what else? This starts a new SDP offer/answer and this might restart the RTP streams.

cspiel1:
> I will start with thread safe sdp handling, since we have already some small race condition issues. I think the easiest way is to hide all public sdp structs (like `sdp_format`) and make the api itself thread safe with setters/getters etc.

Do you have already something? I am currently looking how the sdp API could be made thread safe. Starting with `sdp_media`. There are quite a lot `sdp_media_...` functions in the API. Should all of them get a mutex lock?

sreimers:
I still try to figure out the best way. It's not easy since there are many parts involved. To keep the scope smaller its maybe easier to only make all udp recv helper functions thread safe. This needs some extra test cases, to ensure that nothing will break in the future. Since timers are not executed very often, it is easy to overlook race conditions here.

Most critical part for sdp_media handling is `stream_pt_handler` I think.

cspiel1:
I will have a look on the `stream_pt_handler`.

For the moment I have a commit that adds a lock for r/w access to sdp_media ldir / rdir. Only for discussion. https://github.com/baresip/re/pull/684

sreimers:
superseded by: https://github.com/baresip/baresip/discussions/2445

PR #2406, opened 2023-01-07
Fixed account debug of mwi and call_transfer

juha-h:
Bug was result of commit https://github.com/baresip/baresip/commit/77c5dcaeeddc9d0cad7e893d9ea6e099d69e4182.


alfredh:
The next release is planned for 11. Jan so I guess the feature merge window is closed.

Since this is a bugfix I guess we should merge it ?

PR #2408, opened 2023-01-09
cmake: increase minimum required version

cspiel1:
There is a fix in cmake v3.13 for sub-folders. With v3.12.5 there are still
warnings like this for each module:

```
CMake Error at CMakeLists.txt:363 (install):
  install TARGETS given target "account" which does not exist in this
  directory.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@
 # Versioning
 #
 
-cmake_minimum_required(VERSION 3.10)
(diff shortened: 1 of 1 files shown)


cspiel1:
I think it should be in the release v2.11.0.

alfredh:
can you please update re,rem,retest so we all use the same version ?

please also update this page:

https://github.com/baresip/baresip/wiki/Supported-platforms#supported-versions-of-cmake

PR #2422, opened 2023-01-20
sndio: remove deprecated module

alfredh:
the module has no makefile and appears unused and unmaintained. openbsd users can use the portaudio module instead.


landryb:
@alfredh sorry i'm just seeing that now, but the OpenBSD port of baresip uses sndio. So i'd like to revive the module, because using pulse or portaudio is gross.... right now im working to update the port from 2.6 to 3.0 and switch the build to cmake.

PR #2427, opened 2023-01-24
magic: use C99 __func__ macro

alfredh:
Since we support C99 and later we can use the macro `__func__` directly.


cspiel1:
Should we merge this now? baresip doesn't build currently.

alfredh:
yes, please go ahead and merge it

PR #2431, opened 2023-01-27
audio: memleak if same decoder is set twice

cspiel1:


alfredh:
this is by design

the audio codec should
check the state

which module is leaking?

cspiel1:
Okay, you mean like in g722:
```c
	if (*adsp)
		return 0;
```

Didn't check the modules. The modules can be written by anybody so maybe it makes sense to do this check in the core. For now I'll close this PR.

PR #2444, opened 2023-02-08
audio: payload type handler thread safe

cspiel1:
Looking into the `stream_pt_handler()`:
- Changing the decoder should be done in the main thread
- Manipulation of `audio->telev` and calling the `audio->eventh()` should be done in the main thread.

This PR handles the second point.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -783,7 +783,7 @@ static int stream_pt_handler(uint8_t pt, struct mbuf *mb, void *arg)
 						  telev_work_destructor);
 		w->a  = a;
 		w->mb = mbuf_dup(mb);
-		re_thread_async(NULL, async_telev_event, w);
+		re_thread_async_main(NULL, async_telev_event, w);
 		return ENODATA;
 	}
 
--- a/src/audio.c
+++ b/src/audio.c
@@ -715,6 +715,20 @@ static void ausrc_error_handler(int err, const char *str, void *arg)
 }
 
 
+struct telev_work {
+	struct audio *a;
+	struct mbuf *mb;
+};
+
+
+static void telev_work_destructor(void *arg)
+{
+	struct telev_work *w = arg;
+
+	mem_deref(w->mb);
+}
+
+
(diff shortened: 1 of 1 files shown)


src/audio.c
```diff
@@ -751,7 +779,11 @@ static int stream_pt_handler(uint8_t pt, struct mbuf *mb, void *arg)
 
 	/* Telephone event? */
 	if (lc && !str_casecmp(lc->name, "telephone-event")) {
-		handle_telev(a, mb);
+		struct telev_work *w = mem_zalloc(sizeof(*w),
+						  telev_work_destructor);
+		w->a  = a;
+		w->mb = mbuf_dup(mb);
+		re_thread_async(NULL, async_telev_event, w);
```
sreimers, line 786: You can use `re_thread_async_main` now. If `RX_THREAD` runs a re poll loop, the callback is not executed by main thread.

cspiel1: Oh, I see.

cspiel1:
This is only one tiny step for the RX "real time" thread. I opened a discussion to have a central place for discussing, listing all steps that have to be done and also not have to be done to become thread safe with a new RX thread.
https://github.com/baresip/baresip/discussions/2445

alfredh:
Are you planning for a "socket" thread or a "decoder" thread ?

src/audio.c
```diff
@@ -715,6 +715,20 @@ static void ausrc_error_handler(int err, const char *str, void *arg)
 }
 
 
+struct telev_work {
+	struct audio *a;
+	struct mbuf *mb;
+};
```
alfredh, line 721: if there is going to be alot of these types, it might be better to define one struct
with event type and event payload union. example:

```c
struct work {
    enum event_type type;
    union {
        struct {
            int foo;
        } telev;
        struct {
            char bar;
        } pt;
   } u;
};
```

cspiel1:
Will be covered by: https://github.com/baresip/baresip/pull/2454

PR #2446, opened 2023-02-08
stream: RX thread - make receiver fields atomic

cspiel1:
- made all RX fields atomic because they are accessed also from main thread
- Move RTP timeout detection timer and timeout field to extra struct. They have
  not much to do with the RX thread

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -6,6 +6,7 @@
 #include <string.h>
 #include <time.h>
 #include <re.h>
+#include <re_atomic.h>
 #include <baresip.h>
 #include "core.h"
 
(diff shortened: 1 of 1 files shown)


sreimers:
for atomic we have to use explicit `re_atomic_rlx/re_atomic_rlx_set` helpers.

cspiel1:
Okay, I try to understand this. Maybe not all of these fields need this. I will check each of them again.

cspiel1:
- [ ] `ts_last` written by RX thread. Read and reset to zero in main thread. Also the RTCP handler (most likely main thread) writes to this data field. (concurrent access: high)  --> atomic
- [ ] `ssrc` written from main thread by the SDP offer/answer processing. Then written only once in RX thread and when the ssrc changes. Indirectly read by RTCP handler in video.c (concurrent access: low) --> atomic
- [ ] `pseq` read/written in `lostcalc()` in RX thread. Only initialized in main thread. --> no atomic
- [ ] `ssrc_set` similar like `ssrc` --> atomic
- [ ] `pseq_set` only RX thread. --> no atomic
- [ ] `rtp_estab` only RX thread. --> no atomic
- [ ] `enabled` written in main thread. Read in RX thread. (concurrency: low) --> atomic

cspiel1:
We decided to keep the development of the RX thread (https://github.com/baresip/baresip/discussions/2445) in the main branch. Deleted:
- PR: https://github.com/baresip/baresip/pull/2451
- release branch: audio_rxthread

src/stream.c
```diff
@@ -33,17 +34,21 @@ struct sender {
...
+	struct metric *metric;         /**< Metrics for receiving            */
+	struct jbuf *jbuf;             /**< Jitter Buffer for incoming RTP   */
+	RE_ATOMIC bool enabled;        /**< True if enabled                  */
+	RE_ATOMIC uint64_t ts_last;    /**< Timestamp of last recv RTP pkt   */
+	RE_ATOMIC uint32_t ssrc;       /**< Incoming synchronization source  */
+	RE_ATOMIC bool ssrc_set;       /**< Incoming SSRC is set             */
+	uint32_t pseq;                 /**< Sequence number for incoming RTP */
+	bool pseq_set;                 /**< True if sequence number is set   */
+	bool rtp_estab;                /**< True if RTP stream established   */
+};
```
alfredh, line 46: it looks like around half the struct needs locking. Instead of adding all these atomic fields,
it might be cleaner to just have one mutex that is valid for the whole struct.

A good example here is AUBUF where the API is thread-safe and all the mutex magic
happens inside the .c file

it might be better to create a new file receiver.c and hide the locking inside it,
and add functions for setters and getters.

cspiel1: Good idea! This will simplify to understand the thread safety. I'll try this. Although we should be careful to keep the lock time very short.

cspiel1: Maybe it would be also good for the understanding if the function that run in the RX thread are also moved to receiver.c. I'll try this.

cspiel1:
Replaced by: https://github.com/baresip/baresip/pull/2454

PR #2451, opened 2023-02-10
stream: RX thread - make receiver fields atomic

cspiel1:
- made concurrently used RX fields atomic
- Move RTP timeout detection timer and timeout field to extra struct. They have not much to do with the RX thread

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -6,6 +6,7 @@
 #include <string.h>
 #include <time.h>
 #include <re.h>
+#include <re_atomic.h>
 #include <baresip.h>
 #include "core.h"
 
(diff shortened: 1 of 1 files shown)


sreimers:
Duplicate? https://github.com/baresip/baresip/pull/2446

cspiel1:
This goes against the feature branch.

> I think a good starting point could be a experimental RX_THREAD config option. So we can test/verify with clang/gcc thread sanitizer which runtime race conditions are still happening. with good test coverage we should also be able to ensure that no new race condition bugs are introduced in the future.

I am going to add the RX_THREAD mode.  Will create a PR against https://github.com/baresip/baresip/tree/audio_rxthread. Then the sanitizer and all other tests will be applied. I expect that these test should fail in the feature branch. Can we force the merge without the tests are green? While we solve the problems step by step the feature branch will become green.

The development of the RX thread will need some month I guess. Do you prefer a different way we can handle this?

sreimers:
If RX_THREAD is a config/compile option marked as experimental this should fine (only for local testing). The problem with a feature branch is, that it needs maintenance (all unrelated PRs needs to be merged too). And if we want to merge this branch to main, we have many merge conflicts to resolve I guess (rebasing public branches is not an option). So maybe better to keep the work in main branch I think.

cspiel1:
ok, then I close this and delete the feature branch.

PR #2454, opened 2023-02-14
stream: RX thread - separate thread safe receiver

cspiel1:
- moved struct receiver to a new source file receiver.c
  - except: RTP timeout detection timer and timeout field
- move functions that will run in RX thread to receiver.c

jbuf thread safety needs: https://github.com/baresip/re/pull/693

TODOs:
- [x] start/stop RX thread
- [x] add config option
- [x] some callback function have to be passed to main thread: see "TODO: pass to main thread"
- [x] rtp_handler in audio.c is not touched so far

Testing:
We have to test everything with default setting and `rxmode thread`.
- [x] SDP changes: Switch on/off video. Change audio direction during call.
- [ ] payload type changes
  - [x] inbound DTMF (audio)
  - [ ] codec changes (audio/video)
- [ ] RTCP rx
  - [ ] RTCP_FIR / RTCP_PSFB for video
  - [ ] RTCP ua events
- [ ] RTPESTAB ua event
- [ ] NAT pinhole timer is stopped
- [ ] MNAT works
 - [ ] STUN
 - [ ] ICE
 - [ ] TURN
- [ ] menc works
  - [ ] srtp
  - [ ] dtls_srtp
  - [ ] gzrtp
- [ ] bundle works


cspiel1:
TODO:
- start/stop RX thread
- combine with https://github.com/baresip/baresip/pull/2444
- `stream_sock_alloc` in RX thread

sreimers:
> * `stream_sock_alloc` in RX thread

It's maybe easier to leave `stream_sock_alloc` in main thread for initialization and move socket if ready with `udp_thread_detach/udp_thread_attach`. for re turn this is fine with thread safe tmr handling: https://github.com/baresip/retest/blob/main/src/turn.c#L509

cspiel1:
> > * `stream_sock_alloc` in RX thread
> 
> It's maybe easier to leave `stream_sock_alloc` in main thread for initialization and move socket if ready with `udp_thread_detach/udp_thread_attach`. for re turn this is fine with thread safe tmr handling: https://github.com/baresip/retest/blob/main/src/turn.c#L509

Okay, I'll compare these two solutions.

cspiel1:
Then we could attach only the RTP socket and let the RTCP socket stay in the main polling loop. Very nice! We will have to pass the `rtcph` calls to the main thread for the RTCP multiplexed case anyway. Does this work?

cspiel1:
TODOs:
- some callback function have to be passed to main thread: see "TODO: pass to main thread"
- rtp_handler in audio.c is not touched so far

cspiel1:
It is running already. But the RTP/RTCP handlers are called from both threads. Is there a `udp_thread_dettach()` missing?

cspiel1:
The `rtcp_msg` has to be duplicated (a deep copy is needed). The current solution is to have a pointer to the original mbuf and use `rtcp_decode()` a second time. Any better idea?

cspiel1:
macos-11 build fails. This seems to be unrelated to the PR.

macos-11: pkg-config not found

macos-12: /usr/local/bin/pkg-config found

src/receiver.c
```diff
@@ -0,0 +1,668 @@
...
+	w = mem_zalloc(sizeof(*w), work_destructor);
+	w->type    = WORK_RTCP;
+	w->rx      = rx;
+	mbuf_set_pos(msg->mb, 0);
+	err = rtcp_decode(&w->u.rtcp, msg->mb);
+	mbuf_skip_to_end(msg->mb);
+	if (err)
+		mem_deref(w);
+	else
+		re_thread_async_main(NULL, async_work_main, w);
```
sreimers, line 97: If you want to cancel, this needs a identifier:
```
		re_thread_async_main_id((intptr_t)rx, NULL, async_work_main, w);
```

src/receiver.c
```diff
@@ -0,0 +1,668 @@
...
+	mtx_lock(rx->mtx);
+	if (rx->run) {
+		join = true;
+		rx->run = false;
+	}
+	mtx_unlock(rx->mtx);
+	if (join)
+		thrd_join(rx->thr, NULL);
+
+	re_thread_async_cancel(0);
```
sreimers, line 526:
```c
re_thread_async_main_cancel((intptr_t)rx);
```

src/receiver.c
```diff
@@ -0,0 +1,668 @@
...
+		stream_process_rtcp(rx->strm, msg);
+		return;
+	}
+
+	w = mem_zalloc(sizeof(*w), work_destructor);
+	w->type    = WORK_RTCP;
+	w->rx      = rx;
+	mbuf_set_pos(msg->mb, 0);
+	err = rtcp_decode(&w->u.rtcp, msg->mb);
+	mbuf_skip_to_end(msg->mb);
```
cspiel1, line 93: @sreimers Should we implement a deep copy function for `rtcp_msg` ?

cspiel1: This solution isn't correct anyway. In `rtp.c` there is the `rtcp_recv_handler()`
```c
...
	while (0 == rtcp_decode(&msg, mb)) {

		/* handle internally first */
		rtcp_handler(rs->rtcp, msg);

		/* then relay to application */
		if (rs->rtcph)
			rs->rtcph(src, msg, rs->arg);

		mem_deref(msg);
	}
```
So it is possible that there are multiple RCTP msg in the `mbuf`. A deep copy function would contain the long `switch` statement similar to `rtcp_decode()`.

sreimers: Is there a write within rtcp handlers or aren't they use rtcp_msg read only and we can simply make it const and ensure lifetime of rtcp_msg is long enough?

cspiel1: Deep copy for `rtcp_msg`: https://github.com/baresip/re/pull/698

cspiel1:
It seems that the mingw build does not checkout the newest `re` which contains already the `retest`.

cspiel1:
@sreimers Should `mnat_connected_handler()` be called from main thread?
- it sets some tx fields (Note: also in current `main` isn't thread safe if TX thread is used)
- indirectly calls `stream_start_mediaenc()` through `mnatconnh`
- calls the bundle `mnatconnh`

A lot of stuff which currently runs in main thread.

cspiel1:
Next commit separates a new `aureceiver.c` from `audio.c`. This tiny PR is a dependency in order to check if `aubuf` frames are compatible to the `auplay` without having to remember the sample format separately.
https://github.com/baresip/re/pull/737

cspiel1:
Newest commit simplifies the struct `aurpipe` and depends on https://github.com/baresip/re/pull/803

cspiel1:
rebased and fixed tests

cspiel1:
```
Thread 8 RX thread:
Invalid read of size 4
   at 0x48CB309: udp_read_handler (in /home/runner/work/baresip/re/build/libre.so.15.1.0)
    by 0x48A1AEF: fd_poll (in /home/runner/work/baresip/re/build/libre.so.15.1.0)
    by 0x48A1F67: re_main (in /home/runner/work/baresip/re/build/libre.so.15.1.0)
    by 0x174E5C: rx_thread (in /home/runner/work/baresip/baresip/build/test/selftest)
```

cspiel1:
This try with the checks valgrind reports a concurrency problem with `aufile_play`. Not sure how this is related to the RX thread.

cspiel1:
Sanitizer reports:
- `auplay_write_handler` vs `auplay_alloc`->`aufile_play_alloc`
- `rtcp_set_srate_rx` vs `rtcp_sess_rx_rtp`

cspiel1:
sanitizer:
- `rtcp_sess_rx_rtp`->`source_update_seq` vs `rtcp_sess_rx_rtp`->`source_update_seq` (audio vs video ??)
- `rtcp_sess_rx_rtp` -> `source_calc_jitter` vs `rtcp_sess_rx_rtp` -> `source_calc_jitter`

docs/examples/config
```diff
@@ -73,6 +73,7 @@ jitter_buffer_type	fixed		# off, fixed, adaptive
 jitter_buffer_delay	5-10		# frames
 rtp_stats		no
 #rtp_timeout		60
+#rxmode			default         # default,thread
```
alfredh, line 76: perhaps add prefix "rtp_" ?

include/baresip.h
```diff
@@ -314,6 +314,11 @@ enum audio_mode {
 	AUDIO_MODE_THREAD,           /**< Use dedicated thread          */
 };
 
+/** RTP receive mode */
+enum rx_mode {
+	RX_MODE_DEFAULT = 0,    /**< RTP RX is processed in main thread      */
+	RX_MODE_THREAD,         /**< RTP RX is processed in separate thread  */
+};
```
alfredh, line 321: perhaps rename "rx" to "receive".

include/baresip.h
```diff
@@ -1404,7 +1411,7 @@ struct stream_param {
 typedef void (stream_mnatconn_h)(struct stream *strm, void *arg);
 typedef void (stream_rtpestab_h)(struct stream *strm, void *arg);
 typedef void (stream_rtcp_h)(struct stream *strm,
-			     struct rtcp_msg *msg, void *arg);
+			     const struct rtcp_msg *msg, void *arg);
```
alfredh, line 1414: this change is not needed, prefer to keep it non-const for callbacks.

include/baresip.h
```diff
@@ -1426,6 +1433,7 @@ int  stream_start_mediaenc(struct stream *strm);
 int  stream_start_rtcp(const struct stream *strm);
 int  stream_enable(struct stream *strm, bool enable);
 int  stream_enable_tx(struct stream *strm, bool enable);
+int  stream_enable_rx(struct stream *strm, bool enable);
```
alfredh, line 1436: this function can return "void" ?

src/aucodec.c
```diff
@@ -74,3 +74,13 @@ const struct aucodec *aucodec_find(const struct list *aucodecl,
...
+
+
+int aucodec_print(struct re_printf *pf, const struct aucodec *ac)
+{
+	if (!ac)
+		return 0;
+
+	return re_hprintf(pf, "%s %uHz/%dch",
+			  ac->name, ac->srate, ac->ch);
+}
```
alfredh, line 86: could be a separate PR

src/core.h
```diff
@@ -442,3 +477,31 @@ struct media_track *mediatrack_lookup_media(const struct list *medial,
...
+uint64_t rx_ts_last(struct receiver *rx);
+void rx_set_ts_last(struct receiver *rx, uint64_t ts_last);
+void rx_flush(struct receiver *rx);
+void rx_set_enable(struct receiver *rx, bool enable);
+int rx_get_ssrc(struct receiver *rx, uint32_t *ssrc);
+int rx_debug(struct re_printf *pf, const struct receiver *rx);
+int rx_start_thread(struct receiver *rx, struct rtp_sock *rtp);
+void rx_mnat_connected_handler(const struct sa *raddr1,
+			       const struct sa *raddr2, void *arg);
+bool rx_running(const struct receiver *rx);
```
alfredh, line 507: perhaps "rx_" prefix is a bit short for the global namespace.

how about `rtp_receiver` or something like that ?


We should consider borrowing API names from the WebRTC JS api:
https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpReceiver

alfredh:
can you rebase this PR against HEAD pls?

alfredh:
Is it possible to rebase the PR now ?

cspiel1:
Okay! And I'll try to work in on your last suggestions the next days if I find time.

include/baresip.h
cspiel1: Compared with `stream_enable_tx`. It is in the API and there are cases where an error makes sense.
Both `stream_enable_tx()` and `stream_enable_rx()` doesn't evaluate the returned error code so far. --> Could be another PR, but is not related to this one.

cspiel1:
With last commit the tests are green. The last commit can be separated: https://github.com/baresip/baresip/pull/2637

cspiel1:
TODO:
- new sanitizer warnings
- these re/baresip warnings which does not let the tests fail:
```
srtp: failed to decrypt RTP packet with 34 bytes (Unknown error 217 [217])
srtp: failed to decrypt RTCP packet with 48 bytes (Unknown error 217 [217])
srtp: failed to decrypt RTCP packet with 48 bytes (Unknown error 217 [217])
```

cspiel1:
The newest sanitizer warning maybe can be solved with atomic: https://github.com/baresip/re/pull/865

cspiel1:
On my dev host I have different sanitizer warnings. I'll have to try with renamed PR: https://github.com/baresip/re/pull/865

cspiel1:
All call tests are ran in both RX modes now and are green. I will do more manual tests mentioned above in the description.

In parallel we are trying the branch now on our devices. Our audio experts then will evaluate the stability and latency of the audio processing.

cspiel1:
I forgot, there are warnings that should be solved first:
- `test_call_mediaenc` -> "srtp: failed to decrypt RTCP packet with 48 bytes"
- `test_call_webrtc` -> "srtp: recv: failed to decrypt RTCP-packet"
- `test_call_bundle` -> "srtp: failed to decrypt RTCP packet with 16 bytes"

cspiel1:
The "faulty" 48 bytes are also sent in the RX default mode. But they are not received by the peer. It seems that the listen socket is closed before.
In contrast to the RX thread mode where the 48 bytes are received at the RTCP socket and can not be decoded.

I try to analyze what these 48 bytes are.

CMakeLists.txt
```diff
@@ -166,6 +167,7 @@ set(SRCS
   src/peerconn.c
   src/play.c
   src/reg.c
+  src/receiver.c
```
alfredh, line 170: what is "received" ?

perhaps "rtprecv.c" is a better name ?

include/baresip.h
```diff
@@ -317,6 +317,11 @@ enum audio_mode {
 	AUDIO_MODE_THREAD,           /**< Use dedicated thread          */
 };
 
+/** RTP receive mode */
+enum receive_mode {
```
alfredh, line 321: add prefix e.g. "media" or "rtp".

include/baresip.h
```diff
@@ -317,6 +317,11 @@ enum audio_mode {
...
+	RX_MODE_DEFAULT = 0,    /**< RTP RX is processed in main thread      */
```
alfredh, line 322: rename DEFAULT to MAIN

include/baresip.h
```diff
@@ -396,6 +401,7 @@ struct config_avt {
 	bool rtp_stats;         /**< Enable RTP statistics          */
 	uint32_t rtp_timeout;   /**< RTP Timeout in seconds (0=off) */
 	bool bundle;            /**< Media Multiplexing (BUNDLE)    */
+	enum receive_mode rxmode;       /**< RTP RX processing mode */
```
alfredh, line 404: is it possible to use one method for audio, and a different mode for video ?

src/audio.c
```diff
@@ -117,61 +117,29 @@ struct autx {
...
- Processing decoder pipeline:
-
-       .--------.   .-------.   .--------.   .--------.
- |\    |        |   |       |   |        |   |        |
- | |<--| auplay |<--| aubuf |<--| aufilt |<--| decode |<--- RTP
- |/    |        |   |       |   |        |   |        |
-       '--------'   '-------'   '--------'   '--------'
-
- \endverbatim
- */
```
alfredh, line 134: this diff is quite large and intrusive.

It is difficult to review the whole patch

src/aureceiver.c
```diff
@@ -0,0 +1,625 @@
...
+
+       .--------.   .-------.   .--------.   .--------.
+ |\    |        |   |       |   |        |   |        |
+ | |<--| auplay |<--| aubuf |<--| aufilt |<--| decode |<--- RTP
+ |/    |        |   |       |   |        |   |        |
+       '--------'   '-------'   '--------'   '--------'
+
+ \endverbatim
+ */
+struct aurpipe {
```
alfredh, line 29: can be renamed to `struct audio_recv`

alfredh:
In general, the refactoring and the change combined is quite large.

is it an option to do the refactoring first, in small steps ?

alfredh:
looks like there are some warnings from Sanitizer.

Is there any updates in this patch ?

CMakeLists.txt
cspiel1: This comes from the `struct receiver` like currently in `main`. Should I rename only the file `receiver.c` to `rtprecv.c` and keep the simple name of the `struct receiver`?

include/baresip.h
cspiel1: Currently not. Same mode for audio and video.
This is simpler. And maybe this helps to make audio+video lip sync later.

cspiel1:
> In general, the refactoring and the change combined is quite large.
> 
> is it an option to do the refactoring first, in small steps ?

I'll try my best to break it into smaller pieces after the renaming like you mentioned and the fixes of the sanitizer warnings.
Also we are performing measurements and analysis of the jitter and latency during high CPU load. We need a clear statement and measure how much the RX thread improves these quality values.

include/baresip.h
cspiel1: I have to correct myself. For now I only worked on thread safety for audio processing. I think for now the RX thread for video has to be turned off.

src/stream.c
```diff
@@ -259,7 +259,8 @@ int stream_enable_rx(struct stream *strm, bool enable)
 	rx_set_enable(strm->rx, true);
 
 	if (strm->rtp && strm->cfg.rxmode == RX_MODE_THREAD &&
-	    strm->type == MEDIA_AUDIO && !rx_running(strm->rx))
+	    strm->type == MEDIA_AUDIO && !stream_bundle(strm) &&
```
sreimers, line 262: If `RX_MODE_THREAD` is configured but disabled for other reasons there should be at least a info message why I think.

cspiel1: okay. I'll add one.

In a final version we should make video thread safe also.

cspiel1:
This are my results on x86 with Full-HD video:

### RX_MODE_MAIN
       SW jitter: **4.19ms**
       maximum deviation from p-time: **29.28ms**

### RX_MODE_THREAD
       SW jitter: **0.91ms**
       maximum deviation from p-time: **3.98ms**

alfredh:
this PR is getting quite big.

Perhaps you can squash all commits and create a new fresh PR ?

cspiel1:
Okay. Will do this next week. Currently I am on vacation.

cspiel1:
Did only a fresh rebase. Now will try to extract a first re-factoring PR.

cspiel1:
This PR https://github.com/baresip/baresip/pull/2685 contains the changes for `stream.c`. It also introduces the new configuration option for the RX thread and the thread function.

 While `audio.c` stays untouched so far.
 
 Will close this PR. The branch will still exist.

PR #2457, opened 2023-02-16
main: fix async init order (after config load)

### LINKED ISSUE #2456: baresip main does not start on Debian
I noticed that rem was merged to re and built baresip using it.  Built went fine, but start of baresip fails:
```
$ baresip
baresip v2.12.0 Copyright (C) 2010 - 2022 Alfred E. Heggestad et al.
main: poll_method_set: already set
config: poll method (epoll) set: Invalid argument [22]
config: configure parse error (Invalid argument [22])
main: configure failed: Invalid argument [22]
ua: stop all (forced=1)
```
It started a couple of days ago when I tried the PRs.

sreimers:
Fixes #2456

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -119,8 +119,6 @@ int main(int argc, char *argv[])
 	if (err)
 		goto out;
 
-	re_thread_async_init(ASYNC_WORKERS);
-
 	tmr_init(&tmr_quit);
 
 #ifdef HAVE_GETOPT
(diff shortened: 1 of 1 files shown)


juha-h:
Is this still up to date?

https://github.com/baresip/baresip/wiki/Using-baresip-as-a-library

There is no tmr_init(&tmr_quit) nor re_thread_async_init(ASYNC_WORKERS)
calls.  Are those really needed?

sreimers:
`re_thread_async_init` is needed if you want explicit control, how many workers are started and when.

PR #2461, opened 2023-02-20
Client cert renegotiation in http

fAuernigg:
tls_set_verify_client has same functionality like tls_set_verify_client_trust_all. The old name was misleading for tls usages other than dtls and will be removed.


sreimers:
Looks like this PR is outdated, can you update this PR?

PR #2467, opened 2023-02-25
Dshow fixes

tomek-o:
1. Memory that is already freed is referenced if camera has no resolution equal to requested but best match is the last media offered by the device. Spotted when trying to use 640x480 resolution with OBS-Camera - unlike physical cameras I own this one list resolutions from highest to lowest and the last one, 640x360 is the best match.
2. If during format enumeration device would return something other than FORMAT_VideoInfo (but I'm not sure if/when it can actually happen) then media type memory is leaked.
3. This I believe is actually OBS-Camera issue, but checking buffer buffer size in BufferCB might prevent crash. I'm not sure if I've used the latest version of OBS / virtual camera, but it was reporting back incorrect video size (not the one set by the user) but then returning frames with the size specified by the user. In my test stream was set to 640x360 (SetFormat inside config_pin function) but camera returned back (GetFormat) information about 1920x1080 resolution. Actual frames in callback had 640x360 resolution resulting in buffer overrun / crash.

### DIFF
--- a/modules/dshow/dshow.cpp
+++ b/modules/dshow/dshow.cpp
@@ -151,8 +151,10 @@ public:
 		struct vidframe vidframe;
 		uint64_t timestamp = (uint64_t)(sample_time * VIDEO_TIMEBASE);
 		if (buf_len != buf_len_RGB32 * 4) {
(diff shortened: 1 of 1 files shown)


sreimers:
Thanks!

PR #2469, opened 2023-02-26
Include also params to MESSAGE URI

juha-h:


juha-h:
@cspiel1 If Alfred is busy, could you check also this one?  Now Request-URI of MESSAGE is constructed the same way as of INVITE, i.e., params are included if they exist.

PR #2470, opened 2023-02-26
Moved adding of custom headers from ua_connect_dir to ua_call_alloc

juha-h:
Now adding of the headers work either ua_connect or call_connect is used.

### DIFF
--- a/src/message.c
+++ b/src/message.c
@@ -212,7 +212,13 @@ int message_send(struct ua *ua, const char *peer, const char *msg,
 	if (err)
 		return err;
 
-	err = pl_strdup(&uri, &addr.auri);
+	if (pl_isset(&addr.params)) {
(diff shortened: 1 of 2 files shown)


cspiel1:
Looks good to me.

juha-h:
Thanks, can I merge the PR or should we wait for other comments?

cspiel1:
Did you test it?

I see no risk here that something breaks. It works for outgoing calls.

Incoming calls: In `sipsess_conn_handler()` the custom headers are set, but then overwritten by the filtered custom headers (from the remote). So the current behavior will not change and I think you may merge this if your tests are okay.

juha-h:
Christian Spielberger writes:

> Did you test it?

Yes, I did.  Customs headers get added to both INVITE and REGISTER
requests both when using ua_connect and when using ua_call_alloc +
call_connect.

PR #2473, opened 2023-03-01
call: Fix delayed (auto) answer if awaiting PRACK

maximilianfridrich:
On an incoming call, if autoanwer is configured and we have sent a 183 containing SDP, we have to await the PRACK before we are allowed to send a 200 OK ([RFC 3262](https://www.rfc-editor.org/rfc/rfc3262) section 3 - last paragraph).

With autoanswer configured, it could happen that we try to call `call_answer` while we are still awaiting a PRACK. This commit fixes this.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -59,6 +59,7 @@ struct call {
 	bool answered;            /**< True if call has been answered       */
 	bool got_offer;           /**< Got SDP Offer from Peer              */
 	bool on_hold;             /**< True if call is on hold (local)      */
+	bool ans_queued;          /**< True if an (auto) answer is queued   */
 	struct mnat_sess *mnats;  /**< Media NAT session                    */
 	bool mnat_wait;           /**< Waiting for MNAT to establish        */
 	struct menc_sess *mencs;  /**< Media encryption session state       */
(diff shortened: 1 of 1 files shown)


maximilianfridrich:
I hate to say it, but I was not able to reproduce the failing thread sanitizer check on my machine.

I am not as familiar with this new check, but the output seems to hint at fakevideo.c, not anything PRACK related.

maximilianfridrich:
Ok, the data race is because the `sdp_media->raddr` is written in `msg.c:media_decode()` (e.g. when an INVITE is received) by the main thread and is read in the fakevideo thread when sending a video in `video.c:encode_rtp_send()->packet_handler()`. In this function, the mutex `vtx->lock_tx` is locked before reading the `sdp_media->raddr`, but the main thread doesn't know anything about this mutex.

Still, this doesn't have anything to do with this small fix for the autoanswer handling.

sreimers:
Should be fixed by: https://github.com/baresip/baresip/pull/2474

maximilianfridrich:
Thanks for the fix @sreimers. I rebased and it is working now.

PR #2486, opened 2023-03-10
Send event when dump file is opened

juha-h:
In the next step, printing of the info message could be moved to menu module.


cspiel1:
Looks good to me.

juha-h:
Thanks for review.  I tested it and got the event.  I'll try during the next couple of days to create another PR to handle the event in menu module.

juha-h:
OK, could we then just remove the info call from sdnfile, since it is duplicate of the event?

PR #2489, opened 2023-03-11
add `net_af` config setting

juha-h:
Add `net_af` config setting similar to existing `net_interface`.  Possible values are `4` and `6`.


juha-h:
Wrong branch, need to make a new one.

PR #2490, opened 2023-03-11
config: add net_af config setting

juha-h:
Similar to `net_interface`.  Possible values are `4` and `6`.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -459,9 +459,9 @@ int config_parse_conf(struct config *cfg, const struct conf *conf)
 	(void)conf_get_str(conf, "net_interface",
 			   cfg->net.ifname, sizeof(cfg->net.ifname));
 	if (0 == conf_get(conf, "net_af", &pl)) {
-		if (0 == pl_strcasecmp(&pl, "4"))
+		if (0 == pl_strcasecmp(&pl, "ipv4"))
 			cfg->net.af = AF_INET;
-		else if (0 == pl_strcasecmp(&pl, "6"))
+		else if (0 == pl_strcasecmp(&pl, "ipv6"))
 			cfg->net.af = AF_INET6;
 		else {
 			warning("unsupported af (%r)\n", &pl);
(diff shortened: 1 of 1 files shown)


src/config.c
```diff
@@ -458,6 +458,15 @@ int config_parse_conf(struct config *cfg, const struct conf *conf)
 			    &cfg->net.use_getaddrinfo);
 	(void)conf_get_str(conf, "net_interface",
 			   cfg->net.ifname, sizeof(cfg->net.ifname));
+	if (0 == conf_get(conf, "net_af", &pl)) {
+		if (0 == pl_strcasecmp(&pl, "4"))
```
sreimers, line 462: Maybe `all` or `any` for default `AF_UNSPEC` would be nice too.

juha-h: Thanks for the comment.  If "any", then may be the other values would need to be "ipv4" and "ipv6", instead of "4" or "6".

BUT, net_interface does not have "default" value either.  In my app, there will be "any" option, but I don't think it is necessary in config.

juha-h: The default value will be explained in wiki/Configuration.

robert-scheck: From my point of view "ipv4" and "ipv6" are more self-explaining compared to "4" and "6".

juha-h: Done.

juha-h:
Can this PR be merged now or after something?

sreimers:
We are currently in feature freeze (2 weeks before release), release day is March 22nd. This PR can be merged after this.

juha-h:
Sebastian Reimers writes:

> We are currently in feature freeze (2 weeks before release), release
> day is March 22nd. This PR can be merged after this.

OK, thanks.

src/config.c
```diff
@@ -537,6 +546,7 @@ int config_print(struct re_printf *pf, const struct config *cfg)
 			 "\n"
 			 "# Network\n"
 			 "net_interface\t\t%s\n"
+			 "net_af\t\t%s\n"
```
sreimers, line 549: The corresponding `cfg->net.af` to string conversion is missing here.

juha-h: Fixed by https://github.com/baresip/baresip/pull/2490/commits/de4cd581857341d38f412fda57f6ba1aedccca3f

Issue #2497, opened 2023-03-17
One way video does not work with STUN


cspiel1:

alfredh:
hi,

could you please try one or more of the following:


- Run a long video call with one-way video, behind NAT. Does the video flow after e.g. 30 minutes ?

- Try to enable RTP/RTCP multiplexing.

- One way video is the same as one way audio, if the audio has no microphone packets, it cannot open the NAT hole.

- Is it possible for you to use ICE instead of STUN ? The STUN protocol for NAT traversal is broken by design.


I think @juha-h has a good setup for this with SER and NatHelper/RTP-Proxy.
Please check with him, but I think his setup is robust and reliable.


Alfred

juha-h:
I use rtpengine to handle media if one or both parties of a call is behind NAT.  rtpengine is sitting on public IP address and there is thus no need for STUN/TURN/ICE.

But rtpengine is not able to send media packets to a party behind NAT before it has received at least one packet from it.  That packets opens the NAT pinhole and tells rtpengine where (address/port) to send the packets to.

If media is one way, DTLS or ZRTP negotiation is two way and thus rtpengine still receives at least one packet from the party behind NAT.

cspiel1:
Thanks! So this will not help to open the NAT pinhole for one way audio or video.

Then I'll try ICE instead of STUN with the hope that this opens a NAT pinhole.

cspiel1:
Tried to reproduce the issue with `main` HEAD. But the STUN responses from `stun.linphone.org` lead to a call termination. Found that we use a patch internally. Here is a PR: https://github.com/baresip/re/pull/823

cspiel1:
> * Try to enable RTP/RTCP multiplexing.

The first sent RTCP packet opens the pinhole. Then incoming RTP passes. After that e.g. the H.264 decoder has to wait for the first keyframe. This not an acceptable solution.

> * One way video is the same as one way audio, if the audio has no microphone packets, it cannot open the NAT hole.

Exactly.

> * Is it possible for you to use ICE instead of STUN ? The STUN protocol for NAT traversal is broken by design.

What exactly do you mean? How such an account should be configured? Btw: We don't want to use a TURN server.

cspiel1:
I updated the PR. Pinhole opening is only started:
- if mnat is not configured (like before) and
- if STUN is configured

alfredh:
Please read about STUN and ICE and study the RFCs.


Here is a section from RFC 5389:

```
[...]

PR #2499, opened 2023-03-20
stream: send nat pinhole packet also if mnat is configured

cspiel1:
Fixes: #2497

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -1611,7 +1611,7 @@ int stream_open_natpinhole(struct stream *strm)
 	if (!strm)
 		return EINVAL;
 
-	if (!strm->mnat && strm->pinhole)
+	if (strm->pinhole)
 		tmr_start(&strm->tmr_natph, 10, natpinhole_handler, strm);
 
 	return 0;
--- a/src/stream.c
+++ b/src/stream.c
@@ -1611,7 +1611,8 @@ int stream_open_natpinhole(struct stream *strm)
 	if (!strm)
 		return EINVAL;
 
-	if (strm->pinhole)
+	if (strm->pinhole && (!strm->mnat ||
+			      !str_cmp(strm->mnat->id, "stun")))
 		tmr_start(&strm->tmr_natph, 10, natpinhole_handler, strm);
 
 	return 0;


cspiel1:
can we merge this?

cspiel1:
Didn't want to set the milestone. Instead was looking where the milestones can be planned to see when I should make the next release.

sreimers:
Milestone set for May, I think we should wait for feedback by @alfredh

alfredh:
regarding medianat, most of the modules we have should do opening of NAT pinholes.
So there should be no need for explicit packet.

Media NAT modules:

* stun
* turn
* ice
* natpmp
* pcp

Please also note that RTP/RTCP mux is the recommended solution from IETF
regarding NAT keepalives.

Adding more complexity to solve a problem, without understanding the underlying problem,
is wrong IMO. We should try to debug the problem and see if it happens for all medianat types,
or just a few.

cspiel1:
I'll try to figure out the problem with the STUN configuration.

cspiel1:
In our pcap we had different IP addresses for
- stun.linphone.org (51.89.71.129)
- sip.linphone.org (147.135.128.132)

Now nslookup tells 51.89.71.129 for both servers. We'll repeat the test again without the commit in this PR.

cspiel1:
We think that at least for STUN a dummy packet has to be sent to the peer to open the NAT pinhole. Because the stun module will not process the SDP offer, nor send something to the IP:port mentioned in it.

Issues currently are:
- Early video can't pass the NAT of the callee.
- For audio+video calls without early video, the callee may miss the key frame.

src/stream.c
```diff
@@ -1611,7 +1611,8 @@ int stream_open_natpinhole(struct stream *strm)
 	if (!strm)
 		return EINVAL;
 
-	if (strm->pinhole)
+	if (strm->pinhole && (!strm->mnat ||
+			      !str_cmp(strm->mnat->id, "stun")))
```
alfredh, line 1615: not sure about this. This means that the core gets knowledge about module-names.

I would prefer to delete the STUN module instead, and force ICE usage.

cspiel1: Ah, okay. Is ICE without a TURN server possible?
Does this open the NAT pinhole?

alfredh:
I think this PR is not needed.

Please use ICE instead of STUN. Please also try to understand *why* your setup is not working.

PR #2501, opened 2023-03-23
call,audio: respect SDP media dir on audio start similar to video

cspiel1:
Parallel call memory issue: #2502 - Part 1

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -143,40 +143,37 @@ static const struct sdp_format *sdp_media_rcodec(const struct sdp_media *m)
 static int start_audio(struct call *call)
 {
 	const struct sdp_format *sc;
-	int err = 0;
+	const struct sdp_media *m = stream_sdpmedia(audio_strm(call->audio));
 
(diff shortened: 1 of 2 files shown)


src/call.c
```diff
@@ -143,40 +143,37 @@ static const struct sdp_format *sdp_media_rcodec(const struct sdp_media *m)
 static int start_audio(struct call *call)
 {
 	const struct sdp_format *sc;
-	int err = 0;
+	const struct sdp_media *m = stream_sdpmedia(audio_strm(call->audio));
 
 	/* Audio Stream */
-	sc = sdp_media_rcodec(stream_sdpmedia(audio_strm(call->audio)));
+	sc = sdp_media_rcodec(m);
 	if (sc) {
```
sreimers, line 150: Unrelated but we can avoid this big if/else with:
```c
if (!sc) {
	info("call: audio stream is disabled..\n");
	return 0;
}
```

Issue #2502, opened 2023-03-23
Memory increases massively for multiple parallel outgoing calls


cspiel1:
The commit b7215edb5b548b25c82f03dec1e358c8b32ad15a already switches off early audio RX if there are multiple outgoing calls.

There was missing something:
- https://github.com/baresip/baresip/pull/2501
- https://github.com/baresip/baresip/pull/2503

sreimers:
Can this issue be closed?

PR #2503, opened 2023-03-23
menu: limit early audio TX streams

cspiel1:
Parallel call memory issue: #2502 - Part 2

- menu,config: limit the number of early audio streams
- menu: optimize count of outgoing calls

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -33,6 +33,7 @@ sip_tos			160 # See TOS fields!
 call_local_timeout	120
 call_max_calls		4
 call_hold_other_calls	yes
+#call_max_earlyaudio	32		# 0 deactivates the limit
 
 # Audio
 #audio_path		/usr/local/share/baresip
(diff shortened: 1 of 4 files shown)


cspiel1:
Q: Why a limit for early audio streams and not directly for the number of outgoing calls?
A: The number of simultaneous outgoing calls might be much higher than would be possible if early-audio is enabled.

TODOs in following PRs:
- limits for the number of simultaneous early video RX and TX
- Maybe also an overall limit for the number of outgoing calls. Currently only number of incoming calls are limited.

docs/examples/config
```diff
@@ -33,6 +33,7 @@ sip_tos			160 # See TOS fields!
 call_local_timeout	120
 call_max_calls		4
 call_hold_other_calls	yes
+#call_max_earlyaudio	32
```
sreimers, line 36: Looks like this is option is only related to menu, maybe better to use `menu_max_earlyaudio`

PR #2504, opened 2023-03-23
call: check if SIP UPDATE is allowed, but always update local media

cspiel1:
When call_modify() is called it is sometimes not allowed to send an UPDATE.
But the local media direction should be updated in any situation.


maximilianfridrich:
Looks good to me. `call_answer` and `sipsess_estab_handler` in call.c both call `update_media` as well, so the media directions should still work as expected once the call is established.

PR #2506, opened 2023-03-28
cmake: avoid include of /usr/local/include

cspiel1:
Including /usr/... is unsafe for cross-compilation.

In yocto build fixes the warning:
```
cc1: warning: include location "/usr/local/include" is unsafe for cross-compilation [-Wpoison-system-directories]
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,7 +101,6 @@ add_definitions(${RE_DEFINITIONS})
 include_directories(
   include
   src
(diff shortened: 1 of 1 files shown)


cspiel1:
macos build seems to depend on static `/usr/local/include` in CMakeLists.txt

sreimers:
Can you try to rebase, should be fixed now, some module include/library paths were wrong.

PR #2523, opened 2023-04-13
variadic function fixes

maximilianfridrich:
Fixes a few occasions of calls to variadic functions that could fail in certain instances.

I used an LLVM pass to check all of baresip and re where the last non-variadic argument passed to a variadic function is not a constant. Such function calls are potentially dangerous, so I checked those function calls manually for potential bugs/vulnerabilities. I have checked the 52 warnings generated by this pass, most of which were not an issue. Still, this is a snapshot in time and such problems could be introduced with any new code change.

### DIFF
--- a/src/sdp.c
+++ b/src/sdp.c
@@ -119,8 +119,8 @@ int sdp_decode_multipart(const struct pl *ctype_prm, struct mbuf *mb)
 		return err;
 
 	err = mbuf_strdup(mb, &buf, mbuf_get_left(mb));
-	if (err || !buf)
-		return err ? err : ENOMEM;
+	if (err)
+		return err;
 
 	/* find 1st boundary */
 	s = strstr(buf, bnd_str);
@@ -140,6 +140,5 @@ int sdp_decode_multipart(const struct pl *ctype_prm, struct mbuf *mb)
 	}
 
 	mem_deref(buf);
-
 	return 0;
 }
--- a/src/sdp.c
+++ b/src/sdp.c
@@ -4,6 +4,7 @@
  * Copyright (C) 2011 Alfred E. Heggestad
  */
 #include <stdlib.h>
+#include <string.h>
 #include <re.h>
 #include <baresip.h>
 #include "core.h"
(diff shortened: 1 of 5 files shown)


src/sdp.c
```diff
@@ -111,27 +114,32 @@ int sdp_decode_multipart(const struct pl *ctype_prm, struct mbuf *mb)
...
+	s = strstr(buf, bnd_str);
+
 	/* iterate over each part */
-	while (s.l > 2) {
-		if (re_regex(s.p, s.l, expr, &e))
-			return 0;
+	while (s) {
+		e = strstr(s + bnd.l, bnd_str);
+		if (!e)
+		break;
```
sreimers, line 132: indentation missing

maximilianfridrich: Done.

src/sdp.c
```diff
@@ -111,27 +114,32 @@ int sdp_decode_multipart(const struct pl *ctype_prm, struct mbuf *mb)
...
+		p.l = e - p.p - 2;
 
-		/* valid part in "p" */
 		decode_part(&p, mb);
 
 		s = e;
 	}
 
+	mem_deref(buf);
+
```
cspiel1, line 143: in my opinion this newline is too much

src/ua.c
```diff
@@ -1213,16 +1214,21 @@ static int append_params(struct mbuf *mb, struct pl *params)
 	if (err || !str)
 		return err ? err : ENOMEM;
 
+	err = mbuf_strdup(mb, &buf, mbuf_get_left(mb));
+	if (err || !buf)
```
cspiel1, line 1218: do we need both checks?

src/sdp.c
```diff
@@ -111,27 +114,32 @@ int sdp_decode_multipart(const struct pl *ctype_prm, struct mbuf *mb)
...
-		return ENOMEM;
-
-	/* find 1st boundary */
-	err = re_regex((char *)mbuf_buf(mb), mbuf_get_left(mb), expr, &s);
+	err = pl_strcpy(&bnd, bnd_str, sizeof(bnd_str));
 	if (err)
 		return err;
 
+	err = mbuf_strdup(mb, &buf, mbuf_get_left(mb));
+	if (err || !buf)
```
cspiel1, line 122: both checks? I guess that `if (err)` is enough

maximilianfridrich: True, removed.

PR #2555, opened 2023-04-28
sndio: re-add sndio module for OpenBSD

landryb:
mostly a revert of #2422 + add cmake plumbing to make it built only on OpenBSD. Module works fine in my basic testing of 3.1.0 (eg placing/receiving calls)

if linux or freebsd users want to use sndio, a proper `FindSndio.cmake` module will be needed.

@alfredh i can migrate the module to C11 threads if needed, per https://github.com/baresip/baresip/discussions/2374#discussioncomment-4625820 ?


landryb:
69bf55c4 shamelessly inspired by #2391, quick test shows it still works.

sreimers:
Thanks, if you would like to maintain the module, please re-add it here too: https://github.com/baresip/baresip/wiki/Modules

landryb:
> Thanks, if you would like to maintain the module, please re-add it here too: https://github.com/baresip/baresip/wiki/Modules

done, thanks _a lot_ for merging !

PR #2558, opened 2023-05-02
alsa: use atomic for run flag

cspiel1:


cspiel1:
There are some more such boolean flags. Do we really want atomic for bool? Does it have a benefit?

sreimers:
> There are some more such boolean flags. Do we really want atomic for bool? Does it have a benefit?

In most cases the atomic `run` bool has no big benefit (depends on platform and use-case), but it prevents race condition warnings like with clang thread sanitizer.

PR #2569, opened 2023-05-05
config: add different options for audio and video jitter buffer

sreimers:
Proposal for `jitter_buffer_*` deprecation, its useful to configure jitter buffer for media type, since video needs a greater max buffer for example.

```conf
audio_jitter_buffer_type	fixed
audio_jitter_buffer_delay	5-10	
video_jitter_buffer_type	fixed
video_jitter_buffer_delay	5-50
```

This way you can also disable `jbuf` for audio and use only the `aubuf` jitter buffer:

```conf
audio_jitter_buffer_type	off
video_jitter_buffer_type	fixed
video_jitter_buffer_delay	5-50
```


juha-h:
So is `fixed` best as default value for both if not set?

juha-h:
In mobile device network quality can vary a lot.  Sometimes it  can be good when connected to WiFi or 4G/5G mobile network with strong signal, and sometimes bad when connected to low speed or busy mobile network or when signal is week.

In my Android app, I have used adaptive, but don't know if that is the best choice.

cspiel1:
We use adaptive for our products for jbuf and aubuf. The adaptive algorithm learns the network quality and tries to keep the latency as short as possible.

This PR is a good idea. Is it possible to keep audio/video sync in mind?

juha-h:
Currently there are these config variables:
```
audio_buffer
audio_buffer_mode

jitter_buffer_type
jitter_buffer_delay
jitter_buffer_wish
```
`audio_buffer` text tells `The audio buffer has to be chosen large enough to compensate the expected network jitter, except the jitter buffer is used.`  So does this mean that `audio_buffer` value does not matter if `jitter_buffer_type` is `adaptive` or `fixed`?

@cspiel1 Could you share your values of the above listed variables?

sreimers:
> Is it possible to keep audio/video sync in mind?

Yes, but this needs another PR for calculating playout times by RTCP SR and NTP timestamps for lip-sync.

juha-h:
While studying this, I noticed that there is a bug in the Wiki regarding jitter_buffer_delay: https://github.com/baresip/baresip/wiki/Configuration#jitter_buffer_delay.  Value is `<min>-<max>`, but default value is `fixed`, not a range.  I tried to figure out from source, what the default value is, but so far no luck.

juha-h:
> While studying this, I noticed that there is a bug in the Wiki regarding jitter_buffer_delay: https://github.com/baresip/baresip/wiki/Configuration#jitter_buffer_delay. Value is `<min>-<max>`, but default value is `fixed`, not a range. I tried to figure out from source, what the default value is, but so far no luck.

OK, found it.  Default is 5-10.  Will edit Wiki accordingly.

cspiel1:
We currently use:
```
audio_buffer		20-300		# ms
audio_buffer_mode	adaptive
jitter_buffer_type	adaptive		# off, fixed, adaptive, lowlatency
jitter_buffer_delay	0-20		# frames 
```

juha-h:
For backwards compatibility it would be good that if audio/video specific variables are not given, the current ones would apply to both.

juha-h:
Christian Spielberger writes:

> We currently use:
> ```
> audio_buffer		20-300		# ms
> audio_buffer_mode	adaptive
> jitter_buffer_type	adaptive		# off, fixed, adaptive, lowlatency
> jitter_buffer_delay	0-20		# frames 
> ```

Thanks, I'll try the same.

cspiel1:
For our product it is okay if settings are renamed.

juha-h:
@cspiel1 With this audio config, no incoming audio can be heard:
```
05-18 21:49:33.866  5725  5725 I Baresip : audio_buffer 20-300
05-18 21:49:33.866  5725  5725 I Baresip : audio_buffer_mode adaptive
05-18 21:49:33.866  5725  5725 I Baresip : audio_jitter_buffer_type adaptive
05-18 21:49:33.866  5725  5725 I Baresip : audio_jitter_buffer_delay 0-20
```
Incoming audio is heard if I change 0-20 to 5-20.

So what is now the correct version of your earlier one?
```
audio_buffer		20-300		# ms
audio_buffer_mode	adaptive
jitter_buffer_type	adaptive	# off, fixed, adaptive, lowlatency
jitter_buffer_delay	0-20		# frames 
```

cspiel1:
Was on vacation. I'll have a look on it.

cspiel1:
> So what is now the correct version of your earlier one?

I think you had it correct. Add also the video settings:
```
audio_buffer 20-300
audio_buffer_mode adaptive
audio_jitter_buffer_type adaptive
audio_jitter_buffer_delay 0-20
video_jitter_buffer_type adaptive
video_jitter_buffer_delay 0-20
```

Do you still have this issue? I can't reproduce this.

cspiel1:
All I can say is that this has to be true: `jb->nf > jb->wish`. See jbuf.c:462!
Then frames will be returned by `jbuf_get()`. Counting of `nf` is correct like I can see and `retest` verifies this also. I am not able to see what difference there should be if `jb->wish` is zero or not.

This PR should https://github.com/baresip/re/pull/821 only improves complexity of `jbuf_put()`. Counting of `nf` still will be correct.

Also I can't see if the PR https://github.com/baresip/re/pull/817 has an influence on the check `jb->nf <= jb->wish` in line jbuf.c:462. As long as `nf` is increased at some time.

PR #2595, opened 2023-05-31
SIP 302 moved temporarily

cspiel1:
- ua, call: support send of 302 Moved Temporarily
- menu: add a command for 302 Moved Temporarily

PR #2596, opened 2023-05-31
added ua_redirect() API function

juha-h:


src/call.c
```diff
@@ -1194,6 +1194,29 @@ int call_modify(struct call *call)
...
+	    scode < 300 || scode > 399) {
+		call_hangup(call, scode, reason);
+		return;
+	}
+
+	info("call: redirecting incoming call from %s (%u %s) to %s\n",
+	     call->peer_uri, scode, reason, uri);
+
+	(void)sipsess_reject(call->sess, scode, reason,
+			     "Contact: %s\r\nContent-Length: 0\r\n\r\n", uri);
```
cspiel1, line 1213: - split into two lines -> better readable
- add expires parameter

```c
	char *estr = NULL;

	uint64_t expiry = ua_moved_expires(ua);
	if (expiry)
		re_sdprintf(&estr, ";expires=%lu", expiry);

	(void)sipsess_reject(call->sess, scode, reason,
			     "Contact: <%s>%s\r\n"
			     "Content-Length: 0\r\n\r\n"
			     ,
			     uri, estr ? estr : "");
```

juha-h:
I didn't study Contact header details yet when I wrote the test code.  Now I did and noticed that the header can have many parameters (URIs) that can contain display name, q-value, etc.  More that one URI is needed if response is `300 Multiple Choices`.  So I'll update the PR and replace `uri` param with `contact_params`.

cspiel1:
Even in `re` `sipreg` we have not this level of general purpose API. Compare with `sipreg/reg.c:302`!
```c
	err = mbuf_printf(mb, "Contact: <sip:%s@%J%s>;expires=%u%s%s",
			  reg->cuser, &reg->laddr, sip_transp_param(reg->tp),
			  reg->expires,
			  reg->params ? ";" : "",
			  reg->params ? reg->params : "");
```
The API should guide the app developer somehow. If someone needs more special parameters they can be added later.

juha-h:
Christian Spielberger writes:

> Even in `re` `sipreg` we have not this level of general purpose
> API. Compare with `sipreg/reg.c:302`!

It makes sense that REGISTER registers only one Contact URI, whereas
`300 Multiple Choices` needs more than one, when there is more than one
alternative that can be tried and q-value could be used to prioritize
them.  Also it makes sense that receiver of 3xx can be shown display
names of the contacts.

cspiel1:
How Contact header could look like if there are multiple choices?

juha-h:
Christian Spielberger writes:

> How Contact header could look like if there are multiple choices?

Params are separated by commas:

Contact        =  ("Contact" / "m" ) HCOLON
    ( STAR / (contact-param *(COMMA contact-param)))
contact-param  =  (name-addr / addr-spec) *(SEMI contact-params)
name-addr      =  [ display-name ] LAQUOT addr-spec RAQUOT
addr-spec      =  SIP-URI / SIPS-URI / absoluteURI
display-name   =  *(token LWS)/ quoted-string

contact-params     =  c-p-q / c-p-expires
                      / contact-extension
c-p-q              =  "q" EQUAL qvalue
c-p-expires        =  "expires" EQUAL delta-seconds
contact-extension  =  generic-param
delta-seconds      =  1*DIGIT

cspiel1:
There are two solutions:
- the user of baresip (app developer) knows exactly all these details of the RFC and puts the Contact string into a single `const char*` parameter, or
- the API offers a higher level function for setting an absence note and/or adding an absence Contact with expires parameter and q-value.

I would prefer the second. Because I don't want to study the RFC each time when I use the API.

juha-h:
Christian Spielberger writes:

> - the user of baresip (app developer) knows exactly all these details
> - of the RFC and puts the Contact string into a single `const char*`
> - parameter, o
> - the API offers a higher level function for setting an absence note
> - and/or adding an absence Contact with expires parameter and
> - q-value.
> 
> I would prefer the second. Because I don't want to study the RFC each
> time when I use the API.

How about having in `ua_redirect` param `struct list *contact_params`,
where each list element is a string containing one param?

What comes to call_redirect, that may not need to be exposed to the API
at all.

cspiel1:
Do you really want the app decide for each incoming call if the call should be redirected?

Alternative: The UA could do this for you with a defined expires value, clearing the redirect state automatically if it expires. A simple API function could add redirect Contacts with optional display name and `struct list *contact_params` and thus support also multiple choice redirects.

juha-h:
juha-h writes:

> How about having in `ua_redirect` param `struct list *contact_params`,
> where each list element is a string containing one param?

If it is still too difficult to remember how to write each param, then I
guess contact param struct needs to be defined that has name-addr,
expires, and q fields.

juha-h:
Christian Spielberger writes:

> Do you really want the app decide for each incoming call if the call
> should be redirected?

For sure yes.  Otherwise the user may end up paying lots of money when
the app automatically forwards the call to an expensive telephone
number.

> Alternative: The UA could do this for you with a defined expires
> value, clearing the redirect state automatically if it expires. A
> simple API function could add redirect Contacts with optional display
> name and `struct list *contact_params` and thus support also multiple
> choice redirects.

I just wrote a message about that before reading this.

juha-h:
juha-h writes:

> 
> > Do you really want the app decide for each incoming call if the call
> > should be redirected?
> 
> For sure yes.  Otherwise the user may end up paying lots of money when
> the app automatically forwards the call to an expensive telephone
> number.

The same is true for REFER.  My app asks if the user really wants the
call to be transferred to the given URI.

cspiel1:
```c
int ua_add_redirect(struct ua *ua, const char *contact, int expires, int qvalue);
int ua_clear_redirect(struct ua *ua);
```
The contact string may contain the Display name or not. This is already possible for other API functions.

cspiel1:
> The same is true for REFER. My app asks if the user really wants the call to be transferred to the given URI.

This argument relates to the receiver of SIP REFER. Okay, the receiver of a SIP 302 should behave like that.

cspiel1:
And maybe a function for overwriting the default reason.
```c
int ua_set_redirect_reason(struct ua *ua, int scode, const char *reason);
```

cspiel1:
@alfredh do you think this is too much application logic in the core? Since this is described in RFC-3261 we could put this into ua.c into my opinion.

juha-h:
Christian Spielberger writes:

> int ua_add_redirect(struct ua *ua, const char *contact, int expires, int qvalue);

> int ua_clear_redirect(struct ua *ua);

> int ua_set_redirect_reason(struct ua *ua, int scode, const char *reason);

In my opinion, the only new baresip API function should be
`ua_redirect`.  All the rest should be internal to the application and
not complicate the API.

cspiel1:
This solution https://github.com/baresip/baresip/pull/2595 is compatible with existing `call_hangup()`. If `call->answered` is true `sipsess_abort()` should be called instead of `sipsess_reject()`.

juha-h:
> This solution #2595 is compatible with existing `call_hangup()`. If `call->answered` is true `sipsess_abort()` should be called instead of `sipsess_reject()`.

Isn't that what my code already does:
```
	if (call->state != CALL_STATE_INCOMING || call->answered ||
	    scode < 300 || scode > 399) {
		call_hangup(call, scode, reason);
		return;
	}
```
i.e., call->state == CALL_STATE_INCOMING and call->answered then call_hangup is called?

juha-h:
> Edit: Also the redirect codes with reason texts should be coded into the baresip core in my opinion. We could let the user overwrite them but the core should check the range 300-399.
> 
>     * 300 (Multiple Choices)
> 
>     * 301 (Moved Permanently)
> 
>     * 302 (Moved Temporarily)

Done.

cspiel1:
maybe it looks better if the
```c
if (call->state != CALL_STATE_INCOMING || call->answered ||
    scode < 300 || scode > 399) {
```
checks goes to ua.c?

cspiel1:
But then still we have a function that is called `ua_redirect()` that has to call in some situations `call_hangup()`.  This is suboptimal.

juha-h:
I tried in the beginning to have the test in ua.c, but didn't find how to access call->answered outside of call.c.

juha-h:
Don't know how to fix failed coverage check.

juha-h:
Alfred E. Heggestad writes:

> personally I think the business of Call Forwarding (sending 302)
> belongs in the SIP proxy, and not the endpoint.

I agree and my SIP Proxy is able to handle that, but there may be some
that can't.  The request for choosing whether to give 4xx or 3xx
response game from a user of my baresip Android app.

PR #2600, opened 2023-06-01
pns: enables adding push notification service config to account

codyit:


codyit:
Apologies for the noise, I found a way to build / test validate loop successfully locally.

src/pns.c
```diff
@@ -0,0 +1,48 @@
...
+{
+	struct pns *pns;
+
+	if (!provider || !prid) {
+		warning("pns: provider and prid are required\n");
+		return EINVAL;
+	}
+
+	pns = mem_zalloc(sizeof(*pns), destructor);
+	pns->pn_provider = provider;
```
sreimers, line 42: Please use `str_dup` here (and for the other fields) to store a copy of the string.

sreimers: Update: Then you have to `mem_deref` in destructor, too.

src/pns.c
```diff
@@ -0,0 +1,48 @@
...
+#include <re.h>
+#include <baresip.h>
+#include "core.h"
+
+
+static void destructor(void *data)
+{
+	struct pns *pn = data;
+
+	mem_deref((void*)pn->pn_provider);
```
sreimers, line 16: `(void*)` Casting is not needed.

codyit:
Adding notes here for the steps (some action > effect not isolated unfortunately) that unblocked myself from running test locally (on Fedora not sure if it matters):
1. `cmake -B build -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DMODULES="menu;account;g711;auconv;ausine;srtp;aufile;fakevideo;dtls_srtp"`
    *  Using CLANG mitigated an openssl error I ran into: ` undefined reference to symbol 'OPENSSL_init_ssl@@OPENSSL_3.0.0'`
    * The modules are the minimum required list to complete test run  
2. Test execution is location sensitive, from `build/` execute `test/selftest -v`
3. Clone `re` into the parent path and build it. I did it because of the CMake hint. Perhaps dnf installing libre-devel would've worked but I went straight to source after installing libre didn't work while the `devel` convention escaped me at the time.

I was able to GDB through the tests earlier but some where in the middle I stopped being able to do it without realizing what changed (yes always build with debug enabled)

Apologies for missing the mark I completely forgot about the memory management aspect of C. After the sudden outbreak of common sense I rewrote some of it, have all the tests passed, just tracking down one last memory leak before another revision. Thanks.

src/pns.c
codyit: Done.

src/pns.c
codyit: Done.

alfredh:
Would it be possible to change this into a module e.g. `pns.so` ?


This approach would be much more clean and modular, and only the people who needs PNS
may load that module...

codyit:
@alfredh Thanks for looking into this. Since I'm still trying to understand the code base, I'd like to ask you the question you asked me as the module system author IIUC. Is it possible?

Illustrated from an example SIP REGISTER in https://www.rfc-editor.org/rfc/rfc8599#page-8 , what needs to happen is to add these new SIP URI to Contact header field URI during REGISTER. (And also feature compatibility flag which is not being done yet). So I'd have to change then account structure, then simply present them when registering. 

Perhaps I should rename the commit to "storing RFC 8599 URI parameters" since my plan is to leave the actual push notification service interaction to consumer of the library, in my case the Android baresip+, similar to https://git.jami.net/savoirfairelinux/jami-client-android/-/issues/990 . Or perhaps add it here in a later PR if desired.

 The other modules I read operating at similar abstraction level (pcp stun ice), they all involve implementing an interface defined in the core structures. So unless there's a heavy rewrite to put each sip extension feature to a module it doesn't seem possible in its current form? At least for storing and presenting URI parameters. I can see how actual PNS interaction (e.g. https://unifiedpush.org/developers/go_c/) can be done in a module if doing it here  is desirable.

Or perhaps it's because of my lack of understanding, please let me know.

juha-h:
According to the RFC, `pn-*` are URI (not header) params and pn-rpid is learned from the push service.  They could thus be written by the app to account's entry in accounts file before baresip service is started and no new code would be needed.  But then the params would also appear in the From URI when a call is made, which does not sound like a good idea.  The params  should only appear in the Contact URI of REGISTER request.

I don't know if baresip/re API would currently supports that.  There exists capability to add new headers to REGISTER request, but perhaps not a way to add new Contact URI parameters.

codyit:
> pn-* are URI (not header) params and pn-rpid is learned from the push service

I think we mean the same thing,  `contact header field URI` is the term I got from the RFC. I was just going to say URI but since I'm new I wanted to make sure I use the right terms..
```
Unless specified otherwise, the normal SIP UA registration procedures
   [[RFC3261](https://www.rfc-editor.org/rfc/rfc3261)] apply.  The additional procedures described in this section
   apply when the REGISTER request contains a 'pn-provider' SIP URI
   parameter in the Contact header field URI (Figure 2).
```

alfredh:
any missing API can be added.

If added as a module, it would prove that the baresip API is flexible and modular.

codyit:
@alfredh 

Sorry it's going to be a quite bit beyond me given my understanding is mostly by heuristic right now, but happy to take a stab if I can pick your brain or get a simple ASCII art flow diagram :)

IIUC what's being discussed is to add something that gets called close to `reg_set_customer_hdrs` in `ua.c` but for params: https://github.com/baresip/baresip/blob/3a8cbae728073fc1daead67275abf64c6b65fe06/src/ua.c#L208-L213] 

But doesn't that just shifts the code change from `account` to `ua`?

Is there any example that show me how to store the state of these `pn-` params without changing core structure but implemented in a module instead?

codyit:
Found a IIUC good example for my own question https://github.com/baresip/baresip/commit/f896ceef31eec60bc36b434002606ac57cae2251 basically just statically allocate one and deref when module unloads.

 `uag` has an `eprm` that serves similar purpose but unfortunately it's added to the aor.
https://github.com/baresip/baresip/blob/3a8cbae728073fc1daead67275abf64c6b65fe06/src/ua.c#L1091-L1097

So I'm thinking of either:
1.  add a `cprm` to `uag`, then add an API through the module to update that, or
2. perhaps a handler to inject strategy that morphs reg behavior

I'll attempt 1 for now given my C skill and code base familiarity, and also it's biggest change that I don't have enough context to say whether it's actually needed.

codyit:
Attempting to make it a module, putting it back into draft mode.  
Linked to the change needed in re as a draft.

There are a couple of todos before the final revision, requesting comments to see if it looks about right.
What I want to achieve is to only expose the API to update a UA's cparams if the module is loaded.

juha-h:
Cody T.-H. Chiu writes:

> So I'm thinking of either:
> 1.  add a `cprm` to `uag`, then add an API through the module to update that, or
> 2. perhaps a handler to inject strategy that morphs reg behavior

As I mentioned earlier, if account has push notification enabled via
account params, then somehow they should be added only to contact URI of
REGISTER request.  Most likely new baresip/re API function(s) would be
needed for that.

codyit:
> Cody T.-H. Chiu writes:
> So I'm thinking of either: 1. add a `cprm` to `uag`, then add an API through the module to update that, or 2. perhaps a handler to inject strategy that morphs reg behavior
> As I mentioned earlier, if account has push notification enabled via account params, then somehow they should be added only to contact URI of REGISTER request. Most likely new baresip/re API function(s) would be needed for that.

Yes the linked re PR draft and this updated draft implent these.

sreimers:
> Yes the linked re PR draft and this updated draft implent these.

Dependent Pull requests with exact same title (re and baresip) will be detected by the CI (we can change commit message on squash merge).

codyit:
> please consider adding this module here instead:
> 
> https://github.com/baresip/baresip-apps

Spent the time to read into your module system, there's no established pattern to expose other symbols.

I've significant rewritten it to get config update from event, should be more sensible now, albeit abusing the channel to get event from the client.

I've rebased + squashed my main.

---

@juha-h  After reading more into the code base I understand what you mean by
> could thus be written by the app to account's entry in accounts file before baresip service is started 

IIUC this would lead to bad UX, updating the params require a restart perhaps is fine but limiting it at the foundation seems wrong? Unless.. restarting baresip fro the app is a common practice already? Please take a look to see if this latest version is lesser evil.

juha-h:
I haven't read your code, but it is possible to change account parameters without restarting baresip.  There are many setters in the API for that purpose.  After change, the account can be re-registered if needed.

juha-h:
If pns parameter values do not change after they the app gets them, then they can be written to accounts file before baresip service is started.  If the app get pns param values after basesip service has started, then there can be account setters to set them after which the account can re-register.

juha-h:
Figure 1 of https://www.rfc-editor.org/rfc/rfc8599 shows how the app subscribes to push service and gets the prid.  It can write that and other pn parameters to account's entry in accounts file and then start baresip service.

Or is there requirement to subscribe to push service on the fly while baresio service is running?  If so, pn params can se added to the account using setters after which the account is registered.

juha-h:
I haven't followed the module discussion and don't understand why a
module would be needed.

Why is it not enough that there is a new account parameter, for example,
like this:

;extra_reg_uri_params="pn-provider=acme;pn-param=acme-param;pn-prid=ZTY4ZDJlMzODE1NmUgKi0K"

and, if needed, a setter to set the param on it on the fly.  Then, if
set, reg_register would add them to reg_uri.

What else is needed?  Why module, since this feature is not push
notification specific and just a few lines of new code?

codyit:
> extra_reg_uri_params="pn-provider=acme;pn-pa ...

That's what's being done in this PR, just `s/extra_reg_uri_params/cparams`

> What else is needed? 

Yes things get a bit muddy when it's not ABI accepting 3 strings. As mentioned above, basically anything can be a serialized string, however
> > Now this change uses events to get not exactly structural, but serialized params in events. Seems a bit silly since it looks exactly the same as the desired cparams. So why add them? Defensive validation, state change tracking. Also unblocks triggering of REGISTER upon receiving notification at this level.

Also despite contact URI parameters being a generic construct, changing it during registration is quite specific, until new RFCs are written, it servers as the only local reference usage.

Have you read the change yet? I suppose for the spirit of encapsulation the event should be 3 space delimited strings instead, that should better highlight responsibilities.

codyit:
I've made the change to take in comma separated strings from the event, should highlight the encapsulation.

juha-h:
Cody T.-H. Chiu writes:

> > > Now this change uses events to get not exactly structural, but
> > > serialized params in events.

I don't understand why any events would be needed.  Just add

char *contact_uri_extra_params;

to `struct account` and then set it in `account_alloc()` and (if dynamic
changes are needed) by a setter API function.

Finally, if `contact_uri_extra_params` is set, add its value to
reg_uri in `start_register()` or `reg_register()`.

Why is that not enough?

codyit:
Is there something lost in translation? The paragraph before and after that line all explained the reasoning, happy to get on whatever IM to have a chat e.g. gitter

juha-h:
> Is there something lost in translation? The paragraph before and after that line all explained the reasoning, happy to get on whatever IM to have a chat e.g. gitter

I don't know which line you mean.  Just give a short reply why it is not enough what I proposed in my previous comment.

codyit:
> > Is there something lost in translation? The paragraph before and after that line all explained the reasoning, happy to get on whatever IM to have a chat e.g. gitter
> 
> I don't know which line you mean. Just give a short reply why it is not enough what I proposed in my previous comment.

Please let me know what part wasn't clear in https://github.com/baresip/baresip/pull/2600#issuecomment-1581706768

juha-h:
I don't understand anything about that comment.  Tell me why it does not work what I have proposed.

codyit:
Let me try to rephrase  it.

Basically, what you described is already being done in the latest revision.

The code in pns module handles encapsulation, validation, state tracking (e.g only triggers a cparam changewhen the values differ, then optionally trigger a register later). Basically semi colon delimited parameters ABI acts as a low level one without domain knowledge, the pns module handles calling that safely. These ought to live somewhere, it it defensive at the lower level or at the library consumer...

Now the judgement call is, is it worth moving away from the norm using event to pass it through from the other direction?
I think it depends on the design intent of the event channel, was it supposed to be limited to single direction in the first place, or just as a general bus.

codyit:
> Or is there requirement to subscribe to push service on the fly while baresio service is running? If so, pn params can se added to the account using setters after which the account is registered.

At least prid could change, see:

https://unifiedpush.org/spec/android/#messaging-broadcast-receiver-1

https://www.rfc-editor.org/rfc/rfc8599#section-4.1.3

juha-h:
Cody T.-H. Chiu writes:

> > Or is there requirement to subscribe to push service on the fly while baresio service is running? If so, pn params can se added to the account using setters after which the account is registered.
> 
> At least prid could change, see:
> 
> https://unifiedpush.org/spec/android/#messaging-broadcast-receiver-1
> 
> https://www.rfc-editor.org/rfc/rfc8599#section-4.1.3

So does the UA need to get access to REGISTER response in order to find
out the change or is the push protocol used for that?

codyit:
> Each account can have different push provider. Does your module support that?

Good point. The event comes with the ua so targeting fine, however the PNS storage as it is only tracks one PNS config.  I noticed that just didn't want to introduce too many talking points / revision before we get consensus it's indeed the way to go at a high level.

* can just store one per UA as list or map, or
* beef up `prm.c` in re so it's a serialized store that handles CRUD...

> So does the UA need to get access to REGISTER response in order to find out the change or is the push protocol used for that?

It's from push service.

juha-h:
I still don't understand why the module is needed.  Why does this not work:

1) before baresip service is started, app subscribes to push service for each account that has a push server configured and subscription has not been done yet
2) app then adds the pn parameters learned from the push service to accounts' entries in accounts file
3) app starts baresip service that registers each account with pn parameters (if available) in Contact URI
4) if app learns from push server that parameters of some account have changed (it gets current values using account getter function), makes the change using account setter function, saves new accounts file, and re-registers the account

alfredh:
```
/home/runner/work/baresip/baresip/src/reg.c:271:10: error: implicit declaration of function ‘sipreg_set_contact_params’ [-Werror=implicit-function-declaration]

```

codyit:
> ```
> /home/runner/work/baresip/baresip/src/reg.c:271:10: error: implicit declaration of function ‘sipreg_set_contact_params’ [-Werror=implicit-function-declaration]
> ```

Already made the commit messages same, seems it wont pick up `re` changes

alfredh:
Please try to make the *title* the same between baresip and re PRs.

codyit:
> I still don't understand why the module is needed. Why does this not work:
>     1. before baresip service is started, app subscribes to push service for each account that has a push server configured and subscription has not been done yet
>     2. app then adds the pn parameters learned from the push service to accounts' entries in accounts file
> 
>     3. app starts baresip service that registers each account with pn parameters (if available) in Contact URI
> 
>     4. if app learns from push server that parameters of some account have changed (it gets current values using account getter function), makes the change using account setter function, saves new accounts file, and re-registers the account


Yes it would work as well, which is the baseline of this change. 

The question is who deal with the simple yet required parameter validation per the spec. Do we enforce safety at the library level versus delegating it to each consumer.

i.e. `pn-provider` and `pn-prid` are required, `pn-param` is optional.  Same goes for URI serialization, should the each consumer know to speak `pn-*`, or in particular there's no `s` for `pn-param`. Or otherwise, when you have an update from the UI with either with different or identical value, how should the UA (as a whole from proxy / registrar perspective) behave? 

Not having to know what to do is the abstraction value provided by libraries..  I don't like hyperbole because it's usually of bad taste but I can't seem to illustrate my point without it.. taken to extreme, it would work if baresip takes raw SIP message to relay or even 0 and 1s.. 

Now note that I talk a lot but I'm not hard on fighting for it, if at the end we decide passing raw string is fine, not great in my book but sure. I'm just trying get us to the same page in terms of understanding of the motive and  problem.

Also 2. above, I take that you mean the account config would also be generic Contact URI, otherwise it's going to contain PNS specific things in account module.

alfredh:
if you want to include `pns.so` in the test, please add it here:

https://github.com/baresip/baresip/blob/main/.github/workflows/valgrind.yml#L43

juha-h:
It is the app, not libbaresip that talks with the push service.  Therefore in my opinion it is the app's responsibility check what it gets from the push service.  I would not make libbaresip at all aware of the push service.  It would be enough that libbaresip is capable of adding parameters to Contact URI without having any knowledge of what the parameters contain.  The only requirement would be that the parameters follow RFC3261 syntax:
```
other-param       =  pname [ "=" pvalue ]
pname             =  1*paramchar
pvalue            =  1*paramchar
paramchar         =  param-unreserved / unreserved / escaped
param-unreserved  =  "[" / "]" / "/" / ":" / "&" / "+" / "$"
```

codyit:
> if you want to include `pns.so` in the test, please add it here:
> 
> https://github.com/baresip/baresip/blob/main/.github/workflows/valgrind.yml#L43

Added to all 3 cmake lines

>  however the PNS storage as it is only tracks one PNS config.

I've removed this check for this PR, can optimize later.

>  the app's responsibility check what it gets from the push service

Yes definitely, per https://unifiedpush.org/developers/android/  these are push domain terminologies:
```
 override fun onNewEndpoint(context: Context, endpoint: String, instance: String)
```
Those SIP parameters are SIP protocol extension on the other side.

Let me illustrate with some ASCII graphs

```
|-------------|                      |-------------|                        |-------------|
| Push service| [ service protocol]  | Lang Library|  [ Library interface]  | Application | 
|-------------|                      |-------------|                        |-------------|
```
In fact most applications won't speak wire protocol (HTTP/REST or otherwise) with  Push service either in order to achieve separation of concern. They consumes a library that expose an interface, in UnifiedPush case so you end up with 2 strings.

```
|-------------|                      |-------------|                  |-----------|
| Application | [ service protocol]  | Lang Library|  [ SIP protocol] | SIP proxy | 
|-------------|                      |-------------|                  |-----------|
```

Similarly on the other side, as the SIP RFC describes itself, it's similar to HTTP. Just like the application would not speak HTTP, with the push service, it should interface through a library which expose an interface that takes 2 strings.

modules/pns/pns.c
```diff
@@ -0,0 +1,146 @@
...
+{
+	struct pns *pns;
+	int err = 0;
+
+	if (!pl_isset(provider) || !pl_isset(prid)) {
+			warning("pns: provider and prid are required\n");
+		return EINVAL;
+	}
+
+	pns = mem_zalloc(sizeof(*pns), destructor);
```
sreimers, line 58: Please add a check if `pns` is allocated here, otherwise return `ENOMEM`;

modules/pns/pns.c
```diff
@@ -0,0 +1,146 @@
...
+
+
+/* Decode URI parameters into push notification service */
+static int pns_decode(struct pns **pns, struct pl *data) {
+	struct pl provider, prid, param;
+	int err;
+
+	if (0 != (err = re_regex(data->p, data->l, "[^,]*,[^,]*[,]*[~]*",
+			&provider, &prid, NULL, &param)))
+		return err;
```
sreimers, line 87: Please separate the `if` to make this more readable.

modules/pns/pns.c
```diff
@@ -0,0 +1,146 @@
...
+	struct pl module, event, data;
+	struct pns *pns;
+	char *ua_cparam = NULL;
+
+	if (ev != UA_EVENT_MODULE ||
+			re_regex(prm, strlen(prm), "[^,]*,[^,]*,[~]*",
+				&module, &event, &data) ||
+			pl_strcmp(&module, "pns") ||
+			pl_strcmp(&event, "config_update"))
+		return;
```
sreimers, line 109: This is also not very readable, combine similar checks to one and separate the others.

modules/pns/pns.c
```diff
@@ -0,0 +1,146 @@
...
+
+	err = pns_decode(&pns, &data);
+
+	if (!err) {
+		err = pns_encode(&ua_cparam, pns);
+		mem_deref(pns);
```
sreimers, line 115: You can use a `goto err;` here to simplify the `if err` handling and use a positive logic like `if (err)`.

modules/pns/pns.c
codyit: Done

modules/pns/pns.c
codyit: Done

modules/pns/pns.c
codyit: Done

modules/pns/pns.c
codyit: Done. Didn't C much before this, `goto` scares me a bit :laughing: , but use case makes sense here.

modules/pns/pns.c
```diff
@@ -0,0 +1,158 @@
...
+	}
+	else {
+		err |= ua_set_contact_params(ua, NULL);
+	}
+
+	mem_deref(pns);
+	mem_deref(ua_cparam);
+
+	if (err)
+	error:
```
sreimers, line 135:
```suggestion
error:
	if (err)
```

juha-h:
Why the two new contact_param API functions are ua functions and not account functions?  If they were account functions, then also ua could use them via ua_account() function.

modules/pns/pns.c
```diff
@@ -0,0 +1,158 @@
...
+ */
+static int pns_alloc(struct pns **pnsp, const struct pl *provider,
+		const struct pl *prid, const struct pl *param)
+{
+	struct pns *pns = NULL;
+	int err = 0;
+
+	if (!pl_isset(provider) || !pl_isset(prid)) {
+		warning("pns: provider and prid are required\n");
+		goto out;
```
sreimers, line 55:
```suggestion
		return EINVAL;
```

codyit: I made that change intentionally and updated the test case for better or worse. Otherwise I can add the verbose logic back in to clear out the config

codyit: all empty is intentionally, invalid is invalid, yeah maybe that's better, change coming in a bit

sreimers: A alloc function should update the double pointer only on success to the new valid object, its up to the caller to handle the error case and maybe cleanup if necessary.

juha-h:
Cody T.-H. Chiu writes:

> > Why the two new contact_param API functions are ua functions and not
> > account functions? If they were account functions, then also ua
> > could use them via ua_account() function. 
> 
> Would it be better? I did this only because both `cuser` and
> `custom_hdrs` were in `ua`, not that I'm 100% clear with each's
> responsibilities.

Is there need to store the pn param values somewhere between baresip
restarts or are they re-learned from push service each time baresip is
started?

If not re-learned, then accounts file could be a place to store them and
for that purpose account API functions would be needed.

codyit:
> Cody T.-H. Chiu writes:
> > Why the two new contact_param API functions are ua functions and not > account functions? If they were account functions, then also ua > could use them via ua_account() function. Would it be better? I did this only because both `cuser` and `custom_hdrs` were in `ua`, not that I'm 100% clear with each's responsibilities.
> Is there need to store the pn param values somewhere between baresip restarts or are they re-learned from push service each time baresip is started? If not re-learned, then accounts file could be a place to store them and for that purpose account API functions would be needed.

The RFC out scope UA interaction, but acknowledges prid can update any time, so more towards relearn.

For Unifiedpush in particular distributor refresh is in the spec
https://unifiedpush.org/spec/android/#messaging-broadcast-receiver-1
> The distributor MUST send this action to the registered application in the following cases:
...
the endpoint for the application changed

And relearn sees to be recommended:
https://unifiedpush.org/developers/android/
```
// Check if a distributor is already registered
if (getDistributor(context).isNotEmpty()) {
    // Re-register in case something broke
    // Options:
    // "instance" to handle multiple registrations
    // "features" empty array or [FEATURE_BYTES_MESSAGE]
    //    to be sure the distributor handles non-UTF-8 input
    // "messageForDistributor" that may be displayed by the distrib.
    registerApp(context)
    return
}
```

codyit:
It is so strange, I replicated the mingw setup and memory leak locally.

It won't leak if I don't load the pns module. It leaks even if I comment out everything so only `init` and `close` are stubbed with `return 0;`

sreimers:
Please rebase against main, I just added cv2pdb, it should help to see which mem objects are currently leaked (with a backtrace).

test/ua.c
```diff
@@ -175,10 +177,23 @@ int test_ua_register(void)
 	err |= reg(SIP_TRANSP_TLS);
 #endif
 
-	ua_stop_all(true);
-	ua_close();
+	err |= ua_alloc(&ua, "<sip:x@127.0.0.1>");
+	err |= module_load(".", "pns.so");
```
codyit, line 181: `.so`...

alfredh:
from my view this has two possible solutions:

1. add generic cparams to struct account that can transport e.g. PNS parameters
2. add experimental pns.so module to "baresip-apps".

codyit:
Seems 1. and 2. are not mutually exclusive? Couple of things to unpack here:

> `account` instead of `ua`

Haven't the bandwidth to read into the relationship, I've put out a baresip-studio draft PR to show how events can trigger update on these params . If any of you've read it I'll just blind trust, move over then figure out on studio side.

> Generic cparam only:

Seems baresip can only ever support contact-params to unblock instead of support RFC8599.
```
Contact        =  ("Contact" / "m" ) HCOLON
                  ( STAR / (contact-param *(COMMA contact-param)))
contact-param  =  (name-addr / addr-spec) *(SEMI contact-params)
```
All consumers have to implement PNS separately. Passing params around smells like last resorts that break abstractions to me.. but sure if that's the consensus..  doesn't block what I'm trying to do at all.

Is there no nice way to support new structures and serialization not matching any current interfaces? IMO the event based one in this PR can be generalized...

More importantly why not both? How about I move these APIs to account, then move pns to baresip-apps? (not sure about the definition of apps)

codyit:
My original [no infra dream](https://gist.github.com/codyit/850a77373cfaa57492a4b38e6350bb39) doesn't pan out, no INVITE from twilio despite getting proper notification triggered register binding refresh. Oh well at least it was fun, thanks for the pointers folks.

Since time is not a factor for myself anymore,  I'd rather do the right thing for an PR numbered 2600.

I can remove all references of RFC8599 or PNS and just wrap everything up with 1. @alfredh mentioned. This way baresip ends up supporting extra contact params during registration. (weird on its own leg but sure)

I'm reluctant to be an author of a PR that directly calls this externally tho..  I think it's wrong, so I'll just cancel https://github.com/juha-h/baresip-studio/pull/384

PR #2602, opened 2023-06-02
Handle UA_EVENT_CALL_REDIRECT (formerly UA_EVENT_CALL_BLIND_TRANSFER)…

juha-h:
… in menu module


juha-h:
Requires re PR https://github.com/baresip/re/pull/829.

modules/menu/menu.c
```diff
@@ -793,6 +793,19 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 		menu_selcall(call);
 		break;
 
+	case UA_EVENT_CALL_REDIRECT:
+		val = pl_null;
+		if (!re_regex(prm, strlen(prm), "sip:"))
+			pl_set_str(&val, "invite");
+
+		(void)menu_param_decode(prm, "method", &val);
```
cspiel1, line 801: Is `;method=invite` a possible parameter of the SIP 302 Contact?
Like for the SIP REFER?

juha-h: No it is not.  I was not sure why that code was there.  Will try to simplify.

include/baresip.h
```diff
@@ -808,7 +808,7 @@ enum ua_event {
 	UA_EVENT_CALL_ESTABLISHED,
 	UA_EVENT_CALL_CLOSED,
 	UA_EVENT_CALL_TRANSFER,
-	UA_EVENT_CALL_BLIND_TRANSFER,
+	UA_EVENT_CALL_REDIRECT,
```
alfredh, line 811: this breaks the API, perhaps better to just add a new enum value ...

alfredh:
this sounds like we need a new API to configure this, or a config option, e.g.:

```
sip_follow_redirect:   yes|no
```

Is there any other SIP product with similar config option ?

modules/menu/menu.c
```diff
@@ -793,6 +793,11 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 		menu_selcall(call);
 		break;
 
+	case UA_EVENT_CALL_REDIRECT:
+		info("menu: redirecting call to %s\n", prm + 4);
+		ua_connect(ua, NULL, NULL, prm + 4, VIDMODE_ON);
```
alfredh, line 798: `prm + 4` should be rewritten with proper parsing

include/baresip.h
juha-h: If UA_EVENT_CALL_BLIND_TRANSFER (which is not currently used in baresip project) cannot be renamed, then it means that two events will need to be sent: UA_EVENT_CALL_BLIND_TRANSFER and UA_EVENT_CALL_REDIRECT.

juha-h:
> this sounds like we need a new API to configure this, or a config option, e.g.:
> 
> ```
> sip_follow_redirect:   yes|no
> ```

My plan was to add that after the backwards compatible basic stuff is first done.

modules/menu/menu.c
juha-h: I was too lazy to do that and felt that parsing is not necessary, since we know how the parameter is formatted there and we know that the response code must have exactly three digits.  But can add parsing when it is sure that the PR is otherwise ready to go.

cspiel1: And should we send here `UA_EVENT_CALL_BLIND_TRANSFER`?

juha-h:
Christian Spielberger writes:

> @cspiel1 commented on this pull request.
> 
> 
> 
> > @@ -808,7 +808,7 @@ enum ua_event {
>  	UA_EVENT_CALL_ESTABLISHED,
>  	UA_EVENT_CALL_CLOSED,
>  	UA_EVENT_CALL_TRANSFER,
> -	UA_EVENT_CALL_BLIND_TRANSFER,
> +	UA_EVENT_CALL_REDIRECT,
> 
> Shouldn't we remove `UA_EVENT_CALL_BLIND_TRANSFER` and increase ABI
> number (on next release)? We just want to rename the enum value.

alfedh suggested that UA_EVENT_CALL_BLIND_TRANSFER is not removed and I
restored it.  See also my suggestion on how we can use both.

cspiel1:
As I mentioned above: In this case menu could send the `UA_EVENT_CALL_BLIND_TRANSFER` if it really redirects the call to another target. But first of all the `re` PR has to be ready and accepted by Alfred. See my thoughts here: https://github.com/baresip/re/pull/829#issuecomment-1576818591

juha-h:
I just pushed commit that removed UA_EVENT_CALL_BLIND_TRANSFER.  It is unused and name does not match with blind redirect.  What do to is decided in menu module:
```
	case UA_EVENT_CALL_REDIRECT:
		if (account_sip_autoredirect(ua_account(ua))) {
			info("menu: redirecting call to %s\n", prm + 4);
			ua_connect(ua, NULL, NULL, prm + 4, VIDMODE_ON);
		} else {
			info("menu: got call redirect response to %s\n",
			     prm + 4);
		}
		break;
```
Suggestions no how to improve the second info are welcome.  Otherwise I don't want to keep going back and forth with this.

include/baresip.h
```diff
@@ -808,8 +810,8 @@ enum ua_event {
 	UA_EVENT_CALL_ESTABLISHED,
 	UA_EVENT_CALL_CLOSED,
 	UA_EVENT_CALL_TRANSFER,
-	UA_EVENT_CALL_BLIND_TRANSFER,
 	UA_EVENT_CALL_TRANSFER_FAILED,
+	UA_EVENT_CALL_REDIRECT,
```
alfredh, line 814: please move REDIRECT value one level up, so that ABI is not changed.

modules/menu/menu.c
```diff
@@ -793,6 +793,17 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 		menu_selcall(call);
 		break;
 
+	case UA_EVENT_CALL_REDIRECT:
+		if (account_sip_autoredirect(ua_account(ua))) {
+			info("menu: redirecting call to %s\n", prm + 4);
+			ua_connect(ua, NULL, NULL, prm + 4, VIDMODE_ON);
```
alfredh, line 799: can you test if just `prm` is working ?

`prm + 4` is a bad hack and has to be fixed.

alfredh:
please also rebase to HEAD to force CI trigger.

modules/menu/menu.c
juha-h: `prm` value is of form `3xx,URI` so not likely that ua_connect would work with it.  Since we know for sure that there is three digits before , followed by the URI, I do not consider `prm + 4` a hack.  But I improved the code a bit by looking for the comma:
```
	case UA_EVENT_CALL_REDIRECT:
		uri = strchr(prm, ',') + 1;
		if (account_sip_autoredirect(ua_account(ua))) {
			info("menu: redirecting call to %s\n", uri);
			ua_connect(ua, NULL, NULL, uri, VIDMODE_ON);
		}
		else {
			info("menu: redirect call to %s\n", uri);
		}
		break;
```
Any better?

cspiel1:
@juha-h : Shouldn't we set default to
```
sip_autoredirect=no
```

juha-h:
Christian Spielberger writes:

> @juha-h : Shouldn't we set default to
> ```
> sip_autoredirect=no
> ```

Yes, I would prefer "no", but thought that it cannot be done due to
backwards compatibility.

PR #2604, opened 2023-06-02
Added ua_redirect API function

juha-h:
This is just a new API function that does not affect existing baresip use.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -1194,6 +1194,45 @@ int call_modify(struct call *call)
 	return err;
 }
 
+void call_redirect(struct call *call, uint16_t scode, const char *reason,
+		   const char *contact_params)
+{
+	if (!call)
+		return;
+
+	if (!str_isset(reason))
+		switch (scode) {
+		case 300:
+			reason = "Multiple Choices";
+			break;
+		case 301:
+			reason = "Moved Permanently";
+			break;
+		case 302:
+			reason = "Moved Temporarily";
+			break;
+		case 305:
+			reason = "Use Proxy";
+			break;
+		default:
+			reason = "Unkown";
+		}
+
+	info("call: redirecting incoming call from %s (%u %s) to %s\n",
+	     call->peer_uri, scode, reason, contact_params);
+
+	(void)sipsess_reject(call->sess, scode, reason,
+			     "Contact: %s\r\n"
+			     "Content-Length: 0\r\n"
+			     "\r\n"
+			     ,
+			     contact_params);
+
+	set_state(call, CALL_STATE_TERMINATED);
+
+	call_stream_stop(call);
+}
+
 
 /**
  * Hangup the call
@@ -1407,6 +1446,15 @@ int call_answer(struct call *call, uint16_t scode, enum vidmode vmode)
 	mem_deref(desc);
 
 	return err;
+}
+
+
+bool call_answered(struct call *call)
+{
+	if (!call)
+		return false;
+
+	return call->answered;
 }
 
 
--- a/src/call.c
+++ b/src/call.c
@@ -1216,6 +1216,7 @@ void call_redirect(struct call *call, uint16_t scode, const char *reason,
 			break;
 		default:
 			reason = "Unkown";
+			break;
 		}
 
 	info("call: redirecting incoming call from %s (%u %s) to %s\n",
@@ -1449,7 +1450,7 @@ int call_answer(struct call *call, uint16_t scode, enum vidmode vmode)
 }
 
 
-bool call_answered(struct call *call)
+bool call_answered(const struct call *call)
 {
 	if (!call)
 		return false;
--- a/src/ua.c
+++ b/src/ua.c
@@ -1369,6 +1369,45 @@ void ua_hangup(struct ua *ua, struct call *call,
 
 
 /**
+ * Redirect an incoming call
+ *
+ * @param ua     User-Agent
+ * @param call   Call to redirect, or NULL for current call
+ * @param scode  Status code
+ * @param reason Optional reason
+ * @param contact_params Body of Contact header
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+void ua_redirect(struct ua *ua, struct call *call,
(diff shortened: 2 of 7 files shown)


src/call.c
```diff
@@ -1194,6 +1194,45 @@ int call_modify(struct call *call)
...
+			reason = "Moved Permanently";
+			break;
+		case 302:
+			reason = "Moved Temporarily";
+			break;
+		case 305:
+			reason = "Use Proxy";
+			break;
+		default:
+			reason = "Unkown";
```
alfredh, line 1218: missing break

src/call.c
```diff
@@ -1410,6 +1449,15 @@ int call_answer(struct call *call, uint16_t scode, enum vidmode vmode)
 }
 
 
+bool call_answered(struct call *call)
```
alfredh, line 1452: input param can be `const`

src/call.c
```diff
@@ -1194,6 +1194,45 @@ int call_modify(struct call *call)
 	return err;
 }
 
+void call_redirect(struct call *call, uint16_t scode, const char *reason,
+		   const char *contact_params)
```
alfredh, line 1198: is this function needed ?

can you call `call_hangup` directly ?

alfredh:
please also add a new testcase for this feature, to increase coverage.



it would be nice if this feature could be added without making any changes
to baresip core...

juha-h:
> please also add a new testcase for this feature, to increase coverage.

Will try.

> it would be nice if this feature could be added without making any changes to baresip core...

It was like that in the beginning, but then @cspiel1 proposed to move test
```
if (call->state != CALL_STATE_INCOMING || call->answered ||
    scode < 300 || scode > 399) {
```
to ua.c.  That was not possible without a new core.h function `call_answered()`.  So which do you want it?

src/call.c
juha-h: > is this function needed ?
> 
> can you call `call_hangup` directly ?

I don't think so, because Contact header need to be added and `call_hangup()` passes NULL as extra SIP headers param.

juha-h:
Tried to add new testcase by modifying ua_hangup() case, but the test hung.  Any pointers on what I'm missing?

test/call.c
```diff
@@ -210,6 +211,13 @@ static void event_handler(struct ua *ua, enum ua_event ev,
 			}
 			break;
 
+		case BEHAVIOUR_REDIRECT:
+			ua_redirect(ua, call, 302, NULL,
+				    "sip:alice@example.com");
```
alfredh, line 216: the testcode should only use internal dummy AORs.

public DNS names should not be used, as the testcode should be able to run
in an isolated environment with no internet access...

in your case you can create 3 UAs, A, B and C.

When doing the redirect, you can redirect to C and then verify that the call was established.

src/ua.c
```diff
@@ -1368,6 +1368,45 @@ void ua_hangup(struct ua *ua, struct call *call,
...
+ * @param call   Call to redirect, or NULL for current call
+ * @param scode  Status code
+ * @param reason Optional reason
+ * @param contact_params Body of Contact header
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+void ua_redirect(struct ua *ua, struct call *call,
+		 uint16_t scode, const char *reason,
+		 const char *contact_params)
```
alfredh, line 1384: is this ua wrapper really needed ?

if the application has the `struct call` state it should have enough knowledge about it
to take the decision about redirect or not.

test/call.c
juha-h: I thought that `ua_redirect()` testcase only needs to test if 3xx response is properly _send_ and that the call is then closed.  The test on what happens when baresip receives 3xx belongs to somewhere else and is an existing feature.

src/ua.c
juha-h: That is good point and would greatly simplify my life, since I would need to hassle with the test case.  The only piece that currently prevents it is that I have not found an API call to test if call is answered.  In the PR there is new core.h function
```
bool call_answered(const struct call *call);
```
Would it be OK if I move it from core.h to baresip.h?

alfredh: there is no hard requirement to add new testcases.

Adding new testcase can be done if you have time and knowledge.
It can also be added later.

moving API functions from core.h to baresip.h is also fine

juha-h:
Will be replaced by new API function `call_answered()`.

juha-h:
I have to reopen this, since also these functions are missing from baresip.h:
```
set_state(call, CALL_STATE_TERMINATED);
call_stream_stop(call);
```
So I would prefer to leave it as is if the testcase of _sending_ of 3xx and closing the call can be somehow made to succeed.

juha-h:
It is OK with me to start using my own forked version of baresip and add add `ua_redirect` plus other stuff there.  Just tell and I'll close this PR.

juha-h:
The root cause why this new API function is needed is that `call_hangup()` should have had

`fmt       Formatted strings with extra SIP Headers`

parameter and proper fix would be to add it.

alfredh:
you can add a new function if you want:

```c
void call_hangup_ex(struct call *call, uint16_t scode, const char *reason, const char *fmt, ...)
```


or just add one new function: `call_redirect`.


Please feel free to skip the testcode

juha-h:
OK, will close this and try instead to add new API function
```
void call_hangup_ex(struct call *call, uint16_t scode, const char *reason, const char *fmt, ...)
```

juha-h:
> you can add a new function if you want:
> 
> ```c
> void call_hangup_ex(struct call *call, uint16_t scode, const char *reason, const char *fmt, ...)
> ```

Looks like that is not enough, also `ua_hangup_ex()` should be added. Otherwise the call does not get closed closed.

juha-h:
This is getting so complicated that after several hours of trials and failures I need to give up.  My original code with new `ua_redirect` API function worked fine, but for some reason was not a good choice.

I'll start using it with my baresip fork unless someone else can provide working API functions and show an example how they can be used to provide 3xx response with Contact header.

PR #2617, opened 2023-06-14
config for activating NAT pinhole opener

cspiel1:
- config: add avt setting rtp_send_ping
- stream: NAT pinhole can be activated via config

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -72,6 +72,7 @@ video_jitter_buffer_delay	5-10	# (min. frames)-(max. packets)
 rtp_stats		no
 #rtp_timeout		60
 #avt_bundle		no
+#rtp_send_ping		no
 
 # Network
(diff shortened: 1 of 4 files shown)


cspiel1:
There is already an account parameter `;natpinhole=on`. Should we name the config setting like that?

This implementation keeps the behavior unchanged for the default configuration of `rtp_send_ping =  off`.

alfredh:
if there is already config for this, it might be good to reuse it.

cspiel1:
For this variant the best name in my opinion is `force_natpinhole`.

Alternative:
If we want the natpinhole setting completely orthogonal to the mnat config can be achieved by means of the already existing account parameter `natpinhole=on/off` then this PR would result in a much simpler configuration: https://github.com/baresip/baresip/pull/2618

PR #2622, opened 2023-06-17
stream: faster video jitter buffer offloading

sreimers:
With higher bitrates video decoding freezes sometimes since jitter buffer offloading is not fast enough.

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -473,9 +473,11 @@ static void rtp_handler(const struct sa *src, const struct rtp_header *hdr,
 
 		if (s->type == MEDIA_VIDEO) {
 			/* Read all ready frames */
-			for (int i = 0; i < 1000; i++) {
+			uint32_t n = jbuf_packets(s->rx.jbuf);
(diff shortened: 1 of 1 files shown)


src/stream.c
```diff
@@ -471,9 +471,14 @@ static void rtp_handler(const struct sa *src, const struct rtp_header *hdr,
 			metric_inc_err(s->rx.metric);
 		}
 
-
-		if (stream_decode(s) == EAGAIN)
-			(void) stream_decode(s);
+		if (s->type == MEDIA_VIDEO) {
+			/* Read all ready frames */
+			while (stream_decode(s) == EAGAIN);
```
alfredh, line 476: Perhaps move semicolon to next line, to indicate that the while-loop is running a no-op.

alfredh: also consider a `for` loop instead of `while` loop to avoid potential deadlocks.
A high number can be chosen, e.g. in the range of hundreds ...

PR #2634, opened 2023-06-22
video: allow stateful encode/decode handling

sreimers:
It can be useful to have access to the timestamp when decoding. For example a codec module can forward packets directly, like a SFU (Selective Forwarding Unit).

It's also useful for a stateful encoder/decoder to have access to `const struct video` (like `vidfilt_encupd_h` and `vidfilt_decode_h`), this way a virtual device name can be used as identifier.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1198,12 +1198,12 @@ struct vidcodec;
 typedef int (videnc_packet_h)(bool marker, uint64_t rtp_ts,
 			      const uint8_t *hdr, size_t hdr_len,
 			      const uint8_t *pld, size_t pld_len,
-			      void *arg);
+			      const struct video *vid);
 
 typedef int (videnc_update_h)(struct videnc_state **vesp,
 			      const struct vidcodec *vc,
 			      struct videnc_param *prm, const char *fmtp,
-			      videnc_packet_h *pkth, void *arg);
+			      videnc_packet_h *pkth, const struct video *vid);
 
 typedef int (videnc_encode_h)(struct videnc_state *ves, bool update,
 			      const struct vidframe *frame,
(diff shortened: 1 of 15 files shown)


include/baresip.h
```diff
@@ -1212,11 +1212,13 @@ typedef int (videnc_encode_h)(struct videnc_state *ves, bool update,
...
-typedef int (viddec_decode_h)(struct viddec_state *vds, struct vidframe *frame,
-                              bool *intra, bool marker, uint16_t seq,
-                              struct mbuf *mb);
+typedef int(viddec_update_h)(struct viddec_state **vdsp,
+			     const struct vidcodec *vc, const char *fmtp,
+			     const struct video *vid);
+
+typedef int(viddec_decode_h)(struct viddec_state *vds, struct vidframe *frame,
+			     bool *intra, bool marker, uint16_t seq,
+			     uint64_t ts, struct mbuf *mb);
```
alfredh, line 1221: one thing to consider: add a struct for input and output arguments, this makes it more flexible
for future changes. Example "struct xxx_packet".

alfredh:
One struct that was added in the past was this:

```c
struct vidpacket {
	uint8_t *buf;        /**< Buffer memory                     */
	size_t size;         /**< Size of buffer                    */
	uint64_t timestamp;  /**< Timestamp in VIDEO_TIMEBASE units */
	bool keyframe;       /**< True=keyframe, False=deltaframe   */
};
```


it would be more elegant if all input/output types for audio+video encode+decode
is a struct, for example "audio_frame" or "audio_packet".

alfredh:
Is there any updates on this patch ?

sreimers:
Sorry had some deadlines, will refactor this PR with your hints next week.

PR #2646, opened 2023-07-10
cmake: link mbedtls if available

cspiel1:


sreimers:
> * in both, and how should we avoid a duplicate?

I think it has to be placed in both. Since we should not install a generic `FindMBEDTLS.cmake` with re.

cspiel1:
MbedTLS can also be built as shared library with `USE_SHARED_MBEDTLS_LIBRARY=on`.  I try to set `STATIC` of baresip automatically.

alfredh:
I would like to suggest that we get this enabled in libre first, and get it working there.

After that we can update the applications.

alfredh:
is this needed?

re can link to mbedtls and baresip to openssl

cspiel1:
Yes I think so. We are going to evaluate baresip/re on Zephyr. So there is no openssl.

alfredh:
building with STATIC=1 and static libre.a is not related.

1. A library can be static archive (.a) or dynamic (.so)
2. Baresip does not need to know if we are using static/dynamic mbedtls
3. If libre.a is linked statically, it does not provide any info about which libs to link to


In theory it should be possible to build libre with OpenSSL and baresip with MbedTLS, or vice versa.

CMakeLists.txt
```diff
@@ -49,6 +49,10 @@ else()
   option(STATIC "Build static" OFF)
 endif()
 
+find_package(MBEDTLSSO)
+if(USE_MBEDTLS AND NOT MBEDTLSSO_FOUND)
+  set(STATIC BOOL ON)
```
alfredh, line 54: I dont agree with this logic

cmake/FindMBEDTLSSO.cmake
```diff
@@ -0,0 +1,19 @@
+find_library(MBEDTLSSO_LIBRARY
```
alfredh, line 1: I dont think there is a need for 2 different FindMbedTLS files -- one should be enough

cspiel1:
You're right. Tried without linking to mbedtls in zephyr and it worked. Since everything is linked statically, we don't need this PR.

Shared linking was only meant for developing on the dev host. But this is not needed.

PR #2653, opened 2023-07-19
dial_number: fix warning array subscript has type 'char'

cspiel1:


src/dial_number.c
```diff
@@ -18,7 +18,7 @@ int clean_number(char* str)
 	 * In other cases trust the user input
 	 */
 	while (str[i]) {
-		if (isalpha(str[i]) != 0)
+		if (isalpha((int) str[i]) != 0)
```
alfredh, line 21: not sure why this is needed ...

passing a char to an int function should be legal, as long as sizeof(char) is less than sizeof(int).

What kind of platform is Zephyr, is it linux based ?

cspiel1: It is RTOS based.

It is a gcc warning:
```
/home/cspiel/dev/zephyrproject/modules/sip/baresip/lib/src/dial_number.c:21:32: warning: array subscript has type 'char' [-Wchar-subscripts]
   21 |                 if (isalpha(str[i]) != 0)
```

Should I turn off the warning instead?

alfredh: Example valid code:

```
int a;
char b = 'X';

a = b;
```

You might want to check the size of these types on your system.

cspiel1:
On Zephyr RTOS I also had troubles that point to memory alignment problems. The result are unexpected values. These patches for `re` solve two problems I found:

- `va_arg( ... , uint...)` failure
  - ee27970cb8ea80c584aec6f167e4c89abb45bf18
- `mbuf_read_uint...` failure
  - d3bba834dba398df940572ab51ce2c2cde292fa5

The gcc warning of this PR seems to relate only to the Zephyr compiler. Thus we could close it for now.

We close our evaluation of baresip on Zephyr RTOS for now. As a result we have a duration measurement of encoding and decoding on a Cortex M7 with opus codec 48kHz mono:
```
encoding 27 % of 20ms
decoding 27 % of 20ms
```
Encoding runs in the audio source thread, where decoding in the main thread. This means that encoding+decoding takes 27% CPU.

PR #2655, opened 2023-07-26
call: logic fix in call_modify. remove call->sent_answer = false

cHuberCoffee:
* call_modify triggers an SIP UPDATE or Re-Invite. Therefore it is an SDP offer. A reset of call->sent_answer is wrong.


cspiel1:
@sreimers : Please could you help with the mingw test?

sreimers:
Is fixed in main, rebase should work.

PR #2665, opened 2023-08-02
video,stream: stop natpinhole timer on re-invite

cspiel1:
- If the video stream direction changes from receive only to send-receive,
send-only or inactive, we should stop sending dummy packets.

- Function `stream_open_natpinhole()` is not needed in the API and it doesn't
need to return an error because it can't fail.


cspiel1:
This is a bugfix. If a re-INVITE switches from one-way video to bi-directional then the dummy packets do not stop. The dummy packets have timestamp zero. Sequence number and payload type do match with the video stream. As a result of the zero timestamp some receivers may flush their decode buffer and loose key frames. The result is a stuttering video playback.

Should we merge this now to have it in the upcoming release on Wednesday?

PR #2667, opened 2023-08-05
test call cancel rules

cspiel1:
Rules can be added to a test fixture when to cancel the re main loop. This should simplify the UA event handler and keep the business logic of the test-case in the test function.

This PR adds a first cancel rule for the CALL_PROGRESS event.  More rules that will simplify the event handler will follow.

- test: call - rules for re_cancel
- test: call - use rule to cancel on UA_EVENT_CALL_PROGRESS
- test: call - print info if rule does not match

Note: We plan to simplify `test_call_change_videodir()`. Then add more early media tests, 100rel tests, audio codec switches, SIP UPDATE.  This should ensure stability of https://github.com/baresip/baresip/pull/2454.

Here is a branch with more commits based on this: https://github.com/cspiel1/baresip/tree/test_call_simplify1

### DIFF
--- a/test/call.c
+++ b/test/call.c
@@ -31,6 +31,25 @@ enum action {
 	ACTION_ATT_TRANSFER,
 };
 
+
+struct cancel_rule {
+	struct le le;
+
+	enum ua_event ev;
+	const char *prm;
+	struct ua *ua;
+
+	unsigned n_incoming;
+	unsigned n_progress;
+	unsigned n_established;
+	unsigned n_audio_estab;
+	unsigned n_video_estab;
+	unsigned n_offer_cnt;
+	unsigned n_answer_cnt;
+	unsigned n_rules_ok;
+};
+
+
 struct agent {
 	struct fixture *fix;    /* pointer to parent */
 	struct agent *peer;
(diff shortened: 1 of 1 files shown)


test/call.c
```diff
@@ -158,6 +213,78 @@ static bool agent_audio_video_estab(const struct agent *ag)
...
+		    ag->ua != rule->ua) {
+			info("test: event %s ua=[%s] (expected [%s]\n",
+			     uag_event_str(ev),
+			     account_aor(ua_account(ag->ua)),
+			     account_aor(ua_account(rule->ua)));
+			continue;
+		}
+
+		if (rule->n_incoming &&
+		    ag->n_incoming != rule->n_incoming) {
```
alfredh, line 244: is the rule to compare exact number of events ?

or can the number of events be equal to or higher ?

remember that the counters should be compared after re_main is done.

cspiel1: For `n_incoming` and the others currently proposed it's better to compare exact number. Later we will add counters for audio/video frames for which we specify lower bounds.

The counters are updated in the `event_handler` before the rules are evaluated. If all specified rules are met, then `re_cancel()` is called and after `re_main` is done further ASSERT_ checks can be done. Then optimally more API calls will trigger further actions and `re_main` is called again.

cspiel1:
Thanks, for review and merge. I'll proceed here soon with more improvements and add more tests.

PR #2670, opened 2023-08-07
video: use const struct video for videnc_update_h and viddec_update_h

sreimers:
With this change it's possible to use `video` getters within codec update callbacks.


alfredh:
does it work with *vidloop* module from baresip-apps ?

sreimers:
I can prepare a PR for baresip-apps modules.

alfredh:
is this ready for merge ?

PR #2678, opened 2023-08-10
stream: declare ext_len when assigned

alfredh:


sreimers:
Yes, that's strange, will try to investigate.

alfredh:
HEAD is also failing with the same error:


```
Assertion failed: 0 && "RE_VA_ARG: arg is not compatible", file /home/runner/work/baresip/baresip/baresip-win32/re/src/fmt/print.c, line 348
```

It could be related to 32-bit platforms.


Is it possible to add some more detailed warnings or a backtrace ?

sreimers:
Yes it is related to the secure [mbuf_printf](https://github.com/baresip/re/commit/b6c16a280e013cef14e86d1a5056ecc1dce85cf2) merge. The failing test is `test_ua_refer`. It should generate a ci backtrace already but looks something is broken too. Try to debug locally.

sreimers:
Ok should be fixed by: https://github.com/baresip/re/pull/903

alfredh:
thanks for investigating and fixing

Issue #2676, opened 2023-08-09
Issue with Refer-To on call_transfer


rodrigodeppe:
Hi,
Just wanted to tell you what happened to us this week.
We were testing our software on a SIP PBX (Coral Tadiran) and were not able to transfer answered calls while connected using baresip.
The issue was solved with 2 fixes on the call_transfer function at src/call.c.
1. Adding "<" and ">" before and after the %s on the arg "Refer-To: %s\r\n"
2. Adding a "Referred-by: <%s>" just after this same arg.

Final result:
(...)
call,
"Refer-To: <%s>\r\nReferred-by: <%s>\r\n", nuri,ua_aor(call->ua));
if (err) {
(...)

What are your thoughts about this?
Best regards,
Rodrigo Deppe

alfredh:
https://github.com/baresip/baresip/wiki/Guidelines-for-reporting-Issues

please include SIP signaling

rodrigodeppe:
Sure thing.

Bad one on WireShark:

Session Initiation Protocol (REFER)
    Request-Line: REFER sip:7528@172.21.26.15:5060 SIP/2.0
    Message Header
        Via: SIP/2.0/UDP 172.21.26.136:60393;branch=z9hG4bK6e36534974067b68;rport
        Contact: <sip:4017@172.21.26.136:60393>
        Max-Forwards: 70
        To: "21999640954"<sip:021999640954@172.21.26.15>;tag=40d0e1d0-ac151a0f-13c4-5459e6-571e3a70-5459e6
        From: <sip:4017@172.21.26.15>;tag=2d1a036a5f7313c3
        Call-ID: 40d49990-ac151a0f-13c4-5459e6-404396eb-5459e6@172.21.26.15
        [Generated Call-ID: 40d49990-ac151a0f-13c4-5459e6-404396eb-5459e6@172.21.26.15]
        CSeq: 19309 REFER
        User-Agent: SIP Phonox Voice
        Refer-To: sip:1133@172.21.26.15
        Content-Length: 0

Session Initiation Protocol (480)
    Status-Line: SIP/2.0 480 Temporarily Unavailible
    Message Header
        From: <sip:4017@172.21.26.15>;tag=2d1a036a5f7313c3
        To: "21999640954"<sip:021999640954@172.21.26.15>;tag=40d0e1d0-ac151a0f-13c4-5459e6-571e3a70-5459e6
[...]

PR #2681, opened 2023-08-15
call: fix Refer-To URI angle brackets

sreimers:
"The Contact, From, and To header fields contain a URI.  If the URI contains a comma, question mark or semicolon, the URI MUST be enclosed in angle brackets (< and >).  Any URI parameters are contained within these brackets.  If the URI is not enclosed in angle brackets, any semicolon-delimited parameters are header-parameters, not URI parameters."

Refs #2676


mbattista:
Just a quick note, that you changed the call_replace_transfer and not the call_transfer function. The first one is not used much in baresip.

sreimers:
Yes I also think its unrelated to the issue, just caught my eye ;-)

The first should be fine I think since `uri_encode` is used, i will double check.

PR #2692, opened 2023-08-30
test: add rtcp_mux test

cspiel1:
- test: add rtcp_mux test
- test: rtcp test wait for RTCP SR instead off RTCP APP

### DIFF
--- a/test/call.c
+++ b/test/call.c
@@ -1770,8 +1770,11 @@ out:
 static int test_call_rtcp_base(bool rtcp_mux)
 {
 	struct fixture fix, *f = &fix;
+	struct cancel_rule *cr;
 	int err = 0;
 
+	err = module_load(".", "ausine");
+	TEST_ERR(err);
 
 	/* Use a low packet time, so the test completes quickly */
 	if (rtcp_mux) {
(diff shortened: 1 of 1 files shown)


cspiel1:
The RTCP_APP packet is received in stream.c `rtp_handler()` because the `rs->rtcp_mux`  in rtp.c was not set fast enough.

cspiel1:
`stream->rtcp_mux` is set if SDP media remote attribute `rtcp-mux` was set. Thus it most likely will be set too late. I am not sure how to fix this.

cspiel1:
So currently the test stops if an RCTP APP packet is received. Which I guess is only meant to open a NAT pinhole for RTCP. Do we really need the RTCP APP at the receiver at the very beginning of the call?

Shouldn't we prefer to test if an RTCP SR is received? 
For that we have to specify an audio source. Otherwise no RTP packets are sent and no RTP tx stat is available.

I will add an audio source and try to activate the RTCP SR. This will fix the RTCP mux test.

cspiel1:
updated the description

cspiel1:
Thanks! This would be another PR for `test_call_rtcp_base()`: https://github.com/baresip/baresip/pull/2687
I have a series of commits that finally adds more tests for RX thread. Do you have time for review?

PR #2702, opened 2023-09-04
Implement play_file_ext with offset parameter in ms

larsimmisch:
This PR allows to start playing a file with an offset, which is useful for resuming audio where it was interrupted

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -775,6 +775,10 @@ typedef void (play_finish_h)(struct play *play, void *arg);
 int  play_file(struct play **playp, struct player *player,
 	       const char *filename, int repeat,
 	       const char *play_mod, const char *play_dev);
+int play_file_ext(struct play **playp, struct player *player,
+	      const char *filename, int repeat,
+	      const char *play_mod, const char *play_dev,
+		  size_t offset_ms);
 int  play_tone(struct play **playp, struct player *player,
 	       struct mbuf *tone,
 	       uint32_t srate, uint8_t ch, int repeat,
@@ -783,6 +787,8 @@ void play_set_finish_handler(struct play *play, play_finish_h *fh, void *arg);
 int  play_init(struct player **playerp);
 void play_set_path(struct player *player, const char *path);
 
+
+int enqueue(const char* molecule, void* arg);
 
 /*
  * User Agent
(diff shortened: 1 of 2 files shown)


src/play.c
```diff
@@ -242,7 +255,7 @@ static int aufile_load(struct mbuf *mb, const char *filename,
 	mem_deref(af);
 
 	if (!err) {
-		mb->pos = 0;
+		mb->pos = (prm.srate * fsize * prm.channels / 1000) / file_size;
```
sreimers, line 258: I think you can use `aufmt_sample_size(prm.fmt)` instead `fsize` here.

include/baresip.h
```diff
@@ -784,6 +788,8 @@ int  play_init(struct player **playerp);
 void play_set_path(struct player *player, const char *path);
 
 
+int enqueue(const char* molecule, void* arg);
```
sreimers, line 791: This looks unrelated?

src/play.c
larsimmisch: Yes, that's much short and more complete (done)

include/baresip.h
larsimmisch: Uh, that needs to go away.

src/play.c
```diff
@@ -207,42 +218,12 @@ static int aufile_load(struct mbuf *mb, const char *filename,
...
-			for (i=0; i<n; i++) {
-				err |= mbuf_write_u16(mb,
-						      g711_ulaw2pcm(buf[i]));
-			}
-			break;
-
-		default:
-			err = ENOSYS;
-			break;
-		}
```
sreimers, line 239: deleted by mistake?

src/play.c
```diff
@@ -207,42 +218,12 @@ static int aufile_load(struct mbuf *mb, const char *filename,
...
 	}
 
 	mem_deref(af);
 
 	if (!err) {
-		mb->pos = 0;
+		mb->pos = (prm.srate * aufmt_sample_size(prm.fmt) * prm.channels / 1000) / file_size;
```
sreimers, line 226: Should `play->offset_ms` involved here?

larsimmisch: Yes - that part is missing entirely and I think I'll have to prepare another PR for `aufile_set_pos` in re. I'll close this PR for now.

PR #2705, opened 2023-09-05
Implement play_file_ext with offset parameter in ms

larsimmisch:
This PR allows to start playing a file with an offset, which is useful for resuming audio where it was interrupted.

It depends on https://github.com/baresip/re/pull/943

### DIFF
--- a/modules/debug_cmd/debug_cmd.c
+++ b/modules/debug_cmd/debug_cmd.c
@@ -143,29 +143,33 @@ static int cmd_play_file(struct re_printf *pf, void *arg)
 {
 	struct cmd_arg *carg = arg;
 	struct config *cfg;
-	const char *filename = carg->prm;
+	char filename[4096] = { 0 };
+	int offset = 0;
 	int err = 0;
-
-	cfg = conf_config();
 
 	/* Stop the current tone, if any */
 	g_play = mem_deref(g_play);
 
-	if (str_isset(filename))
-	{
-		err = re_hprintf(pf, "playing audio file \"%s\" ..\n",
-				 filename);
+	if (str_isset(carg->prm) && sscanf(carg->prm, "%4095s %d", filename, &offset)) {
+
+		cfg = conf_config();
+
+		err = re_hprintf(pf, "playing audio file \"%s\" with offset %d..\n",
+				filename, offset);
 		if (err)
 			return err;
 
-		err = play_file(&g_play, baresip_player(), filename, 0,
-                        cfg->audio.alert_mod, cfg->audio.alert_dev);
+		err = play_file_ext(&g_play, baresip_player(), filename, 0,
+						cfg->audio.alert_mod, cfg->audio.alert_dev, (size_t)offset);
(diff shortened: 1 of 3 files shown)


src/play.c
```diff
@@ -516,6 +523,26 @@ static void parse_play_settings(char *file, int *repeat, int *delay)
...
+ * @param playp    Pointer to allocated player object
+ * @param player   Audio-file player
+ * @param filename Name of WAV file to play
+ * @param repeat   Number of times to repeat
+ * @param play_mod Audio player module
+ * @param play_dev Audio player device
+ * @param offset_ms Offset in milliseconds
+ *
+ * @return 0 if success, otherwise errorcode
+ */
```
larsimmisch, line 542: If the offset is beyond the file size (in milliseconds), the file will not be played. This could be added to the docstring

modules/debug_cmd/debug_cmd.c
```diff
@@ -143,30 +143,34 @@ static int cmd_play_file(struct re_printf *pf, void *arg)
 {
 	struct cmd_arg *carg = arg;
 	struct config *cfg;
-	const char *filename = carg->prm;
+	char filename[4096] = { 0 };
```
larsimmisch, line 146: This could be rewritten using `char filename[strlen(carg->prm)]` if `carg->prm` is set, but that triggers a warning about initialising a variable length array.

cspiel1: Pls avoid such large stack buffers!

modules/debug_cmd/debug_cmd.c
```diff
@@ -143,30 +143,34 @@ static int cmd_play_file(struct re_printf *pf, void *arg)
...
-	cfg = conf_config();
-
 	/* Stop the current tone, if any */
 	g_play = mem_deref(g_play);
 
-	if (str_isset(filename))
-	{
-		err = re_hprintf(pf, "playing audio file \"%s\" ..\n",
-				 filename);
+	if (str_isset(carg->prm) && sscanf(carg->prm, "%4095s %d", filename, &offset)) {
```
cspiel1, line 153: use `re_regex()` instead of `sscanf()`

With `pl_strdup()` you get a C string from a `pl`

Then there is no need for a large buffer.

cspiel1:
The `src/play.c` code is meant for playing ringtones/warning tones for SIP. I am not sure if an offset is useful for this use case.

`debug_cmd` is meant only for debug purposes. Not as a general WAV player.

This would be alternatives:
- a tiny program based on `re`, that makes use of https://github.com/baresip/re/pull/943
- a module in baresip-apps that uses `play_file()` or `play_tone()` of the baresip API

larsimmisch:
> The src/play.c code is meant for playing ringtones/warning tones for SIP. I am not sure if an offset is useful for this use case.

That's a good point (I noticed it will read the entire file into memory).

I'm writing a baresip module anyway, so I can do it there.

PR #2720, opened 2023-09-14
avcodec/decode: refactor hw_frame handling

sreimers:
With hardware decoding I had crashes after `av_hwframe_transfer_data`. It looks like a clean `dst` frame is needed and can not be reused with some hardware codecs.

### DIFF
--- a/modules/avcodec/decode.c
+++ b/modules/avcodec/decode.c
@@ -202,6 +202,7 @@ static int ffdecode(struct viddec_state *st, struct vidframe *frame,
 		    bool *intra)
 {
 	AVFrame *hw_frame = NULL;
+	AVFrame *sw_frame = st->pict;
 	AVPacket *avpkt;
 	int i, got_picture, ret;
 	int err = 0;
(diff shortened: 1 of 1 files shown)


modules/avcodec/decode.c
```diff
@@ -251,42 +259,44 @@ static int ffdecode(struct viddec_state *st, struct vidframe *frame,
...
+				sw_frame->format,
+				av_get_pix_fmt_name(sw_frame->format));
 			goto out;
 		}
 
 		for (i=0; i<4; i++) {
-			frame->data[i]     = st->pict->data[i];
-			frame->linesize[i] = st->pict->linesize[i];
+			frame->data[i]     = sw_frame->data[i];
+			frame->linesize[i] = sw_frame->linesize[i];
```
alfredh, line 283: please note that the frame memory has to be available after this function has returned,
because the vidframe is point to those data buffers ...

sreimers: `av_free` should only `free` the allocated AVFrame not the data itself I think.

Issue #2700, opened 2023-09-02
Data race in WebRTC demo when log_level is LEVEL_DEBUG


paresy:
During my search for some TURN server quirks (https://github.com/baresip/baresip/discussions/2696) i discovered a data race in the WebRTC demo, when log_level is LEVEL_DEBUG.

To reproduce you just need to raise the `log_level` to `LEVEL_DEBUG` in `webrtc/src/main.c`.

```
  re_thread_async_init(ASYNC_WORKERS);
+ log_level_set(LEVEL_DEBUG);
  sys_coredump_set(true);
```

Start the demo (no special parameters required) and press connect to start the stream inside the browser. After seeing the test-video press disconnect. Repeat 1-2 times and the data race should occour.

MacOS Ventura, Clang 14, Thread Sanitizer enabled. re/baresip from main branch of this evening.

```
==================
WARNING: ThreadSanitizer: data race (pid=59742)
  Write of size 8 at 0x000109303148 by thread T18:
    #0 encode_rtp_send <null>:136320512 (libbaresip.9.4.0.dylib:arm64+0x156a8) (BuildId: ac7ddfbb36d733c0b627225dc374d15732000000200000000100000000000d00)
    #1 poll_aubuf_tx <null>:136320512 (libbaresip.9.4.0.dylib:arm64+0x14914) (BuildId: ac7ddfbb36d733c0b627225dc374d15732000000200000000100000000000d00)
    #2 ausrc_read_handler <null>:136320512 (libbaresip.9.4.0.dylib:arm64+0x1281c) (BuildId: ac7ddfbb36d733c0b627225dc374d15732000000200000000100000000000d00)
    #3 play_thread <null>:136321408 (ausine.so:arm64+0x3aa0) (BuildId: f74f4b561a32377280bf09d450def28332000000200000000100000000000d00)
    #4 handler <null>:136320576 (libre.16.4.0.dylib:arm64+0x80608) (BuildId: d9c63719fe4931c4b57ddf1aca473db932000000200000000100000000000d00)
    #5 handler <null>:136320576 (libre.16.4.0.dylib:arm64+0xec630) (BuildId: d9c63719fe4931c4b57ddf1aca473db932000000200000000100000000000d00)

  Previous read of size 8 at 0x000109303148 by main thread (mutexes: write M0):
[...]

PR #2721, opened 2023-09-18
video: protect shared resources in video_debug

alfredh:


alfredh:
we can use patch from paresy:

https://github.com/baresip/baresip/issues/2700


or this patch

PR #2725, opened 2023-09-21
config: jitter buffer size given in packets

cspiel1:
relates to: https://github.com/baresip/re/pull/960
Please see re PR for description!


sreimers:
I think we should keep the frame definition. Since you want to pre-calculate how many min. frames are stored within the jitter buffer. Especially if you use fixed size and manually adjust for audio/video sync. Min. packets are meaningless for video and time calculation. You want to keep `n` frames, so you know your jitter buffer keeps `n` ms (depends on fps).

cspiel1:
Have to separate the plot tools.

PR #2728, opened 2023-09-22
Fixed ua_update_account ("norefersub" extension was not added)

juha-h:


maximilianfridrich:
Maybe it would be better to move the call of `add_extension(ua, "norefersub");` to the `create_register_clients` function where the rest of the extensions are added. Then you don't have to add it to the `ua_update_account` function.

juha-h:
Replaced by https://github.com/baresip/baresip/pull/2731.

PR #2731, opened 2023-09-27
ua, static_menu: Fix 100rel cmd

### LINKED ISSUE #2727: 100rel yes/no issue
While testing 100rel/PRAC with parallel forked calls, I noticed that when I start baresip with `;100rel=yes` in account config and then give menu command
```
/100rel no
```
INVITE does not contain `PRAC` in Allow header, but still contains `100rel` in Supported header:
```
INVITE sip:bar@test.tutpro.com SIP/2.0
...
Allow: INVITE,ACK,BYE,CANCEL,OPTIONS,NOTIFY,SUBSCRIBE,INFO,MESSAGE,UPDATE,REFER
Supported: gruu,path,outbound,replaces,100rel,norefersub
```
If I start baresip with `;100rel=no`, there is no  `100rel` in Supported header.

It looks like a bug to me that `100rel` is still in Supported header after `/100 rel no` command.

maximilianfridrich:
Fixes #2727.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -900,6 +900,8 @@ int  ua_call_alloc(struct call **callp, struct ua *ua,
 struct call *ua_find_call_state(const struct ua *ua, enum call_state st);
 int ua_raise(struct ua *ua);
 int ua_set_autoanswer_value(struct ua *ua, const char *value);
+void ua_add_extension(struct ua *ua, const char *extension);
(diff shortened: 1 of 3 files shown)


juha-h:
Yes, it is best to add  "norefersub" extension in create_register_clients because it is called in ua_alloc and in ua_update_account.

PR #2732, opened 2023-09-27
jbuf frame completeness

cspiel1:
relates to: https://github.com/baresip/re/pull/963

- config: remove jbuf mode adaptive
- config,stream: configure jbuf packet pool size for video
- add plots for jbuf


tools/jbuf/audio/config
```diff
@@ -7,8 +7,8 @@ audio_buffer   10-50
 audio_buffer_mode	adaptive
 audio_silence		0.0
 
-audio_jitter_buffer_type	adaptive
-audio_jitter_buffer_delay   0-500
+audio_jitter_buffer_type	fixed
```
cspiel1, line 10: I have to check if `on` does also work.

We could also set `..._jitter_buffer_type` to deprecated now.

PR #2733, opened 2023-09-27
tools: jbuf plots

cspiel1:
- tools: add jbuf plot tools
- tools: add jbuf audio call plots
- tools: jbuf support png plots
- tools: jbuf add README.md
- tools: jbuf README.md add note about units
- tools: jbuf typo in jbuf.plot
- tools: jbuf README.md add note about peer

### DIFF
--- a/tools/jbuf/generate_plot.sh
+++ b/tools/jbuf/generate_plot.sh
@@ -23,7 +23,7 @@ function gen_datfile() {
     filename=$2
 
     jqc='.traceEvents[] | select (.ph == "'"${ph}"'") | .args.line'
-    cat jbuf.json | jq -r "${jqc}" > $filename
+    jq -r "${jqc}" jbuf.json > $filename
 }
 
 source ./jitter.sh
--- a/tools/jbuf/generate_plot.sh
+++ b/tools/jbuf/generate_plot.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+
+if [ ! -f .env ]; then
+    echo ".env is missing. Copy and edit env-template!"
+    exit 1
+fi
+
+export `cat .env`
+
+echo "target: $target"
+echo "netif:  $netif"
+echo "once:   $once"
+
+if ! which jq; then
+    echo "Install jq"
+    exit 1
+fi
+
+trap "./jitter-off.sh; killall -q baresip" EXIT
+
+function gen_datfile() {
+    ph=$1
+    filename=$2
+
+    jqc='.traceEvents[] | select (.ph == "'"${ph}"'") | .args.line'
+    cat jbuf.json | jq -r "${jqc}" > $filename
+}
+
+source ./jitter.sh
+init_jitter $netif
+
+strm="audio"
+for jmin in 0 10 20; do
+    for i in 0 1; do
+        if [ "$i" == "0" ]; then
+            strm="audio"
+        else
+            strm="video"
+        fi
+
+        echo "########### jitter buffer $strm $jmin ###############"
+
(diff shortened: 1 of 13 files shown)


tools/jbuf/generate_plot.sh
```diff
@@ -0,0 +1,85 @@
...
+fi
+
+trap "./jitter-off.sh; killall -q baresip" EXIT
+
+function gen_datfile() {
+    ph=$1
+    filename=$2
+
+    jqc='.traceEvents[] | select (.ph == "'"${ph}"'") | .args.line'
+    cat jbuf.json | jq -r "${jqc}" > $filename
```
sreimers, line 26:
```suggestion
    jq -r "${jqc}" jbuf.json > $filename
```

tools/jbuf/generate_plot.sh
```diff
@@ -0,0 +1,85 @@
...
+
+        sleep 1
+
+        gen_datfile "P" jbuf.dat
+        gen_datfile "O" overrun.dat
+        gen_datfile "U" underflow.dat
+        gen_datfile "L" toolate.dat
+        gen_datfile "D" duplicate.dat
+        gen_datfile "S" oosequence.dat
+        gen_datfile "T" lost.dat
```
sreimers, line 68: I think it would be nice if generating and testing are separated. So you can manually test easily without test loop.

cspiel1: The loop is in order to generate plots for different jbuf min settings.

tools/jbuf/generate_plot.sh
cspiel1: This didn't work.
```
jq: error: Could not open file jbuf.json: Permission denied
```

So I reverted the change.

sreimers: Strange, maybe depends on version, works fine here.

cspiel1:
### Some Results

## Video with jbuf `adaptive` `0-500`
jbuf_video_0

## Video with jbuf `fixed` `2-500`
jbuf_video_2

tools/jbuf/generate_plot.sh
sreimers: I know, but you can extract generating to a separate function so this can be used to run/perform manually tests and generate plots after this.

sreimers:
> Note: Underflows like counted here are not really a problem. Real underflows of the decoder can't be counted in jbuf so far.

Yes, `waiting` or something like this would be better. But's fine for now. We can optimize wording after refactoring.

tools/jbuf/generate_plot.sh
cspiel1: I guess I understood. Like last commit?

sreimers: Thanks!

tools/jbuf/jbuf.plot
```diff
@@ -0,0 +1,53 @@
...
+#
+# How to generate a plot
+# ======================
+# This gnuplot script plots DEBUG_LEVEL 6 output of jbuf.c. You have to
+# increment the DEBUG_LEVEL in ajb.c if you want to get the table for
+# jbuf.dat. Then call baresip like this:
+#
+# ./baresip 2>&1 | grep -Eo "jbuf.*" > jbuf.dat
+#
+# Call this script. Then compare the plot legend with the variables in jbuf.c!
```
sreimers, line 11: Looks outdated

cspiel1:
Thanks! Please also merge the related PR https://github.com/baresip/re/pull/964.

PR #2735, opened 2023-09-28
Moved adding of "norefersub" extension from ua_alloc() to create_register_clients()

juha-h:
Otherwise "norefersub" is lost from Supported header after ua_update_account() call.


juha-h:
I'll close this because it is the same as https://github.com/baresip/baresip/pull/2734.

PR #2739, opened 2023-10-01
call: include Referred-by: tag in REFERs

rodrigodeppe:
include Referred-by: in call_transfer and call_replace_transfer

PR #2741, opened 2023-10-02
Update README.md

rodrigodeppe:
Adds (partial) support for RFC-3892.


sreimers:
This should be part of this PR: https://github.com/baresip/baresip/pull/2739

Simply add another commit to the PR, will be squash merged.

PR #2749, opened 2023-10-04
Add a list of error handlers for audio errors

larsimmisch:
This is useful when using `audio_set_source` with the `aufile` module and a filename: when the file has played to the end, the error handler is called with a 0 error code and the message `end of file`. This is sometimes needed by an application/module.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,9 @@ Win32/*
 # Visual studio config
 mk/win32/baresip.vcxproj.user
 
+# VSCode config
+.vscode/*
+
 # clangd
 .cache
 compile_commands.json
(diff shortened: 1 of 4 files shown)


alfredh:
this PR looks okay, but have you tried to use UA events:

https://github.com/baresip/baresip/blob/main/include/baresip.h#L800

larsimmisch:
I observe the UA events, but the `end of file` will not be sent as a `UA_EVENT_AUDIO_ERROR`, because the error code is 0 in this case and `UA_EVENT_AUDIO_ERROR` is only sent for real errors.

larsimmisch:
It would be an alternative to also send `UA_EVENT_AUDIO_ERROR` for error code 0, but that might break existing code. If you'd be ok with that breakage, your suggestion seems a better solution.

paresy:
As far as i remember this was the case in older baresip versions and was "fixed" to only send this message when real errors happen

larsimmisch:
Closing this in favor of #2755 (this is even simpler and more in line with the existing code)

PR #2754, opened 2023-10-06
jbuf frame completeness

cspiel1:
This PR improves jbuf in order to
- ensure frame completeness for video and correct order of frames,
- reduce the buffered frames with return EAGAIN if it is possible, but not too early.
- Correct order of frames/packets for audio streams.
- Does not increase `O`-notation complexity for `jbuf_put()` nor `jbuf_get()`.
- Makes adaptive mode obsolete and removes it.

This PR still is a draft and will be compared to https://github.com/baresip/baresip/pull/2757.


juha-h:
PR description reads:
```
config: remove jbuf mode adaptive
```
but [docs/examples/config](https://github.com/baresip/baresip/pull/2754/files#diff-39c18196cf83b2f820313774fd9109ab5d5dd2c40e87ad7fec59e905e8caaf6b) still has it for audio and video.

Have I misunderstood something?

cspiel1:
### Plots for video
- without packet loss
jbuf-video
- with packet loss
jbuf-video-loss

cspiel1:
Plots for audio
- min size 1 **without** packet loss
jbuf-audio
ajb
- min size 1 **with** packet loss
jbuf-audio-loss
ajb-loss
- min size 0 **without** packet loss
jbuf-audio-zero
ajb-zero
- min size 0 **with** packet loss
jbuf-audio-zero-loss
ajb-zero-loss

sreimers:
Looks like for audio, the buffers does not play well together, if I see correctly we have 2-3 packets in jbuf (20-60ms) + up to 80ms within aubuf and this does not compensate a 50ms jitter (leads to real aubuf underruns)?

cspiel1:
Looking at the `ajb` plots auf aubuf it looks like that the computed jitter does not increase fast enough to the expected value of 50ms. Ideally we would expect that there only underruns at the beginning of the simulated jitter. For these plots I used this jitter simulation:
```
sudo tc qdisc add dev ifb1 root netem delay 0ms 50ms loss 0.7%
```

Note that `ajb` is only active if configured with `audio_buffer_mode  adaptive`.

This PR should touch only `jbuf` and
- ensure frame completeness for video and correct order of frames.
- Reduce the buffered frames with `return EAGAIN` if it is possible, but not too early.

After this we could think about moving `ajb` from aubuf to jbuf and use a timer for `jbuf_get()` and the decoding.

alfredh:
please resolve the conflicts ...

cspiel1:
I will resolve the conflicts tomorrow.
@sreimers and me plan to compare this with: https://github.com/baresip/baresip/pull/2757

src/conf.c
```diff
@@ -337,7 +337,7 @@ enum jbuf_type conf_get_jbuf_type(const struct pl *pl)
 {
 	if (0 == pl_strcasecmp(pl, "off"))      return JBUF_OFF;
 	if (0 == pl_strcasecmp(pl, "fixed"))    return JBUF_FIXED;
-	if (0 == pl_strcasecmp(pl, "adaptive")) return JBUF_ADAPTIVE;
```
alfredh, line 340: Perhaps add a warning about the config option has been removed?

cspiel1:
Replaced by: https://github.com/baresip/baresip/pull/2757

PR #2755, opened 2023-10-06
Add UA_EVENT_END_OF_FILE 

larsimmisch:
This will have the side effect of calling `ausrc_error_h` handlers with a 0 error code. I looked at the code (also in baresip-apps) and I didn't see any usages where this would break existing code.

I like the extra UA event `UA_END_OF_FILE`, which makes this more explicit.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -825,6 +825,7 @@ enum ua_event {
 	UA_EVENT_CALL_REMOTE_SDP,     /**< param: offer or answer */
 	UA_EVENT_REFER,
 	UA_EVENT_MODULE,
+	UA_EVENT_END_OF_FILE,
 	UA_EVENT_CUSTOM,
 
 	UA_EVENT_MAX,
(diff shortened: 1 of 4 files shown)


include/baresip.h
```diff
@@ -825,6 +825,7 @@ enum ua_event {
 	UA_EVENT_CALL_REMOTE_SDP,     /**< param: offer or answer */
 	UA_EVENT_REFER,
 	UA_EVENT_MODULE,
+	UA_EVENT_END_OF_FILE,
```
alfredh, line 828: Is the event related to audio, video or both ?

Perhaps add some info in the event name about what it is related to ...?

cspiel1: There is a `vidsrc_error_handler()` in video.c but no callback in call.c. For now only audio files can be handled. Video files could be added in a future PR.
I suggest to add Doxygen which mentions only audio files. This can be replaced if someone adds video file support.

cspiel1: Maybe its better to put the video stuff in a separate PR. Normally we merge with "squash and merge", then it becomes one commit. Also its easier to review if the PR is small.

PR #2757, opened 2023-10-09
jbuf: adaptive playout time calculation

sreimers:
Experimental playout time calculation for jitter buffer. It replaces frame calculation, since it uses a time based reference offset.

TODO:

- [ ] Add unit tests
- [ ] Add tmr fallback to prevent jbuf overflow on jitter spike
- [ ] Compare with current solution and https://github.com/baresip/baresip/pull/2754

It's based on suggestions from this book: https://csperkins.org/standards/rtp-book.html

image


src/jbuf.c
```diff
@@ -515,62 +622,27 @@ int jbuf_get(struct jbuf *jb, struct rtp_header *hdr, void **mem)
...
-				   seq_diff, f->hdr.seq, jb->seq_get);
-		}
-	}
-#endif
+	uint32_t next_playout = (uint32_t)(tmr_jiffies() * jb->p.srate / 1000);
 
-	/* Update sequence number for 'get' */
-	jb->seq_get = f->hdr.seq;
+	/* Check playout time */
+	if (f->playout_time > next_playout) {
```
cspiel1, line 630: Frame completeness is not checked. Is it already ensured by skew and jitter?

sreimers: Yes, it's ensured by jitter compensation, it's now a time based buffering delay, this should give the lowest delay which is needed to buffer.

cspiel1: okay, we can compare the plots if you are ready.

sreimers: Sure, give me some time to prepare and ensure the algorithms are working correctly (wrap-around calculations are tough ;-))

cspiel1:
When `jitter` and/or `skew` reduces, then jbuf should return `EAGAIN` --> seems to be done already in this branch.

Then `aubuf` will increase number of buffered audio frames. Since `jbuf` already does the jitter compensation and packet re-ordering these extra frames can be dropped from `aubuf`. Maybe immediately dropped, or after some short time. This should reduce the latency.

Threads are involved for audio:
Thread 1 - main thread, or the RX thread in near future
- jbuf_put() is called
- jbuf_get(), maybe by a timer. Or maybe by a wait condition (mqueue) from Thread 2.
- does also decoding
- pushes to aubuf

Thread 2 - audio playback thread
- only reads the audio frames from aubuf

sreimers:
> * jbuf_get(), maybe by a timer. Or maybe by a wait condition (mqueue) from Thread 2.

I implemented a fallback tmr solution here (fallback since we should avoid to many event loop wakeups): https://github.com/baresip/baresip/pull/2757/commits/80b2758a2c9b0de29fe1eea244a95a333718973e

alfredh:
what is the status of this PR ?

Can you please rebase it on top of GIT HEAD ?

src/core.h
```diff
@@ -515,3 +515,4 @@ void rtprecv_enable_mux(struct rtp_receiver *rx, bool enable);
 int  rtprecv_debug(struct re_printf *pf, const struct rtp_receiver *rx);
 void rtprecv_mnat_connected_handler(const struct sa *raddr1,
 				    const struct sa *raddr2, void *arg);
+void rtprecv_srate(struct rtp_receiver *rx, uint32_t srate);
```
alfredh, line 518: Perhaps `rtprecv_set_srate` is better ?

src/rtprecv.c
```diff
@@ -126,41 +127,53 @@ static int handle_rtp(struct rtp_receiver *rx, const struct rtp_header *hdr,
...
+		if (err != EAGAIN)
+			break;
+	} while (n--);
 
-	return err;
+	delay = jbuf_next_play(rx->jbuf);
+	if (delay < 0)
+		delay = 10; /* TODO: calculate ptime */
+
+	tmr_start(&rx->tmr_decode, delay, decode_tmr, rx);
```
alfredh, line 176: please note that the accuracy of the timer system is not very accurate.

If you ask a timer to sleep 10 ms it will sleep 10 ms or more.
A system that is overloaded can easily sleep 20,30,40 ms or even more.

sreimers: This is not a big problem I think, `jbuf_next_play` is calculated by the receive time and `aubuf` should cover this. `tmr` has the same accuracy as the previous rtp event calls, but can catch up faster if needed to prevent jbuf overflows.

cspiel1:
I'll try to fix the sanitizer warning. I think it is not related to this PR.

sreimers:
The failing valgrind looks also unrelated: https://github.com/baresip/baresip/issues/2873

src/config.c
```diff
@@ -470,31 +472,24 @@ int config_parse_conf(struct config *cfg, const struct conf *conf)
...
-	(void)conf_get_range(conf, "jitter_buffer_delay",
-			     &cfg->avt.video.jbuf_del);
-	(void)conf_get_range(conf, "jitter_buffer_delay",
-			     &cfg->avt.audio.jbuf_del);
-
 	if (0 == conf_get(conf, "audio_jitter_buffer_type", &jbtype))
 		cfg->avt.audio.jbtype = conf_get_jbuf_type(&jbtype);
 
-	(void)conf_get_range(conf, "audio_jitter_buffer_delay",
+	(void)conf_get_range(conf, "audio_jitter_buffer_delay_ms",
```
cspiel1, line 478: Maybe a little bit shorter? e.g.:
- audio_jitter_buffer_ms
- audio_jbuf_ms

Similar for video.

cspiel1:
All tests are green. Should I generate some plots with jitter simulation? Maybe I could do this next week. Or do you like to try this?

sreimers:
> Or do you like to try this?

Yes, I'm trying to do that right now, I will update this PR: https://github.com/baresip/tools/pull/2/files
So we have repeatable test cases and can capture also some other real scenarios (bad WLAN, LTE etc.).

sreimers:
Some first plots with 50ms qdisc Jitter and
`audio_jitter_buffer_ms 20-300`: 

image
image
image

sreimers:
This how it looks like with no jitter and if receiver clock is faster than sender (drifting), I think there is room for improvement:

image
image

include/baresip.h
```diff
@@ -1506,21 +1508,28 @@ void stream_enable_rtp_timeout(struct stream *strm, uint32_t timeout_ms);
 struct jbuf;
 struct rtp_header;
 
+typedef uint64_t (jbuf_next_play_fn)(const struct jbuf *jb);
```
alfredh, line 1511: You can add the `_h` suffix here ...

modules/multicast/receiver.c
```diff
@@ -797,6 +797,7 @@ int mcreceiver_alloc(struct sa *addr, uint8_t prio)
 	struct mcreceiver *mcreceiver = NULL;
 	struct config_avt *cfg = &conf_config()->avt;
 	struct range jbuf_del;
+	uint32_t jbuf_sz;
```
alfredh, line 800: Should we consider moving multicast module to baresip-apps ?

include/baresip.h
```diff
@@ -419,11 +419,13 @@ struct config_avt {
...
-		struct range jbuf_del;  /**< Delay, number of frames*/
+		struct range jbuf_del;  /**< Max./Min. Delay [ms]   */
+		uint32_t jbuf_sz;       /**< Max. buffer  [packets] */
 	} audio;
 	struct {
 		enum jbuf_type jbtype;  /**< Jitter buffer type     */
-		struct range jbuf_del;  /**< Delay, number of frames*/
+		struct range jbuf_del;  /**< Max./Min. Delay [ms]   */
+		uint32_t jbuf_sz;       /**< Max. buffer  [packets] */
 	} video;
```
alfredh, line 429: The audio and video structs look similar ...

sreimers: Yes, but audio and video need different parameters (audio/video latency or max packet settings). Or do you mean to create a struct for reuse?

alfredh: It is possible to write the code like this for reuse:

```
struct {
    int foo;
    int bar;
    /* etc.. */
} audio, video;
```

sreimers:
image

Here are some stats from production with this branch. The adaption works smoothly and as expected. I will try to find/test some edge cases, but mostly this is ready now.

cspiel1:
Is it possible to adapt the plot scripts in `tools/jbuf` for this branch?

cspiel1:
Build `re` with
```sh
cmake -B build -DUSE_TRACE=ON -DCMAKE_C_FLAGS="-DRE_JBUF_TRACE"
cmake --build build
cmake --install build
```
and tried to build baresip with
```
cmake -B build -DUSE_TRACE=ON -DCMAKE_C_FLAGS="-DRE_JBUF_TRACE"
cmake --build build
```
gives:
```
/home/cspiel/src/baresip/baresip/src/jbuf.c: In function ‘plot_jbuf’:
/home/cspiel/src/baresip/baresip/src/jbuf.c:152:54: warning: passing argument 6 of ‘re_trace_event’ makes pointer from integer without a cast [-Wint-conversion]
  152 |         re_trace_event("jbuf", "plot", 'P', NULL, 0, RE_TRACE_ARG_STRING_COPY,
      |                                                      ^~~~~~~~~~~~~~~~~~~~~~~~
      |                                                      |
      |                                                      int
...
```
and more such errors

sreimers:
Can not reproduce, the but it's JBUF_TRACE instead of RE_JBUF_TRACE now:

`cmake -B build -DUSE_TRACE=ON -DCMAKE_C_FLAGS="-DJBUF_TRACE"`

Plotting scripts can be found here: https://github.com/baresip/tools/pull/2

cspiel1:
You're right. Sorry, wrong branch.

cspiel1:
Can we compare some how with
- https://github.com/baresip/baresip/pull/2754#issuecomment-1750452090 for audio and
- https://github.com/baresip/baresip/pull/2754#issuecomment-1750441498 for video ?

sreimers:
If there are no more comments, I would like to merge this PR this week.

juha-h:
Thanks for the PR.  Currently I have
```
audio_buffer 20-300
audio_buffer_mode adaptive
audio_jitter_buffer_type adaptive
audio_jitter_buffer_delay 0-20
```
Is `audio_jitter_buffer_delay` now gone?  If yes, migration guide would be nice.

sreimers:
Yes, will add a config migration hint within a second PR.

landryb:
Fwiw, with this PR 4.1.0 fails to build on OpenBSD:
```
/usr/obj/ports/baresip-4.1.0/bin/cc -DARCH=\"amd64\" -DHAVE_ARC4RANDOM -DHAVE_ATOMIC -DHAVE_EXECINFO -DHAVE_FORK -DHAVE_GETIFADDRS -DHAVE_GETOPT -DHAVE_KQUEUE -DHAVE_PTHREAD -DHAVE_PWD_H -DHAVE_ROUTE_LIST -DHAVE_SELECT -DHAVE_SELECT_H -DHAVE_SETRLIMIT -DHAVE_SIGNAL -DHAVE_STRERROR_R -DHAVE_STRINGS_H -DHAVE_SYSLOG -DHAVE_SYS_TIME_H -DHAVE_UNAME -DHAVE_UNISTD_H -DHAVE_UNIXSOCK=1 -DMOD_PATH=\"/usr/local/lib/baresip/modules\" -DOPENBSD -DOS=\"OpenBSD\" -DRELEASE -DSHARE_PATH=\"/usr/local/share/baresip\" -DUSE_DTLS -DUSE_OPENSSL -DUSE_OPENSSL_AES -DUSE_OPENSSL_HMAC -DUSE_TLS -DUSE_ZLIB -DVERSION=\"4.1.0\" -DVER_MAJOR=4 -DVER_MINOR=1 -DVER_PATCH=0 -D_GNU_SOURCE -Dbaresip_EXPORTS -I/usr/obj/ports/baresip-4.1.0/baresip-4.1.0/include -I/usr/obj/ports/baresip-4.1.0/baresip-4.1.0/src -I/usr/local/include/re -O2 -pipe -DNDEBUG -std=gnu11 -fPIC -Wall -Wextra -pedantic -Wcast-align -Wbad-function-cast -Wmissing-declarations -Wmissing-prototypes -Wnested-externs -Wno-strict-aliasing -Wold-style-definition -Wshadow -Waggregate-return -Wstrict-prototypes -Wuninitialized -Wvla -Watomic-implicit-seq-cst -Wshorten-64-to-32 -Wno-gnu-zero-variadic-macro-arguments -Wno-c2x-extensions -MD -MT CMakeFiles/baresip.dir/src/jbuf.c.o -MF CMakeFiles/baresip.dir/src/jbuf.c.o.d -o CMakeFiles/baresip.dir/src/jbuf.c.o -c /usr/obj/ports/baresip-4.1.0/baresip-4.1.0/src/jbuf.c

/usr/obj/ports/baresip-4.1.0/baresip-4.1.0/src/jbuf.c:630:9: error: use of undeclared identifier 'ETIME'                                 
  630 |                 err = ETIME;                                                                                                     
      |                       ^                                                                                                          
1 error generated.     
```
i dunno for linux but on OpenBSD that's not a default errno, apparently that's only one in c++ headers (found in /usr/include/c++/v1/errno.h) ? including `<errno.h>` doesnt fix the build.

sreimers:
Should be fixed by: https://github.com/baresip/re/pull/1420

landryb:
> Should be fixed by: [baresip/re#1420](https://github.com/baresip/re/pull/1420)

definitely, many thanks for the super quick fix :)

PR #2758, opened 2023-10-10
test: call - count video frames in videodir tests

cspiel1:
- test: call - wait for three video frames in video test
- test: call - count video frames in videodir test


alfredh:
The next release (3.6.0) is scheduled for Wed 18. October.
Should we merge this before or after the release ?

cspiel1:
There is no risk because the changes are only in tests. So please!

PR #2763, opened 2023-10-13
test call - fix logical and

cspiel1:
- test: call - add cancel_rules_debug()
- test: call - reset all cancel rules and cancel only once


cspiel1:
The bug appears only in more complicated tests like will follow with pr https://github.com/baresip/baresip/pull/2762.
Anyway, I suggest to put it in v3.6.0 to have it correct already in upcoming release.

PR #2764, opened 2023-10-13
call: delay for the initial re-invite after call established

cspiel1:
The initial re-invite should be postponed for a random time in order to avoid
SIP 491 Request Pending error situations with concurrent re-invites.

The issue was 100% reproducible in our Asterisk setup with activated 100rel.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -1915,6 +1915,13 @@ static void set_established_mdir(void *arg)
 }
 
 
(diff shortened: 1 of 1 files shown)


cspiel1:
@alfredh you may decide if it should go to the v3.6.0 or later. It solved the problem for our installations.

PR #2765, opened 2023-10-16
Implement OPTIONS ping

maximilianfridrich:
call: add call_msg_src getter

Related:
- https://github.com/baresip/baresip-apps/pull/30


juha-h:
Is `call_peer_addr` proper name for the getter, because when the peer is behind a SIP proxy `msg->src` address is that of the proxy and not of the peer?

maximilianfridrich:
@juha-h Renamed to `call_msg_src`.

juha-h:
Maximilian Fridrich writes:

> @juha-h Renamed to `call_msg_src`.

Good.

PR #2770, opened 2023-10-19
test: call - improve tests for call progress

cspiel1:
- test: call - beautify cancel rule debug output
- test: call - replace BEHAVIOUR_PROGRESS by answermode early
- test: call - check also value zero for counter
- further, re_main timeout of 5 seconds should be enough

### DIFF
--- a/test/call.c
+++ b/test/call.c
@@ -222,6 +222,12 @@ static struct cancel_rule *cancel_rule_add_and(struct cancel_rule *cr,
 	return r;
 }
 
+#define UINTSET(u) (u) != (unsigned) -1
+
+#define cr_debug_nbr(fld) \
+	(UINTSET(cr->fld) ? \
+		re_hprintf(pf, "    " #fld ":    %u\n", cr->fld) : 0)
+
 
 static int cancel_rule_debug(struct re_printf *pf,
 			     const struct cancel_rule *cr)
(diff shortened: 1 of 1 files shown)


test/call.c
```diff
@@ -197,72 +196,141 @@ static struct cancel_rule *fixture_add_cancel_rule(struct fixture *f,
 	r->n_progress    = n_progress;
 	r->n_established = n_established;
 
-	list_append(&f->rules, &r->le, r);
+	r->n_audio_estab = (unsigned) -1;
+	r->n_video_estab = (unsigned) -1;
+	r->n_offer_cnt   = (unsigned) -1;
+	r->n_answer_cnt  = (unsigned) -1;
+	r->n_vidframe    = (unsigned) -1;
```
sreimers, line 203: Maybe better use `int` as type?

cspiel1: ah, yes.

Currently I am confused why the SIP 183 is coming twice in valgrind retest check. Maybe the SIP retry mechanism causes this?

Not sure if we should check `n_progress==1` or `n_progress>=1`.

cspiel1: Today I already found an issue because of the `==` check. The `call_progress()` was called twice. So I could fix this by
dropping `BEHAVIOUR_PROGRESS`. So `==` check might be useful in general.

cspiel1:
In `test_call_100rel_video` the 180 Ringing is received twice although:
- `call_accept()` is called only once
- It takes less than 4ms between `call_accept` , first and second 180 Ringing. Which is less than T1 = 500ms. So no retransmission takes place.

I can't reproduce this on my dev host.

test/call.c
```diff
@@ -1203,15 +1265,15 @@ int test_call_change_videodir(void)
...
+	f->behaviour = BEHAVIOUR_NOTHING;
 	f->estab_action = ACTION_NOTHING;
 
 	/* Make a call from A to B */
 	err = ua_connect(f->a.ua, 0, NULL, f->buri, VIDMODE_ON);
 	TEST_ERR(err);
 
 	/* wait for CALL_PROGRESS */
-	err = re_main_timeout(10000);
+	err = re_main_timeout(5000);
```
alfredh, line 1276: I would prefer to skip this change.

The reason is that sometimes the test need close to 10 seconds to complete.
For example when running with with valgrind or qemu, and multiple parallel builds.

cspiel1: Okay, the reason was only because of faster debugging. I remove all of this changes in this PR.

Issue #2772, opened 2023-10-20
no ACK to re-INVITE 200 OK


juha-h:
Just noticed that when today's baresip main send re-INVITE (hold) and gets back 200 OK, it does not confirm it with ACK (see below).  Has some changes made recently that could explain this?
```
[0:00:26] audio=28357/10005 video=0/0 (bit/s) efps=0.0/0.0    call: hold sip:bar@test.tutpro.com
17:27:35.726#
TCP 192.168.171.45:53262 -> 192.168.171.45:5060
INVITE sip:bar@test.tutpro.com;gr=urn:uuid:24432fb8-562a-296b-f409-473a010260c1 SIP/2.0
Via: SIP/2.0/TCP 192.168.171.45:53262;branch=z9hG4bKcc980b57fb6cbb7b;rport
Contact: <sip:test@test.tutpro.com;gr=urn:uuid:18ffb445-ce14-78c2-8c2e-6ffd34e4a275>
Max-Forwards: 70
Route: <sip:192.168.171.45;transport=tcp;lr>
To: "Bar" <sip:bar@test.tutpro.com>;tag=7bbf032083534e4c
From: <sip:test@test.tutpro.com>;tag=4ad3f03fd7042a74
Call-ID: ec6d412ac36c7f48
CSeq: 1118 INVITE
User-Agent: baresip v3.6.0 (x86_64/Linux)
Content-Type: application/sdp
Content-Length: 1144

v=0
o=- 2707730650 834682369 IN IP4 192.168.171.45
s=-
c=IN IP4 192.168.171.45
t=0 0
m=audio 8954 RTP/AVP 96 97 98 9 99 18 0 8 10 9 100
a=rtpmap:96 opus/48000/2
a=fmtp:96 stereo=0;sprop-stereo=0;maxaveragebitrate=28000;cbr=0;useinbandfec=1
a=rtpmap:97 AMR-WB/16000
a=rtpmap:98 AMR/8000
a=rtpmap:9 G722/8000
a=rtpmap:99 G7221/16000
a=fmtp:99 bitrate=32000
a=rtpmap:18 G729/8000
a=rtpmap:0 PCMU/8000
a=rtpmap:8 PCMA/8000
a=rtpmap:10 telephone-event/8000
a=fmtp:10 0-15
a=rtpmap:9 G722/8000
a=rtpmap:100 CODEC2/8000
a=sendonly
a=ssrc:1489257764 cname:sip:test@test.tutpro.com
[...]

PR #2775, opened 2023-10-23
test: call - add call on-hold/resume test

cspiel1:
relates to: https://github.com/baresip/re/pull/990
try to reproduce: https://github.com/baresip/baresip/issues/2772

- adds call on hold and resume test
- adds function to wait for SIP ACK

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -221,6 +221,7 @@ bool call_has_audio(const struct call *call);
 bool call_has_video(const struct call *call);
 bool call_early_video_available(const struct call *call);
 bool call_refresh_allowed(const struct call *call);
+bool call_ack_pending(const struct call *call);
 int  call_transfer(struct call *call, const char *uri);
 int  call_replace_transfer(struct call *target_call, struct call *source_call);
 int  call_status(struct re_printf *pf, const struct call *call);
--- a/src/call.c
+++ b/src/call.c
@@ -1557,6 +1557,12 @@ bool call_refresh_allowed(const struct call *call)
 }
 
 
+bool call_ack_pending(const struct call *call)
+{
+	return call ? sipsess_ack_pending(call->sess) : false;
+}
+
+
 /**
  * Get the session call-id for the call
  *
--- a/src/call.c
+++ b/src/call.c
@@ -1557,6 +1557,14 @@ bool call_refresh_allowed(const struct call *call)
 }
 
 
+/**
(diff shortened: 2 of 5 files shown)


cspiel1:
I added a function to re: https://github.com/baresip/re/pull/990

Does somebody have a better idea?

cspiel1:
The last commit replaces ACK check from UA a to b. Because b sends the SIP Reply also b has to wait for the ACK.

Now the ACK check fails because it was not received during re_main. A reason could be that the loop was canceled to early.

cspiel1:
For a correct cancel rule that waits for an ACK maybe we need an event/callback function. Maybe the new getter function is not enough. I have to think if we can avoid this.

cspiel1:
@sreimers do you have a better idea to implement a wait for ACK?
- At least we should avoid the timeout warnings somehow. 
- The last commit also adds an ugly sleep of 1ms.

cspiel1:
re-based to main HEAD after merge of 33047e60cf14324603d2effb79a1b85140557b17

sreimers:
> @sreimers do you have a better idea to implement a wait for ACK?
> 
> * At least we should avoid the timeout warnings somehow.
> * The last commit also adds an ugly sleep of 1ms.

Not sure if I understand all details of the problem, is it maybe possible to add/register or reuse any callback instead of the active wait loop?

cspiel1:
> Not sure if I understand all details of the problem, is it maybe possible to add/register or reuse any callback instead of the active wait loop?
> 
> edit: Ah I see, you suggested something similar already.

Yes, but as we know another callback increases the complexity. Currently there is no callback that can be used. Except if we put an SDP into the ACK. This would test something different, a special case where the offer is contained in the SIP replay and the answer in the ACK, if I understood `ack_handler()` in listen.c correctly.

sreimers:
Maybe a active tmr callback works better in this case, it's not perfect but should prevent timeout warnings.

cspiel1:
I tried to add an `sipsess_ack_h` callback. This is too much only for unit tests. BTW: Found that we have a redundancy use of `call->arg`. It can be overwritten by ABI function `call_set_handlers()`.

The timer is a good idea for checking the ACK.

alfredh:
Thanks for writing this test. This looks like a good addition :)

cspiel1:
Your welcome! In following PRs we could add the wait for ACK condition to more tests. Maybe other re-INVITE tests.

alfredh:
please resolve the conflicts ..

PR #2779, opened 2023-10-25
test: call - add 100rel test for audio

cspiel1:


cspiel1:
This is also ready for merge. After this we are able to activate the RX thread.

cspiel1:
@sreimers @alfredh sorry, please can we merge this test? I want to proceed with the RX thread PR.

PR #2783, opened 2023-10-27
fvad: Add a voice activity detector

larsimmisch:
Support for a VAD filter module. The VAD library used is https://github.com/dpirch/libfvad, which is from the WebRTC code.

It seems to work well.


modules/fvad/fvad.c
```diff
@@ -0,0 +1,317 @@
...
+				goto out;
+			}
+		}
+	}
+
+	if (pos != af->sampc) {
+		warning("fvad_process: samples left over: %d",
+			af->sampc - pos);
+	}
+
```
larsimmisch, line 208: I don't like this. When the ptime is smaller than 10ms, the VAD library cannot process frames.

I see two possibilities:
- return EINVAL when the ptime is smaller than 10ms (this would require writing something like `int audio_get_ptime(struct audio *au, bool tx)`
- store leftover data and process it with the next frame

Thoughts?

include/baresip.h
```diff
@@ -827,6 +827,8 @@ enum ua_event {
 	UA_EVENT_CALL_MENC,
 	UA_EVENT_VU_TX,
 	UA_EVENT_VU_RX,
+	UA_EVENT_VAD_TX,
+	UA_EVENT_VAD_RX,
```
cspiel1, line 831: I suggest to
- Use `UA_EVENT_MODULE` with specific parameter for vad.
- Use function `module_event()` in the module `fvad`. It is not necessary that the event is generated in `call.c`.

It should be possible to implement everything in the module, without any changes in the baresip core. @alfredh should this go to baresip-apps?

The application could evaluate the specific module event.

modules/fvad/fvad.c
cspiel1: You could try with `mbuf` to collect 20ms frames.

include/baresip.h
larsimmisch: Ah, ok. I'll try to rewrite it using module events.

larsimmisch: When sending a module event, I need to pass a `struct call *` or a `struct ua *` to `module_event` so I can make sure the right ua/call processes the event, but all I get is a `struct audio *` and I can't see how to get back at the call from there. Any ideas?

cspiel1: Yes. Currently this is not directly possible. But the list of calls can be searched for the right audio stream. Use `uag_filter_calls()` and pass your list and match handlers!

larsimmisch: Ok, in this case, the `module_event` function would have to get an optional `struct audio *` (and a `struct video *` too) so I can filter in the app.

Or a single `void *` pointer is passed and an informational enum, what sort of pointer it is. Something like:

```
enum ptr_type {
    PTR_CALL,
    PTR_UA,
    PTR_AUDIO,
    PTR_VIDEO
};

void module_event(const char *module, const char *event, ptr_type t, void *sender,
		const char *fmt, ...);
```
I don't like the idea with the back pointer much either.

cspiel1: `encode_update()` and `decode_update()` has a pointer to audio. `uag_filter_calls()` only needs the handlers that your module could provide.

modules/fvad/fvad.c
```diff
@@ -0,0 +1,365 @@
...
+		vad.decupdh = NULL;
+	}
+
+	if (!tx_enabled) {
+		vad.ench = NULL;
+		vad.encupdh = NULL;
+	}
+
+	if (!tx_enabled && !rx_enabled) {
+		warning("neither vad_rx nor vad_tx are enabled"
```
cspiel1, line 340: Use prefix "fvad" and add "\n" at end!
```c
warning("fvad: neither ... \n");
```

cmake/FindFVAD.cmake
```diff
@@ -0,0 +1,29 @@
+find_path(FVAD_INCLUDE_DIR
```
cspiel1, line 1: Maybe someone else has more cmake knowledge, but I suggest to start with pkg-config. Compare with FindAAC.cmake!

```cmake
find_package(PkgConfig QUIET)
pkg_search_module(AAC fdk-aac)
...
```

modules/fvad/fvad.c
```diff
@@ -0,0 +1,365 @@
...
+ *
+ * Copyright (C) 2023 Lars Immisch
+ */
+#include <string.h>
+#include <stdlib.h>
+#include <re.h>
+#include <rem.h>
+#include <baresip.h>
+#include <fvad.h>
+#include "fvad.h"
```
cspiel1, line 14: A name collision? You don't need the header file. Remove it!

modules/fvad/fvad.c
```diff
@@ -0,0 +1,365 @@
...
+ */
+
+
+struct vad_enc {
+	struct aufilt_enc_st af;  /* inheritance */
+	struct le le;
+	const struct audio *au;
+	bool vad_tx;
+	Fvad *fvad;
+	const struct call *call;
```
cspiel1, line 32: You can set the pointer to the call object in the first `..._update()` function. If the pointer is already set, then you don't need another search for the call. Note, that you should not assume any order between the call of `encode_update()` and `decode_update()`.

Then you don't need the pointer to audio.

modules/fvad/fvad.c
```diff
@@ -0,0 +1,365 @@
...
+};
+
+
+struct vad_dec {
+	struct aufilt_enc_st af;  /* inheritance */
+	struct le le;
+	const struct audio *au;
```
cspiel1, line 39: I think you can remove `au`.
`call` should be enough.

cmake/FindFVAD.cmake
larsimmisch: I'll take a look.

modules/fvad/fvad.c
larsimmisch: Done (I'd rather have one, but I dislike the name-collision too)

modules/fvad/fvad.c
larsimmisch: I don't see how I can do that. The `..._update` functions receive an `audio` pointer, but not a `call` pointer.

So the app needs to call `uag_filter_calls` when it first receives the first `UA_EVENT_MODULE` event. The search functions will then set the `call` pointer and this will be filled in on subsequent invocations of `module_event`.

cmake/FindFVAD.cmake
larsimmisch: Done

modules/fvad/fvad.c
```diff
@@ -0,0 +1,365 @@
...
+		for (size_t i = 0; i < ms / chunk_time; ++i) {
+			int err = fvad_process(fvad, buf + pos, sampc);
+			pos += sampc;
+			ms -= chunk_time;
+			if (err > 0) {
+				detected = true;
+				goto out;
+			}
+			else if (err < 0) {
+				warning("fvad_process(%d) failed", sampc);
```
cspiel1, line 212: prefix and "\n" termination

modules/fvad/fvad.c
```diff
@@ -0,0 +1,365 @@
...
+				goto out;
+			}
+		}
+	}
+
+	if (pos != af->sampc) {
+		warning("fvad_process: samples left over: %d",
```
cspiel1, line 219: same here

modules/fvad/fvad.c
```diff
@@ -0,0 +1,365 @@
...
+
+		module_event("fvad", "vad", NULL, (struct call*)vad->call,
+			"%d", vad_rx);
+	}
+
+	return 0;
+}
+
+/* To be used with uag_filter_calls */
+bool find_vad_rx(const struct call *call, void *arg)
```
cspiel1, line 276: Should be static.

Use `uag_filter_calls()` in `decode_update()` and `encode_update()`.  You need a match handler and a list handler, which you can share between decode and encode.

The lists `vad_rxl` and `vad_txl` are not needed.

As I already wrote. You don't need to remember pointer `au` to audio object, only the `call`.

larsimmisch: Doh. I get it now. I'd misunderstood `uag_filter_calls`. It's fixed now.

modules/fvad/fvad.c
```diff
@@ -160,9 +174,9 @@ static int decode_update(struct aufilt_dec_st **stp, void **ctx,
 		return EINVAL;
 	}
 
-	st->au = au;
+	uag_filter_calls(find_first_call, find_call, &fa);
```
cspiel1, line 177: This look good now.

Add a check in both update functions if call is NULL!

modules/fvad/fvad.c
```diff
@@ -0,0 +1,348 @@
...
+	if (!st || !af)
+		return EINVAL;
+
+	bool vad_tx = auframe_vad(vad->fvad, af);
+
+	if (vad_tx != vad->vad_tx) {
+		vad->vad_tx = vad_tx;
+
+		if (vad_stderr)
+			print_vad(61, 32, false, vad_tx);
```
cspiel1, line 258: @sreimers what do you think about this pretty logging? Should we prefer `re_trace`? Or simply `debug()`?

modules/fvad/fvad.c
```diff
@@ -0,0 +1,348 @@
...
+	Fvad *fvad;
+	const struct call *call;
+};
+
+
+struct vad_dec {
+	struct aufilt_enc_st af;  /* inheritance */
+	bool vad_rx;
+	Fvad *fvad;
+	const struct call *call;
```
cspiel1, line 38: `const` does not have benefits here because it has to be cast away for `uag_filter_calls()`.

modules/fvad/fvad.c
```diff
@@ -0,0 +1,348 @@
...
+
+	bool vad_tx = auframe_vad(vad->fvad, af);
+
+	if (vad_tx != vad->vad_tx) {
+		vad->vad_tx = vad_tx;
+
+		if (vad_stderr)
+			print_vad(61, 32, false, vad_tx);
+
+		module_event("fvad", "vad", NULL, (struct call*)vad->call,
```
cspiel1, line 260: Instead NULL pass `call_ua(vad->call)`.

Remove the `struct call*`. `const` for the pointer to call does help here.

modules/fvad/fvad.c
```diff
@@ -0,0 +1,348 @@
...
+	bool vad_rx = auframe_vad(vad->fvad, af);
+
+	if (vad_rx != vad->vad_rx) {
+		vad->vad_rx = vad_rx;
+
+		if (vad_stderr)
+			print_vad(64, 32, false, vad_rx);
+
+		module_event("fvad", "vad", NULL, (struct call*)vad->call,
+			"%d", vad_rx);
```
cspiel1, line 284: This are very many UA events. For each audio frame?

Should we use an interval for the events? Could be set via config.

Maybe a floating average, if it is not already computed as average by the libfvad.

modules/fvad/fvad.c
```diff
@@ -0,0 +1,348 @@
...
+	int16_t *allocated = NULL;
+	bool detected = false;
+
+	/* convert to 16 bit linear */
+	if (af->fmt == AUFMT_S16LE) {
+		buf = af->sampv;
+	}
+	else {
+		allocated = (int16_t*)mem_alloc(sizeof(int16_t) * af->sampc,
+			NULL);
```
cspiel1, line 201: Please allocate only once in `decode_update()` and once in `encode_update()`!

modules/fvad/fvad.c
larsimmisch: Yes, it will be sent on state changes.

modules/fvad/fvad.c
```diff
@@ -0,0 +1,348 @@
...
+	/* process all chunk_sizes that fvad accepts */
+	for (size_t chunk_time_index = 0;
+		chunk_time_index < RE_ARRAY_SIZE(chunk_times_ms);
+		++chunk_time_index) {
+
+		const size_t chunk_time = chunk_times_ms[chunk_time_index];
+		const size_t sampc = af->srate / 1000 * chunk_time;
+
+		for (size_t i = 0; i < ms / chunk_time; ++i) {
+			int err = fvad_process(fvad, buf + pos, sampc);
```
cspiel1, line 218: not sure if this is correct

If `af->srate` is 16000 and `chunk_time` is 60, then
- `sampc == 480` but
- `af->sampc == 320`

The function `fvad_process()` reads to far. Please solve this and check with valgrind that there is no illegal read!
-

modules/fvad/fvad.c
```diff
@@ -132,6 +138,10 @@ static int encode_update(struct aufilt_enc_st **stp, void **ctx,
 	}
 
 	uag_filter_calls(find_first_call, find_call, &fa);
+	if (!fa.call) {
+		mem_deref(st);
+		return EINVAL;
+	}
```
cspiel1, line 144: What I meant was:
```c
if (!st->call) {
    uag_filter_calls(...);
   st->call = fa.call;
}
```

The other check and `return EINVAL` is not needed.

modules/fvad/fvad.c
```diff
@@ -184,26 +198,32 @@ static int decode_update(struct aufilt_dec_st **stp, void **ctx,
...
+
+		if (!*buffer) {
+			*buffer = (int16_t*)mem_alloc(sizeof(int16_t) * af->sampc,
 			NULL);
-		buf = allocated;
+
+			if (!*buffer) {
+				warning("fvad: cannot allocate buffer\n");
+				return false;
+			}
```
cspiel1, line 216: Sorry, I just realized that in case of unsupported sample format module `auconv` should be used. It converts to s16le. So you can remove the else path. Just print a warning in `..._update()` functions and return EINVAL!

There is a config option to bind to s16le for the audio devices. `ausrc_format` and `auplay_format`.

modules/fvad/fvad.c
sreimers: It looks more like a `debug()` case to me. Since it's a extra module and not core related this should be fine.

modules/fvad/fvad.c
```diff
@@ -126,6 +133,7 @@ static int encode_update(struct aufilt_enc_st **stp, void **ctx,
 
 	int err = fvad_set_sample_rate(st->fvad, prm->srate);
 	if (err < 0) {
+		warning("fvad: sample rate %d is not supported\n", prm->srate);
```
cspiel1: use %u

modules/fvad/fvad.c
```diff
@@ -171,6 +186,7 @@ static int decode_update(struct aufilt_dec_st **stp, void **ctx,
...
```
cspiel1: %u

modules/fvad/fvad.c
```diff
@@ -0,0 +1,360 @@
...
+static bool vad_stderr = false;
+
+
+static void enc_destructor(void *arg)
+{
+	struct vad_enc *st = arg;
+
+	if (st->fvad) {
+		fvad_free(st->fvad);
+	}
```
cspiel1, line 55: If there is only one (not wrapped) line we prefer the if body without brackets.
```c
if (st->fvad)
		fvad_free(st->fvad);
```

modules/fvad/fvad.c
```diff
@@ -0,0 +1,360 @@
...
+}
+
+
+static void dec_destructor(void *arg)
+{
+	struct vad_dec *st = arg;
+
+	if (st->fvad) {
+		fvad_free(st->fvad);
+	}
```
cspiel1, line 67:
```c
if (st->fvad)
		fvad_free(st->fvad);
```

modules/fvad/fvad.c
```diff
@@ -0,0 +1,360 @@
...
+	bool vad_tx = auframe_vad(vad->fvad, af);
+
+	if (vad_tx != vad->vad_tx) {
+		vad->vad_tx = vad_tx;
+
+		if (vad_stderr)
+			print_vad(61, 32, false, vad_tx);
+
+		module_event("fvad", "vad", call_get_ua(vad->call), vad->call,
+			"%d", vad_tx);
```
cspiel1, line 273: Please align under `"fvad"`

modules/fvad/fvad.c
```diff
@@ -0,0 +1,360 @@
...
+	bool vad_rx = auframe_vad(vad->fvad, af);
+
+	if (vad_rx != vad->vad_rx) {
+		vad->vad_rx = vad_rx;
+
+		if (vad_stderr)
+			print_vad(64, 32, false, vad_rx);
+
+		module_event("fvad", "vad", call_get_ua(vad->call),
+			(struct call*)vad->call, "%d", vad_rx);
```
cspiel1, line 296: pls align

modules/fvad/fvad.c
cspiel1: @larsimmisch The `vad_stderr` and `print_vad()` looks for me like debug code. If we want to keep debug logging then normally we use `debug()` function. Starting baresip in verbose mode (`-v`) enables the debug output in general.

Alternative: If you want to keep a switch for vad, then call it `fvad_log` and use `info()`. Avoid the special control sequences in the print string. Colors are only used for warnings.

larsimmisch: This was more or less copied from the vumeter module. It's nice when using the console app.

I can change it though.

modules/fvad/fvad.c
larsimmisch: What do you mean with _align_ here? Rename `print_vad` to `print_fvad`?

larsimmisch: It makes sense to distinguish between rx ad tx, so I intend to do:
```
module_event("fvad", "vad_tx", call_get_ua(vad->call), vad->call, vad_tx ? "on" : "off");
```

(and `rx` for the decoding side)

cspiel1: Align:
```c
		module_event("fvad", "vad_tx", call_get_ua(vad->call), vad->call,
			     vad_tx ? "on" : "off");
```

alfredh:
perhaps this new module
belongs in baresip-apps?

larsimmisch:
> perhaps this new module belongs in baresip-apps?

I can move it there, but it is not an application module, it's a filter

cspiel1:
Anyway, it can be moved to baresip-apps. But please first make the last changes we discussed above!

cspiel1:
Okay. Look good now. Only this minor change please:
```patch
diff --git a/modules/fvad/fvad.c b/modules/fvad/fvad.c
index 6249de4d..0ad0f2f8 100644
--- a/modules/fvad/fvad.c
+++ b/modules/fvad/fvad.c
@@ -250,13 +250,11 @@ static int encode(struct aufilt_enc_st *st, struct auframe *af)
 
        if (vad_tx != vad->vad_tx) {
                const char* desc = vad_tx ? "on" : "off";
-
                vad->vad_tx = vad_tx;
 
                debug("vfad: vad_tx: %s\n", desc);
-
                module_event("fvad", "vad_tx", call_get_ua(vad->call),
-                       vad->call, desc);
+                            vad->call, desc);
        }
 
        return 0;
@@ -274,13 +272,11 @@ static int decode(struct aufilt_dec_st *st, struct auframe *af)
 
        if (vad_rx != vad->vad_rx) {
                const char* desc = vad_rx ? "on" : "off";
-
                vad->vad_rx = vad_rx;
 
                debug("vfad: vad_rx: %s\n", desc);
-
                module_event("fvad", "vad_rx", call_get_ua(vad->call),
-                       vad->call, desc);
+                            vad->call, desc);
        }
 
        return 0;
```

cspiel1:
okay, please move this to baresip-apps. Then we can merge it.

Issue #2785, opened 2023-10-28
null pointer dereference when handling OPTIONS request


juha-h:
My Android baresip app crashes due to null pointer dereference when it handles OPTIONS request.  `200 OK` response including sdp body is returned fine.   Finally `ua_handle_options` calls `mem_deref(call);` which calls `mem_deref(call->audio);` which calls `mem_deref(a->strm);`  which calls `print_rtp_stats(s);`, and which calls `metric_n_packets(rtprecv_metric(s->rx));`:
```
struct metric *rtprecv_metric(struct rtp_receiver *rx)
{
	/* it is allowed to return metric because it is thread safe */
	return rx->metric;
}
```
Crash happens, because in above `rx` is NULL.  Crash can be avoided by adding NULL check to it:
```
struct metric *rtprecv_metric(struct rtp_receiver *rx)
{
	/* it is allowed to return metric because it is thread safe */w
	if (!rx) {
		debug("rtprecv: rx is NULL\n");
		return NULL;
	}

	return rx->metric;
}
```
Would it be OK to create a PR that adds the NULL check or is there something wrong with the app?  What does `/* it is allowed to return metric because it is thread safe */` mean?

Full backtrace is below.
```
********** Crash dump: **********
Build fingerprint: 'samsung/a04xx/a04:13/TP1A.220624.014/A045FXXU3CWI2:user/release-keys'
#00 0x00000000002c2fac /data/app/~~FEqqnPIJHjgplNKLQ4-VfA==/com.tutpro.baresip-wlS-mw0Usg3MBOSX_jpm_w==/base.apk!libbaresip.so (rtprecv_metric+0) (BuildId: c89761280bcdfed82a7526df9f7564aca91ec6b2)
                                                                                                                                rtprecv_metric
                                                                                                                                /usr/src/libbaresip-android/baresip/src/rtprecv.c:502:13
#01 0x00000000002c0d34 /data/app/~~FEqqnPIJHjgplNKLQ4-VfA==/com.tutpro.baresip-wlS-mw0Usg3MBOSX_jpm_w==/base.apk!libbaresip.so (BuildId: c89761280bcdfed82a7526df9f7564aca91ec6b2)
                                                                                                                                print_rtp_stats
                                                                                                                                /usr/src/libbaresip-android/baresip/src/stream.c:96:43
                                                                                                                                stream_destructor
                                                                                                                                /usr/src/libbaresip-android/baresip/src/stream.c:133:3
[...]

PR #2786, opened 2023-10-28
rtprecv: fix possible rtprecv_metric null pointer deref

juha-h:
Fixes https://github.com/baresip/baresip/issues/2785. Please commit or provide better fix.


cspiel1:
Thanks!

`stream->rx` is NULL if `use_rtp` is false which is the case for incoming OPTIONS.

I found more possible NULL pointer derefs. PR: https://github.com/baresip/baresip/pull/2787

PR #2777, opened 2023-10-25
RTP audio: activate RX thread via config

cspiel1:
Has to be rebased after
- https://github.com/baresip/baresip/pull/2776
- https://github.com/baresip/baresip/pull/2775
- An one more test PR.

commits:
- test: call - replace re_cancel in audio_sample_handler() by cancel rule
- test: call - for test_call_aulevel() move new auframe_handler()
- test: play - update to auframe in auframe_handler
- test: call - fix integer precision warning
- stream,rtprecv: activate RX thread via config
- test: run call tests with RX_MODE_THREAD
- test: main - init async workers
- rtprecv: detach from rtp/rtcp sockets on stop

PR #2791, opened 2023-10-31
RTP audio: activate RX thread

cspiel1:
- stream,rtprecv: activate RX thread via config
- test: run call tests with RX_MODE_THREAD
- test: main - init async workers
- rtprecv: detach from RTP/RTCP sockets before re_cancel
- stream,rtprecv: detach from RTP/RTCP also in RECEIVE_MODE_MAIN
- test: call - add AUDIO_MODE_THREAD for test_call_aufilt
- test: call - reset to RECEIVE_MODE_MAIN
- rtprecv: use enable flag to prevent data race on stream termination
- test: call - do not test unsupported combination

replaces: https://github.com/baresip/baresip/pull/2777

### DIFF
--- a/.github/workflows/valgrind.yml
+++ b/.github/workflows/valgrind.yml
@@ -43,4 +43,4 @@ jobs:
       run: cmake -B build -DSTATIC=1 -DMODULES="g711;ausine;fakevideo;auconv;dtls_srtp;srtp;aufile" && cmake --build build -j
 
     - name: valgrind test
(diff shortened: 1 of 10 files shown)


cspiel1:
I got stuck here. It seems that `rx_thread_activate()` in `rtprecv_thread()` does not always work with slow timing (e.g. valgrind).

cspiel1:
I forgot the re-factoring part of audio.c.

cspiel1:
The next step for RX thread is: https://github.com/baresip/baresip/pull/2795

PR #2795, opened 2023-11-02
rx thread: refactoring part 2 - separate audio receiver

cspiel1:
- audio: separate aureceiver
- aureceiver: simplify logging of over/underruns
- aup: concurrency safe check if aubuf exists
- aup: use atomic instead mtx lock for latency
- audio: fix sanitizer warning in auplay_write_handler
- aucodec: move aucodec_print to internal API
- audio/video: disable stream in destructor
- aup: correctly read/write atomic
- aup: check atomic ready flag instead of aubuf pointer
- aup: replace atomic by mtx avoids thread sanitizer warnings
- aur: rename aurpipe to audio_recv
- aur: rename audio_recv pointer from rp to ar
- aur: rename functin prefix aup_ to aur_
- audio,aur: measure SW decoding jitter
- aur: print times in ms float numbers
- audio: rename pointer aup to aur
- rtprecv: correct error handling

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -224,6 +224,7 @@ static void audio_destructor(void *arg)
 
 	debug("audio: destroyed (started=%d)\n", a->started);
 
+	stream_enable(a->strm, false);
 	stop_tx(&a->tx, a);
 	stop_rx(&a->rx);
 
--- a/src/audio.c
+++ b/src/audio.c
@@ -127,12 +127,12 @@ struct aurx {
 	int pt;                       /**< Payload type for incoming RTP   */
 	enum aufmt play_fmt;          /**< Sample format for audio playback*/
 
-	struct aurpipe *aup;          /**< Audio receive pipeline          */
+	struct audio_recv *aup;       /**< Audio receive pipeline          */
 	bool first_write;             /**< First write to auplay           */
 };
 
 
-struct aurpipe;
+struct audio_recv;
 
 
 /** Generic Audio stream */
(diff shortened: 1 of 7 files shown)


src/audio.c
```diff
@@ -117,61 +117,29 @@ struct autx {
...
-
-       .--------.   .-------.   .--------.   .--------.
- |\    |        |   |       |   |        |   |        |
- | |<--| auplay |<--| aubuf |<--| aufilt |<--| decode |<--- RTP
- |/    |        |   |       |   |        |   |        |
-       '--------'   '-------'   '--------'   '--------'
-
- \endverbatim
- */
 struct aurx {
```
alfredh, line 120: is it possible to refactor the whole struct ?

src/core.h
```diff
@@ -112,6 +112,36 @@ void audio_sdp_attr_decode(struct audio *a);
...
 
 
+/*
+ * Audio Receiver Pipeline
+ */
+
+struct audio_recv;
+
+int aur_alloc(struct audio_recv **aupp, const struct config_audio *cfg,
+	      size_t sampc, uint32_t ptime);
```
alfredh, line 122: for a new person it is difficult to understand what `aur` means.

perhaps a bit longer, for example `aurecv`.

alfredh:
please note that it is not _only_ refactoring, you are also adding usage of mutex.

Adding a mutex might have implications for timing and race conditions.

src/audio.c
cspiel1: The fields that not moved to `struct audio_recv` have no r/w with the RX thread.

cspiel1:
This PR does not activate the RX thread. Thus the mutex is not need now and does not influence the behavior because everything still is running in the main thread.
Should I remove it for now and add it with the final "RX thread activation" PR?

cspiel1:
I removed the new `aubuf_mtx` for now.

cspiel1:
`aurx->mtx` locks in main:
- for `rx->aubuf_started`
- for `rx->filtl`

I remove all other locks for now.

src/audio.c
cspiel1: I try to move them also to `struct audio_recv` in aureceiver.c

cspiel1:
@alfredh I moved the whole struct rx to aureceiver.c. You're right! Now it looks better.

From my side it's ready for the merge.

alfredh:
Perhaps you can ask some other users/developers to test the patch ?

cspiel1:
@juha-h please could you test this with the Android app?
@mbattista please could you test this with the gtk module?
@robert-scheck do you have also a baresip application and could you test this branch?
@larsimmisch please could you test this in your application and with the vad filter?

The question is does audio work reliable like before? RTCP, audio filter, inbound DTMF, different codecs

larsimmisch:
@cspiel1 I took a brief look and tested recording with VAD and it worked as before

juha-h:
@cspiel1 Tried this PR in my Android app and received audio was fine.

PR #2796, opened 2023-11-02
Debian version upgrade

juha-h:


debian/changelog
```diff
@@ -1,3 +1,9 @@
+baresip (3.6.1) UNRELEASED; urgency=medium
```
sreimers, line 1: The current version in main is 3.6.0

juha-h:
Sebastian Reimers writes:

> > @@ -1,3 +1,9 @@
> +baresip (3.6.1) UNRELEASED; urgency=medium
> 
> The current version in main is 3.6.0

I know, but since tag 3.6.0 there has been many commits to main.  Are
you planning to re-tag 3.6.0?

sreimers:
I'm not sure if we need a extra release for baresip. I just released re v3.6.1 with the relevant fix. We should not bump debian changelog version independent from official version I think. We always increase version numbers only on releases. So fixing the debian version number in main should be fine.

juha-h:
Sebastian Reimers writes:

> I'm not sure if we need a extra release for baresip. I just released
> re v3.6.1 with the relevant fix. We should not bump debian changelog
> version independent from official version I think. We always increase
> version numbers only on releases. So fixing the debian version number
> in main should be fine.

So this didn't fix any bug in 3.6.0?

https://github.com/baresip/baresip/commit/8c2f58336bb585ca11e3fe36c0501b0f04c47701

sreimers:
> So this didn't fix any bug in 3.6.0? [8c2f583](https://github.com/baresip/baresip/commit/8c2f58336bb585ca11e3fe36c0501b0f04c47701)

Not sure, maybe @cspiel1 can decide if this is critical enough for a extra release.

juha-h:
> > So this didn't fix any bug in 3.6.0? [8c2f583](https://github.com/baresip/baresip/commit/8c2f58336bb585ca11e3fe36c0501b0f04c47701)
> 
> Not sure, maybe @cspiel1 can decide if this is critical enough for a extra release.

If not, then I can edit the PR so that Debian version is 3.6.0.  Current one 3.5.0 is not correct.

cspiel1:
It was not a critical bug. It fixes only that too many UA events are thrown. At least it is not worth to force a creation of a release branch `3.6.x`.

alfredh:
Thanks!

Unrelated: it looks like Homebrew has deprecated macOS 11, do we need to build for both macOS 11 and 12 ?

PR #2800, opened 2023-11-06
test/call.c: extend test_call_hold_resume

maximilianfridrich:
Updates the `test_call_hold_resume` test by initiating a call and offering the streams as `sendonly`. The test verifies that this does not trigger an `UA_EVENT_CALL_HOLD` on the receiving side of this offer. It then does a few hold/resumes from both sides and verifies the events `UA_EVENT_CALL_HOLD` and `UA_EVENT_CALL_RESUME` are only triggered when they should.

### DIFF
--- a/test/call.c
+++ b/test/call.c
@@ -77,6 +77,8 @@ struct agent {
 	unsigned n_video_estab;
 	unsigned n_offer_cnt;
 	unsigned n_answer_cnt;
+	unsigned n_hold_cnt;
+	unsigned n_resume_cnt;
 	unsigned n_vidframe;
 	unsigned n_auframe;
 	double aulvl;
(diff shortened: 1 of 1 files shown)


cspiel1:
Looks good to me.

cspiel1:
In a future PR we should make the counter `n_...` to be integer instead of uint.
`-1` means do not compare the value.

alfredh:
This test became unstable after the merge:

https://github.com/baresip/baresip/actions/runs/6783024753/job/18436437302

can you please investigate ?

maximilianfridrich:
@alfredh I'm not sure if I'm interpreting the linked output correctly, but it looks like it found a data race when running `test_call_100rel_audio`. I don't know how I could've caused a data race with this change - I didn't touch `test_call_100rel_audio`.

cspiel1:
I will have a look.

cspiel1:
> This test became unstable after the merge:
> 
> https://github.com/baresip/baresip/actions/runs/6783024753/job/18436437302
> 
> can you please investigate ?

This https://github.com/baresip/baresip/pull/2809 should directly solve the sanitizer warning.

PR #2802, opened 2023-11-10
audio: fix inbound dtmf END event

cspiel1:
DTMF END is missing when registered to Asterisk.

Asterisk sends less RTP EVENT packets than baresip and does not mix audio RTP packets and DTMF packets like baresip. If the "DTMF end" follows directly a normal DTMF packet than it was not recognized.

- call `stream_pt_handler()` always
- `stream_pt_handler()` in audio and video have a lightweight check at top
- NULL pointer check of audio.rx is not needed

Alternative check with less lines, but more difficult to read:
```c
	if (pt == rx->pt && (!rx->pt_tel || pt != rx->pt_tel))
		return 0;
```

Does somebody prefer the shorter check?

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -150,6 +150,7 @@ struct aurx {
 	void *sampv;                  /**< Sample buffer                   */
 	uint32_t ptime;               /**< Packet time for receiving       */
 	int pt;                       /**< Payload type for incoming RTP   */
+	int pt_tel;                   /**< Payload type for tel event      */
 	double level_last;            /**< Last audio level value [dBov]   */
 	bool level_set;               /**< True if level_last is set       */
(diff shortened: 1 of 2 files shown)


src/audio.c
```diff
@@ -738,25 +739,34 @@ static int stream_pt_handler(uint8_t pt, struct mbuf *mb, void *arg)
...
 	if (lc && !str_casecmp(lc->name, "telephone-event")) {
+		rx->pt_tel = pt;
 		handle_telev(a, mb);
 		return ENODATA;
 	}
 
 	if (!lc)
 		return ENOENT;
 
+	if (rx->pt == pt)
```
cspiel1, line 763: This check is not needed. But simpler validation.

PR #2810, opened 2023-11-14
uag: fallback for registrar-less NAT setups

cspiel1:
Fixes SIP 404 for NAT setups.

In registrar less setups the IP address in the Request-URI might differ from
the local IP address if the SIP request comes from NAT. The UA fallback should
not be restricted to local IP addresses.

### DIFF
--- a/src/uag.c
+++ b/src/uag.c
@@ -245,38 +245,6 @@ static bool uri_user_and_host(const struct uri *uri)
 }
 
 
-static bool uri_host_local(const struct uri *uri)
-{
-
-	const char *hostv[] = {
-		"localhost",
-		"127.0.0.1",
-		"::1"
-	};
-	struct sa sap;
-	size_t i;
-	int err;
-
-	if (!uri)
-		return false;
-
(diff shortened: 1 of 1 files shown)


cspiel1:
@juha-h please could you have a look and test this with a registrar-less NAT setup, SIP outbound?

juha-h:
Sorry, I'm currently on short vacation without my laptop. Also,  the setup would be difficult for me to arrange.

cspiel1:
Have a nice vacation!

Then let me give some arguments that can't break anything. This commit makes the fallback UA selection less restrictive.
```patch
-                       if (!uri_host_local(&msg->uri))
-                               continue;
-
```

- Choosing the wrong UA can't happen, because the removed check does not consider the current account.
- The exact match `0 == pl_casecmp(cuser, &acc->luri.user)` follow it overwrites the fallback UA.

So what changes:
Some cases where baresip answered with SIP 404 will be answered with SIP 180 Ringing including the Contact header that is configured in the accounts.
This cases are registrar-less SIP requests sent to the NAT and forwarded to baresip.

PR #2812, opened 2023-11-21
menu: add follow up invite timer

cspiel1:
This change fixes an issue that occurs by calling `ua_connect()` in the UA
event handler. The OUTGOING event was sent to the application before the
REFER or REDIRECT events that should trigger a follow up SIP call.

The UA events should be sent in this order to the application:
- REFER, then OUTGOING
- REDIRECT, then OUTGOING

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -461,6 +461,33 @@ static void redial_handler(void *arg)
 }
 
 
+static void invite_handler(void *arg)
+{
+	(void)arg;
+
+	const char *uri = menu.invite_uri;
+
+	if (!str_isset(uri))
+		return;
+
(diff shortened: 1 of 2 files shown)


modules/menu/menu.c
```diff
@@ -461,6 +461,33 @@ static void redial_handler(void *arg)
...
+static void invite_handler(void *arg)
+{
+	(void)arg;
+
+	const char *uri = menu.invite_uri;
+
+	if (!str_isset(uri))
+		return;
+
+	ua_connect(uag_find_requri(uri), NULL, NULL, uri, VIDMODE_ON);
```
alfredh, line 473: Return value must be checked here.

PR #2813, opened 2023-11-22
jbuf: fix memory leak in jbuf_debug()

cspiel1:


src/jbuf.c
```diff
@@ -750,11 +750,11 @@ int jbuf_stats(const struct jbuf *jb, struct jbuf_stat *jstat)
 int jbuf_debug(struct re_printf *pf, const struct jbuf *jb)
 {
 	int err = 0;
-	struct mbuf *mb = mbuf_alloc(512);
 
 	if (!jb)
 		return 0;
 
+	struct mbuf *mb = mbuf_alloc(512);
```
sreimers, line 757: We should null check `mb` here too.

cspiel1: ah, your right. I'll make another PR.

cspiel1: PR: https://github.com/baresip/baresip/pull/2814

PR #2815, opened 2023-11-22
release v3.7.0

cspiel1:


cspiel1:
@sreimers @juha-h  I didn't get the outcome of your discussions about the debian changelog.

 `UNRELEASED` or unstable?

sreimers:
Unstable looks fine to me.

juha-h:
I don't know if there was an outcome.  I just used `dch` command to create new entry and by default it produces `UNRELEASED` entry.  That could then be changed by the maintainer to something else when he/she makes the version available in a Debian distribution.

cspiel1:
ABI version incremented, because new values added to `enum ua_event`

PR #2816, opened 2023-11-24
Decode url in custom header

nltd101:
**Problem:**
utilize /uaaddheader to add a custom header
usage: /uaaddheader <key>=<value> <ua-idx>
So that it is impossible to custom header containing space (' '). 
**My solution:**
implement decode URL format before pushing to custom headers

### DIFF
--- a/modules/menu/static_menu.c
+++ b/modules/menu/static_menu.c
@@ -1275,6 +1275,7 @@ static int cmd_addheader(struct re_printf *pf, void *arg)
 	const char *usage = "usage: /uaaddheader <key>=<value> <ua-idx>\n";
 	struct pl n, v;
 	int err;
+	int ret;
 
 	if (!ua) {
 		re_hprintf(pf, usage);
@@ -1289,13 +1290,16 @@ static int cmd_addheader(struct re_printf *pf, void *arg)
 	}
 
 	struct mbuf mbe;
-	err = mbuf_printf(&mbe, "%H", uri_header_unescape, &v);
-	if (err == 0) {
+	mbuf_init(&mbe);
+	ret = mbuf_printf(&mbe, "%H", uri_header_unescape, &v);
+	if (ret == 0) {
 		v.p = (const char *)mbe.buf;
 		v.l = mbe.end;
 	}
 
-	return ua_add_custom_hdr(ua, &n, &v);
+	ret = ua_add_custom_hdr(ua, &n, &v);
+	mem_deref(mbe.buf);
+	return ret;
 }
 
 


modules/menu/static_menu.c
```diff
@@ -1266,6 +1266,40 @@ static int cmd_uareg(struct re_printf *pf, void *arg)
 	return 0;
 }
 
+/**
+ * Decode url format
+ * 
+ * @param dst  destination
+ * @param src  source
```
alfredh, line 1273: check doxygen comments

modules/menu/static_menu.c
```diff
@@ -1266,6 +1266,40 @@ static int cmd_uareg(struct re_printf *pf, void *arg)
...
+ *
+ */
+void urldecode(struct pl w)
```
alfredh, line 1276: this is pass by value

better to use a pointer?

alfredh:
thanks

please take a look at the failing CI jobs.

modules/menu/menu.c
```diff
@@ -1012,6 +1013,39 @@ struct ua   *menu_ua_carg(struct re_printf *pf, const struct cmd_arg *carg,
...
+/**
+ * Decode url format
+ *
+ * @param w  Pl correspond the string is in URL format will be decoded
+ */
+void url_decode(struct pl* w)
+{
+	char *dst = (char*)w->p;
+	char *src = (char*)w->p;
+	int len = w->l;
```
sreimers, line 1025: w->l is `size_t` so len should be `size_t` too.

nltd101: Thank you

modules/menu/menu.c
```diff
@@ -1012,6 +1013,39 @@ struct ua   *menu_ua_carg(struct re_printf *pf, const struct cmd_arg *carg,
...
+ */
+void url_decode(struct pl* w)
+{
+	char *dst = (char*)w->p;
+	char *src = (char*)w->p;
+	size_t len = w->l;
+	char a, b;
+	while (*src && src < w->p+len) {
+		if ((*src == '%') &&
+				((a = src[1]) && (b = src[2])) &&
```
sreimers, line 1029: src[1]) and src[2] can be out of bound here if url is malformed, right? a struct pl is not necessary null terminated. so pl length has to be checked.

nltd101: I got it. Thanks for correcting me.

alfredh:
Is it an option to use this function for decode and/or escaping/un-escaping:

https://github.com/baresip/re/blob/main/include/re_uri.h#L37

alfredh:
can you please clarify what is that function doing ? is it URL encode or decode ?

can you add some sample URLs to the documentation ?

modules/menu/menu.c
```diff
@@ -1012,6 +1013,40 @@ struct ua   *menu_ua_carg(struct re_printf *pf, const struct cmd_arg *carg,
 	return ua;
 }
 
+/**
+ * Decode url format
+ *
+ * @param w  Pl correspond the string is in URL format will be decoded
+ */
+void url_decode(struct pl* w)
```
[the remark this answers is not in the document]
nltd101: Thanks, I have tried. I will use escaping/un-escaping in re_uri.h as your suggestion above.

modules/menu/static_menu.c
```diff
@@ -1288,6 +1288,13 @@ static int cmd_addheader(struct re_printf *pf, void *arg)
 		return EINVAL;
 	}
 
+	struct mbuf mbe;
+	err = mbuf_printf(&mbe, "%H", uri_header_unescape, &v);
+	if (err == 0) {
+		v.p = (const char *)mbe.buf;
```
sreimers, line 1294: - `mbe` has to be initialized
- `mbuf_printf` allocates memory in `mbe.buf` and has to be derefed after `ua_add_custom_hdr`
- instead of `err` please use `ret` here

PR #2818, opened 2023-11-27
sipsess: refactor and simplify SDP negotiation state

maximilianfridrich:
fix and simplify SDP negotiation state handling

The bool sent_answer is removed and instead the sdp_neg_state of the sipsess is used which accurately tracks the state of the SDP negotiation.

Further, call_set_mdir was removed and unified in call_set_media_direction.

Related re PR:

- https://github.com/baresip/re/pull/1016

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -264,9 +264,9 @@ const char   *call_user_data(const struct call *call);
 int call_set_user_data(struct call *call, const char *user_data);
 void call_set_evstop(struct call *call, bool stop);
 bool call_is_evstop(struct call *call);
-bool call_sent_answer(const struct call *call);
 int call_msg_src(const struct call *call, struct sa *sa);
 enum sip_transp call_transp(const struct call *call);
+enum sdp_neg_state call_sdp_neg_state(const struct call *call);
 
 /*
  * Custom headers
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -267,6 +267,7 @@ bool call_is_evstop(struct call *call);
 int call_msg_src(const struct call *call, struct sa *sa);
 enum sip_transp call_transp(const struct call *call);
 enum sdp_neg_state call_sdp_neg_state(const struct call *call);
+bool call_sdp_change_allowed(const struct call *call);
 
 /*
  * Custom headers
--- a/modules/menu/static_menu.c
+++ b/modules/menu/static_menu.c
@@ -158,9 +158,10 @@ static int cmd_answerdir(struct re_printf *pf, void *arg)
 		ua = call_get_ua(call);
 	}
 
-	if (call_sent_answer(call))
-		(void)call_set_media_estdir(call, adir, vdir);
-	else
+	(void)call_set_media_estdir(call, adir, vdir);
+
+	enum sdp_neg_state sdp_state = call_sdp_neg_state(call);
+	if (sdp_state == SDP_NEG_NONE || sdp_state == SDP_NEG_REMOTE_OFFER)
 		(void)call_set_media_direction(call, adir, vdir);
 
 	err = answer_call(ua, call);
--- a/modules/menu/static_menu.c
+++ b/modules/menu/static_menu.c
@@ -160,8 +160,7 @@ static int cmd_answerdir(struct re_printf *pf, void *arg)
 
 	(void)call_set_media_estdir(call, adir, vdir);
 
-	enum sdp_neg_state sdp_state = call_sdp_neg_state(call);
-	if (sdp_state == SDP_NEG_NONE || sdp_state == SDP_NEG_REMOTE_OFFER)
+	if (call_sdp_change_allowed(call))
 		(void)call_set_media_direction(call, adir, vdir);
 
 	err = answer_call(ua, call);
--- a/src/call.c
+++ b/src/call.c
@@ -3232,10 +3232,30 @@ enum sip_transp call_transp(const struct call *call)
  *
  * @param call Call object
  *
- * @return 0 on success, non-zero otherwise
  * @return SDP negotiation state
  */
 enum sdp_neg_state call_sdp_neg_state(const struct call *call)
 {
 	return call ? sipsess_sdp_neg_state(call->sess) : SDP_NEG_NONE;
+}
+
+
+/**
+ * Check if an SDP change is allowed currently
+ *
+ * @param call Call object
+ *
+ * @return true if SDP change is currently allowed, false otherwise
+ */
+bool call_sdp_change_allowed(const struct call *call)
+{
+	if (!call)
+		return false;
+
+	enum sdp_neg_state sdp_state = call_sdp_neg_state(call);
+
+	return (call->state == CALL_STATE_ESTABLISHED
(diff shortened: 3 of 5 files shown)


test/call.c
```diff
@@ -1703,6 +1705,8 @@ static int test_100rel_audio_base(enum audio_mode txmode)
 
 	f->a.n_auframe=0;
 	f->b.n_auframe=0;
+	err = call_set_media_estdir(ua_call(f->a.ua), SDP_INACTIVE,
+				    SDP_INACTIVE);
 	err = call_set_media_direction(ua_call(f->a.ua), SDP_SENDRECV,
```
cspiel1, line 1710: use `|=`

modules/menu/static_menu.c
```diff
@@ -158,9 +159,10 @@ static int cmd_answerdir(struct re_printf *pf, void *arg)
 		ua = call_get_ua(call);
 	}
 
-	if (call_sent_answer(call))
-		(void)call_set_media_estdir(call, adir, vdir);
-	else
+	(void)call_set_media_estdir(call, adir, vdir);
+
+	sdp_state = call_sdp_neg_state(call);
```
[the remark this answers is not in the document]
maximilianfridrich: Done.

test/call.c
maximilianfridrich: Done.

src/call.c
```diff
@@ -1159,7 +1162,7 @@ int call_connect(struct call *call, const struct pl *paddr)
 		if (err)
 			return err;
 
-		call_set_mdir(call, call->estadir, call->estvdir);
+		call_set_media_direction(call, call->estadir, call->estvdir);
```
cspiel1, line 1165: returns err, but we are not interested in it.

Since `stream_set_ldir()` also returns void, I think `call_set_media_direction()` should also return void.

maximilianfridrich: Done.

cspiel1:
Looks good for me.

modules/menu/static_menu.c
```diff
@@ -158,10 +158,11 @@ static int cmd_answerdir(struct re_printf *pf, void *arg)
...
 	}
 
-	if (call_sent_answer(call))
-		(void)call_set_media_estdir(call, adir, vdir);
-	else
-		(void)call_set_media_direction(call, adir, vdir);
+	(void)call_set_media_estdir(call, adir, vdir);
+
+	enum sdp_neg_state sdp_state = call_sdp_neg_state(call);
+	if (sdp_state == SDP_NEG_NONE || sdp_state == SDP_NEG_REMOTE_OFFER)
```
cspiel1, line 164: Is it possible to move this checks to the baresip core?

juha-h: If a change applies to all apps and not just the menu app, I would prefer to make the change in the core so that all apps would benefit.

maximilianfridrich: I added the `call_sdp_change_allowed` function to call.c.

I can not move this check into the `call_set_media_direction` function because this function **only** sets the local directions of the streams and this could for example break the `/medialdir` command in `dynamic_menu.c`.

maximilianfridrich:
Note: The bool `got_offer` in the call struct can not safely be removed and replaced by using the SDP state because this bool is used in places where we are not guaranteed that a sipsess exists yet (e.g. `call_alloc`, `call_streams_alloc`).

This is different for the `sent_answer` bool, it can be safely replaced by the SDP state.

test/call.c
```diff
@@ -2934,8 +2938,8 @@ static int test_call_hold_resume_base(bool tcp)
 	ASSERT_TRUE(!call_ack_pending(ua_call(f->b.ua)));
 
 	/* set media inactive from B */
-	err = call_set_media_direction(ua_call(f->b.ua), SDP_INACTIVE,
-				       SDP_INACTIVE);
+	call_set_media_direction(ua_call(f->b.ua), SDP_INACTIVE,
+				 SDP_INACTIVE);
 	TEST_ERR(err);
```
cspiel1, line 2943: Is this TEST_ERR obsolete?

test/call.c
```diff
@@ -2951,8 +2955,8 @@ static int test_call_hold_resume_base(bool tcp)
 	ASSERT_EQ(1, f->b.n_hold_cnt);
 
 	/* set call to resume from B */
-	err = call_set_media_direction(ua_call(f->b.ua), SDP_SENDRECV,
-				       SDP_SENDRECV);
+	call_set_media_direction(ua_call(f->b.ua), SDP_SENDRECV,
+				 SDP_SENDRECV);
 	TEST_ERR(err);
```
cspiel1, line 2960: This TEST_ERR also obsolete?

test/call.c
maximilianfridrich: Yes, done.

src/ua.c
```diff
@@ -1317,11 +1317,8 @@ int ua_connect_dir(struct ua *ua, struct call **callp,
 		goto out;
 
 	if (adir != SDP_SENDRECV || vdir != SDP_SENDRECV) {
-		err = call_set_media_direction(call, adir, vdir);
-		if (err) {
-			mem_deref(call);
-			goto out;
-		}
+		(void)call_set_media_estdir(call, adir, vdir);
+		call_set_media_direction(call, adir, vdir);
```
cspiel1, line 1321: We removed returned error because we don't care about the media direction if `call==NULL`, generally.
In my option we should also remove the error code for `call_set_media_estdir()`.
This PR already breaks the ABI because it replaces `call_sent_answer()` by `call_sdp_neg_state()` and `call_sdp_change_allowed()`.
Any other opinions?

cspiel1:
Looks good now.

include/baresip.h
```diff
@@ -255,17 +255,18 @@ struct call  *call_find_linenum(const struct list *calls, uint32_t linenum);
 struct call  *call_find_id(const struct list *calls, const char *id);
 void call_set_current(struct list *calls, struct call *call);
 const struct list *call_get_custom_hdrs(const struct call *call);
-int call_set_media_direction(struct call *call, enum sdp_dir a,
+void call_set_media_direction(struct call *call, enum sdp_dir a,
```
alfredh, line 258: is it possible to split up the patch into logical blocks ?

alfredh:
We also have this type, is it possible to reuse it:


```
/*
 * Session Description
 */

/* RTCSdpType */
enum sdp_type {
	SDP_NONE,
	SDP_OFFER,
	SDP_ANSWER,
	SDP_ROLLBACK  /* special type */
};
```

include/baresip.h
maximilianfridrich: I tried to split it up a bit and got rid of one of the commits.

maximilianfridrich:
> I would like to suggest that you make a separate PR for the change that changes the return value from int to void..

Will do.

alfredh:
what is the status here ?

cspiel1:
It is ready for merge. First https://github.com/baresip/re/pull/1016 then this. We tested and use it already in our products.

cspiel1:
Thanks!

PR #2828, opened 2023-12-12
misc: rx thread activate

cspiel1:
The last part for the audio RX "real-time" thread.

Enables RX thread via config, uses worker for passing work to the main thread and adds tests to ensure thread safety.

- stream,rtprecv: activate RX thread via config
- test: run call tests with RX_MODE_THREAD
- test: main - init async workers
- rtprecv: detach from RTP/RTCP sockets before re_cancel
- stream,rtprecv: detach from RTP/RTCP also in RECEIVE_MODE_MAIN
- test: call - add AUDIO_MODE_THREAD for test_call_aufilt
- test: call - reset to RECEIVE_MODE_MAIN
- rtprecv: use enable flag to prevent data race on stream termination
- test: call - do not test unsupported combination
- test: call - failure debug for test_media_base()
- rtprecv: detach sockets before attach
- rtprecv: warnings for UDP thread attach
- aur: add mutex locks
- aur: add aubuf_mtx

### DIFF
--- a/src/aureceiver.c
+++ b/src/aureceiver.c
@@ -292,9 +292,11 @@ void aurecv_receive(struct audio_recv *ar, const struct rtp_header *hdr,
 	if (!mb)
 		goto out;
 
+	mtx_lock(ar->mtx);
 	if (hdr->pt != ar->pt) {
+		mtx_unlock(ar->mtx);
 		*ignore = true;
-		return;
+		goto unlock;
 	}
 
 	*ignore = false;
@@ -349,6 +351,9 @@ void aurecv_receive(struct audio_recv *ar, const struct rtp_header *hdr,
 /*                (void)aurecv_stream_decode(ar, hdr, mb, lostc, drop);*/
 
 	(void)aurecv_stream_decode(ar, hdr, mb, 0, drop);
+
+unlock:
+	mtx_unlock(ar->mtx);
 }
 
 
@@ -357,7 +362,9 @@ void aurecv_set_extmap(struct audio_recv *ar, uint8_t aulevel)
 	if (!ar)
 		return;
 
+	mtx_lock(ar->mtx);
 	ar->extmap_aulevel = aulevel;
+	mtx_unlock(ar->mtx);
 }
 
 
@@ -435,10 +442,10 @@ void aurecv_flush(struct audio_recv *ar)
 	if (!ar)
 		return;
 
+	mtx_lock(ar->mtx);
 	aubuf_flush(ar->aubuf);
 
 	/* Reset audio filter chain */
-	mtx_lock(ar->mtx);
 	list_flush(&ar->filtl);
 	mtx_unlock(ar->mtx);
 }
@@ -455,6 +462,7 @@ int aurecv_decoder_set(struct audio_recv *ar,
 	info("audio: Set audio decoder: %s %uHz %dch\n",
 	     ac->name, ac->srate, ac->ch);
 
+	mtx_lock(ar->mtx);
 	if (ac != ar->ac) {
 		ar->ac = ac;
 		ar->dec = mem_deref(ar->dec);
@@ -471,6 +479,7 @@ int aurecv_decoder_set(struct audio_recv *ar,
 	ar->pt = pt;
 
 out:
+	mtx_unlock(ar->mtx);
 	return err;
 }
 
@@ -517,7 +526,9 @@ bool aurecv_level_set(const struct audio_recv *ar)
 	if (!ar)
 		return false;
 
+	mtx_lock(ar->mtx);
 	set = ar->level_set;
+	mtx_unlock(ar->mtx);
 
 	return set;
 }
@@ -529,7 +540,9 @@ double aurecv_level(const struct audio_recv *ar)
 	if (!ar)
 		return 0.0;
 
+	mtx_lock(ar->mtx);
 	v = ar->level_last;
+	mtx_unlock(ar->mtx);
 
 	return v;
 }
@@ -542,7 +555,9 @@ const struct aucodec *aurecv_codec(const struct audio_recv *ar)
 	if (!ar)
 		return NULL;
 
+	mtx_lock(ar->mtx);
 	ac = ar->ac;
+	mtx_unlock(ar->mtx);
 	return ac;
 }
 
@@ -567,7 +582,9 @@ void aurecv_stop(struct audio_recv *ar)
 		return;
 
 	ar->auplay = mem_deref(ar->auplay);
+	mtx_lock(ar->mtx);
 	ar->ac = NULL;
+	mtx_unlock(ar->mtx);
 }
 
 
@@ -713,6 +730,7 @@ int aurecv_debug(struct re_printf *pf, const struct audio_recv *ar)
 		goto out;
 	}
 
+	mtx_lock(ar->mtx);
 	bpms = (double)ar->srate * ar->ch * aufmt_sample_size(ar->fmt) /
 	       1000.0;
 	err  = mbuf_printf(mb,
@@ -750,6 +768,8 @@ int aurecv_debug(struct re_printf *pf, const struct audio_recv *ar)
 			  ar->ap ? ar->ap->name : "none",
 			  ar->device,
 			  aufmt_name(ar->play_fmt));
+	mtx_unlock(ar->mtx);
+
 	if (err)
 		goto out;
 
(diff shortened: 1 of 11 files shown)


cspiel1:
When everything is green the commit https://github.com/baresip/baresip/pull/2828/commits/ee740e3ebf8777e7a62e5b5af0220ff3e7d2fefd will be reverted.

cspiel1:
This can be tested by enabling the RX thread with this line in your `~/.baresip/config`:
```
rtp_rxmode		thread
```

@juha-h Please could you test this with the Android App?

@larsimmisch Please could you test this with the vad filter?

larsimmisch:
I saw a problem on `main` yesterday with fvad and my application. I play a file using `audio_set_source("aufile", filename)` and then start a recording with `audio_set_source("aufile", "record.wav")`. Starting the recording returns 0, but the file is empty. Looking into it a bit, I saw that in `aurecv_start_player`, `aurecv_codec(ar);` returns `NULL`, so it exits early.

On `v.3.7.0`, the recording is created successfully. I don't understand why I didn't see that earlier.

Any hints what to look for? Creating a minimal test application would obviously be good, but I wouldn't get to it before the weekend.

larsimmisch:
Ah, it can be reproduced through the debug interface.

Something like:
```
/insmod menu
/uanew <sip:test@immisch-macbook-pro.local>
```

Then place an incoming call, accept it and:
```
/auplay aufile,record.wav
```

My config has:
```
audio_path		/usr/local/share/baresip
audio_player		aufile,/dev/null
audio_source		aufile,/usr/local/share/baresip/ringback_s16.wav
module			fvad.so
```

juha-h:
@cspiel1 I tried the PR in my Android app and receiving audio worked OK.  I saw these in the log:
```
12-14 09:43:05.870 25379 25426 D Baresip Lib: rtp_receiver: RTP RX thread started
...
12-14 09:43:06.388 25379 25426 D Baresip Lib: audio_recv: create audio buffer [20 - 300 ms] [640 - 9600 bytes]
```

cspiel1:
Thanks @juha-h for the test. The log is exactly what I would expect.

@larsimmisch I understood that this problem is in `main` branch HEAD. So not directly related to this PR. I will try to analyse+solve this immediately. Thanks!

cspiel1:
@larsimmisch The issue you found has nothing to do with `fvad`. Here is the fix: https://github.com/baresip/baresip/pull/2830

larsimmisch:
Thanks for the speedy fix on `main`!

I've now tested this branch (with the fix cherry-picked) with the config `rtp_rxmode thread`  and I did see `rtp_receiver: RTP RX thread started`.

The test was fine.

cspiel1:
Thanks for the test!

src/aureceiver.c
```diff
@@ -722,16 +746,19 @@ int aurecv_debug(struct re_printf *pf, const struct audio_recv *ar)
 		err |= mbuf_printf(mb, "       time = (not started)\n");
 	}
 
-	err |= re_hprintf(pf, "       player: %s,%s %s\n",
+	err |= mbuf_printf(mb, "       player: %s,%s %s\n",
```
alfredh, line 749: this looks like a separate bugfix ? perhaps a separate PR for this ?

src/config.c
```diff
@@ -470,6 +472,14 @@ int config_parse_conf(struct config *cfg, const struct conf *conf)
 	(void)conf_get_u32(conf, "rtp_timeout", &cfg->avt.rtp_timeout);
 
 	(void)conf_get_bool(conf, "avt_bundle", &cfg->avt.bundle);
+	if (0 == conf_get(conf, "rtp_rxmode", &rxmode)) {
+
+		if (0 == pl_strcasecmp(&rxmode, "thread")) {
+			cfg->avt.rxmode = RECEIVE_MODE_THREAD;
+			warning("rtp_rxmode thread is currently "
+				"experimental\n");
+		}
```
alfredh, line 481: please compare all values, and print a warning if it matches none

src/config.c
```diff
@@ -633,6 +644,8 @@ int config_print(struct re_printf *pf, const struct config *cfg)
 			 cfg->avt.rtp_stats ? "yes" : "no",
 			 cfg->avt.rtp_timeout,
 			 cfg->avt.bundle ? "yes" : "no",
+			 cfg->avt.rxmode == RECEIVE_MODE_THREAD ? "thread" :
+								  "main",
```
alfredh, line 648: What if someone adds a new mode called e.g. "foo" ?

juha-h:
I have not seen a rational for rx thread.  What are its advantages/disadvantages?  In which situations it should he activated?  Why is it not the default mode?

src/aureceiver.c
cspiel1: https://github.com/baresip/baresip/pull/2852

src/config.c
cspiel1: I added `rtp_receive_mode_str()`.

cspiel1:
> I have not seen a rational for rx thread. What are its advantages/disadvantages? In which situations it should he activated? Why is it not the default mode?

We had this discussion earlyer in this year. To sum it up: Audio processing should be done in an independent thread. The main thread handles everything from SIP signaling, parts of video processing and also processing of arbitrary baresip modules.

In the meanwhile we already have test results that show that the RX thread reduces the maximum deviation from the packet time that is caused by long blocking calls at the beginning of a call. For our products we already switched to `rtp_rxmode thread`.

For now `rtp_rxmode main` will be the default mode which will not change the behavior. In the future we could make `rtp_rxmode thread` to the default or maybe the only supported mode.

cspiel1:
TODO:
```
$ ./selftest -r main
$ ./selftest -r thread
```

cspiel1:
TODO:
- [ ] check if new selftest command arguments work
- [ ] check if github actions run both RX modes

juha-h:
@cspiel1 Thanks for the explanation.

cspiel1:
Now `test_call_rtcp` produces a sanitizer warning (not in single test mode).

In `re/src/rtp/sess.c` when the sender report is created in `mk_sr()` which runs in the main thread.

cspiel1:
It is a mystery to me why this didn't appear earlier. The output of  sanitizer is:
```
./test/selftest -r thread
...
Write of size 8 at 0x7b4c00034428 by thread T59 (mutexes: write M0):
    #0 handle_incoming_sr /home/cspiel/src/baresip/re/src/rtp/sess.c:181
    ...
Previous read of size 8 at 0x7b4c00034428 by main thread (mutexes: write M1):
    #0 sender_apply_handler /home/cspiel/src/baresip/re/src/rtp/sess.c:400
```

@sreimers solutions that I can think of
- (A) Run the RTCP sender report timer in the RX thread.
- (B) Add a lock to `rtp_source()`.

(A) most likely will produce new concurrency problems that would have to be solved.
While with (B) we have to be careful to hold lock time short and don't forget any r/w cases.

cspiel1:
Most of the RTCP work in `re` is done from RX thread if activated.
These functions are called from main thread:
- `rtcp_debug()`
- `rtcp_stats()`
- `send_rtcp_report()`, but this could be changed with solution (A)

cspiel1:
Last commit implements solution (A).

TODO:
- lock `rtcp_debug()`
- lock `rtcp_stats()`

cspiel1:
Rebased and dropped the reverted changes in `tests`.

The following commit will add delayed calls to `audio_debug()`. There are further concurrency problems with RTCP.

cspiel1:
PR https://github.com/baresip/re/pull/1039 solves the concurrency problems with RTCP vs audio_debug.

cspiel1:
Rebased, added the commit 46729c94e6eede25ed3a8a9aee1803417c7924a4 and ran all the selftest-s 50 times successfully with clang sanitizer.

cspiel1:
I further did again call tests with PCMU/PCMA, opus, G.722 with `rtp_rxmode ` `main` and `thread`. It is ready for the merge.

Are there any further suggestions?

cspiel1:
- Rebased after merge of https://github.com/baresip/baresip/pull/2867.
- made test_call_rtcp more stable and faster with  https://github.com/baresip/baresip/compare/26136d611b10227b01cbf71438336c2e5bf4cf17..a5f8af44e73a3e75802b52dc91d314a37b9a0d45

alfredh:
it should be ok to merge this now..

cspiel1:
Thank you for your support, reviews and interest on this work!

PR #2830, opened 2023-12-14
audio,aur: audio_set_player() only stops auplay

cspiel1:
This fixes audio_set_player()


larsimmisch (approved):
This fixes my problem as far as I can see, thanks!

PR #2831, opened 2023-12-14
cmake: Fix rpath on MacOS

larsimmisch:
Without this change, baresip does not run and complains about not finding libbaresip.dylib in /usr/local/openssl/lib (which is the wrong location for it)


CMakeLists.txt
```diff
@@ -54,6 +54,9 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(CMAKE_C_STANDARD 11)
 set(CMAKE_C_EXTENSIONS OFF)
 
+set(CMAKE_MACOSX_RPATH 1)
```
sreimers, line 57: Should not be needed, the default is True, I think.

https://cmake.org/cmake/help/latest/policy/CMP0042.html#policy:CMP0042

PR #2845, opened 2023-12-25
ci: use actions/checkout@v4

robert-scheck:
Should be mainly Node20


robert-scheck:
The build failure is not related to this change, could somebody re-run the failed CI check, please?

PR #2847, opened 2023-12-27
magic: use assert() instead of BREAKPOINT

alfredh:


src/magic.h
```diff
@@ -19,7 +21,7 @@
 	if (MAGIC != s->magic) {					\
 		warning("%s: wrong magic struct=%p (magic=0x%08x)\n",	\
 			__func__, s, s->magic);			\
-		BREAKPOINT;						\
+		assert(false);						\
```
sreimers, line 24: should we use `re_assert(false)` instead?

PR #2867, opened 2024-01-09
aur: a mutex for aubuf allocation

cspiel1:
The aubuf is created when the first audio frame is pushed. It can't be created during initialization because e.g. `srate` and `channels` may be changed in the filters.

Since the auplay thread (often) is started before the first audio frame is decoded, it has to check if the aubuf was already created.
Thus a mutex is needed.


cspiel1:
You're welcome!

PR #2869, opened 2024-01-10
menu: an incoming call should not change the current call

cspiel1:
Additionally, the ringtone is only started for the first incoming call.
The ringtone is not stopped if another call is incoming.

If there are multiple incoming calls, then the first incoming call is answered with command `accept` by default.

Current behavior in main: The last incoming call is answered.

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -349,8 +349,6 @@ static bool menu_play(const struct call *call,
 
 static void play_incoming(const struct call *call)
 {
-	/* stop any ringtones */
-	menu_stop_play();
 
 	if (call_state(call) != CALL_STATE_INCOMING)
 		return;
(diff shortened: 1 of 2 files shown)


cspiel1:
@alfredh you wrote once into `call.c`

```c
/* NOTE: The new call must always be added to the tail of list,
 *       which indicates the current call.
 */
 list_append(lst, &call->le, call);
```

Is this still wanted?
As a workaround I added to menu.c in this PR:
```
menu_selcall(menu.curcall);
```

alfredh:
the tail is used to indicate current call.

This logic could be changed or improved, if you think it is faulty ...

cspiel1:
> the tail is used to indicate current call.
> 
> This logic could be changed or improved, if you think it is faulty ...

Okay, I will think about this.

Issue #2873, opened 2024-01-12
test_call_change_videodir fails sometimes.


sreimers:
With valgrind this test fails sometimes:

```c
selftest: ASSERT_EQ: /home/sreimers/projekte/baresip/baresip-linux/baresip/test/call.c:1431: test_call_change_videodir(): expected=0(0x0), actual=3(0x3)

Cancel Rules:                                                                                                                                               
  --- ? ---                                                                                                                                                 
    prm:  vidframe                                                                                                                                          
    ua:   sip:b@127.0.0.1                                                                                                                                   
    n_incoming:    1                                                                                                                                        
    n_progress:    0                                                                                                                                        
    n_established:    1                                                                                                                                     
    n_vidframe:    3                                                                                                                                        
    met:  no                                                                                                                                                
  AND -->                                                                                                                                                   
  --- ? ---                                                                                                                                                 
    prm:  vidframe                                                                                                                                          
    ua:   sip:a@127.0.0.1                                                                                                                                   
    n_incoming:    0                                                                                                                                        
    n_progress:    1                                                                                                                                        
    n_established:    1                                                                                                                                     
    n_vidframe:    3                                                                                                                                        
    met:  no                                                                                                                                                
  --- CALL_REMOTE_SDP ---                                                                                                                                   
    prm:  offer                                                                                                                                             
    ua:   sip:b@127.0.0.1                                                                                                                                   
    n_incoming:    1                                                                                                                                        
    n_progress:    0                                                                                                                                        
    n_established:    1                                                                                                                                     
    met:  no                                                                                                                                                
  AND -->                                                                                                                                                   
  --- CALL_REMOTE_SDP ---                                                                                                                                   
    prm:  answer                                                                                                                                            
    ua:   sip:a@127.0.0.1                                                                                                                                   
    n_incoming:    0                                                                                                                                        
    n_progress:    1                                                                                                                                        
    n_established:    1                                                                                                                                     
    met:  no                                                                                                                                                
Agents:                                                                                                                                                     
  --- Agent a ---          
    close_scode: 0                                                                                                                                          
    failed: no                                                                                                                                              
    n_incoming: 0                                                                                                                                           
    n_progress: 1                                                                                                                                           
    n_established: 1                                                                                                                                        
    n_closed: 0                                                                                                                                             
    n_transfer_fail: 0                                                        
    n_dtmf_recv: 0                                                                                                                                          
    n_transfer: 0                                                             
[...]

PR #2876, opened 2024-01-12
call: fix race condition for call_set_video_dir()

cspiel1:
fixes https://github.com/baresip/baresip/issues/2873

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -1499,6 +1499,7 @@ int call_set_video_dir(struct call *call, enum sdp_dir dir)
 	if (!call)
 		return EINVAL;
 
+	call->estvdir = dir;
 	stream_set_ldir(video_strm(call_video(call)), dir);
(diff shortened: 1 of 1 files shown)


src/call.c
```diff
@@ -1499,6 +1499,7 @@ int call_set_video_dir(struct call *call, enum sdp_dir dir)
 	if (!call)
 		return EINVAL;
 
+	call->estvdir = dir;
```
sreimers, line 1502: Can this happen for `call_set_audio_ldir` too?

cspiel1: The functions `call_set_..._ldir()` sets only the local direction. It does not generate a re-INVITE or UPDATE, nor change the media direction for the established state.

`call_set_video_dir()` should change the video direction by sending a re-INVITE or UPDATE and remember also the direction locally.

PR #2877, opened 2024-01-15
account,message,ua: secure incoming SIP MESSAGEs

maximilianfridrich:
Currently, BareSIP accepts all incoming SIP MESSAGE requests as long as they are addressed to a valid UA. This poses a security risk and is not conformant to the security considerations in [RFC 3428 section 11](https://www.rfc-editor.org/rfc/rfc3428#section-11).

With this change, SIP MESSAGEs can be enabled per account using the new account option `uas_req` (the default is off) and can be set to require TLS. If `uas_user` or `uas_pass` are set, authentication is required.

Further, a foundation was built to extend this to other out-of-dialog requests like REFER.

### DIFF
--- a/docs/examples/accounts
+++ b/docs/examples/accounts
@@ -41,7 +41,7 @@
 #    ;stunpass=STUN/TURN/ICE-password
 #    ;stunserver=stun:[user:pass]@host[:port]  # see RFC 3986: Use of the format "user:password" in the userinfo field is deprecated.
 #    ;tcpsrcport=6060
-#    ;uas_req={yes, no, tls}  # default: no
+#    ;inreq_allowed={yes, no}  # default: no
 #    ;uas_user=username
 #    ;uas_pass=password
 #    ;video_codecs=h264,vp8,...
--- a/docs/examples/accounts
+++ b/docs/examples/accounts
@@ -41,6 +41,7 @@
 #    ;stunpass=STUN/TURN/ICE-password
 #    ;stunserver=stun:[user:pass]@host[:port]  # see RFC 3986: Use of the format "user:password" in the userinfo field is deprecated.
 #    ;tcpsrcport=6060
+#    ;uas_req={yes, no, tls}  # default: no
 #    ;uas_user=username
 #    ;uas_pass=password
 #    ;video_codecs=h264,vp8,...
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -79,11 +79,10 @@ enum jbuf_type {
 	JBUF_ADAPTIVE
 };
 
-/** Defines the UAS request mode */
-enum uas_req_mode {
-	UAS_REQ_MODE_OFF = 0,
-	UAS_REQ_MODE_ON,
-	UAS_REQ_MODE_TLS
+/** Defines the incoming out-of-dialog request mode */
+enum inreq_mode {
+	INREQ_MODE_OFF = 0,
+	INREQ_MODE_ON,
 };
 
 struct account;
@@ -99,7 +98,7 @@ int account_set_sipnat(struct account *acc, const char *sipnat);
 int account_set_answermode(struct account *acc, enum answermode mode);
 int account_set_rel100_mode(struct account *acc, enum rel100_mode mode);
 int account_set_dtmfmode(struct account *acc, enum dtmfmode mode);
-int account_set_uas_req_mode(struct account *acc, enum uas_req_mode mode);
+int account_set_inreq_mode(struct account *acc, enum inreq_mode mode);
 int account_set_display_name(struct account *acc, const char *dname);
 int account_set_regint(struct account *acc, uint32_t regint);
 int account_set_stun_uri(struct account *acc, const char *uri);
@@ -128,7 +127,7 @@ uint32_t account_prio(const struct account *acc);
 enum answermode account_answermode(const struct account *acc);
 enum rel100_mode account_rel100_mode(const struct account *acc);
 enum dtmfmode account_dtmfmode(const struct account *acc);
-enum uas_req_mode account_uas_req_mode(const struct account *acc);
+enum inreq_mode account_inreq_mode(const struct account *acc);
 const char *account_display_name(const struct account *acc);
 const char *account_aor(const struct account *acc);
 const char *account_auth_user(const struct account *acc);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -79,6 +79,13 @@ enum jbuf_type {
 	JBUF_ADAPTIVE
 };
 
+/** Defines the UAS request mode */
+enum uas_req_mode {
+	UAS_REQ_MODE_OFF = 0,
+	UAS_REQ_MODE_ON,
+	UAS_REQ_MODE_TLS
+};
+
 struct account;
 
 int account_alloc(struct account **accp, const char *sipaddr);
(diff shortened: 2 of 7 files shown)


alfredh:
another approach would be to create a new module called e.g. `message.so`.

When the module is not loaded, SIP MESSAGE is not supported.
When the module is loaded, SIP MESSAGE is supported.

cspiel1:
Maybe we find a unique solution for all incoming out of dialog SIP requests. E.g. SIP REFER, SIP MESSAGE, ...

For SIP REFER we use `uas_user`/`uas_pass`

juha-h:
> Maybe we find a unique solution for all incoming out of dialog SIP requests. E.g. SIP REFER, SIP MESSAGE, ...
> 
> For SIP REFER we use `uas_user`/`uas_pass`

Yes, unique solution would be good.  REFER without user's consent is especially dangerous.

maximilianfridrich:
@juha-h Yes, the proposed solution and new option `uas_req` is intended to be used for other out-of-dialog requests like REFER too. This will be done in a simple follow-up PR that uses the new functionality from this PR.

juha-h:
REFER is different from MESSAGE.  When out-of-dialog REFER comes in, user of the account should be asked if he/she is willing to let the REFER to happen or at least there should be an option to require such a consent request.

maximilianfridrich:
That feature definitely makes sense, but I don't think it is within the scope of this PR. Some functionality of securing MESSAGEs could still potentially be reused for REFER, but this should probably happen in a separate discussion.

docs/examples/accounts
```diff
@@ -41,6 +41,7 @@
 #    ;stunpass=STUN/TURN/ICE-password
 #    ;stunserver=stun:[user:pass]@host[:port]  # see RFC 3986: Use of the format "user:password" in the userinfo field is deprecated.
 #    ;tcpsrcport=6060
+#    ;uas_req={yes, no, tls}  # default: no
```
alfredh, line 44: the current behaviour is to accept all incoming.

alfredh:
I think this kind of logic should be handled in the SIP proxy.

What if the transport changes from TLS to TCP or vice versa ?

juha-h:
To me, param name `uas_req` does not tell anything.  Something more obvious should be used as the name.  Also, I don't understand what tls transport protocol choice is and how it is related to yes and no.

cspiel1:
Maybe
```
uas_allowed={yes, no, tls}  # default: no
```
`tls` means that accepting incoming out of dialog requests is only allowed with transport TLS. In combination with `uas_user` and `uas_pass` it is secure.

cspiel1:
> I think this kind of logic should be handled in the SIP proxy.
> 
> What if the transport changes from TLS to TCP or vice versa ?

This PR is about adding security for incoming SIP REFER and SIP MESSAGE which are out-of-dialog SIP requests. Each request is checked independently for the transport protocol.

We'll also think about your proposal for a message.so module.

cspiel1:
We set this to draft and concentrate first on TLS client verification:
- https://github.com/baresip/re/pull/1059
- https://github.com/baresip/baresip/pull/2888

maximilianfridrich:
@alfredh Currently, the BareSIP message logic is part of the core and there are five modules which rely on those public API functions. We looked into moving the messaging logic into its own `message.so` module, here's what we found:

- The BareSIP API would change and five public functions would be removed from BareSIP.
- Five modules rely on these API functions and all of those would have to be refactored.

Our approach would have been to have a MESSAGE listener in the message module which would trigger a new UA event for incoming messages. However, this would require a lot of refactoring in the modules and additional (de)serialization logic of the message. Further, the `message_send` method should probably be moved to `ua.c` as `ua_message_send` (similar to `ua_options_send`).

We made this current approach flexible because we think it can be useful to extend it to all incoming out-of-dialog requests in the future to provide more security in BareSIP.

> What if the transport changes from TLS to TCP or vice versa ?

We will remove the `tls` option and simply make the option yes or no.

Overall, we believe the large refactoring effort does not provide major benefits while it comes with the risk of changing behavior or causing regressions and it also changes the public API. We are open to feedback and suggestions (including changes to the `uas_allowed`/`uas_req` approach).

alfredh:
next release is scheduled for Wed 31. Jan -- hence we are in the freeze window...

We can merge the PRs after the release is out ...

docs/examples/accounts
alfredh, line 44: the word `UAS` is coming from RFC 3261 and is a technical term that should not be
exposed to the enduser.

Perhaps a better term is "incoming request" or "inbound request" ...?

maximilianfridrich: Works for me. Would you prefer it if I renamed it for all internal cases as well? E.g. `account_set_incoming_req_mode` instead of `account_set_uas_req_mode`, `UAS_REQ_MODE_OFF` ->`INCOMING_REQ_MODE_OFF`.

cspiel1: Should we also change the current `uas_user` and `uas_pass` ?
```
uas_req  --> inreq_allowed = {yes, no}
uas_user --> inreq_user
uas_pass --> inreq_pass
```
We already removed `inreq_allowed=tls` and will replace this use-case with `sip_transports=tls`.

maximilianfridrich: We would like to propose to change the current behavior to not accept incoming MESSAGE requests per default.

Our reasoning is that we see this as non-RFC compliant. According to [RFC 3428 section 11](https://www.rfc-editor.org/rfc/rfc3428#section-11):
```
   This implies that MESSAGE requests have a greater need
   for security than most other SIP requests.  In particular, UAs that
   support the MESSAGE request MUST implement end-to-end authentication,
   body integrity, and body confidentiality mechanisms.
```
Some form of authentication of SIP MESSAGEs is a **MUST**, hence we don't think it should be enabled per default. If it is enabled, the user is responsible for configuring BareSIP to authenticate the SIP MESSAGEs, e.g. by setting `uas_user`/`uas_pass` or by setting TLS as the only transport protocol and enabling TLS client verification.

Let me know if this is okay. If not, we will change the default to `yes`.

cspiel1:
@alfredh do you have any further suggestions? From our side this is ready for the merge.

alfredh:
please rebase on top of main HEAD ...

Looks okay for merging ...

maximilianfridrich:
@alfredh Thank you! Rebase is done.

juha-h:
In my opinion it was bad idea to, by default, break years long behavior.  Also, it should be possible for user to choose based on contacts, where to accept messages from.

juha-h:
When SIP proxy is used, it can verify the sender of the message.  Then it is enough that the user can list, from which other users to accept messages from.  That can be done either at the SIP proxy or UAS.  There is no need for uas_user/uas_pass.

juha-h:
Also, the response code `488 Not Acceptable Here` is not correct. It should be `403 Forbidden`.

juha-h:
Does inreq_allow  apply to INVITE requests?  If no, it is not logical.

This change is far too destructive, when the default is no.

juha-h:
My Android app users will not anymore get MESSAGEs unless they explicitly allow them by changing their account settings.  It is far too much to ask them to do.  Please change default to 'yes'.  Also, in my opinion this kind of thing should be application specific, not in baresip core.

PR #2888, opened 2024-01-25
transp,tls: add TLS client verification

maximilianfridrich:
config,uag: add sip_verify_client for TLS connections

Per default, client certificates are neither requested nor verified.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -12,6 +12,7 @@ sip_capath		/etc/ssl/certs
 #sip_transports		udp,tcp,tls,ws,wss
 #sip_trans_def		udp
 sip_verify_server	yes
+#sip_verify_client	no
 sip_tos			160 # See TOS fields!
 
 ## TOS fields ##
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -348,6 +348,7 @@ struct config_sip {
 	uint32_t transports;    /**< Supported transports mask      */
 	enum sip_transp transp; /**< Default outgoing SIP transport protocol */
 	bool verify_server;     /**< Enable SIP TLS verify server   */
+	bool verify_client;     /**< Enable SIP TLS verify client   */
 	uint8_t tos;            /**< Type-of-Service for SIP        */
 };
 
--- a/src/config.c
+++ b/src/config.c
@@ -36,6 +36,7 @@ static struct config core_config = {
 		0,
 		SIP_TRANSP_UDP,
 		false,
+		false,
 		0xa0,
 	},
 
@@ -371,6 +372,9 @@ int config_parse_conf(struct config *cfg, const struct conf *conf)
 	(void)conf_get_bool(conf, "sip_verify_server",
 			&cfg->sip.verify_server);
 
+	(void)conf_get_bool(conf, "sip_verify_client",
+			&cfg->sip.verify_client);
+
 	if (!conf_get(conf, "sip_trans_def", &tr))
 		cfg->sip.transp = sip_transp_decode(&tr);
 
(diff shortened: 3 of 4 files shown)


juha-h:
Is this for the case when baresip UA is not using a SIP proxy/registrar?  If baresip UA is using a SIP proxy/registrar, then it would suffice that it only accepts incoming requests from its SIP proxy.

maximilianfridrich:
@juha-h Exactly, it is mostly meant for peer-to-peer cases.

docs/examples/config
```diff
@@ -11,6 +11,7 @@ sip_cafile		/etc/ssl/certs/ca-certificates.crt
 #sip_transports		udp,tcp,tls,ws,wss
 #sip_trans_def		udp
 sip_verify_server	yes
+#sip_verify_client	no
```
alfredh, line 14: is it an option to change this to "sip_verify_peer" that can be common for client/server ?

does the instance know if it is a client or server ?

maximilianfridrich: I think it would make sense to have both cases separately.

- `sip_verify_server`: If BareSIP connects to a peer, it checks the peer's (server) certificate and validates the hostname it is connecting to with the CN/SAN of the certificate presented.
- `sip_verify_client` If someone wants to connect to BareSIP, BareSIP sends a CertificateRequest to the peer and then verifies the peer's (client) certificate.

Client certificate verification is much less common, and most of the time only server verification is done. I don't think we should suddenly change this behavior and merge the two options. I could see users wanting one of the verifications done but not both.

juha-h:
I would like to see option sip_allow_clients and it should be per
account.  In case when outbound is defined, no clients should be allowed
for that account.

maximilianfridrich:
@juha-h Sounds sensible. I think it is a different feature that should be implemented in a separate PR. In this PR we introduce an exact equivalent for `sip_verify_server` to change how BareSIP acts when it receives a TLS ClientHello (instead of sending one).

cspiel1:
> I would like to see option sip_allow_clients and it should be per account. In case when outbound is defined, no clients should be allowed for that account.

TLS Client has not much to do with a SIP Client.

juha-h:
Christian Spielberger writes:

> > I would like to see option sip_allow_clients and it should be per
> > account. In case when outbound is defined, no clients should be
> > allowed for that account. 
> 
> TLS Client has not much to do with a SIP Client.

When SIP request comes in to a given SIP client (account), the account
could require that it must come over (verified) TLS connection.

maximilianfridrich:
@juha-h Understood. This feature makes sense (it would most likely require post-handshake authentication). However, I see it as a distinct feature - with this PR we want an exact equivalent for `sip_verify_server` to change how BareSIP acts when it receives a TLS ClientHello. If you want your feature, you are welcome to open a PR for it.

This new option is turned off by default, so it won't change the current BareSIP behavior for anyone.

cspiel1:
The Fedora build fails. It guess that the wrong `re` branch is cloned. @sreimers please could you have a look?
```
Run pr_title=$(gh pr view $PR_NUMBER -q .title --json title || echo 'no_pull_request_found')
failed to run git: fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

Cloning into 're'...
```

cspiel1:
The rebase didn't help. @robert-scheck please could you have a look?
See: https://github.com/baresip/baresip/pull/2888#issuecomment-1935887799

sreimers:
Please try to rebase. I commited a fix: https://github.com/baresip/baresip/pull/2912

robert-scheck:
> The rebase didn't help. @robert-scheck please could you have a look? See: [#2888 (comment)](https://github.com/baresip/baresip/pull/2888#issuecomment-1935887799)

I used the GitHub action "sreimers/pr-dependency-action", because the regular CI was doing so (in my very first tests, I used a bare `git clone …`, but the action seems to honor different branches?). Running `git config --global --add safe.directory "${GITHUB_WORKSPACE}"` before the action might help for the error message (not tested).

maximilianfridrich:
That worked, thank you!

alfredh:
I think this PR can be merged now ?

PR #2905, opened 2024-02-05
config: support distribution-specific/default CA paths

robert-scheck:
Especially Linux distributions support, additionally to the already existing CA file, also CA paths, such as `/etc/ssl/certs` (Debian, openSUSE, SLES), `/etc/pki/tls/certs` (Fedora, CentOS/RHEL) or `/system/etc/security/cacerts` (Android).

This makes a default CA path configurable for downstreams during build-time (and provides a proper default aligned with CA file).

PR #2913, opened 2024-02-10
config: add default CA path for Android

robert-scheck:
Thanks to @juha-h in https://github.com/baresip/baresip/pull/2905#issuecomment-1937099354, we can add the default CA path for Android, too.


juha-h:
Looks good to me.

robert-scheck:
> Looks good to me.

No, sorry, I forgot one part; added with commit 8afaa3f now.

PR #2917, opened 2024-02-12
aufile: avoid race condition in case of fast destruction

cspiel1:
In case of very short live time of aufile the thread got started after the
destructor was called.

Also the `started` flag is only read/written from main thread and doesn't need
to be atomic.


cspiel1:
TODO: Found another problem in `main` module aufile. There is a wrong samplerate used for playback of `ring.wav`
```
/usr/local/share/baresip/ring.wav: RIFF (little-endian) data, WAVE audio, ITU G.711 mu-law, mono 8000 Hz
```
It is played much too fast. I have to verify that this PR has nothing to do with it. And then create a second PR.

PR #2925, opened 2024-02-16
account: read catchall flag from accounts file

cspiel1:


src/account.c
```diff
@@ -1110,6 +1111,26 @@ int account_set_rtcp_mux(struct account *acc, bool value)
...
+/**
+ * Sets catchall flag for the User-Agent account. A catch all account catches
+ * all inbound SIP requests
+ *
+ * @param acc      User-Agent account
+ * @param value    true or false
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int account_set_catchall(struct account *acc, bool value)
```
alfredh, line 1123: Perhaps it is better to return `void` here ?

cspiel1: Note, that there are many similar functions that return EINVAL only for acc==NULL.

PR #2936, opened 2024-03-01
vp8/encode: optimizations and target_bitrate fix

sreimers:


juha-h:
Yesterday I build my Android baresip app with the latest vpx libs.  When I made tests against Debian baresip, I noticed that with vp9 calls worked fine, but not with vp8. On Debian, no video window shows up and console reports lots of these:
```
Set video encoder: VP8  (1000000 bit/s, 15.00 fps)
Set video decoder: VP8 
v4l2: /dev/video0: found valid V4L2 device (960 x 540) pixfmt=YUYV
v4l2: Current input: "Camera 1"
v4l2: current framerate is 15.00 fps
video tx pipeline:       v4l2 ---> selfview_pip ---> snapshot ---> VP8
video rx pipeline:        x11 <--- selfview_pip <--- snapshot <--- VP8
audio: Set audio decoder: opus 16000Hz 1ch
stream: Enable RTP timeout (60000 milliseconds)
stream: Enable RTP timeout (60000 milliseconds)
test@test.tutpro.com: Call established: sip:bar@test.tutpro.com
jbuf: drop 1 old frame seq=3636bit/s) efps=0.0/0.0    
jbuf: drop 1 old frame seq=3637
jbuf: drop 1 old frame seq=3638
jbuf: drop 1 old frame seq=3639
jbuf: drop 1 old frame seq=3640
jbuf: drop 1 old frame seq=3641
...
```
jbuf settings were the same in both tests:
```
video_jitter_buffer_type adaptive
video_jitter_buffer_delay 1-50
jitter_buffer_wish	6
```
Could this issue have something to do with this PR?

sreimers:
Looks like the max. packets within jitter buffer is to small. Try to increase:

`video_jitter_buffer_delay 1-200`

juha-h:
Sebastian Reimers writes:

> Looks like the max. packets within jitter buffer is to small. Try to increase:
> 
> `video_jitter_buffer_delay 1-200`

That helped, but wondering why only VP8 has the problem.

sreimers:
> That helped, but wondering why only VP8 has the problem.

This can be related to the wrong bitrate calculation.

juha-h:
I tried again after this was merged and now  `jbuf: drop 1 old frame` messages are gone. Thanks.

Issue #2939, opened 2024-03-03
answered call has video even when not asked for


juha-h:
This kind of call answer API sequence has earlier (don't remember how long ago) produced in my Android app a non-video call:
```
call_set_media_direction(call, Api.SDP_SENDRECV, Api.SDP_INACTIVE)
sdp_media_set_disabled(stream_sdpmedia(video_strm(call_video(call)), true);
ua_answer(ua, call, Api.VIDMODE_ON)
```
Now the call starts with 2-way video.  The log is below.  Why does call modify enable video even when `check_rtp: not checking "video" RTP (dir=inactive)`?  What has changed that broke the earlier behavior where the API calls shown in above were enough to have an audio-only call established?
```
03-03 16:27:14.075  3152  3223 D Baresip+ Lib: ua: sipsess connect via TLS 192.168.243.45:5061 --> 192.168.243.151:33940
03-03 16:27:14.076  3152  3223 D Baresip+ Lib: bar@test.tutpro.com: selected for bar-0xb40000743871b510
03-03 16:27:14.076  3152  3223 D Baresip+ Lib: ua: using connection-address 192.168.243.45 of SDP offer
03-03 16:27:14.085  3152  3223 D Baresip+ Lib: call: alloc with params laddr=192.168.243.151, af=AF_INET, use_rtp=1
03-03 16:27:14.086  3152  3223 D Baresip+ Lib: call: use_video=1
03-03 16:27:14.094  3152  3222 D Baresip+ Lib: ua event (CALL_INCOMING) sip:test@test.tutpro.com
03-03 16:27:16.837  3152  3152 D Baresip+: AoR sip:bar@test.tutpro.com answering call from test
03-03 16:27:16.839  3152  3152 D Baresip+ Lib: call set audio/video media direction of call -5476376646239205360 to 3/0
03-03 16:27:16.851  3152  3152 D Baresip+: Received onStartCommand action Call Answer
03-03 16:27:16.880  3152  3152 D Baresip+ Lib: answering ua/call -5476376647719340784/-5476376646239205360 with video mode 1
03-03 16:27:16.881  3152  3152 D Baresip+ Lib: ua event (CALL_LOCAL_SDP) answer
03-03 16:27:16.881  3152  3223 D Baresip+ Lib: call: answering call on line 1 from sip:test@test.tutpro.com with 200
03-03 16:27:16.881  3152  3223 D Baresip+ Lib: stream: update 'audio'
[...]

PR #2940, opened 2024-03-04
call: set estdir in call_set_media_direction

maximilianfridrich:
This reverts the behavior of `call_set_media_direction` to how it used to be. Further, re-introduce `call_set_mdir` which only sets the current media direction.

The change in behavior was introduced in #2818 in order to simplify and clarify the setting of media directions.

Resolves: #2939.


juha-h:
Could this be merged now so that I can get my app working again?

PR #2943, opened 2024-03-05
audio: cleanup start/stop redundancy part 1

cspiel1:
- call: use sdp_media_rformat() of re


cspiel1:
A static `sdp_media_rcodec()` in baresip makes sense to exclude telephone-events.

PR #2945, opened 2024-03-05
call,audio: cleanup audio start/stop redundancy

cspiel1:
mentioned here: https://github.com/baresip/baresip/pull/2941

- Similar to `video_update()`
  - Replace `start_audio()` and `audio_start()` by `audio_update()`.
  - start/stop NAT pinhole for recv-only audio
  - start/stop audio rx/tx for re-INVITE, 183 Session Progress, SIP UPDATE which respects media direction
- Remove expendable `stop_aur()`
- start audio source and player at unique point

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1383,7 +1383,7 @@ struct stream *audio_strm(const struct audio *au);
 uint64_t audio_jb_current_value(const struct audio *au);
 int  audio_set_bitrate(struct audio *au, uint32_t bitrate);
 bool audio_rxaubuf_started(const struct audio *au);
-int  audio_start(struct audio *a);
+int  audio_update(struct audio *a);
 int  audio_start_source(struct audio *a, struct list *ausrcl,
 			struct list *aufiltl);
 void audio_stop(struct audio *a);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1384,6 +1384,7 @@ uint64_t audio_jb_current_value(const struct audio *au);
 int  audio_set_bitrate(struct audio *au, uint32_t bitrate);
 bool audio_rxaubuf_started(const struct audio *au);
 int  audio_update(struct audio *a);
+int  audio_start(struct audio *a);
 int  audio_start_source(struct audio *a, struct list *ausrcl,
 			struct list *aufiltl);
 void audio_stop(struct audio *a);
--- a/src/audio.c
+++ b/src/audio.c
@@ -194,16 +194,6 @@ static void stop_tx(struct autx *tx, struct audio *a)
 }
 
 
-static void stop_aur(struct audio_recv *aur)
-{
-	if (!aur)
-		return;
-
-	/* audio player must be stopped first */
-	aurecv_stop(aur);
-}
-
-
 static void audio_destructor(void *arg)
 {
 	struct audio *a = arg;
@@ -212,7 +202,7 @@ static void audio_destructor(void *arg)
 
 	stream_enable(a->strm, false);
 	stop_tx(&a->tx, a);
-	stop_aur(a->aur);
+	aurecv_stop(a->aur);
 
 	mem_deref(a->tx.enc);
 	mem_deref(a->tx.aubuf);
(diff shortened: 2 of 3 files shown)


include/baresip.h
```diff
@@ -1381,7 +1381,7 @@ struct stream *audio_strm(const struct audio *au);
 uint64_t audio_jb_current_value(const struct audio *au);
 int  audio_set_bitrate(struct audio *au, uint32_t bitrate);
 bool audio_rxaubuf_started(const struct audio *au);
-int  audio_start(struct audio *a);
+int  audio_update(struct audio *a);
```
alfredh, line 1384: this will break the API. I think we should try to avoid breaking the API in the future ...

Perhaps keep audio_start() and add audio_update() ?

src/audio.c
```diff
@@ -1338,17 +1364,15 @@ int audio_encoder_set(struct audio *a, const struct aucodec *ac,
 	if (ac->ptime)
 		tx->ptime = ac->ptime;
 
-	if (!tx->ausrc) {
-		err |= audio_start(a);
-	}
-
```
alfredh, line 1344: Is this a good idea ?

src/video.c
```diff
@@ -1275,9 +1275,8 @@ int video_update(struct video *v, const char *peer)
 
 	if (!sc) {
 		info("video: video stream is disabled..\n");
-		video_stop_source(v);
-		video_stop_display(v);
-		return err;
+		video_stop(v);
+		return 0;
```
alfredh, line 1279: perhaps this can be split into two PRs, one for audio and one for video ?

juha-h:
Alfred E. Heggestad writes:

> > @@ -1381,7 +1381,7 @@ struct stream *audio_strm(const struct audio *au);
>  uint64_t audio_jb_current_value(const struct audio *au);
>  int  audio_set_bitrate(struct audio *au, uint32_t bitrate);
>  bool audio_rxaubuf_started(const struct audio *au);
> -int  audio_start(struct audio *a);
> +int  audio_update(struct audio *a);
> 
> this will break the API. I think we should try to avoid breaking the
> API in the future ... 
> 
> Perhaps keep audio_start() and add audio_update() ?

Yes, I currently have:

JNIEXPORT void JNICALL Java_com_tutpro_baresip_Api_call_1start_1audio(
        JNIEnv *env, jobject thiz, jlong call)
{
    LOGD("starting audio of call %ld\n", (long)call);
    re_thread_enter();
    struct audio *a = call_audio((struct call *)call);
    if (!audio_started(a))
        audio_start(a);
    re_thread_leave();
}

and would have no idea how it would need to be changed if audio_start()
disappears.

include/baresip.h
cspiel1: The reason for this PR was to drop all unnecessary redundancy about starting/updating the audio streams.
Also I want to make video and audio API similar.
I'll try to find a way to keep the backward compatibility with minimal redundancy.

src/audio.c
cspiel1: yes, I suggest to define a clear hierarchy
- `audio_update` / `mediatrack_start_audio`
  - `aufilt_setup`
  - `audio_decoder_set`
  - `audio_encoder_set`
  - `start_source`

And each of this functions behave according to their name

cspiel1: For the decoding stream it is like that:
- `audio_update` / `media_start_audio`
  - `aufilt_setup`
  - `audio_decoder_set`
    - `aurecv_start_player`

Because the decoder can be changed during a call. This makes it sometimes necessary to restart the audio player when `audio_decoder_set` is called.

src/video.c
cspiel1: - https://github.com/baresip/baresip/pull/2982

cspiel1:
- Rebased and squashed the video changes with the revert.
- Tested: https://github.com/baresip/baresip/pull/2945#issuecomment-1991077344

alfredh:
IMO this patch contains too many unrelated changes, and is a bit large.

It would be nice to split it up into logical parts (one PR per logical part).

cspiel1:
Will split up this PR.

PR #2949, opened 2024-03-06
stream: avoid sanitizer warnings for strm->tx

cspiel1:


sreimers:
Maybe we should postpone after release. RX Thread mode is experimental and this is a big change before release.

cspiel1:
I think it has not much or nothing to do with the RX thread. Also nothing with the TX send thread.

The TX data is driven by the Microphone thread (ausrc thread). And SIP is handled by the main thread. Should we merge it now, or later? The concurrency failure is very old.

sreimers:
> I think it has not much or nothing to do with the RX thread. Also nothing with the TX send thread.

Ah I see, yes let's postpone after release.

src/stream.c
```diff
@@ -1502,7 +1525,14 @@ enum media_type stream_type(const struct stream *strm)
 
 int stream_pt_enc(const struct stream *strm)
 {
-	return strm ? strm->tx.pt_enc : -1;
+	int pt;
+
+
+	mtx_lock(strm->tx.lock);
```
sreimers, line 1531: `strm` null check is missing here, possible null pointer deref.

src/stream.c
```diff
@@ -1595,7 +1626,6 @@ int stream_debug(struct re_printf *pf, const struct stream *s)
 	err |= re_hprintf(pf, " tx.enabled: %s\n",
 			  re_atomic_rlx(&s->tx.enabled) ? "yes" : "no");
 	err |= rtprecv_debug(pf, s->rx);
```
sreimers, line 1628: Should we reorder `rtprecv_debug` and `rtp_debug`? So we can unlock earlier? Or is `tx.lock` needed for `rtprecv_debug` too?

sreimers:
Please rebase against main, this should fix:
```
FATAL: ThreadSanitizer: unexpected memory mapping 0x59f14d472000-0x59f14d4a3000
```

PR #2963, opened 2024-03-14
account: use correct format %zu for printing outbound

maximilianfridrich:
This caused a `SIGABRT` when an outbound proxy is set and account debug information is printed (because of the `RE_VA_ARG` usage in `print.c`).

### DIFF
--- a/src/account.c
+++ b/src/account.c
@@ -2005,7 +2005,7 @@ int account_debug(struct re_printf *pf, const struct account *acc)
 			  acc->pinhole ? "yes" : "no");
 	for (i=0; i<RE_ARRAY_SIZE(acc->outboundv); i++) {
 		if (acc->outboundv[i]) {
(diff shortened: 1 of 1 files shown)


cspiel1:
Looks correct. Could find the reason for the SEGV in sanitizer workflow.

src/account.c
```diff
@@ -2005,7 +2005,7 @@ int account_debug(struct re_printf *pf, const struct account *acc)
 			  acc->pinhole ? "yes" : "no");
 	for (i=0; i<RE_ARRAY_SIZE(acc->outboundv); i++) {
```
sreimers, line 2006:
```suggestion
	for (size_t i=0; i<RE_ARRAY_SIZE(acc->outboundv); i++) {
```

sreimers:
Please rebase against main, this should fix:
```
FATAL: ThreadSanitizer: unexpected memory mapping 0x59f14d472000-0x59f14d4a3000
```

PR #2964, opened 2024-03-15
dialog,account: add module for subscribe to dialog events

vlitvinov:
Added new module that give a possibility subscribe to the Event:dialog.
To do this you need to create a contact with 'dialog=p2p' parameter (as for presence) and enable the dialog subscriptions on the account (it disabled by default).
After the registration module will subscribe to all dialog contacts from this UA. Body of received Notify will be printed using info() function.


modules/dialog/dialog.c
```diff
@@ -0,0 +1,272 @@
...
+
+		if (0 == msg_param_decode(&addr->params, "dialog", &val) &&
+		    0 == pl_strcasecmp(&val, "p2p")) {
+
+			if (dialog_find(ua, c))
+				continue;
+
+			dialog = mem_zalloc(sizeof(*dialog), destructor);
+			if (!dialog)
+				err |= ENOMEM;
```
sreimers, line 164:
```suggestion
			if (!dialog) {
				err = ENOMEM;
				return err;
			}
```

alfredh:
I would prefer if new modules are added to baresip-apps as a first step,
so that people can play around with it. If the new module becomes very useful
and popular, we can merge it to baresip.

alfredh:
there are some ccheck warnings:

```
./modules/dialog/dialog.c:97: line is too wide (85 - max 79)
./modules/dialog/dialog.c:128: line is too wide (82 - max 79)
./modules/dialog/dialog.c:180: line is too wide (119 - max 79)
Statistics:
~~~~~~~~~~~
Number of files processed:     c: 246  cpp: 12  h: 53  mk: 0  m4: 0  py: 1  m: 1  s: 0  java: 0  php: 0 
Number of lines with errors:   3
```

vlitvinov:
Hello. Thanks for suggestions. Branch has been updated.
Unfortunately it's not possible to add this module to baresip-app project because it usage requires changes in account module.

docs/examples/accounts
```diff
@@ -26,6 +26,7 @@
 #    ;medianat={stun,turn,ice}
 #    ;natpinhole={yes, no}  # default: no
 #    ;mwi=no
+#    ;dialog={yes, no} # default: no
```
alfredh, line 29: Is this needed ? If the module is loaded, then the feature is enabled?

alfredh:
I suggest you start with "baresip-apps" and see if there are other people interested ...

Issue #1089, opened 2020-09-19
avfilter module is missing explicit license


jonassmedegaard:
Hi,

Newly added avfilter module has copyright Mikhail Kurkov but no license statement.

I assume it has same license as the project in general (BSD-3-Clause) but would be better if explicitly stated in each copyright-protected file.

alfredh:
adding @mkurkov 

the license is defined in docs/COPYING (BSD license).

all files and the whole project has the same license.

should we add the BSD license text to all .c files ?

jonassmedegaard:
Well, not all code is BSD-3-Clause licensed: test/ccheck.py is GPL-2.

I think it is best if both copyright and license is stated in each file: That way the information is less likely to be lost if code is borrowed by other projects (like apparently you did with modules/ctrl_tcp/netstring).

As I understand it (I am not a lawyer), *nothing* is required for legal purposes - it is (nowadays, since [the US signed the Berne Convention in 1989](https://en.wikipedia.org/wiki/Copyright#International_copyright_treaties)) all done purely as a curtesy for your users, to clearly inform them of the freedoms that you have granted them.

alfredh:
we could change the license for ccheck.py to BSD also, it is just a small test tool.

adding the BSD license text to all .c file is IMHO not really needed.
it will be a lot of redundant text in all .c files.

jonassmedegaard:
Quoting Alfred E. Heggestad (2020-09-28 14:51:33)
> we could change the license for ccheck.py to BSD also, it is just a small test tool.
> 
> adding the BSD license text to all .c file is IMHO not really needed.
> it will be a lot of redundant text in all .c files.

there is _already_ a license notice in all code files _except_ the few I filed as an issue here.

I am not talking about expanding to write _more_ everywhere.

jonassmedegaard:
Quoting Jonas Smedegaard (2020-09-28 15:58:27)
> Quoting Alfred E. Heggestad (2020-09-28 14:51:33)
> > we could change the license for ccheck.py to BSD also, it is just a small test tool.
> > 
> > adding the BSD license text to all .c file is IMHO not really needed.
> > it will be a lot of redundant text in all .c files.
> 
> there is _already_ a license notice in all code files _except_ the few I filed as an issue here.
> 
> I am not talking about expanding to write _more_ everywhere.

oh, my mistake: Indeed most files lack license.

I recommend to add just a minimum license hint.

...and I am not alone - see e.g. here: https://reuse.software/

[...]

PR #2975, opened 2024-03-20
srtp: allow rekeying of running streams

cHuberCoffee:
It's allowed to re-key and change the cipher-suite of streams at any point in time. While the new information got processed, the decryption objects never got updated.
For completeness the srtp got a command to re-key outgoing streams of a given call, selected by call id.
The new testcase depends on an re PR [#1089](https://github.com/baresip/re/pull/1089) for checking the streaming keys.

### DIFF
--- a/modules/srtp/srtp.c
+++ b/modules/srtp/srtp.c
@@ -274,21 +274,46 @@ static int start_crypto(struct menc_st *st, const struct pl *key_info)
 {
 	size_t olen, len;
 	char buf[64] = "";
+	uint8_t *new_key = NULL;
 	int err;
 
 	len = get_master_keylen(resolve_suite(st->crypto_suite));
 
 	/* key-info is BASE64 encoded */
+	new_key = mem_zalloc(len, NULL);
+	if (!new_key)
+		return ENOMEM;
 
-	olen = sizeof(st->key_rx);
-	err = base64_decode(key_info->p, key_info->l, st->key_rx, &olen);
-	if (err)
+	err = base64_decode(key_info->p, key_info->l, new_key, &olen);
+	if (err) {
+		mem_deref(new_key);
 		return err;
+	}
 
 	if (len != olen) {
-		warning("srtp: %s: srtp keylen is %u (should be %zu)\n",
-			st->crypto_suite, olen, len);
+		warning("srtp: %s: %s: srtp keylen is %u (should be %zu)\n",
+			stream_name(st->strm), st->crypto_suite, olen, len);
+		mem_deref(new_key);
+		return err;
 	}
+
+	if (olen > sizeof(st->key_rx)) {
+		warning("srtp: %s: received key exceeds max key length\n",
+			stream_name(st->strm));
+		mem_deref(new_key);
+		return ERANGE;
+	}
+
+	/* receiving key-info changed -> reset srtp_rx */
+	if (st->srtp_rx && memcmp(st->key_rx, new_key,
+		sizeof(st->key_rx) > olen ? olen : sizeof(st->key_rx))) {
+		info("srtp: %s: re-keying in progress\n",
+			stream_name(st->strm));
+		st->srtp_rx = mem_deref(st->srtp_rx);
+	}
+
+	memcpy(st->key_rx, new_key, olen);
+	new_key = mem_deref(new_key);
 
 	err = start_srtp(st, st->crypto_suite);
 	if (err)
@@ -327,6 +352,13 @@ static bool sdp_attr_handler(const char *name, const char *value, void *arg)
 
 	if (!cryptosuite_issupported(&c.suite))
 		return false;
+
+	/* receiving crypto-suite changed -> reset srtp_rx */
+	if (st->srtp_rx && pl_strcmp(&c.suite, st->crypto_suite)) {
+		info ("srtp (%s-rx): cipher suite changed from %s to %r\n",
+			stream_name(st->strm), st->crypto_suite, &c.suite);
+		st->srtp_rx = mem_deref(st->srtp_rx);
+	}
 
 	st->crypto_suite = mem_deref(st->crypto_suite);
 	pl_strdup(&st->crypto_suite, &c.suite);
--- a/modules/srtp/srtp.c
+++ b/modules/srtp/srtp.c
@@ -307,7 +307,7 @@ static int start_crypto(struct menc_st *st, const struct pl *key_info)
 	}
 
 	/* receiving key-info changed -> reset srtp_rx */
-	if (st->srtp_rx && memcmp(st->key_rx, new_key,
+	if (st->srtp_rx && mem_seccmp(st->key_rx, new_key,
 		sizeof(st->key_rx) > olen ? olen : sizeof(st->key_rx))) {
 		info("srtp: %s: re-keying in progress\n",
 			stream_name(st->strm));
@@ -315,6 +315,7 @@ static int start_crypto(struct menc_st *st, const struct pl *key_info)
 	}
 
 	memcpy(st->key_rx, new_key, olen);
+	mem_secclean(new_key, olen);
 	new_key = mem_deref(new_key);
 
 	err = start_srtp(st, st->crypto_suite);
@@ -492,9 +493,9 @@ static int media_alloc(struct menc_media **stp, struct menc_sess *sess,
 }
 
 
-static int cmd_tx_rekey(struct re_printf *pf, void *arg) {
+static int cmd_tx_rekey(struct re_printf *pf, void *arg)
+{
 	const struct cmd_arg *carg = arg;
-	struct call *selcall = NULL;
 	int err = 0;
 
 	if (!str_isset(carg->prm)) {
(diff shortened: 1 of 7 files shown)


modules/srtp/srtp.c
```diff
@@ -272,24 +273,50 @@ static int sdp_enc(struct menc_st *st, struct sdp_media *m,
...
+	/* receiving key-info changed -> reset srtp_rx */
+	if (st->srtp_rx && memcmp(st->key_rx, new_key,
+		sizeof(st->key_rx) > olen ? olen : sizeof(st->key_rx))) {
+		info("srtp: %s: re-keying in progress\n",
+			stream_name(st->strm));
+		st->srtp_rx = mem_deref(st->srtp_rx);
 	}
 
+	memcpy(st->key_rx, new_key, olen);
+	new_key = mem_deref(new_key);
```
alfredh, line 318: crypto key is privacy sensitive, and should be securely erased. See mem_secclean()

modules/srtp/srtp.c
```diff
@@ -458,6 +492,49 @@ static int media_alloc(struct menc_media **stp, struct menc_sess *sess,
 }
 
 
+static int cmd_tx_rekey(struct re_printf *pf, void *arg) {
```
alfredh, line 495: nit: move curly brackets to next line ...

modules/srtp/srtp.c
```diff
@@ -458,6 +492,49 @@ static int media_alloc(struct menc_media **stp, struct menc_sess *sess,
...
+	const struct cmd_arg *carg = arg;
+	struct call *selcall = NULL;
```
alfredh, line 497: move to line 505

modules/srtp/srtp.c
```diff
@@ -458,6 +492,49 @@ static int media_alloc(struct menc_media **stp, struct menc_sess *sess,
...
+	if (!selcall) {
+		re_hprintf(pf, "srtprekey: call id \"%s\" not found\n",
+			carg->prm);
+		return EINVAL;
+	}
+
+	re_hprintf(pf, "srtprekey: rekey srtp transmission keys of"
+		" call \"%s\"\n", carg->prm);
+
+	struct stream *astrm = audio_strm(call_audio(selcall));
```
alfredh, line 515: Perhaps better to use the list of streams ?

modules/srtp/srtp.c
```diff
@@ -458,6 +492,49 @@ static int media_alloc(struct menc_media **stp, struct menc_sess *sess,
...
+	if (err) {
+		re_hprintf(pf, "srtprekey: call update failed %m\n", err);
+	}
+
+	return err;
+}
+
+
+static const struct cmd cmdv[] = {
+{"srtprekey", 0, CMD_PRM, "Change outgoing streaming keys", cmd_tx_rekey},
```
alfredh, line 534: Is this command useful for endusers ?

src/stream.c
```diff
@@ -955,6 +955,15 @@ int stream_update(struct stream *s)
 }
 
 
+void stream_remove_menc_media(struct stream *strm)
+{
+	if (!strm)
+		return;
+
+	strm->mes = mem_deref(strm->mes);
```
alfredh, line 963: should also set strm->menc and strm->mencs to NULL ?
otherwise it will be left in a half undefined state.

test/call.c
```diff
@@ -3066,3 +3066,172 @@ int test_call_hold_resume(void)
...
+		NULL, NULL, &key_prms, NULL, NULL);
+	if (err)
+		return false;
+
+	err = re_regex(key_prms.p, key_prms.l, "[^:]+:[^|]+[|]*[^|]*[|]*[^|]*",
+		NULL, &key_info, NULL, NULL, NULL, NULL);
+	if (err)
+		return false;
+
+	return pl_strdup(key, &key_info) ? false : true;
```
alfredh, line 3092: can also return 0==pl_strdup()

test/call.c
```diff
@@ -3066,3 +3066,172 @@ int test_call_hold_resume(void)
...
+
+
+int test_call_srtp_tx_rekey(void)
+{
+	struct fixture fix, *f = &fix;
+	struct cancel_rule *cr = NULL;
+	struct auplay *auplay = NULL;
+
+	char rekeycmd [128];
+	struct mbuf *cmd_resp = mbuf_alloc(128);
```
alfredh, line 3110: check for NULL ?

test/call.c
```diff
@@ -3066,3 +3066,172 @@ int test_call_hold_resume(void)
...
+	sdp_media_rattr_apply(m, "crypto", sdp_crypto_handler, &a_rx_key_new);
+
+	m = stream_sdpmedia(audio_strm(call_audio(ua_call(f->b.ua))));
+	ASSERT_EQ(SDP_SENDRECV, sdp_media_ldir(m));
+	ASSERT_EQ(SDP_SENDRECV, sdp_media_rdir(m));
+	sdp_media_lattr_apply(m, "crypto", sdp_crypto_handler, &b_tx_key_new);
+	sdp_media_rattr_apply(m, "crypto", sdp_crypto_handler, &b_rx_key_new);
+
+	/* transmission key of a must change */
+	ASSERT_TRUE(str_casecmp(a_tx_key, a_tx_key_new));
```
alfredh, line 3199: I think str_casecmp() returns int ?

alfredh:
Thanks, this is great work!

src/stream.c
cHuberCoffee: The `strm->mencs` holds the event callback to `call.c` which gets called after the crypto parameters are replaced already. Also the stream stays secure.
The `strm->menc` can still be in place since the encryption module stays the same.

My understanding is the session stays the same while the crypto parameter changes. So the `strn->mes` changes.
Due to network timing it is possible to drop a few frames in case the receiver gets packets encrypted with the new key but has not processed the new parameters yet. It would lead to a bigger discussion about key handling and key management in the module.

modules/srtp/srtp.c
cHuberCoffee: I need the transmission re-keying for the testcase anyway. I had in mind why not and its easy callable in the testcase.
To be fair its unlikely a user will ever re-key the outgoing stream with this command.

test/call.c
cHuberCoffee: Thats true and wanted here. We want a `!= 0` value which evaluates to true. A unchanged key would evaluate to `0` and the `ASSERT_TRUE` would fail.
We have no direct macro to check inequality of strings.

src/stream.c
alfredh: perhaps add the name "state" to the function name, to be explicit.

Please note that encryption state consist of 1 session state and N media states per media stream.
You should make sure that all states are destroyed atomically.

test/call.c
```diff
@@ -3066,3 +3066,177 @@ int test_call_hold_resume(void)
...
+	err |= module_load(".", "ausine");
+	TEST_ERR(err);
+
+	err = mock_auplay_register(&auplay, baresip_auplayl(),
+		auframe_handler, f);
+	TEST_ERR(err);
+
+	fixture_init_prm(f, ";mediaenc=srtp-mand"
+		";ptime=1;audio_player=mock-auplay,a");
+	mem_deref(f->b.ua);
```
alfredh, line 3133: also set f->b.ua to NULL

test/call.c
alfredh: better to write it like this:

```
ASSERT_TRUE(0 != str_casecmp())
```

alfredh:
Does this also apply to other crypto modules such as ZRTP or DTLS-SRTP ?

I think the new command can be skipped, and the code can be written directly in the testcode.

cHuberCoffee:
For now this only applies to the encryption via the SRTP module. I will include a check in the `stream_remove_menc_media` (in the future `stream_remove_menc_media_state`) to ensure it only works with media encryption via the SRTP module.

I will remove the command and put the code into the testcode.

src/stream.c
```diff
@@ -955,6 +955,31 @@ int stream_update(struct stream *s)
...
+void stream_remove_menc_media_state(struct stream *strm)
+{
+	if (!strm)
+		return;
+
+	if ((str_casecmp(strm->menc->id, "srtp") != 0) &&
+	    (str_casecmp(strm->menc->id, "srtp-mand") != 0) &&
+	    (str_casecmp(strm->menc->id, "srtp-mandf") != 0)) {
+		return;
+	}
```
alfredh, line 977: this goes against the modular design. The core should work with any module.
The core should not have any knowledge about module names or how they work.

test/call.c
```diff
@@ -3066,3 +3066,163 @@ int test_call_hold_resume(void)
 out:
 	return err;
 }
+
+
+static bool sdp_crypto_handler(const char *name, const char *value, void*arg)
```
alfredh, line 3071: add space after "void" and before star.

src/stream.c
cHuberCoffee: ok i see your point and remove the checks again.

as cleanup i would move it into core.h . So the function is only a private interface function. The testcases would have access to it since they include the core.h. And the function is only used in the corresponding testcase.

at least for now (until testcases are written) i can not say anything about the zrtp & dtls_srtp behaviour.

PR #2976, opened 2024-03-20
sndfile: write correct sample rate to WAV header

cspiel1:
When the update functions are called the sample rate and format can't be
predicted. Thus the header has to be written when the first audio frame reaches
the filter.

Issue mentioned here: https://github.com/baresip/baresip/discussions/2938#discussioncomment-8849657

### DIFF
--- a/modules/sndfile/sndfile.c
+++ b/modules/sndfile/sndfile.c
@@ -17,19 +17,23 @@
  *
  * Example Configuration:
  \verbatim
-  snd_path 					/tmp/
+  snd_path					/tmp/
  \endverbatim
  */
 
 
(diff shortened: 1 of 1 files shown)


alfredh:
The decoder and filter should be set based on the codec negotiated in SDP.

When the first incoming RTP packet arrives, it will check the payload-type and possibly reset the decoder.

So I am not sure if this change is really needed.

cspiel1:
The sample rate can change if e.g. `auplay_srate` is set and `auresamp` module is loaded. It also depends on the order of the filters `sndfile` - `auresamp`, or `auresamp` - `sndfile`.

PR #2979, opened 2024-03-27
ua: ua_event for DnD rejected call

cspiel1:
Adds a UA event for rejected incoming calls due to activated DnD.

The UA event contains
- Contact header: Can be used by the application in order to call back.
- Display name of the From header: Can be used by the application to display the peer of the rejected call.
- From Uri: May also be used to display the peer.

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -704,6 +704,8 @@ void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
 	if (uag_dnd()) {
 		(void)sip_treply(NULL, uag_sip(), msg,
 			480,"Temporarily Unavailable");
+
+		ua_event_rejected(ua, "DnD", msg);
 		return;
 	}
 
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1652,8 +1652,8 @@ void ua_event(struct ua *ua, enum ua_event ev, struct call *call,
 	      const char *fmt, ...);
 void module_event(const char *module, const char *event, struct ua *ua,
 		struct call *call, const char *fmt, ...);
-void ua_event_rejected(struct ua *ua, const char *reason,
-		       const struct sip_msg *msg);
+int ua_event_rejected(struct ua *ua, const char *reason,
+		      const struct sip_msg *msg);
 const char  *uag_event_str(enum ua_event ev);
 
 
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -858,6 +858,7 @@ enum ua_event {
 	UA_EVENT_CALL_REMOTE_SDP,     /**< param: offer or answer */
 	UA_EVENT_CALL_HOLD,           /**< Call put on-hold by peer          */
 	UA_EVENT_CALL_RESUME,         /**< Call resumed by peer              */
+	UA_EVENT_CALL_REJECTED,       /**< Incoming call rejected            */
 	UA_EVENT_REFER,
 	UA_EVENT_MODULE,
 	UA_EVENT_END_OF_FILE,
@@ -1651,6 +1652,8 @@ void ua_event(struct ua *ua, enum ua_event ev, struct call *call,
 	      const char *fmt, ...);
 void module_event(const char *module, const char *event, struct ua *ua,
 		struct call *call, const char *fmt, ...);
+void ua_event_rejected(struct ua *ua, const char *reason,
+		       const struct sip_msg *msg);
 const char  *uag_event_str(enum ua_event ev);
 
 
--- a/src/event.c
+++ b/src/event.c
@@ -71,6 +71,7 @@ static const char *event_class_name(enum ua_event ev)
 	case UA_EVENT_CALL_REMOTE_SDP:
 	case UA_EVENT_CALL_HOLD:
 	case UA_EVENT_CALL_RESUME:
+	case UA_EVENT_CALL_REJECTED:
 		return "call";
 	case UA_EVENT_VU_RX:
 	case UA_EVENT_VU_TX:
@@ -411,6 +412,26 @@ out:
 
 
 /**
+ * Send a call rejected User-Agent event to all UA event handlers
+ *
+ * @param ua     User-Agent object (optional)
+ * @param reason Call object (optional)
+ * @param msg    SIP Message
+ */
+void ua_event_rejected(struct ua *ua, const char *reason,
+			  const struct sip_msg *msg)
+{
+	const struct sip_hdr *hdr;
+	hdr = sip_msg_hdr(msg, SIP_HDR_CONTACT);
+	ua_event(ua, UA_EVENT_CALL_REJECTED, NULL,
+		 "{\"reason\":\"%s\",\"contact\":\"%r\","
+		 "\"display\":\"%r\",\"from\":\"%H\"}",
+		 reason, hdr ? &hdr->val : NULL,
+		 &msg->from.dname, uri_encode, &msg->from.uri);
+}
+
+
+/**
  * Get the name of the User-Agent event
  *
  * @param ev User-Agent event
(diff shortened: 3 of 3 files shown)


cspiel1:
With module ctrl_dbus the tool dbus-monitor shows this DBUS signal:
```
signal time=1712060692.114381 sender=:1.7928 -> destination=(null destination) serial=7 path=/baresip; interface=com.github.Baresip; member=event
   string "call"
   string "CALL_REJECTED"
   string "{"type":"CALL_REJECTED","class":"call","accountaor":"sip:A@localhost","param":"{\"reason\":\"DnD\",\"contact\":\"<sip:od5-0xae0988@192.168.110.182:5060>\",\"display\":\"OD5\",\"from\":\"sip:od5@192.168.110.182\"}"}"
```

cspiel1:
The current solution introduces a new event type `UA_EVENT_CALL_REJECTED` which has the event class "call", but we don't want to allocate a call object only for encoding the UA event. So we pass `NULL` as call pointer and encode the wanted data into the `prm` parameter of the UA event handlers.

If someone has a better idea, please let me know! We think that the Contact header + From header + Display name is useful for rejected calls. At least the Contact header currently is not part of real call UA events.

TODOs:
- Decode+encode the Contact header for security reason.
- I'll try to use the JSON API of re for encoding. Maybe it looks better.

juha-h:
Christian Spielberger writes:

> The current solution introduces a new event type
> `UA_EVENT_CALL_REJECTED` which has the event class "call", but we
> don't want to allocate a call object only for encoding the UA
> event. So we pass `NULL` as call pointer and encode the wanted data
> into the `prm` parameter of the UA event handlers. 
> 
> If someone has a better idea, please let me know! We think that the
> Contact header + From header + Display name is useful for rejected
> calls. At least the Contact header currently is not part of real call
> UA events.

Why don't you implement DnD in the app (menu module)?

cspiel1:
The DnD feature is already in the baresip core. It is checked before a call is allocated and `call_accept()` is called. The 480 "Temporarily Unavailable" is sent before call allocation, like
- 404 "Not Found"
- 486 "Max Calls"
- 420 "Bad Extension"
- 421 "Extension required"

I think we should not change this behavior.

cspiel1:
I tried `odict_alloc`, and `odict_entry_add` and I think that it does make it better. An `odict_entry_add` with type `struct pl` is missing.

Anyway, I add a commit that corrects the escaping of the inner JSON object.

juha-h:
Christian Spielberger writes:

> The DnD feature is already in the baresip core. It is checked before a call is allocated and `call_accept()` is called. The 480 "Temporarily Unavailable" is sent before call allocation, like
> - 404 "Not Found"
> - 486 "Max Calls"
> - 420 "Bad Extension"
> - 421 "Extension required"
> 
> I think we should not change this behavior.

Keep the behavior, but don't use it.  Instead, as I mentioned, implement
it in the app per account. which is a much better choice.

cspiel1:
Maybe this would be possible. I have to think about it. For now I can say, with the solution you propose we get these unwanted behavior if the device is in DnD mode:
- the call is allocated,
- `call_accept()` is called,
- 180 Ringing is sent back,
- the UA_EVENT_CALL_INCOMING event is sent out and these will pass on our devices through
  - intercom.so
  - multicast.so
  - menu.so

The UA_EVENT_CALL_INCOMING will result in unwanted behavior for the user. And on the peer there will be also unwanted behavior.

And if we add a new application module with another UA event handler that is processed first, we still have the problem that 180 Ringing is sent to the peer.

cspiel1:
When visualizing the new UA event with autotest, the JSON looks correct. Thus I guess that unwanted backslashes are printed by the dbus-monitor tool.

```
autotest: [ ua=sip:A@localhost call= ] event: CALL_REJECTED ({"reason":"DnD","contact":"sip:od5-0xae0988@192.168.110.182:5060","display":"OD5","from":"sip:od5@192.168.110.182"})
```

I set this to "Ready for review".

juha-h:
Christian Spielberger writes:

> Maybe this would be possible. I have to think about it. For now I can
> say, with the solution you propose we get these unwanted behavior if
> the device is in DnD mode: 
> - the call is allocated,
> - `call_accept()` is called,
> - 180 Ringing is sent back,
> - the UA_EVENT_CALL_INCOMING event is sent out and these will pass on our devices through

When call comes in, why is 180 Ringing sent back before the app has had
a chance to decide what to do with the call?  In my opinion, the app
should get CALL_INCOMING event and then be able to response with 180
Ringing, 486 Busy Here, ..., or 200 Answering.

juha-h:
Christian Spielberger writes:

> And if we add a new application module with another UA event handler
> that is processed first, we still have the problem that 180 Ringing is
> sent to the peer.

That exactly is the problem.  There should be no automatic 180 Ringing
response.  For example, my app accepts only one simultaneous call.  When
the second call comes in, my app responds with 486 Busy Here.  Now that
I checked with SIP trace, there was unwanted 180 Ringing before the 486:

...
04-02 17:44:41.946  6471  6549 D Baresip Lib: call: alloc with params laddr=10.0.0.2, af=AF_INET, use_rtp=1
04-02 17:44:41.946  6471  6549 D Baresip Lib: call: use_video=0
04-02 17:44:41.953  6471  6549 D Baresip Lib: 17:44:41.953#
04-02 17:44:41.954  6471  6549 D Baresip Lib: TLS 10.0.0.2:60304 -> xxx.xxx.xxx.xxx:yyyy
04-02 17:44:41.954  6471  6549 D Baresip Lib: SIP/2.0 180 Ringing
...
04-02 17:44:41.957  6471  6548 D Baresip Lib: ua event (CALL_INCOMING)
***@***.***
...
04-02 17:44:41.977  6471  6549 D Baresip Lib: call: rejecting incoming call from ***@***.*** (486 Busy Here)
...

There must be some way to prevent the automatic 180 Ringing response.

juha-h:
Other examples include: user does not want to accept calls if caller is not in contacts or if caller is in spammers.  In all these cases, there should not be a 180 Ringing response.

cspiel1:
That makes sense for me. So I put this PR back to draft and I try to make a new one for further discussion.

What about the 183 Session Progress? Baresip sends always
1. 180 Ringing
2. 183 Session Progress

so both in this order.

src/ua.c
```diff
@@ -704,6 +704,8 @@ void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
 	if (uag_dnd()) {
 		(void)sip_treply(NULL, uag_sip(), msg,
 			480,"Temporarily Unavailable");
+
+		ua_event_rejected(ua, "DnD", msg);
```
alfredh, line 708: Can we move this to menu module ?

The less logic in the core, the better ...

juha-h:
Alfred E. Heggestad writes:

> Can we move this to menu module ?

That was my suggestion too.

> The less logic in the core, the better ...

Yes.

juha-h:
When call comes in and user does not want/is not able to receive it, use might want to response with `302 Moved Temporarily` to somewhere, e.g., to voicemail, but `ua_hangup` does not have contact URI argument.

cspiel1:
Currently the state `CALL_STATE_INCOMING` and the event `CALL_EVENT_INCOMING` mean that the call was already answered with 180 "Ringing". There are already several modules (maybe external apps) that depend on this.
- module `menu`
- module `gtk`
- modules in baresip apps

Should we add a state/event `PRE_INCOMING`?

Better wording would be `INCOMING` and `ACCEPTED`, but that would probably break applications.

juha-h:
How about CALL_[STATE|EVENT]_ARRIVED?

cspiel1:
Will be replaced. DnD moved to module menu.

PR #2982, opened 2024-03-29
video: stream enable/disable for re-INVITE/UPDATE

cspiel1:
Use of `stream_enable_tx(false)` when video sending is switched off may avoid concurrency problems with the Video TX thread.
Use of `stream_enable_rx(false)` when video receiving is switched off is a first step for enabling Video RX thread.


cspiel1:
- [ ] Tested early-video on/off during ringing state with SIP UPDATE. 
- [ ] Tested during established state all video direction changes with SIP INVITE.

PR #2986, opened 2024-04-03
ci/build: add and fix baresip-apps build

cspiel1:
Currently baresip-apps does not build.

This PR should add baresip-apps build to the workflows.
A baresip-apps PR with same name will fix the build

edit:
For discussion. Is this a good idea?
In re workflows, baresip is not built.

fix: https://github.com/baresip/baresip-apps/pull/45

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,6 +105,12 @@ jobs:
         repo: https://github.com/baresip/re
         secret: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


cspiel1:
I merged https://github.com/baresip/baresip-apps/pull/45

alfredh:
the dependencies goes like this:

* baresip depends on libre
* baresip-apps depends on baresip

If you add now a dependency from baresip to baresip-apps we will get a circular dependency which is bad.

It would be better if you add the build in the baresip-apps project.

PR #3001, opened 2024-04-18
uag: fix initializer

cspiel1:
part 1 of DnD feature: https://github.com/baresip/baresip/pull/2979

The initializer of uag is not correct.

### DIFF
--- a/src/uag.c
+++ b/src/uag.c
@@ -11,26 +11,7 @@
 
 /* One instance */
 
-static struct uag uag = {
-	NULL,
-	LIST_INIT,
-	NULL,
-	NULL,
-	NULL,
-	NULL,
-	false,
-	0,
-	NULL,
-	NULL,
-	false,
-	false,
-	NULL,
-	NULL,
-#ifdef USE_TLS
-	NULL,
-	NULL
-#endif
-};
+static struct uag uag = {.ual = LIST_INIT};
 
 
 /* This function is called when all SIP transactions are done */


cspiel1:
Problems in  MaxOS build:
- `avcodec_close` is deprecated
- libavformat.dylib was built for newer macOS version (12.0) than being linked (10.12)

@sreimers should I make a PR for the first? For the second I have no idea.

sreimers:
> * libavformat.dylib was built for newer macOS version (12.0) than being linked (10.12)

This can be ignored.

cspiel1:
Here is the fix of the deprecated warning: https://github.com/baresip/baresip/pull/3002

src/uag.c
```diff
@@ -12,23 +12,23 @@
...
+	NULL,         /**< exith             */
+	false,        /**< nodial            */
+	false,        /**< dnd               */
+	NULL,         /**< arg               */
+	NULL,         /**< eprm              */
 #ifdef USE_TLS
-	NULL,
-	NULL
+	NULL,         /**< tls               */
+	NULL          /**< wss_tls           */
```
cspiel1, line 31: Should we replace this by `memset(&uag, 0, sizeof(uag))` in `ua_init()`?

sreimers: You can replace this by:

`static struct uag uag = {.ual = LIST_INIT};`

All none specified members are initialized by 0 or similar zero type value.

PR #3003, opened 2024-04-19
ua: enforce magic cookie in Via branch

maximilianfridrich:
According to [RFC 3261 section 8.1.1.7](https://www.rfc-editor.org/rfc/rfc3261#section-8.1.1.7) the branch parameter in the Via header MUST start with the magic cookie "z9hG4bK". This is now enforced.

Some UAs don't comply with this and send multiple requests with the same branch values which can cause issues.

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -681,6 +681,7 @@ out:
 void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
 {
 	struct config *config = conf_config();
+	const char magic_branch[] = RE_RFC3261_BRANCH_ID;
 	const struct sip_hdr *hdr;
 	struct ua *ua;
 	struct call *call = NULL;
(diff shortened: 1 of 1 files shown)


src/ua.c
```diff
@@ -681,6 +681,7 @@ static int sdp_connection(struct mbuf *mb, int *af, struct sa *sa)
 void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
 {
 	struct config *config = conf_config();
+	const char magic_branch[] = "z9hG4bK";
```
sreimers, line 684: Maybe:

```suggestion
	const char magic_branch[] = RE_RFC3261_BRANCH_ID;
```
and define:

```c
#define RE_RFC3261_BRANCH_ID         "z9hG4bK"
```

src/ua.c
```diff
@@ -693,6 +694,13 @@ void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
 	      sip_transp_name(msg->tp),
 	      &msg->src, &msg->dst);
 
+	if (pl_strncmp(&msg->via.branch, magic_branch, strlen(magic_branch))
```
sreimers, line 697: Size can be calculated at compile time in this case:
```suggestion
	if (pl_strncmp(&msg->via.branch, magic_branch, sizeof(magic_branch)-1)
```

maximilianfridrich: Thank you! Done.

src/ua.c
maximilianfridrich: Thank you! Done.

PR #3011, opened 2024-04-22
core: move DnD feature to menu

cspiel1:
This removes application code from baresip core.

    core: move DnD feature to menu
    
    Adds `CALL_EVENT_ARRIVED`, `CALL_STATE_ARRIVED` and `UA_EVENT_CALL_ARRIVED` to
    the API enums what allows the application to decide if baresip should answer
    with 180 Ringing or with an error status.


juha-h:
I didn't find in the API how to respond with `180 Ringing` when UA_EVENT_CALL_ARRIVED arrives.  What call is it?

Also, it should be possible to respond with `302 Moved Temporarily`. It needs contact URI parameter in the API call.

juha-h:
Thanks for the `call_accept`/`call_reject` explanation.  Fine with another PR for `302 Moved Temporarily`.

juha-h:
@cspiel1 I'm still confused about call_accept() automatically called from the core:
```
	err = call_arrived(call, msg);
	if (err)
		return;

	err = call_accept(call, uag_sipsess_sock(), msg);
	if (err)
		goto error;
```
What if CALL_EVENT_ARRIVED is handled by another thread?  Is it then possible that call_accept() gets called before call_reject() is called?

juha-h:
Noticed that line
```
case UA_EVENT_CALL_ARRIVED:        return "CALL_ARRIVED";
```
is missing from `const char *uag_event_str(enum ua_event ev)`:
```
04-23 10:04:10.575  9521  9648 D Baresip Lib: ua event (?) sip:test@test.tutpro.com
```

include/baresip.h
```diff
@@ -184,11 +185,13 @@ enum call_event {
 /** Call States */
 enum call_state {
 	CALL_STATE_IDLE = 0,
+	CALL_STATE_ARRIVED,
```
alfredh, line 188: please add new values at the end of the enum ... for better binary compat ...

include/baresip.h
```diff
@@ -219,6 +222,7 @@ int  call_answer(struct call *call, uint16_t scode, enum vidmode vmode);
 int  call_progress_dir(struct call *call,
 		       enum sdp_dir adir, enum sdp_dir vdir);
 int  call_progress(struct call *call);
+void call_reject(struct call *call, uint16_t scode, const char *reason);
```
alfredh, line 225: is this the same as call_hangup() ?

include/baresip.h
cspiel1: okay.

Should we keep `CALL_STATE_UNKNOWN` as last value?

include/baresip.h
cspiel1: No, there is a big difference.

`call_hangup()` needs already a SIP session `call->sess` to send the BYE.

`call_reject()` marks the call as REJECTED, which triggers in `call_arrived()` a SIP reply with `sip_treply()`. The status code and reason can be specified by the application (menu, ...). The SIP session is not allocated. Also no streams are allocated.

juha-h:
Test using the latest PR version went fine even when I don't understand how it is guaranteed that `call_accept` does not get called before the app has called `call_reject` when it handles `CALL_ARRIVED`.

cspiel1:
Thanks for testing!

In the `sipsess_conn_handler()` in ua.c we have:

```c
	err = call_arrived(call, msg);
	if (err)
		return;

	err = call_accept(call, uag_sipsess_sock(), msg);
	if (err)
		goto error;
```

The first emits the `CALL_EVENT_ARRIVED`. In the event handler of the app it can be decided to call `call_reject()`. If so `call_arrived()` returns with ENOENT. Then the `sipsess_conn_handler()` returns immediately.

juha-h:
I understand the logic

> The first emits the `CALL_EVENT_ARRIVED`. In the event handler of the app it can be decided to call `call_reject()`. If so `call_arrived()` returns with ENOENT. Then the `sipsess_conn_handler()` returns immediately.

but was just wondering if the app (where the decision on what to do is made) is running the same thread as the native code from which the event is sent.

cspiel1:
> but was just wondering if the app (where the decision on what to do is made) is running the same thread as the native code from which the event is sent.

Yes, the event handler runs always in the `re_main` thread.

Note, that this solution does not allow to reject a call from a different process or thread. It has to be done by a baresip module or an application that is linked to libbaresip and where the UA event handler runs in the main thread of `re_main`.

It is a solution that does not change the default behavior. Thus any existing application keeps working without the need of any adaption.

cspiel1:
You see there are reasons why we should not remove the `call_accept()` from the `sipsess_conn_handler()`. Another one is that `call_accept()` needs the `sip_msg` which currently is not hold in any baresip struct. It is kept in the `sipsess`, which we don't want to allocate when a call is rejected.

cspiel1:
Should we rename `call_reject()` to `call_set_rejected()`, because it does not directly send the SIP reply but only sets the state?

juha-h:
@cspiel1 wrote 
> Edit: An alternative would be to remove the `call_accept()` from the `sipsess_conn_handler()`. Then all existing application would have to add the call to `call_accept()` (resp. a changed form of `call_reject()`), when processing the new `UA_EVENT_CALL_ARRIVED`.
> 
> Application code for alternative solution:
> 
> ```c
> static void ua_event_handler(struct ua *ua, enum ua_event ev,
> 			     struct call *call, const char *prm, void *arg)
> {
> ...
> 	switch (ev) {
> 
> 	case UA_EVENT_CALL_ARRIVED:
> 		if (menu.dnd)
> 			call_reject(call, 480, "Temporarily Unavailable");
> 		else
> 			call_accept(call, ...);
> 
> 		break;
> ...
> }
> ```
This is exactly how I thought it would work when I asked `I didn't find in the API how to respond with 180 Ringing when UA_EVENT_CALL_ARRIVED arrives. What call is it?`.

Would it somehow be possible to get is both ways, i.e., preserve existing behavior, but also allow the above implementation?

juha-h:
On the other hand. this PR already breaks existing behavior, since DnD s moved to menu.  It would be cleaner to explicitly either to accept or reject the call as in the above example.

cspiel1:
Yes, removing `call_accept()` from `sipsess_conn_handler()` would simplify the API and the code, but break existing applications.

@alfredh what do you prefer?

include/baresip.h
```diff
@@ -179,6 +179,7 @@ enum call_event {
 	CALL_EVENT_TRANSFER,
 	CALL_EVENT_TRANSFER_FAILED,
 	CALL_EVENT_MENC,
+	CALL_EVENT_ARRIVED,
```
alfredh, line 182: I think "arrived" is the wrong wording to use here. What has arrived ?

If the event is sent before incoming, perhaps pre-incoming is a better word ?

include/baresip.h
```diff
@@ -191,6 +192,8 @@ enum call_state {
 	CALL_STATE_ESTABLISHED,
 	CALL_STATE_TERMINATED,
 	CALL_STATE_TRANSFER,
+	CALL_STATE_ARRIVED,
+	CALL_STATE_REJECTED,
```
alfredh, line 196: Is there a difference between TERMINATED and REJECTED ?

modules/menu/menu.c
```diff
@@ -678,6 +678,13 @@ static void ua_event_handler(struct ua *ua, enum ua_event ev,
 
 	switch (ev) {
 
+	case UA_EVENT_CALL_ARRIVED:
+		if (menu.dnd) {
+			call_set_rejected(call, 480,
+					  "Temporarily Unavailable");
+		}
```
alfredh, line 685: is it possible to decline the call directly here, instead of setting it to rejected and replying later ?

modules/menu/static_menu.c
```diff
@@ -877,6 +878,7 @@ static int cmd_hangupall(struct re_printf *pf, void *arg)
 	}
 	else if (!pl_strcmp(&pldir, "in")) {
 		hangup_callstate(CALL_STATE_INCOMING);
+		hangup_callstate(CALL_STATE_ARRIVED);
```
alfredh, line 881: this is the drawback of adding a new state ... lots of code have to be checked and updated ...

modules/multicast/receiver.c
```diff
@@ -195,7 +195,7 @@ static void resume_uag_state(void)
 	}
 
 	if (h > multicast_callprio()) {
-		uag_set_dnd(false);
+		/* TODO: set dnd=false */
```
alfredh, line 198: is this a good time to move `multicast` to baresip-apps ?

src/call.c
```diff
@@ -1253,6 +1257,15 @@ void call_hangup(struct call *call, uint16_t scode, const char *reason)
 }
 
 
+void call_set_rejected(struct call *call, uint16_t scode, const char *reason)
+{
```
alfredh, line 1261: please check input arguments

src/call.c
```diff
@@ -1253,6 +1257,15 @@ void call_hangup(struct call *call, uint16_t scode, const char *reason)
...
+	set_state(call, CALL_STATE_REJECTED);
+	call->scode = scode;
+	mem_deref(call->reason);
+	(void)str_dup(&call->reason, reason);
```
alfredh, line 1265: if str_dup fails it will leave the call->reason pointer dangling

src/call.c
```diff
@@ -2221,6 +2234,26 @@ static bool valid_addressfamily(struct call *call, const struct stream *strm)
...
+	int err = 0;
+	set_state(call, CALL_STATE_ARRIVED);
+	call_event_handler(call, CALL_EVENT_ARRIVED, "%s", call->peer_uri);
+
+	if (call->state == CALL_STATE_REJECTED) {
+		(void)sip_treply(NULL, uag_sip(), msg, call->scode,
+				 call->reason ? call->reason : "Rejected");
+
+		set_state(call, CALL_STATE_TERMINATED);
+		mem_deref(call);
```
alfredh, line 2249: is it possible to avoid calling mem_deref(call) from within call.c ?

cspiel1:
> Yes, removing `call_accept()` from `sipsess_conn_handler()` would simplify the API and the code, but break existing applications.
> 
> @alfredh what do you prefer?

I think that we don't want to break existing applications. Now I have an idea for another approach that doesn't need  to extend emus call_event and call_state. I'll try a `call_incoming_h` handler.

```pseudo
void sipsess_conn_handler(...)
{
if (call_incoming_h(call, msg)==Reject)
    sip_treply(..., msg, scode, reason);
}
```

modules/menu/menu.c
cspiel1: here we don't have the `sip_msg`

juha-h:
It would be nice to get the issue resolved that this PR is fixing.  Any solution is fine with me. 

Just for curiosity, I checked what kind of states pjsip has.  Turned out that there is a separate set of stated for matching various SIP concepts:
```
/**
 * This enumeration describes invite session state.
 */
typedef enum pjsip_inv_state
{
    PJSIP_INV_STATE_NULL,           /**< Before INVITE is sent or received  */
    PJSIP_INV_STATE_CALLING,        /**< After INVITE is sent               */
    PJSIP_INV_STATE_INCOMING,       /**< After INVITE is received.          */
    PJSIP_INV_STATE_EARLY,          /**< After response with To tag.        */
    PJSIP_INV_STATE_CONNECTING,     /**< After 2xx is sent/received.        */
    PJSIP_INV_STATE_CONFIRMED,      /**< After ACK is sent/received.        */
    PJSIP_INV_STATE_DISCONNECTED,   /**< Session is terminated.             */
} pjsip_inv_state;
```
```
/**
 * This enumeration represents transaction state.
 */
typedef enum pjsip_tsx_state_e
{
    PJSIP_TSX_STATE_NULL,       /**< For UAC, before any message is sent.   */
    PJSIP_TSX_STATE_CALLING,    /**< For UAC, just after request is sent.   */
    PJSIP_TSX_STATE_TRYING,     /**< For UAS, just after request is received.*/
    PJSIP_TSX_STATE_PROCEEDING, /**< For UAS/UAC, after provisional response.*/
    PJSIP_TSX_STATE_COMPLETED,  /**< For UAS/UAC, after final response.     */
    PJSIP_TSX_STATE_CONFIRMED,  /**< For UAS, after ACK is received.        */
    PJSIP_TSX_STATE_TERMINATED, /**< For UAS/UAC, before it's destroyed.    */
    PJSIP_TSX_STATE_DESTROYED,  /**< For UAS/UAC, will be destroyed now.    */
    PJSIP_TSX_STATE_MAX         /**< Number of states.                      */
} pjsip_tsx_state_e;
```
```
/**
 * Dialog state.
 */
typedef enum pjsip_dialog_state
{
    /** Dialog is not established. */
    PJSIP_DIALOG_STATE_NULL,

    /** Dialog has been established (probably early) */
    PJSIP_DIALOG_STATE_ESTABLISHED
} pjsip_dialog_state;
```
```
/**
 * This enumeration describes SDP negotiation state. 
 */
enum pjmedia_sdp_neg_state
{
    /** 
     * This is the state of SDP negoator before it is initialized. 
     */
    PJMEDIA_SDP_NEG_STATE_NULL,

    /** 
     * This state occurs when SDP negotiator has sent our offer to remote and
     * it is waiting for answer. 
     */
    PJMEDIA_SDP_NEG_STATE_LOCAL_OFFER,

    /** 
     * This state occurs when SDP negotiator has received offer from remote
     * and currently waiting for local answer.
     */
    PJMEDIA_SDP_NEG_STATE_REMOTE_OFFER,

    /**
     * This state occurs when an offer (either local or remote) has been 
     * provided with answer. The SDP negotiator is ready to negotiate both
     * session descriptors. Application can call #pjmedia_sdp_neg_negotiate()
     * immediately to begin negotiation process.
     */
    PJMEDIA_SDP_NEG_STATE_WAIT_NEGO,

    /**
     * This state occurs when SDP negotiation has completed, either 
     * successfully or not.
     */
    PJMEDIA_SDP_NEG_STATE_DONE
};
```

alfredh:
here is an idea for how to solve this problem in a more clean way...
you add an event specific parameter to the UA event, which for example can be a struct or union.

Example:

```c
union event_param {
    int foo;
   int bar;
   struct sip_msg *msg;
    /* ... etc ... */
};
```


the function sending the event will be responsible for setting the right parameters.

the function reading the event will know which params to use, based on the event type.

juha-h:
So, for example, `UA_EVENT_CALL_INCOMING` handler could get access to msg and thus would be able to call `call_accept` or `call_reject`?

alfredh:
Can you please rebase to git main HEAD ?

cspiel1:
Also I try to get rid of the need for another CALL_STATE

alfredh:
Also another option is to save the sip_msg pointer on the call object:


```diff
diff --git a/include/baresip.h b/include/baresip.h
index 9e4df231..ed17cf37 100644
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -279,6 +279,8 @@ int call_msg_src(const struct call *call, struct sa *sa);
 enum sip_transp call_transp(const struct call *call);
 enum sdp_neg_state call_sdp_neg_state(const struct call *call);
 bool call_sdp_change_allowed(const struct call *call);
+void call_set_sipmsg(struct call *call, struct sip_msg *msg);
+struct sip_msg *call_sipmsg(const struct call *call);
 
 /*
  * Custom headers
```

juha-h:
Alfred E. Heggestad writes:

> Also another option is to save the sip_msg pointer on the call object:

That would be cleaner, since there would then be no need to change
ua_event_handler arguments.

juha-h:
There need to be `call_set_accepted` in the API so that the app can explicitly either accept or reject the call.

juha-h:
I still don't understand why the new event UA_EVENT_CALL_ARRIVED is needed.  Why can't handler of UA_EVENT_CALL_INCOMING either reject the call by sending the reply and setting state to TERMINATED or accept the call?

cspiel1:
Let's first finish https://github.com/baresip/baresip/pull/3018

PR #3012, opened 2024-04-22
srtp: lock possible re-keying against usage in receive handler

cHuberCoffee:
* while the main thread may remove the st->srtp_rx struct, the rx thread accessed the freed memory.
 * concurrent access of st->srtp_rx is mitigated with lock.
 * critical section is between first possible removal of st->srtp_rx, until new st->srtp_rx is ready. for the rx thread the access to srtcp_decrypt and srtp_decrypt.


cspiel1:
I am afraid that there is a mutex locking failure in `src/udp/udp.c` where the udp helpers are processed.

sreimers:
I just re-run the thread sanitizers multiple times, looks good so far.

PR #3015, opened 2024-04-26
mc: move multicast to baresip-apps

cspiel1:


alfredh:
there is an unrelated warning:

```
/opt/homebrew/include/spandsp/noise.h:72:45: error: invalid UTF-8 in comment [-Werror,-Winvalid-utf8]
The tolerance for each 1/3rd octave band is <A1><D3>3dB.
```

The compiler should not print warnings from system headers.
System header should be included with angle brackets `<>` and 
directories with system headers should be included using `-isystem`

sreimers:
> The compiler should not print warnings from system headers.

Should be fixed by this PR: https://github.com/baresip/baresip/pull/3017

alfredh:
can you please rebase this PR ?

PR #3018, opened 2024-04-29
event refactoring

cspiel1:
Idea got from:
- https://github.com/baresip/baresip/pull/3011#issuecomment-2082666509
- Initially from https://github.com/baresip/baresip/pull/3011#issuecomment-2081420514

This PR involved now to a backwards compatible solution. Old and new event API exist in parallel.

For now only updated three modules.


include/baresip.h
```diff
@@ -870,6 +870,17 @@ enum ua_event {
...
 };
 
+
+struct event_param {
+	const char *txt;
+	union {
+		struct ua *ua;
+		struct call *call;
+		struct sip_msg *msg;
+	} u;
```
sreimers, line 880: Maybe adding a direct `type` field for the union would be better, a wrong assumption can easily lead to segfaults or strange bugs.

src/event.c
```diff
@@ -333,29 +424,27 @@ void uag_event_unregister(ua_event_h *h)
  * @param fmt  Formatted arguments
  * @param ...  Variable arguments
  */
-void ua_event(struct ua *ua, enum ua_event ev, struct call *call,
-	      const char *fmt, ...)
+void ua_event(enum ua_event ev, const char *fmt, ...)
```
sreimers, line 427: Unsure if mixing formatting args and optional args is a good idea. It's very confusing.

juha-h: SInce this is `ua_event`, I would prefer to keep `struct ua *ua` param and add `event_param` (without `struct ua *ua;` in union).

cspiel1: Yes.

And maybe we should not directly drop the current API.

I'll try a multi function solution for emitting events:
```c
/**
 * @deprecated
 */
void ua_event(struct ua *ua, enum ua_event ev, struct call *call, const char *fmt, ...);

void ua_event_app(enum ua_event ev, const char *fmt, ...);
void ua_event_ua(enum ua_event ev, struct ua *ua, const char *fmt, ...);
void ua_event_call(enum ua_event ev, struct call *call, const char *fmt, ...);
void ua_event_sip_msg(enum ua_event ev, struct sip_msg *msg, const char *fmt, ...);
```

This could be extended without breaking the API.

cspiel1:
I completely replaced the commits

cspiel1:
This commit shows how DnD could be solved: 8b5c3b17326ec323159277e25e1aa0fc93950d37

juha-h:
Still don't understand why new event (UA_EVENT_SIP_INCOMING)  is needed.  Also, better name would be UA_EVENT_INVITE_INCOMING.

Now when I read ua.c code, I noticed that call blocking based on contacts does not belong to core either and should be moved to menu.

src/event.c
```diff
@@ -410,6 +636,153 @@ void module_event(const char *module, const char *event, struct ua *ua,
...
+ * @return 0 if success, otherwise errorcode
+ */
+int event_sip_msg_emit(enum ua_event ev, const struct sip_msg *msg,
+		       const char *fmt, ...)
+{
+	struct event event = {.ev = ev};
+	va_list ap;
+	int err;
+
+	if (event_class(ev) != EVENT_CLASS_CALL)
```
sreimers, line 773: Is `EVENT_CLASS_CALL` right here?

src/event.c
```diff
@@ -410,6 +636,153 @@ void module_event(const char *module, const char *event, struct ua *ua,
...
+/**
+ * Emit an application event
+ *
+ * @param ev   User-agent event
+ * @param fmt  Formatted arguments
+ * @param ...  Variable arguments
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int event_app_emit(enum ua_event ev, const char *fmt, ...)
```
sreimers, line 681: Maybe for App events a optional void * arg could be useful.

src/event.c
cspiel1: Good point!
```c
	if (event_class(ev) != EVENT_CLASS_SIP)
		return EINVAL;
```
is correct.

Explanation:
The event class selects the union field.

| event class | union field |
| --- | --- |
| EVENT_CLASS_UA | event.u.ua |
| EVENT_CLASS_CALL | event.u.call |
| EVENT_CLASS_APP | event.nothing |
| EVENT_CLASS_SIP  |  event.u.msg |

Problem
---
The function `ua_event_class_name()` returns some incongruous class names
- `register`, better would be `ua`
- `mwi` --> better `ua`
- `UA_EVENT_CREATE` returns `application` --> better `ua`
- `VU_REPORT` --> better `call`

But maybe there are already applications which event parsing rely on the current texts. Thus I didn't touch this function. The function is used in `event_encode_dict()` which I would mark as deprecated.

The new function `odict_encode_event()` uses `event_class_name()` that returns logical fitting class names. This should also make parsing of events simpler for applications.

cspiel1:
Module events
---
The last commit should be discussed, and maybe dropped.

A replacement for function `module_event()` could be `event_module_emit()`.
Does an extra API function `event_module_emit` for Module events really have enough benefit for the extra code?

```c
	# new module event
	event_module_emit("ctrl_dbus", modev->event, NULL, "%s", modev->txt);
	# or we use the other event functions, like
	event_app_emit(UA_EVENT_MODULE, NULL, "ctrl_dbus,%s,%s",
		       modev->event, modev->txt);
```

Note: `event_module_emit()` currently untested

cspiel1:
The tests are green now.

Now everything is backwards compatible, the old functions are marked as deprecated and a warning is printed if they are used.

The unit tests already use the new event handler.

In order to keep the PR as small as possible I will
- remove the commit that adds the DnD example,
- remove the new module event API mentioned here https://github.com/baresip/baresip/pull/3018#issuecomment-2089769648

juha-h:
Regarding this:
```
err = event_sip_msg_emit(UA_EVENT_SIP_INCOMING, msg, "");
```
I would prefer names `event_sip_req_emit` and `UA_EVENT_SIP_REQ_INCOMING`.

include/baresip.h
```diff
@@ -866,10 +866,25 @@ enum ua_event {
...
+struct event {
+	enum ua_event ev;
+	const char *txt;
+	int err;
+	union {
+		struct ua *ua;
+		struct call *call;
+		const struct sip_msg *msg;
+		void *arg;
+	} u;
```
alfredh, line 884: union means that only one of the pointers are valid, for example ua OR call.
you cannot cast a ua pointer to a call pointer, or msg pointer etc.

is it better to use a struct here ?

most of the events have a valid UA pointer

include/baresip.h
```diff
@@ -1652,14 +1668,25 @@ void module_app_unload(void);
...
 void uag_event_unregister(ua_event_h *eh);
+int  event_register(event_h *eh, void *arg);
+void event_unregister(event_h *eh);
 void ua_event(struct ua *ua, enum ua_event ev, struct call *call,
 	      const char *fmt, ...);
+int event_app_emit(enum ua_event ev, void *arg, const char *fmt, ...);
+int event_ua_emit(enum ua_event ev, struct ua *ua, const char *fmt, ...);
+int event_call_emit(enum ua_event ev, struct call *call, const char *fmt, ...);
+int event_sip_msg_emit(enum ua_event ev, const struct sip_msg *msg,
+		       const char *fmt, ...);
```
alfredh, line 1683: is the `call` valid in this case ?

cspiel1:
> Regarding this:
> 
> ```
> err = event_sip_msg_emit(UA_EVENT_SIP_INCOMING, msg, "");
> ```
> 
> I would prefer names `event_sip_req_emit` and `UA_EVENT_SIP_REQ_INCOMING`.

It is more the type of the second parameter we want to address with the name. If we had C++ it would be best like this:
```c
int event_emit(enum ua_event ev, void *arg, const char *fmt, ...);
int event_emit(enum ua_event ev, struct ua *ua, const char *fmt, ...);
int event_emit(enum ua_event ev, struct call *call, const char *fmt, ...);
int event_emit(enum ua_event ev, const struct sip_msg *msg, const char *fmt, ...);
```

And it is not only for incoming SIP requests.
`event_sip_msg_emit()` could also be used for SIP replies, for every event for which we want to add a SIP message.

For the event I tried to keep it short with `UA_EVENT_SIP_INCOMING`. We could also name it `UA_EVENT_SIPMSG_INCOMING`.

include/baresip.h
```diff
@@ -866,10 +866,25 @@ enum ua_event {
 	UA_EVENT_MODULE,
 	UA_EVENT_END_OF_FILE,
 	UA_EVENT_CUSTOM,
+	UA_EVENT_SIP_INCOMING,
 
 	UA_EVENT_MAX,
 };
 
+
+struct event {
```
alfredh, line 875: perhaps add a one or two letter prefix to "event", like `xxevent`.

include/baresip.h
cspiel1: I also thought about the choice between union and struct. And decided to try first a union. Reasons for that:
- In the current UA event API we often pass a `call` and its `ua` which is redundant.
- Then we had often UA events where `ua` was set, but `call` was NULL.
- The new events which add a `sip_msg`  pointer should be used before a call is allocated.

It seems to work with a union. An application would have no benefit to have multiple of these pointers.

include/baresip.h
cspiel1: Yes. Here the table from above extended with the function names:

| function name | event class | union field |
| --- | --- | --- |
| event_ua_emit | EVENT_CLASS_UA | event.u.ua |
| event_call_emit | EVENT_CLASS_CALL | event.u.call |
| event_app_emit | EVENT_CLASS_APP | event.u.arg |
| event_sip_msg_emit | EVENT_CLASS_SIP  |  event.u.msg |

cspiel1:
Currently there are a lot of deprecated warnings when the unit tests are ran. Thus before I'll set this PR green, I will remove the calls to the deprecated API in the baresip core.

juha-h:
`For the event I tried to keep it short with UA_EVENT_SIP_INCOMING. We could also name it UA_EVENT_SIPMSG_INCOMING.`

OK, but I would prefer the same as in the event_sip_msg_emit, i.e., UA_EVENT_SIP_MSG_INCOMING.

cspiel1:
> `For the event I tried to keep it short with UA_EVENT_SIP_INCOMING. We could also name it UA_EVENT_SIPMSG_INCOMING.`
> 
> OK, but I would prefer the same as in the event_sip_msg_emit, i.e., UA_EVENT_SIP_MSG_INCOMING.

The "emit" function names are related to event classes and to the union fields.
The event types (which still are `enum ua_event`) are grouped into event classes. @alfredh already suggested to name the new event `UA_EVENT_SIPSESS_CONN` which makes sense, because it is emitted in the sipsess connection handler. This name is both, perfectly specific and general enough. The event class is `EVENT_CLASS_SIP`.

cspiel1:
Please compare with `list_apply()` implementation! It also stops the loop when the handler returns `true`.
`true` means stop processing.

We don't have a `call` object where `event_sip_msg_emit()`, `event_ua_emit()` or `event_app_emit()` is called. It should become a general purpose event API.

The err code of the "emit" function can be used to have influence on further processing after the event was emitted and processed by the handler list. As an example see Dnd commit: https://github.com/baresip/baresip/commit/881af62a630ee603b6daa36c0ba054884830bd09

juha-h:
Christian Spielberger writes:

> The err code of the "emit" function can be used to have influence on
> further processing after the event was emitted and processed by the
> handler list. As an example see Dnd commit:
> https://github.com/baresip/baresip/commit/881af62a630ee603b6daa36c0ba054884830bd09

My point is that the information regarding possible further processing
should be carried by some other means than the boolean result, e.g., by
call state change.  So, for example, in case of DnD, after event
processing, the call state could be either CALL_STATE_TERMINATED
or CALL_STATE_RINGING.

cspiel1:
In the sipsess connection handler further processing is stopped by the error code ENOENT.

Further processing of the event handler list is stopped by return true.

The DnD implementation I proposed with `https://github.com/baresip/baresip/commit/881af62a630ee603b6daa36c0ba054884830bd09` has the benefit, that there is no need to allocate a call object if DnD is switched on.

In my opinion err = ENOENT is an adequate error code for preventing the allocation of a call object and leaving the sipsess connection handler immediately.

Please see `https://github.com/baresip/baresip/commit/881af62a630ee603b6daa36c0ba054884830bd09`!

juha-h:
All this should be done in app (menu module)  UA_EVENT_CALL_INCOMING (or whatever it will be called) event  handler, not in core:
```
case CALL_EVENT_INCOMING:

		if (contact_block_access(baresip_contacts(),
					 peeruri)) {

			info("ua: blocked access: \"%s\"\n", peeruri);

			ua_event(ua, UA_EVENT_CALL_CLOSED, call, "%s", str);
			mem_deref(call);
			break;
		}

		ua_event(ua, UA_EVENT_CALL_INCOMING, call, "%s", peeruri);
		switch (ua->acc->answermode) {

		case ANSWERMODE_EARLY:
		case ANSWERMODE_EARLY_AUDIO:
			(void)call_progress(call);
			break;

		case ANSWERMODE_EARLY_VIDEO:
			if (!call_early_video_available(call)) {
				info ("ua: peer is not capable of early "
					"video. proceed as normal call\n");
				break;
			}

			(void)call_progress(call);
			break;

		default:
			break;
		}
		break;
```

cspiel1:
To be also backwards compatible with `module_event()` https://github.com/baresip/baresip/pull/3018/commits/e384139da9cb2fbaee052f9c3430a074404f8d8f

include/baresip.h
```diff
@@ -866,10 +866,25 @@ enum ua_event {
 	UA_EVENT_MODULE,
 	UA_EVENT_END_OF_FILE,
 	UA_EVENT_CUSTOM,
+	UA_EVENT_SIPSESS_CONN,
 
 	UA_EVENT_MAX,
 };
 
+
+struct bevent {
```
alfredh, line 875: you should also consider to make this struct opaque, i.e. defined in core.h or a .c file
it will make the api more robust and future proof.
and it will be more difficult for API users to make accidental mistakes

juha-h:
My Android app cannot use the new `event_handler`, because it returns `bool`.  App's `ua_event_handler` forwards the event to Kotlin application that decides what to do with the event and it does not return anything back to `ua_event_handler`.  So if use of the new `event_handler` becomes mandatory, my app stops working.

I have many times tried for vain to tell there must be some other means that the return value to tell what the result of event handling was.

juha-h:
I have not found anywhere any documentation when this event handler should return true and when false:
```
typedef bool (bevent_h)(struct bevent *event, void *arg);
```
All other handlers return void.

juha-h:
The handler of incoming INVITE requests could work like this:

- if INVITE cannot for some reason or other accepted, response with 4xx
- if called party is currently relocated, response with 3xx
- if INVITE needs to be answered automatically, accept it with 2xx response
- otherwise, accept INVITE with 18x response

So it simply does the job and does not need to return anything.

juha-h:
One more example of code that needs to be moved from core to the handler is this:
```
        if (config->call.max_calls &&
	    uag_call_count() + 1 > config->call.max_calls) {
		info("ua: rejected call from %r (maximum %d calls)\n",
		     &msg->from.auri, config->call.max_calls);
		(void)sip_treply(NULL, uag_sip(), msg, 486, "Max Calls");
		return;
	}
```

cspiel1:
If an incoming INVITE is rejected by an event handler, then no further event handler should process the event. My proposition is to return true if the event processing should be stopped. We could also add a stop flag to `struct bevent`.

Isn't it possible to return a bool from a Kotlin callback function?

cspiel1:
Instead of the return value I put the stop flag to `struct bevent`. Now there are the options:
- setting `bevent.err` to an error code -> stops event processing with an error,
- setting `bevent.stop=true` -> stop event processing without an error,
- by default the next event handler is called. This removes also the need for `return false;` at multiple locations.

juha-h:
Christian Spielberger writes:

> If an incoming INVITE is rejected by an event handler, then no further
> event handler should process the event. My proposition is to return
> true if the event processing should be stopped. We could also add a
> stop flag to `struct bevent`.

That doesn't change the situation.  If event processing should be
stopped (for example when the handler rejects the INVITE due to dnd),
the state will be CALL_STATE_TERMINATED.  Why checking of the state is
not enough?  Why do you need a return value or stopped flag, which in my
opinion is bad design and breaks the idea of handler?  After short
search I have not found in baresip any other handler that returns a
value or has stopped flag.

> Isn't it possible to return a bool from a Kotlin callback function?

Yes it could be possible, but see above.

juha-h:
Why these are not done before `bevent_sip_msg_emit(UA_EVENT_SIPSESS_CONN, msg, "")` call?
```
        /* handle multiple calls */
	if (config->call.max_calls &&
	    uag_call_count() + 1 > config->call.max_calls) {

		info("ua: rejected call from %r (maximum %d calls)\n",
		     &msg->from.auri, config->call.max_calls);
		(void)sip_treply(NULL, uag_sip(), msg, 486, "Max Calls");
		return;
	}

	/* Handle Require: header, check for any required extensions */
	hdr = sip_msg_hdr_apply(msg, true, SIP_HDR_REQUIRE,
				require_handler, ua);
	if (hdr) {
		info("ua: call from %r rejected with 420"
			     " -- option-tag '%r' not supported\n",
			     &msg->from.auri, &hdr->val);

		(void)sip_treplyf(NULL, NULL, uag_sip(), msg, false,
				  420, "Bad Extension",
				  "Unsupported: %r\r\n"
				  "Content-Length: 0\r\n\r\n",
				  &hdr->val);
		return;
	}

	if (ua->acc->rel100_mode == REL100_REQUIRED &&
	    !(sip_msg_hdr_has_value(msg, SIP_HDR_SUPPORTED, "100rel") ||
	      sip_msg_hdr_has_value(msg, SIP_HDR_REQUIRE, "100rel"))) {

		info("ua: call from %r rejected with 421"
			     " -- option-tag '%s' not supported by remote\n",
			     &msg->from.auri, "100rel");
		(void)sip_treplyf(NULL, NULL, uag_sip(), msg, false,
				  421, "Extension required",
				  "Require: 100rel\r\n"
				  "Content-Length: 0\r\n\r\n");
		return;
	}
```

juha-h:
So when INVITE comes in, `ipsess_conn_handler` does the checks and (if they succees) sends UA_EVENT_SIPSESS_CONN event.  When the app receives it, it `calls ua_hangup`, `ua_answer` (when auto answering), or `ua_accept`.  That would be clean without any implicit actions.

juha-h:
... but I'm ready to give up as long there API documentation is clear.

cspiel1:
I'll add Doxygen to the API

juha-h:
Once you move these https://github.com/baresip/baresip/pull/3018#issuecomment-2095311947, I can add new API function `ua_accept` and can always stop processing after UA_EVENT_SIPSESS_CONN event.

cspiel1:
updated the description

juha-h:
Could you move this
```
err = bevent_sip_msg_emit(UA_EVENT_SIPSESS_CONN, msg, "");
	if (err == ENOENT)
		return;
```
down after all checks?

include/baresip.h
```diff
@@ -881,6 +886,7 @@ enum answer_method {
 /** Defines the User-Agent event handler */
 typedef void (ua_event_h)(struct ua *ua, enum ua_event ev,
 			  struct call *call, const char *prm, void *arg);
+typedef void (bevent_h)(struct bevent *event, void *arg);
```
alfredh, line 889: suggested signature:

```
typedef void (bevent_h)(struct ua_event ev, struct bevent_payload *event, void *arg);
```

All events must have a valid EV event set.

The payload is optional and type dependent.

src/event.c
```diff
@@ -14,6 +14,20 @@ enum {
...
+	bool stop;
+	union {
+		struct ua *ua;
+		struct call *call;
+		const struct sip_msg *msg;
+		void *arg;
+	} u;
+};
+
+
```
alfredh, line 30: perhaps rename this file to bevent.c ?

include/baresip.h
cspiel1: I'll try this.

cspiel1: Do you mean?
```c
typedef void (bevent_h)(enum ua_event ev, struct bevent_payload *event, void *arg);
```

Rename `struct bevent` -> `struct bevent_payload`?

cspiel1:
@alfredh Now I changed to
```c
typedef void (bevent_h)(enum ua_event ev, struct bevent *event, void *arg);
```

Can we keep the shorter name `struct bevent`?

> All events must have a valid EV event set.
> The payload is optional and type dependent.

The `.._emit(enum ua_event ev, ...)` functions ensure that a valid ev is given. Everything else is optional.

alfredh:
I would like to suggest that we split this up into logical blocks (many PRs).

1. add core bevent
2. add backwards wrapper
3. update function calls
4. add new features like DND

For example a small PR can be to just rename event.c to bevent.c

Is there a need to update all the calls that are sending the event ?
It would be nice if the PR is small and easy to read.

CMakeLists.txt
```diff
@@ -167,7 +167,7 @@ set(SRCS
   src/custom_hdrs.c
   src/descr.c
   src/dial_number.c
-  src/event.c
+  src/bevent.c
```
alfredh, line 170: this can be a separate PR

include/baresip.h
```diff
@@ -866,10 +866,15 @@ enum ua_event {
 	UA_EVENT_MODULE,
 	UA_EVENT_END_OF_FILE,
 	UA_EVENT_CUSTOM,
+	UA_EVENT_SIPSESS_CONN,
```
alfredh, line 869: this change and related changes can be separate PR

cspiel1:
- squashed the commits and pushed to https://github.com/cspiel1/baresip/tree/bevent_complete_refactoring as backup
- will push force the first PR in order to keep the discussions here

Issue #2918, opened 2024-02-12
aufile: ITU G.711 mu-law with 8000 sample rate are played much too fast


cspiel1:
Reproduce this with:

config
```
module			aufile.so
file_ausrc		aufile
ring_aufile		ring.wav
```

Trigger an incoming call.

alfredh:
Here are the media files:

```
share/autoanswer.wav:  RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 8000 Hz
share/busy.wav:        RIFF (little-endian) data, WAVE audio, ITU G.711 mu-law, mono 8000 Hz
share/callwaiting.wav: RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 8000 Hz
share/error.wav:       RIFF (little-endian) data, WAVE audio, ITU G.711 mu-law, mono 8000 Hz
share/message.wav:     RIFF (little-endian) data, WAVE audio, ITU G.711 mu-law, mono 8000 Hz
share/notfound.wav:    RIFF (little-endian) data, WAVE audio, ITU G.711 mu-law, mono 8000 Hz
share/ringback.wav:    RIFF (little-endian) data, WAVE audio, ITU G.711 mu-law, mono 8000 Hz
share/ring.wav:        RIFF (little-endian) data, WAVE audio, ITU G.711 mu-law, mono 8000 Hz
share/sound0.wav:      RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 44100 Hz
share/sound1.wav:      RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 44100 Hz
share/sound2.wav:      RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 44100 Hz
share/sound3.wav:      RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 44100 Hz
share/sound4.wav:      RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 44100 Hz
share/sound5.wav:      RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 44100 Hz
share/sound6.wav:      RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 44100 Hz
[...]

PR #3020, opened 2024-05-01
aufile: use correct audio format S16LE for aubuf frames

alfredh:


cspiel1:
Verified that this fixes the issue #2918. But still I try to understand the fix. Because `AUFMT_S16LE` is already set in `src/play.c:449`

PR #3030, opened 2024-05-17
debug_cmd: command aufileinfo return answer

cspiel1:
- If the chosen audio source supports blocking mode, the result is returned now
  directly with the given print function.
- Otherwise the result is printed to the log like before.


alfredh:
Would it not be simpler to just use this API:

```
size_t aufile_get_length(struct aufile *af, const struct aufile_prm *prm);
```

cspiel1:
The command supports any audio source.

`aufile_get_length()` supports only WAV files.

cspiel1:
e.g. gst can be used to read arbitrary audio file formats.

Related PR: https://github.com/baresip/baresip/pull/3032

alfredh:
most audio file containers support reading the duration.

I would suggest adding the duration in seconds to the audio_source parameter.
Modules that support duration can set this parameter.

Streaming audio will not have any known duration.

The current approach is to play the whole file and calculate number of samples.
What is the file is 100 hours long ?
It would be much more robust and simple to just read out the "duration" from the container.

alfredh:
here is a proposal:

```diff
diff --git a/include/baresip.h b/include/baresip.h
index ab7b1b38..3d895a33 100644
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -569,6 +569,7 @@ struct ausrc_prm {
        uint8_t    ch;          /**< Number of channels         */
        uint32_t   ptime;       /**< Wanted packet-time in [ms] */
        int        fmt;         /**< Sample format (enum aufmt) */
+       float duration;
 };
 
 typedef void (ausrc_read_h)(struct auframe *af, void *arg);
diff --git a/modules/aufile/aufile_src.c b/modules/aufile/aufile_src.c
index a0103e8b..2438650c 100644
--- a/modules/aufile/aufile_src.c
+++ b/modules/aufile/aufile_src.c
@@ -270,6 +270,9 @@ int aufile_src_alloc(struct ausrc_st **stp, const struct ausrc *as,
                st->errh(0, NULL, st->arg);
        }
 
+       size_t duration_ms = aufile_get_length(st->aufile, &fprm);
+       prm->duration = .001f * duration_ms;
+
  out:
        if (err)
                mem_deref(st);
```

cspiel1:
Thanks, will try this in a few days.

modules/debug_cmd/debug_cmd.c
```diff
@@ -172,7 +177,9 @@ struct fileinfo_st {
 	struct ausrc_prm prm;
 	size_t sampc;
 	struct tmr tmr;
-	bool   finished;
+	RE_ATOMIC bool finished;
+	struct re_printf *pf;
```
alfredh, line 181: taking a copy of this pointer may lead to undefined behaviour.

In this module, we dont know how `pf` is declared in the application.

modules/debug_cmd/debug_cmd.c
```diff
@@ -193,18 +200,23 @@ static void fileinfo_timeout(void *arg)
 	if (st->prm.ch && st->prm.srate)
 		s = ((double) st->sampc)  / st->prm.ch / st->prm.srate;
 
-	if (st->finished) {
-		info("debug_cmd: length = %1.3lf seconds\n", s);
+	if (re_atomic_rlx(&st->finished)) {
```
alfredh, line 203: Can you make the `atomic` related changes into a separate PR ?

alfredh:
Have you figured out what we should do about this one ?

cspiel1:
I'd give this https://github.com/baresip/baresip/pull/3030#issuecomment-2119185712 a try.

And we could replace the "blocking mode" of `ausrc_alloc()` by
```c
int ausrc_prm_query(struct ausrc_prm *prm, struct list *ausrcl,
		const char *name, const char *device);
```

modules/debug_cmd/debug_cmd.c
cspiel1: The bool `finished` may become obsolete.

modules/debug_cmd/debug_cmd.c
cspiel1: Also, may become obsolete.

cspiel1:
> And we could replace the "blocking mode" of `ausrc_alloc()` by
> 
> ```c
> int ausrc_prm_query(struct ausrc_prm *prm, struct list *ausrcl,
> 		const char *name, const char *device);
> ```

Another API function would need also a callback handler. With last commit I replaced the "blocking mode" with `ptime==0` by a mode auf `ausrc_alloc()` where the read and error handlers are NULL pointers. I think this is more intuitive.

cspiel1:
TODO: A memleak on baresip shutdown. Not caught by the unit tests.

cspiel1:
Should I add a unit test for `module/aufile` and command "aufileinfo"? Module aufile doesn't have dependencies that go beyond `re`.

alfredh:
this looks good. local testing works:

```
$ ./baresip -e "/aufileinfo sound0.wav" -eq
baresip v3.12.1 Copyright (C) 2010 - 2024 Alfred E. Heggestad et al.
Local network addresses:
       eno1:  10.0.1.2
       eno1:  fe80::96c6:91ff:fead:cdc9
ua: adding SIP CA file: /etc/ssl/certs/ca-certificates.crt
ua: adding SIP CA path: 
aucodec: PCMU/8000/1
aucodec: PCMA/8000/1
ausrc: aufile
auplay: aufile
vidcodec: H264
vidcodec: H264
vidcodec: H265
avcodec: using H.264 encoder 'libx264' -- libx264 H.264 / AVC / MPEG-4 AVC / MPEG-4 part 10
avcodec: using H.264 decoder 'h264' -- H.264 / AVC / MPEG-4 AVC / MPEG-4 part 10
avcodec: using H.265 encoder 'libx265' -- libx265 H.265 / HEVC
avcodec: using H.265 decoder 'hevc' -- HEVC (High Efficiency Video Coding)
vidsrc: fakevideo
vidisp: fakevideo
Populated 1 account
Populated 2 contacts
Populated 2 audio codecs
Populated 0 audio filters
Populated 3 video codecs
Populated 0 video filters
baresip is ready.
/aufileinfo sound0.wav
aufile: loading input file '/usr/local/share/baresip/sound0.wav'
aufile: /usr/local/share/baresip/sound0.wav: 44100 Hz, 1 channels, S16LE
debug_cmd: length = 0.400 seconds
Quit
ua: stop all (forced=0)
```

cspiel1:
For now I have to remove the test. Is somebody interested in this test with aufile?

alfredh:
> For now I have to remove the test. Is somebody interested in this test with aufile?

Adding test is optional.

It might be added if it is easy to do, and the test covers the core, and the
PR author has extra energy to do it :)

alfredh:
Thanks! merged to main.

Issue #3034, opened 2024-05-21
auconv not loading


bacterik:
macOS Sonoma. Just compiled and update to 3.12.0, but the module auconv dosnt seem to load. And I get distorted audio ofcourse.

```
ausrc_channels          1-2
auplay_channels         1-2
#audio_txmode           poll            # poll, thread
audio_level             no
ausrc_format            float           # s16, float, ..
auplay_format           float           # s16, float, ..
auenc_format            s16             # s16, float, ..
audec_format            s16             # s16, float, ..
```

...

```
# Audio codec Modules (in order)
module                  opus.so
module                  amr.so
#module                 g7221.so
#module                 g722.so
#module                 g726.so
module                  g711.so
module                  l16.so
#module                 mpa.so
#module                 codec2.so

# Audio filter Modules (in encoding order)
module                  auconv.so
module                  auresamp.so
module                  vumeter.so
#module                 sndfile.so
#module                 plc.so
#module                 webrtc_aec.so
```
...

```
aucodec: opus/48000/2
aucodec: AMR/8000/1
aucodec: PCMU/8000/1
aucodec: PCMA/8000/1
aucodec: L16/48000/2
aucodec: L16/44100/2
aucodec: L16/32000/2
aucodec: L16/16000/2
aucodec: L16/8000/2
aucodec: L16/48000/1
aucodec: L16/44100/1
aucodec: L16/32000/1
aucodec: L16/16000/1
aucodec: L16/8000/1
aufilt: auconv
aufilt: auresamp
aufilt: vumeter
auplay: coreaudio
ausrc: coreaudio
auplay: audiounit
ausrc: audiounit
auplay: jack
ausrc: jack
ausrc: aubridge
auplay: aubridge
ausrc: aufile
auplay: aufile
```
...

`All 1 useragent registered successfully! (48 ms)`

call: connecting to 'sip:xxx@xxx'..](url)

**audio tx pipeline:   coreaudio ---> aubuf ---> opus
audio rx pipeline:   coreaudio <--- aubuf <--- opus**

^ no auconv

calling-client@xxx: Call established: sip:calling-destination@xxx
audio: tx: invalid sample formats (FLOAT -> S16LE). Use module auconv!
audio: tx: invalid sample formats (FLOAT -> S16LE). Use module auconv!
audio: tx: invalid sample formats (FLOAT -> S16LE). Use module auconv!
audio: tx: invalid sample formats (FLOAT -> S16LE). Use module auconv!

[...]

PR #3046, opened 2024-05-27
audio: fix setup of audio filters

cspiel1:
Currently audio filters do not work.

Enc/Dec need to be setup first.


cspiel1:
Thanks for quick  review!

sreimers:
Fixes #3034

PR #3048, opened 2024-05-28
webrtc_aec: Add support for extended filter option

marcel-behlau-elfin:
The module currently contains a description for the `webrtc_aec_extended_filter` parameter, but didn't provide an implementation.

The PR implements the `webrtc_aec_extended_filter` option.

### DIFF
--- a/modules/webrtc_aec/aec.cpp
+++ b/modules/webrtc_aec/aec.cpp
@@ -113,6 +113,7 @@ int webrtc_aec_alloc(struct aec **stp, void **ctx, struct aufilt_prm *prm)
 			new webrtc::ExtendedFilter(true)
 		);
 	}
+
 	aec->inst->SetExtraOptions(config);
 
  out:
--- a/modules/webrtc_aec/aec.cpp
+++ b/modules/webrtc_aec/aec.cpp
@@ -109,8 +109,10 @@ int webrtc_aec_alloc(struct aec **stp, void **ctx, struct aufilt_prm *prm)
 	aec->inst->gain_control()->Enable(true);
 
 	if (webrtc_aec_extended_filter) {
-		config.Set<webrtc::ExtendedFilter>(new webrtc::ExtendedFilter(true));
-      }
+		config.Set<webrtc::ExtendedFilter>(
+			new webrtc::ExtendedFilter(true)
+		);
+	}
 	aec->inst->SetExtraOptions(config);
 
  out:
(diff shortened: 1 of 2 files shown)


alfredh:
Is this an option that can be enabled all the time (i.e. no need for config) ?

marcel-behlau-elfin:
From the documentation (https://www.freedesktop.org/wiki/Software/PulseAudio/Notes/9.0/):

> There's a new "extended filter" mode in the WebRTC echo canceller. The extended filter is more complex and less sensitive to incorrect delay reporting from the hardware than the regular filter. The extended filter mode is disabled by default, because it seemed produce worse results during double-talk (based on some very quick testing). Pass "extended_filter=true" in aec_args when loading module-echo-cancel to enable this feature.

So passing this as an option should be fine.

modules/webrtc_aec/aec.cpp
```diff
@@ -31,6 +31,7 @@
 
 using namespace webrtc;
 
+static bool webrtc_aec_extended_filter;
```
cspiel1, line 34: Shouldn't this be added to the struct aec?

modules/webrtc_aec/aec.cpp
```diff
@@ -106,6 +108,13 @@ int webrtc_aec_alloc(struct aec **stp, void **ctx, struct aufilt_prm *prm)
 
 	aec->inst->gain_control()->Enable(true);
 
+	if (webrtc_aec_extended_filter) {
+		config.Set<webrtc::ExtendedFilter>(
+			new webrtc::ExtendedFilter(true)
+		);
+	}
```
cspiel1, line 115: please add an empty line

marcel-behlau-elfin: Before line 112? 

Sorry, i didn't understand currently the coding style.

cspiel1: After a `}` normally we have an empty line.

marcel-behlau-elfin: Fixed with c3c6e88b

modules/webrtc_aec/aec.cpp
marcel-behlau-elfin: I used the implementation of `vumeter_stderr` from `modules/vumeter/vumeter.c` as template, which uses a static, global variable.

I would also prefer to avoid the global variable, but i'm not sure, how to access `aec` within `module_init`.

cspiel1:
Pls could you take this https://github.com/baresip/baresip/pull/3048#discussion_r1617174798 in account?

And @alfredh tried to ask if this option should/can be enabled in general, for better aec performance (what ever). We try to keep the configuration as small as possible.

@alfredh should we avoid adding module config settings to the config example in the future in general? A new setting can be added to modules, anyway. The config example should be as short as possible.

alfredh:
The current patch is fine.

It is fine to add the config option, since it adds extra processing ...

PR #3047, opened 2024-05-27
test: call - complete test for auconv

cspiel1:
This PR adds a test for audio filters. More concretely:
- auconv
- auresamp

```
audio tx pipeline:      ausine ---> aubuf ---> auconv ---> auresamp ---> aucmock
audio rx pipeline:  mock-auplay <--- aubuf <--- auconv <--- auresamp <--- aucmock
```

It tests these combinations
| ausrc/auplay | codec |
| -------- | --------- |
| s16le | s16le |
| s16le | float |
| float | s16le |
| float | float |

resampler test:
```
auresamp: resample encoder 16000/1 --> 48000/2
auresamp: resample decoder 48000/2 --> 16000/1
```

PR #3053, opened 2024-05-29
auresamp: fix size of sampv if aufmt conversion needed

cspiel1:
Second "try". Now it should be correct.

The test is activated here in this draft: https://github.com/baresip/baresip/pull/3047

`st->sampv` is used for:
- audio format conversion from input format to s16le and for
- conversion from s16le to the output format (= input format)

The maximum size needs to be allocated.


cspiel1:
All tests are green now for linux systems in this draft: https://github.com/baresip/baresip/pull/3047

PR #3057, opened 2024-06-06
ua,uag: support SUBSCRIBE only if handler is set

maximilianfridrich:
BareSIP always has the SUBSCRIBE method in its Allow header, so other UAs correctly assume that we will respond to those. However, if no `subh` is set, these SUBSCRIBE messages will not be responded to.

Now, only if `subh` is set, we add SUBSCRIBE to the Allow header and if it is not set, we respond with 405 Method Not Allowed (see [RFC 3261 section 8.2.1](https://www.rfc-editor.org/rfc/rfc3261#section-8.2.1)).

### DIFF
--- a/src/uag.c
+++ b/src/uag.c
@@ -476,6 +476,9 @@ static bool sub_handler(const struct sip_msg *msg, void *arg)
 
 	if (uag.subh)
 		uag.subh(msg, ua);
+	else
+		(void)sip_treply(NULL, uag_sip(), msg, 405,
+				 "Method Not Allowed");
 
 	return true;
 }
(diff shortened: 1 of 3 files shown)


src/uag.c
```diff
@@ -1267,3 +1270,13 @@ bool uag_delayed_close(void)
 	return uag.delayed_close;
 }
 
+
+/**
+ * Get the subscribe handler
+ *
+ * @return Subscribe handler
+ */
+ sip_msg_h *uag_subh(void)
```
cspiel1, line 1279: leading blank?

maximilianfridrich: Fixed.

PR #3060, opened 2024-06-07
ci/coverage: increase min coverage

sreimers:


cspiel1:
Currently we compute only over `src`, the baresip core.
```
cov=$(~/.local/bin/gcovr -r . -f src -s | grep lines | awk '{ print $2 }' | sed 's/%//'
```
Should we add some tests for modules without dependency, like
- g711
- ausine
- fakevideo
- auconv
- auresamp
- aufile

Of course step by step.

Issue #3066, opened 2024-06-19
Several tests fail, in single test mode


cspiel1:
@cHuberCoffee found that several test are failing if they are run in single test mode.

E.g.
```
$ ./test/selftest test_call_multiple
running baresip selftest version 3.12.1 with 1 tests
[ RUN      ] test_call_multiple (rx main)
selftest: re_main() loop timed out -- test hung..

TEST_ERR: /home/cspiel/src/baresip/baresip/test/call.c:1094: (Connection timed out [110])
test_call_multiple (rx main): test failed (Connection timed out [110])
test failed (Connection timed out [110])
re main loop:
  maxfds:       1024
  nfds:         3
  method:       epoll
  polling:      0
  sig:          0
  timers:       1
  mutex:        0x5af455de24f0
  tid:          0x752617452740
  thread_enter: 0
  async:        0x5af455de27b0

```

cspiel1:
For `test_call_multiple`
```
test: [ sip:a@127.0.0.1 ] event: CALL_CLOSED (486 Max Calls)
sip:b@127.0.0.1:37698: session closed: 486 Max Calls
```

cspiel1:
@alfredh @sreimers I suggest to fix this before v3.13.0. This should not be very difficult.

I could make the release today on the afternoon.

cspiel1:
How to run each test in single test mode:
```sh
$ tests=`grep -E "TEST.test_" test/main.c | sed -e "s_TEST(__" -e "s_),__"`
$ for test in $tests; do test/selftest $test; done
```

cspiel1:
First PR fixes "Max Calls" issues: https://github.com/baresip/baresip/pull/3067

We need another PR for `test_ua_register_dns` which fails only in single test mode in combination with RX thread.

cspiel1:
Good case with two test cases:
```
./test/selftest test_ua_register_auth_dns test_ua_register_dns
```

Bad case with only this one
```
./test/selftest test_ua_register_dns
```

PR #3067, opened 2024-06-19
test: call - reset max_calls to default value

cspiel1:
Fixes: https://github.com/baresip/baresip/issues/3066


cspiel1:
`test_ua_register_dns` still fails in single test mode

PR #3069, opened 2024-06-19
release v3.13.0

cspiel1:


cspiel1:
Here I made a mistake to press "Enter" after filling the release name into the input field in the "New Release" form. This selects the default button, which is "Publish Release". The result is that the tag is set to the last commit before the "release v3.13.0" commit which contains the CHANGELOG.md and the version number update. Should I make a v3.13.1?

sreimers:
I just deleted the release + tag and re-created.

cspiel1:
Maybe someone else now has this problem like me:
```
$ git pull --tags
From https://github.com/baresip/baresip
 ! [rejected]            v3.13.0    -> v3.13.0  (would clobber existing tag)
```

cspiel1:
I think the tag has to be deleted locally also and then git pull:

```
git tag -d v3.13.0
git pull
```

cspiel1:
Thanks @sreimers !

PR #3072, opened 2024-06-27
feature: in-band dtmf detection and generation

marcel-behlau-elfin:
The following PR provides a new in_band_dtmf module. The module allows to encode in-band DTMF tones by given string input. Is an in-band DTMF tone detected in decoding stream, a `UA_EVENT_CALL_DTMF_START` and a `UA_EVENT_CALL_DTMF_END` event are triggered. 

The decoding part is based on the work of @philippbachmann08 (sa. https://github.com/philippbachmann08/baresip/tree/feat/inband-dtmf-detection)

Within the PR some todos are written. Can someone provide some feedback about my questions?

I would also like to remove the `in_band_dtmf_send` function, by using the `call_send_digit` function. In this way, other modules like the GTK frontend can also use in-band DTMF tones. 

As far as i see, there is no option to inject an user handler or similar, to provide a module based implementation? 
Should i implement there a user handler? Are there already some ideas, how to do this?


modules/in_band_dtmf/in_band_dtmf.c
```diff
@@ -0,0 +1,329 @@
...
+	.dech    = decode
+};
+
+
+/**
+ * \struct cmdv
+ * The commands for this module.
+ * in_band_dtmf_send expects a single parameter.
+ *	- A string that will be splitted into single characters.
+ *	  Each valid DTMF character will be send as in-band DTMF tone.
```
cspiel1, line 264: typo: `sent`

marcel-behlau-elfin: Fixed by 93f80aa1

modules/in_band_dtmf/in_band_dtmf.c
```diff
@@ -0,0 +1,329 @@
...
+		return EINVAL;
+
+	st = mem_zalloc(sizeof(*st), enc_destructor);
+	if (!st)
+		return ENOMEM;
+
+	st->mb = mbuf_alloc(0);
+	if (!st->mb) {
+		mem_deref(st);
+		err = ENOMEM;
```
sreimers, line 106: You can simply return ENOMEM here and avoid the else part.

modules/in_band_dtmf/in_band_dtmf.c
```diff
@@ -0,0 +1,329 @@
...
+static void in_band_dtmf_dec_handler(char digit, void *arg)
+{
+	(void)arg;
+	char key_str[2];
+	struct in_band_dtmf_filt_dec *st;
+
+	key_str[0] = digit;
+	key_str[1] = '\0';
+
+	if (!list_count(&decs)) {
```
alfredh, line 52: It is more efficient to use `list_isempty` here.

modules/in_band_dtmf/in_band_dtmf.c
```diff
@@ -0,0 +1,329 @@
...
+
+
+static int encode(struct aufilt_enc_st *aufilt_enc_st, struct auframe *af)
+{
+	struct in_band_dtmf_filt_enc *st = (void *)aufilt_enc_st;
+	uint16_t* data = af->sampv;
+	uint16_t i;
+
+	if (mbuf_get_left(st->mb)) {
+		af->fmt = AUFMT_S16LE; /* TODO: Take care about format? */
```
alfredh, line 125: You can add checks to only support AUFMT_S16LE in this module.
The format conversion should be done by other modules, e.g. `auconv` ...

modules/in_band_dtmf/in_band_dtmf.c
```diff
@@ -0,0 +1,329 @@
...
+
+
+static int decode(struct aufilt_dec_st *st, struct auframe *af)
+{
+	struct in_band_dtmf_filt_dec *sf = (struct in_band_dtmf_filt_dec *)st;
+
+	if (!st || !af)
+		return EINVAL;
+
+	/* TODO: Take care of float format? */
```
alfredh, line 176: check that input frame is S16LE

modules/in_band_dtmf/in_band_dtmf.c
marcel-behlau-elfin: Fixed by ffe1ed65

modules/in_band_dtmf/in_band_dtmf.c
marcel-behlau-elfin: Fixed by 5d151b85

modules/in_band_dtmf/in_band_dtmf.c
marcel-behlau-elfin: Fixed by 2d1cc6d6

modules/in_band_dtmf/in_band_dtmf.c
marcel-behlau-elfin: Fixed by 2d1cc6d6

modules/in_band_dtmf/in_band_dtmf.c
```diff
@@ -0,0 +1,337 @@
...
+struct in_band_dtmf_filt_enc {
+	struct aufilt_enc_st af;  /* inheritance */
+	struct mbuf *mb;
+	unsigned srate;
+	struct le le_priv;
+};
+
+
+static struct list encs;
+static struct list decs;
```
alfredh, line 40: are these needed ?

modules/in_band_dtmf/in_band_dtmf.c
```diff
@@ -0,0 +1,337 @@
...
+	key_str[1] = '\0';
+
+	if (!list_count(&decs)) {
+		warning("in_band_dtmf: no active call\n");
+		return;
+	}
+
+	st = decs.head->data;
+
+	ua_event(st->ua, UA_EVENT_CALL_DTMF_START, st->call, "%s", key_str);
```
alfredh, line 59: Perhaps add some delay between these two events ?

modules/in_band_dtmf/in_band_dtmf.c
marcel-behlau-elfin: I need to pass some information to the related functions (`in_band_dtmf_send` and `in_band_dtmf_dec_handler`). The `in_band_dtmf_dec_handler` provides the user argument `arg` which could be an option. But i also need to pass the current `ua` and `call` argument, which i get within `ua_event_handler`.

I'm not sure if the list approach is the most easiest implementation. I copied the idea from `modules/mixausrc/mixausrc.c`.  I guess i could also invent an active call flag to my structs, and use then two static variables and omit the list.

Which way do you prefer?

modules/in_band_dtmf/in_band_dtmf.c
marcel-behlau-elfin: Fixed by bc677829

alfredh:
please take a look at the pointer to `struct audio` which is the parent audio session:

```c
typedef int (aufilt_encupd_h)(struct aufilt_enc_st **stp, void **ctx,
			      const struct aufilt *af, struct aufilt_prm *prm,
			      const struct audio *au);
```


it might be possible to save this pointer to your objects, and then use it to forward incoming
DTMF events to core/audio.

if missing, new functions may be added to audio.c


Here is the matching handling for incoming telephone events:


```
typedef void (audio_event_h)(int key, bool end, void *arg);
```

marcel-behlau-elfin:
I applied changes to `audio.c` with b03b43c1 and used the new interface with b29a231c

modules/in_band_dtmf/in_band_dtmf.c
```diff
@@ -0,0 +1,324 @@
...
+	struct in_band_dtmf_filt_dec *st;
+
+	if (list_isempty(&decs)) {
+		warning("in_band_dtmf: no active call\n");
+		return;
+	}
+
+	st = decs.head->data;
+
+	st->au->eventh(digit, true, st->au->arg);
```
alfredh, line 55: Perhaps it is better to add a wrapper function for this in audio.c ?

modules/in_band_dtmf/in_band_dtmf.c
alfredh: Is the `decs` state still needed ?

modules/in_band_dtmf/in_band_dtmf.c
marcel-behlau-elfin: I introduced now a `audio_get_event_handler` and also a `audio_get_argument`. 

This could be replaced by a `audio_call_event_handler(struct audio *au, int key, bool end)`.
In this case, the `au->arg` argument would be used for `au->eventh` arg argument.
But i'm not aware about the meaning of this argument. Is this every time the `au->arg` or can it be different?

modules/in_band_dtmf/in_band_dtmf.c
marcel-behlau-elfin: Yes, since the dec also manage the `struct tmr tmr_dtmf_end`, used to delay the dtmf end event.

src/audio.c
```diff
@@ -511,6 +511,18 @@ bool audio_txtelev_empty(const struct audio *au)
...
+  return au->eventh;
+}
+
+
+void* audio_get_argument(const struct audio *au)
+{
+  return au->arg;
+}
+
+
```
alfredh, line 525: perhaps this is better:

```c
void audio_call_telev_handler(struct audio *au, int key, bool end)
{
	if (!au)
		return;

	if (au->eventh)
		au->eventh(key, end, au->arg);
}
```

marcel-behlau-elfin: I introduced `audio_call_telev_handler`

modules/in_band_dtmf/in_band_dtmf.c
```diff
@@ -0,0 +1,323 @@
...
+	struct le le_priv;
+};
+
+
+static struct list encs;
+static struct list decs;
+
+
+static void dtmfend_handler(void *arg)
+{
```
alfredh, line 43: Please use `arg` to pass the state: `struct in_band_dtmf_filt_dec`

The current digit can be saved on the state and read here.

modules/in_band_dtmf/in_band_dtmf.c
```diff
@@ -0,0 +1,323 @@
...
+	st = decs.head->data;
+
+	audio_call_telev_handler(st->au, digit, true);
+}
+
+
+static void in_band_dtmf_dec_handler(char digit, void *arg)
+{
+	(void)arg;
+	struct in_band_dtmf_filt_dec *st;
```
alfredh, line 61: Please get the state from the `void *arg` variable.

Then the decs list is not needed anymore.

modules/in_band_dtmf/in_band_dtmf.c
```diff
@@ -0,0 +1,323 @@
...
+
+	if (list_isempty(&decs)) {
+		warning("in_band_dtmf: no active call\n");
+		return;
+	}
+
+	st = decs.head->data;
+
+	tmr_start(&st->tmr_dtmf_end, 50,
+		dtmfend_handler, (void*)(uintptr_t)digit);
```
alfredh, line 71: Please use `st` as the last argument.

modules/in_band_dtmf/in_band_dtmf.c
```diff
@@ -0,0 +1,323 @@
...
+
+	if (!stp || !prm)
+		return EINVAL;
+
+	st = mem_zalloc(sizeof(*st), enc_destructor);
+	if (!st)
+		return ENOMEM;
+
+	st->mb = mbuf_alloc(0);
+	if (!st->mb) {
```
alfredh, line 116: here is a memory leak. Declare a variable `int err`, if mbuf_alloc fails set err to ENOMEM and goto out.

alfredh:
Perhaps you can split the PR into two Pull Requests, one for decode and one for encode ?

modules/in_band_dtmf/in_band_dtmf.c
marcel-behlau-elfin: fixed by 1b0cf18c

modules/in_band_dtmf/in_band_dtmf.c
marcel-behlau-elfin: How to manage the memory then? Should `decs` become a `dec` or is there a way to avoid the global variable? 

```
 static struct list encs;
-static struct list decs;
+static struct dec;
```

alfredh:
Please take a look at the plc module:

https://github.com/baresip/baresip/blob/main/modules/plc/plc.c

The state is created in the update function and passed back to the core for ownership.
The core will delete the object when the session is done. There is no need for you to also
track the state in this module.

marcel-behlau-elfin:
> Please take a look at the plc module:
> 
> https://github.com/baresip/baresip/blob/main/modules/plc/plc.c
> 
> The state is created in the update function and passed back to the core for ownership. The core will delete the object when the session is done. There is no need for you to also track the state in this module.

I removed the decs list by 72d9c297



> Perhaps you can split the PR into two Pull Requests, one for decode and one for encode ?

I don't see that this approach is pretty forward.
Currently the implementation of encoding and decoding can be tested in a loop by starting 2 baresip instances. Splitting the functions in two PR will this complicate.

Will splitting up hurry up the process?

modules/in_band_dtmf/in_band_dtmf.c
marcel-behlau-elfin: fixed by 72d9c297

modules/in_band_dtmf/in_band_dtmf.c
marcel-behlau-elfin: fixed by https://github.com/baresip/baresip/commit/72d9c297ce45b8bfece361889761fd05919be46b

modules/in_band_dtmf/in_band_dtmf.c
marcel-behlau-elfin: fixed by https://github.com/baresip/baresip/commit/72d9c297ce45b8bfece361889761fd05919be46b

alfredh:
this one look good now.

@sreimers and @cspiel1 please review and merge if you agree

cspiel1:
Please, could you explain what is the difference to current DTMF solution
- `stream_pt_handler()` for incoming DTMF and
- `audio_send_digit()` for sending DTMF ?

alfredh:
in traditional SIP there are three ways to send DTMF:

1. Telephone event
2. SIP INFO
3. In-band (in the audio stream)


Point 1 and 2 is already implemented. This PR adds support for point 3.

PR #3077, opened 2024-07-03
test: add TLS SNI selftest

maximilianfridrich:
Mentioned in https://github.com/baresip/re/pull/1136


test/call.c
```diff
@@ -3251,3 +3259,142 @@ int test_call_srtp_tx_rekey(void)
...
+
+	ASSERT_EQ(0, fix.c.n_incoming);
+	ASSERT_EQ(0, fix.c.n_established);
+	ASSERT_EQ(0, fix.c.n_closed);
+	ASSERT_EQ(0, fix.c.close_scode);
+
+	/* 2nd test. CA set. Call from A to C. TLS handshake must fail because
+	 * certificate of C is selected which is from an unknown CA. */
+	re_snprintf(s, sizeof(s), "%s/sni/root-ca.pem", dp);
+	err = tls_add_cafile_path(uag_tls(), s, NULL);
```
cspiel1, line 3340: We may also set `sip_certificate` to a wrong certificate in order to check if the account certificate is chosen by SNI.

maximilianfridrich: Good idea. Done.
The global `sip_certificate` is set at the beginning of the test. The final call must still succeed because SNI chooses the correct UA certificate and does not choose the global certificate.

test/call.c
```diff
@@ -3251,3 +3259,147 @@ int test_call_srtp_tx_rekey(void)
...
+{
+	int err = 0;
+	struct fixture fix, *f = &fix;
+	struct dns_server *dns_srv = NULL;
+	struct dnsc *dnsc = NULL;
+	char buri_tls[256], curi_tls[256];
+	const char *dp = test_datapath();
+	char s[256];
+
+	/* Set global certificate. */
```
cspiel1, line 3275: Maybe clearer:
/* Wrong certificate as global setting */

maximilianfridrich: Yes, done.

cspiel1:
Looks good. Should we merge this?

PR #3079, opened 2024-07-03
srtp: Fix base64 key decoding for AEAD_AES_256_GCM

weili-jiang:
The base64 decode routine requires a buffer size that is somewhat larger than the expected key size by olen < 3 * (ilen/4). When given just the expected key size for AEAD_AES_256_GCM as the base64 output length, EOVERFLOW is returned.

This was previously discussed in https://github.com/baresip/baresip/pull/790 when AEAD_AES_256_GCM support was added, but has regressed due to #2975.

Found the issue and tested the fix by making calls both to other Baresip instances and to FreeSWITCH.


sreimers:
Thanks!

PR #3080, opened 2024-07-11
add filter_registrar option

maximilianfridrich:
If the new config option `filter_registrar` is set to some transport (default: `""`), any incoming request of that transport is verified to have originated from a registered server's IP address (only makes sense for registered accounts `regint>0`). Further, no listening socket is opened for the connection-oriented transport protocols listed in `filter_registrar` (i.e. a UDP socket is always opened if UDP is configured). BareSIP's default behavior stays the exact same.

Requires:
- https://github.com/baresip/re/pull/1160

### DIFF
--- a/src/reg.c
+++ b/src/reg.c
@@ -15,6 +15,7 @@ struct reg {
 	struct sipreg *sipreg;       /**< SIP Register client                */
 	int id;                      /**< Registration ID (for SIP outbound) */
 	int regint;                  /**< Registration interval              */
+	struct sa paddr;             /**< Peer Address                       */
 
 	/* status: */
 	uint16_t scode;              /**< Registration status code           */
@@ -131,6 +132,8 @@ static void register_handler(int err, const struct sip_msg *msg, void *arg)
 	if (200 <= msg->scode && msg->scode <= 299) {
 
 		uint32_t n_bindings;
+
+		reg->paddr = msg->src;
 
 		n_bindings = sip_msg_hdr_count(msg, SIP_HDR_CONTACT);
 		reg->af    = sipmsg_af(msg);
@@ -421,4 +424,10 @@ const struct sa *reg_laddr(const struct reg *reg)
 		return NULL;
 
 	return sipreg_laddr(reg->sipreg);
+}
+
+
+const struct sa *reg_paddr(const struct reg *reg)
+{
+	return reg ? &reg->paddr : NULL;
 }
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -15,6 +15,7 @@ sip_verify_server	yes
 #sip_verify_client	no
 #sip_tls_resumption	all		# none, ids, tickets
 sip_tos			160 # See TOS fields!
+#filter_registrar   udp,tcp
 
 ## TOS fields ##
 #    7     6     5     4     3     2     1     0
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -362,6 +362,7 @@ struct config_sip {
 	bool verify_client;     /**< Enable SIP TLS verify client   */
 	enum tls_resume_mode tls_resume; /** TLS resumption mode    */
 	uint8_t tos;            /**< Type-of-Service for SIP        */
+	uint32_t reg_filt;	/**< Registrar filter transport mask*/
 };
 
 /** Call config */
@@ -945,6 +946,7 @@ int ua_set_autoanswer_value(struct ua *ua, const char *value);
 void ua_add_extension(struct ua *ua, const char *extension);
 void ua_remove_extension(struct ua *ua, const char *extension);
 bool ua_req_allowed(const struct ua *ua, const struct sip_msg *msg);
+bool ua_req_check_origin(const struct ua *ua, const struct sip_msg *msg);
 
 
 /* One instance */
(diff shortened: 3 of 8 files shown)


juha-h:
If UDP is used, nothing can be verified.  Also, it would make more sense to use TLS and verify that server's certificate matches.

juha-h:
> The source IP address of the request is still set and we can verify it. I'm not sure what you mean.

Yes, but since in UDP there is no connection setup, sender can put whatever source address in the IP packet.  So baresip can check that the source address of the IP packet matches, but cannot verify where it was sent from.

maximilianfridrich:
I see, you're right. This of course does not protect against IP address spoofing. We all know that TLS is the solution, but we still have a feature request of a customer who wants to verify IP addresses and maybe other people would like that too. They just have to be aware of the limitations of this approach, but I believe it could still be a useful feature for some people.

juha-h:
Add clearly to the documentation that in case of UDP this setting cannot verify that a SIP request originated from the server, but only that source IP address of the request matched.

juha-h:
Another security setting that  could be added is "Send all in-dialog requests and replies to the outbound proxy".

src/reg.c
```diff
@@ -422,3 +422,12 @@ const struct sa *reg_laddr(const struct reg *reg)
...
 	return sipreg_laddr(reg->sipreg);
 }
+
+
+const struct sa *reg_paddr(const struct reg *reg)
+{
+	if (!reg)
+		return NULL;
+
+	return sipreg_paddr(reg->sipreg);
```
alfredh, line 432: Probably better to save `paddr` locally. Then there is no changes to libre needed ...

docs/examples/accounts
```diff
@@ -43,6 +43,7 @@
 #    ;stunserver=stun:[user:pass]@host[:port]  # see RFC 3986: Use of the format "user:password" in the userinfo field is deprecated.
 #    ;tcpsrcport=6060
 #    ;inreq_allowed={yes, no}  # default: yes
+#    ;server_only={yes, no}  # default: no
```
alfredh, line 46: TCP and TLS already have a connection to the server, which is difficult to spoof.

What happens if you are using SIP outbound with 2 edge proxies and load balancing,
will you accept incoming request from both ?

juha-h:
Commonly outbound proxy or proxies are configured using domain name of the host(s) in the URI(s). Will this PR make DNS query to find their current IP addresses?

juha-h:
Starts to looks to me that baresip account config is not a good way to solve the problem.  Better to use firewall configuration in the host.

src/reg.c
maximilianfridrich: You're right, that's much better. Done.

docs/examples/accounts
maximilianfridrich: You're right, now the peer address is stored in the `reg` struct in BareSIP and now all of the register clients are checked (i.e. all proxies defined in outbound will be allowed). Done.

maximilianfridrich:
> Commonly outbound proxy or proxies are configured using domain name of the host(s) in the URI(s). Will this PR make DNS query to find their current IP addresses?

That is something I considered, however I believe it is quite unlikely that the case occurs where we receive a REGISTER response from some IP A, and then receive an incoming request from the same "server" (i.e. URI) with a different IP. It's not impossible of course, but I think it could be enough to document this limitation if anyone wants to use this feature.

> Starts to looks to me that baresip account config is not a good way to solve the problem. Better to use firewall configuration in the host.

That's a valid point. If the maintainers agree and don't want to see this feature in BareSIP then we will have to accept that.

maximilianfridrich:
Grandstream and Snom for example have such a setting:
- "filter_registrar": https://service.snom.com/display/wiki/filter_registrar
- "Only Accept SIP Requests from Known Servers": https://documentation.grandstream.com/knowledge-base/gvc32xx-series-security-guidelines/

juha-h:
When baresip uses TCP or TLS to connect to outbound proxy/proxies, baresip should accept SIP requests only when they come via those connections.  That is, there is no reason why baresip should accept incoming TCP or TLS connections from anywhere.  And in case of UDP, source IP address does not prove anything as I have mentioned earlier.

alfredh:
The SIP stack has a TCP server that is listening for incoming TCP/TLS and WS/WSS connections.
In the context of this PR you might want to disable this server to increase the perceived security.

Reusing naming from other products is nice for people already familiar with Snom products:

Example:

```
filter_registrar    udp
```

```
filter_registrar    udp,tcp
```

maximilianfridrich:
Sounds good, thank you for the feedback. I will start working on that.

cspiel1:
This is ready for merge in our opinion. Any further suggestions?

alfredh:
this look good. Please go ahead and merge it.

juha-h:
Is there now the possibility to prevent all incoming connection setups when outbound proxy is defined using a connection oriented transport?

maximilianfridrich:
Yes, @juha-h. For example with the following config:

```
sip_transports      tcp,tls
filter_registrar    tcp,tls
```

No listening ports will be opened and only outgoing TCP/TLS connections will work. The proxy/registrar will have to use the existing TCP/TLS connection.

juha-h:
> Yes, @juha-h. For example with the following config:
> 
> ```
> sip_transports      tcp,tls
> filter_registrar    tcp,tls
> ```
> 
> No listening ports will be opened and only outgoing TCP/TLS connections will work. The proxy/registrar will have to use the existing TCP/TLS connection.

Very good.

Issue #3064, opened 2024-06-11
impl_ext_end_proxy called from wrong context, check thread and locking: Operation not permitted


robert-scheck:
After updating from Red Hat Enterprise Linux 9.3 to 9.4, which contained an update of PipeWire from 0.3.67 to 1.0.1, I see the following failures:

```
$ baresip 
baresip v3.12.1 Copyright (C) 2010 - 2024 Alfred E. Heggestad et al.
Local network addresses:
 enp0s13f0u1u4u4:  192.0.2.111
  wlp0s20f3:  198.51.100.184
 enp0s13f0u1u4u4:  2001:db8:1:21:4860:3096:f7ab:aaae
 enp0s13f0u1u4u4:  fe80::6672:8f70:a634:b29d
  wlp0s20f3:  2001:db8:1:24:1024:f3d2:8074:eaa4
  wlp0s20f3:  fe80::c2be:122c:6a87:77dc
ua: adding SIP CA file: /etc/pki/tls/certs/ca-bundle.crt
ua: adding SIP CA path: 
aucodec: G722/16000/1
aucodec: PCMU/8000/1
aucodec: PCMA/8000/1
aufilt: auconv
aufilt: auresamp
pipewire: headers 1.0.1 library 1.0.1 
*** impl_ext_end_proxy called from wrong context, check thread and locking: Operation not permitted
*** impl_ext_end_proxy called from wrong context, check thread and locking: Operation not permitted
pipewire: connected to pipewire
auplay: pipewire
ausrc: pipewire
*** impl_ext_end_proxy called from wrong context, check thread and locking: Operation not permitted
medianat: stun
medianat: turn
medianat: ice
Populated 1 account
Populated 34 contacts
aufilt: gtk_vumeter
Populated 3 audio codecs
Populated 3 audio filters
Populated 0 video codecs
Populated 0 video filters
baresip is ready.
tux@sip.example.net: selected for tux-0x556188e4f2e0
tux@sip.example.net: (prio 0) {0/UDP/v4} 200 OK (STARFACE PBX) [1 binding]
All 1 useragent registered successfully! (4 ms)
gtk_menu starting
```

I can not say with absolute certainty that it's related to PipeWire, but it seems to be a fair match. Note that baresip and libre have been already since 3.12.0 built against RHEL 9.4 in the Fedora buildsystem, so it's not simply a missing rebuild.

Just let me know if you need more information, and depending on what you need…maybe also how to gather these data.

[...]

PR #3084, opened 2024-07-22
pipewire: lock loop while starting and registry scan

cspiel1:
Fixes: https://github.com/baresip/baresip/issues/3064

### DIFF
--- a/modules/pipewire/pipewire.c
+++ b/modules/pipewire/pipewire.c
@@ -183,6 +183,7 @@ static struct pw_stat *pw_stat_alloc(void)
 	if (!pw->loop)
 		goto errout;
 
+	pw_thread_loop_lock(pw->loop);
 	err = pw_thread_loop_start(pw->loop);
 	if (err)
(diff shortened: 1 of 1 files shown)


cspiel1:
@HectorGarciaMiguel please could you test this solution with pipewire >= 0.3.68?

HectorGarciaMiguel:
> @HectorGarciaMiguel please could you test this solution with pipewire >= 0.3.68?
> 
> Is there still a "check thread and locking: Operation not permitted"?

Ok I'll try, let me have a few minutes

HectorGarciaMiguel:
> @HectorGarciaMiguel please could you test this solution with pipewire >= 0.3.68?
> 
> Is there still a "check thread and locking: Operation not permitted"?

Ok, I have also tested with the 1.2.0 pipewire version, the last one, and the message is no longer displayed.

cspiel1:
Thanks! Any further suggestions? Should we merge this?

Issue #3061, opened 2024-06-08
app has no way to prevent 180 Ringing response


juha-h:
I don't know what happened, but I didn't anymore find any PR that would address the issue where the app has no way to reject INVITE request without `180 Ringing` response from  baresip core.

cspiel1:
I still have in on my TODO list. Don't we have already an issue for this?

juha-h:
For sure there was no open issue and I didn't find a closed one either.

alfredh:
this logic is done in `call_accept`:

```c
	err = sipsess_accept(&call->sess, sess_sock, msg, 180, "Ringing",
			     account_rel100_mode(call->acc),
			     ua_cuser(call->ua), "application/sdp", NULL,
			     auth_handler, call->acc, true,
			     sipsess_offer_handler, sipsess_answer_handler,
			     sipsess_estab_handler, sipsess_info_handler,
			     call->acc->refer ? sipsess_refer_handler : NULL,
			     sipsess_close_handler,
			     call, "Allow: %H\r\n%H",
			     ua_print_allowed, call->ua,
			     ua_print_require, call->ua);
```

You can try to send in "0" instead of "180" and see if the sending of 180 is skipped...

cspiel1:
I am on vacation the next two weeks. The plan is to proceed after it with this branch
https://github.com/cspiel1/baresip/tree/bevent_complete_refactoring

juha-h:
I changed 180 to 0 and when call came in, baresip gave `SIP/2.0 500 Call Error` response.

juha-h:
@alfredh Turned out that `500 Call Error` came because of this test 
```
if (!sessp || !sock || !msg || scode < 101 || scode > 299 |
```
in front of `sipsess_accept`.

I made this change in accept.c:
```
diff --git a/src/sipsess/accept.c b/src/sipsess/accept.c
index 1f506dc..4b28675 100644
--- a/src/sipsess/accept.c
+++ b/src/sipsess/accept.c
@@ -74,7 +74,7 @@ int sipsess_accept(struct sipsess **sessp, struct sipsess_sock *sock,
    va_list ap;
    int err;
 
-   if (!sessp || !sock || !msg || scode < 101 || scode > 299 ||
+   if (!sessp || !sock || !msg || scode < 100 || scode > 299 ||
        !cuser || !ctype)
        return EINVAL;
 
[...]

PR #3088, opened 2024-07-24
Moved 180 Ringing response from core to app (menu module)

juha-h:
Depends on https://github.com/baresip/re/pull/1167
Solves https://github.com/baresip/baresip/issues/3061

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -216,6 +216,7 @@ typedef void (call_list_h)(struct call *call, void *arg);
 
 int  call_connect(struct call *call, const struct pl *paddr);
 int  call_answer(struct call *call, uint16_t scode, enum vidmode vmode);
+int  call_ringing(struct call *call);
 int  call_progress_dir(struct call *call,
 		       enum sdp_dir adir, enum sdp_dir vdir);
 int  call_progress(struct call *call);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -216,7 +216,7 @@ typedef void (call_list_h)(struct call *call, void *arg);
 
 int  call_connect(struct call *call, const struct pl *paddr);
 int  call_answer(struct call *call, uint16_t scode, enum vidmode vmode);
-int  call_ringing(struct call *call);
+int  call_ringing(const struct call *call);
 int  call_progress_dir(struct call *call,
 		       enum sdp_dir adir, enum sdp_dir vdir);
 int  call_progress(struct call *call);
--- a/src/call.c
+++ b/src/call.c
@@ -2233,7 +2233,7 @@ int call_accept(struct call *call, struct sipsess_sock *sess_sock,
 		mem_deref(rid);
 	}
 
-	err = sipsess_accept(&call->sess, sess_sock, msg, 180, "Trying",
+	err = sipsess_accept(&call->sess, sess_sock, msg, 100, "Trying",
 			     account_rel100_mode(call->acc),
 			     ua_cuser(call->ua), "application/sdp", NULL,
 			     auth_handler, call->acc, true,
--- a/src/call.c
+++ b/src/call.c
@@ -2137,7 +2137,7 @@ static bool valid_addressfamily(struct call *call, const struct stream *strm)
  *
  * @return 0 if success, otherwise errorcode
  */
-int call_ringing(struct call *call)
+int call_ringing(const struct call *call)
 {
 	if (!call)
 	   return EINVAL;
--- a/src/call.c
+++ b/src/call.c
@@ -2233,7 +2233,7 @@ int call_accept(struct call *call, struct sipsess_sock *sess_sock,
 		mem_deref(rid);
 	}
 
-	err = sipsess_accept(&call->sess, sess_sock, msg, 100, "Trying",
+	err = sipsess_accept(&call->sess, sess_sock, msg, 0, "",
 			     account_rel100_mode(call->acc),
 			     ua_cuser(call->ua), "application/sdp", NULL,
 			     auth_handler, call->acc, true,
--- a/src/call.c
+++ b/src/call.c
@@ -2130,6 +2130,27 @@ static bool valid_addressfamily(struct call *call, const struct stream *strm)
 }
 
 
+/**
+ * Send a SIP 180 Ringing response
+ *
+ * @param call Call to send ringing response
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int call_ringing(struct call *call)
+{
+	if (!call)
+	   return EINVAL;
+
+	return sipsess_progress(call->sess, 180, "Ringing",
+				account_rel100_mode(call->acc),
+				NULL,
+				"Allow: %H\r\n%H",
+				ua_print_allowed, call->ua,
+				ua_print_require, call->ua);
+}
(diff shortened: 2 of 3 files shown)


src/call.c
```diff
@@ -2212,17 +2233,16 @@ int call_accept(struct call *call, struct sipsess_sock *sess_sock,
 		mem_deref(rid);
 	}
 
-	err = sipsess_accept(&call->sess, sess_sock, msg, 180, "Ringing",
+	err = sipsess_accept(&call->sess, sess_sock, msg, 180, "Trying",
```
maximilianfridrich, line 2236: 180 Trying? That seems unusual. What's the reasoning behind this?

src/call.c
```diff
@@ -2212,17 +2233,16 @@ int call_accept(struct call *call, struct sipsess_sock *sess_sock,
...
 			     account_rel100_mode(call->acc),
 			     ua_cuser(call->ua), "application/sdp", NULL,
 			     auth_handler, call->acc, true,
 			     sipsess_offer_handler, sipsess_answer_handler,
 			     sipsess_estab_handler, sipsess_info_handler,
 			     call->acc->refer ? sipsess_refer_handler : NULL,
 			     sipsess_close_handler,
-			     call, "Allow: %H\r\n%H",
```
maximilianfridrich, line 2223: What's the reasoning behind the removal of the Allow and Require header?

According to [RFC 3261 section 20](https://www.rfc-editor.org/rfc/rfc3261#section-20) they are optional in INVITE responses.

src/call.c
cspiel1: I guess @juha-h wants to reply with 100 Trying and put the 180 Ringing into the app/module menu. But what I remember is that 100 Trying is sent already maybe internally from `re`. We have to be sure not to send it twice.

juha-h:
There was typo in Trying response code.  Fixed now.  In my tests, 100 Trying is sent only once. I left Allow and Require headers out from Trying, since I didn't see any reason why they should be in Trying response.

juha-h:
I don't know why selftest is hungs.

sreimers:
The PR Title has to match exactly the corresponding re PR, otherwise main is used.

juha-h:
Sebastian Reimers writes:

> The PR Title has to match exactly the corresponding re PR, otherwise main is used.

OK, I renamed the re PR.  Would it make more sense that PR branch names
must match?

cspiel1:
### Alternative
A complete event refactoring
https://github.com/baresip/baresip/pull/3089

juha-h:
Still selftest fails.

juha-h:
> ### Alternative
> 
> A complete event refactoring #3089

That has been closed and I need to get the issue fixed.

cspiel1:
Yes, because I'll split it up into smaller PRs. This is the first step https://github.com/baresip/baresip/pull/3090.

If we merge this PR first, we'll have to revert it later.

include/baresip.h
```diff
@@ -216,6 +216,7 @@ typedef void (call_list_h)(struct call *call, void *arg);
 
 int  call_connect(struct call *call, const struct pl *paddr);
 int  call_answer(struct call *call, uint16_t scode, enum vidmode vmode);
+int  call_ringing(const struct call *call);
```
cspiel1, line 219: If we add an API function now, it will be difficult to remove it later after https://github.com/baresip/baresip/pull/3090 an following bevent PRs are done.

src/call.c
```diff
@@ -2212,17 +2233,16 @@ int call_accept(struct call *call, struct sipsess_sock *sess_sock,
...
-			     ua_print_allowed, call->ua,
-			     ua_print_require, call->ua);
+			     call, NULL);
```
cspiel1, line 2244: This changes the default behavior of the baresip core. Whereas, the bevent solution will keep the default behavior and allow the application to reject an incoming call.

juha-h:
It changes buggy behavior which is fine.

juha-h:
Then revert later. This has taken too long time.

juha-h:
No need to remove later. Just don't use it later.

juha-h:
It is a bug in baresip core when it sends 180 response.  It should not send any response (not even 100).  The response needs to be at the control of the app and can be 100 Trying, 180 Ringing, 301/302 Moved Permanently/Temporarily, 403 Forbidden, 403 Forbidden, 486 Busy, etc.  Fixing of a bug does not need to be backwards compatible with the bug.  Regarding preventing also 100 Trying, that could be done, for example,  by scode value 0 in sipsess_accept.

cspiel1:
I discussed with @maximilianfridrich. We agree that both
- 180 Ringing and
- 183 Session Progress
better should be sent from menu (or other app).

The bevent refactoring may be useful additionally for cleanup of the core and for future call decline use cases, like blacklist. So, not an alternative but a different improvement.

We should not break the default behavior.
How about a config setting `call_accept`?
```c
     if (config.call_accept)
          call_accept(call, );
```
Set `config.call_accept = true` by default.

Can we let `call_accept()` like it is in main, no "100 Trying" and no need for a new function `call_ringing()`.

cspiel1:
You'll could add like already discussed a new event `UA_EVENT_SIPSESS_CONN`. Thus finally:
```c
if (config.call_accept)
    call_accept(call, );
else
   ua_event(ua, UA_EVENT_SIPSESS_CONN, ...);
```

In menu event handler then call `call_accept(call)`

juha-h:
I don't know where call_accept should be called from.  The solution of this PR is looks much simpler and cleaner to me.  Also, default behavior should be that no reply is sent automatically from the core.

I'll update this PR so that no 100 Trying is sent from the core.

juha-h:
Now now 100 Trying is sent from the core.

juha-h:
In my opinion this PR now produces a very clean and simple fix to the issue.  Thanks to @alfredh for the call_ringing hint.

src/call.c
cspiel1: Is it okay to allocate a SIP session without sending any reply? If so, is it possible that we use `sipsess_alloc()` here directly?

An API function called `call_accept()` should send a reply in my opinion.

cspiel1:
I'll try a minimal PR based on this https://github.com/baresip/baresip/pull/3088#issuecomment-2252147838

cspiel1:
With current `ua_event()` API it is not possible to pass `sip_msg` to the event handler. Thus it is not possible to make `call_accept()` public.

So I suggest to proceed with https://github.com/baresip/baresip/pull/3090

juha-h:
> With current `ua_event()` API it is not possible to pass `sip_msg` to the event handler. Thus it is not possible to make `call_accept()` public.
> 
> So I suggest to proceed with #3090

Fine with me as long as by default `180 Ringing` response is not sent and it does not take many weeks to get the PR done and the bug fixed.

juha-h:
Christian Spielberger writes:

> I'll try a minimal PR based on this https://github.com/baresip/baresip/pull/3088#issuecomment-2252147838

OK, if default value of config.call_accept == false, i.e., that no 180
Ringing is sent by default.

juha-h:
Replaced by https://github.com/baresip/baresip/pull/3123.

PR #3089, opened 2024-07-25
core: refactor events

cspiel1:


cspiel1:
@alfredh suggested

> I would like to suggest that we split this up into logical blocks (many PRs).
> 
> 1. add core bevent
> 2. add backwards wrapper
> 3. update function calls
> 4. add new features like DND
> 
> For example a small PR can be to just rename event.c to bevent.c
> 
> Is there a need to update all the calls that are sending the event ? It would be nice if the PR is small and easy to read.

I'll try my best.

cspiel1:
Step 1 of this PR: add core bevent
https://github.com/baresip/baresip/pull/3090

PR #3090, opened 2024-07-25
bevent: add core bevent API

cspiel1:
Step one of https://github.com/baresip/baresip/pull/3089 which replaces https://github.com/baresip/baresip/pull/3088

- adds bevent core
- adds tests for bevent core

### Full description copied from https://github.com/baresip/baresip/pull/3089

The new event system allows e.g. an application module to return `ENOENT` in order to stop processing of incoming calls.

```c
	err = bevent_sip_msg_emit(UA_EVENT_SIPSESS_CONN, msg, "");
	if (err == ENOENT)
		return;
```

Based on https://github.com/baresip/baresip/pull/3018, where the discussion can be found.

- allows to reject a call before the call object is allocated
- allows further simplification of baresip core by moving some of the 4xx replies to app modules (menu)
- e.g. support DnD feature in module menu
- allows simple extension of events without breaking the API

### Steps for bevent
- [x] add core bevent
- [ ] add event class filter mask to `bevent_register()`
- [ ] add backwards wrapper
- [ ] update function calls
- [ ] provide alternative to https://github.com/baresip/baresip/pull/3088
- [ ] move DnD feature from core to menu


cspiel1:
Because the new API is not used so far it will become difficult to reach the coverage.

include/baresip.h
```diff
@@ -1654,14 +1660,31 @@ void module_app_unload(void);
...
+int bevent_sip_msg_emit(enum ua_event ev, const struct sip_msg *msg,
+			const char *fmt, ...);
 void module_event(const char *module, const char *event, struct ua *ua,
 		struct call *call, const char *fmt, ...);
 const char  *uag_event_str(enum ua_event ev);
+struct call    *bevent_get_call(const struct bevent *event);
+struct ua      *bevent_get_ua(const struct bevent *event);
+const struct sip_msg *bevent_get_msg(const struct bevent *event);
+void *bevent_get_apparg(const struct bevent *event);
+enum ua_event bevent_get_enum(const struct bevent *event);
```
maximilianfridrich, line 1684: Should we rename this to `bevent_get_type` ?

src/bevent.c
```diff
@@ -324,6 +615,57 @@ void uag_event_unregister(ua_event_h *h)
...
 
+/**
+ * Register an Event handler
+ *
+ * @param eh  Event handler
+ * @param arg Handler argument
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int  bevent_register(bevent_h *eh, void *arg)
```
maximilianfridrich, line 626: Would it make sense to allow an event handler to only register for a certain event class? E.g. an event handler is only interested in UA events, or only CALL events (or CLASS_UNDEFINED for all events) and the handler is only called for those event types.

src/bevent.c
```diff
@@ -410,6 +752,168 @@ void module_event(const char *module, const char *event, struct ua *ua,
...
+ * Emit an application event
+ *
+ * @param ev   User-agent event
+ * @param arg  Application specific argument (optional)
+ * @param fmt  Formatted arguments
+ * @param ...  Variable arguments
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int bevent_app_emit(enum ua_event ev, void *arg, const char *fmt, ...)
```
maximilianfridrich, line 810: What is the advantage to having different API functions for the different event classes? They all pass pointers as arguments anyways, wouldn't a single API function `bevent_emit` be enough?

src/bevent.c
maximilianfridrich: Or allow flags? E.g. `BEVENT_CLASS_UA | BEVENT_CLASS_CALL`

cspiel1: Good idea. Let's do this in a separate PR. Because `bevent_register()` currently is only used in the selftest, it is no problem to change the API function later. I add this to the TODO list  in the description.

src/bevent.c
cspiel1: Different event classes use different union fields. 
And the different emit functions have different parameters. This is the major improvement.

E.g. If the user wants to emit an application event, he doesn't want to take care about a call object.

cspiel1: Also it is very easy to add new event classes, without breaking the backwards compatibility.

maximilianfridrich: Okay, so the advantage I see is that if the API is used correctly, type checking is enforced and it is more difficult to incorrectly cast some void pointers to a wrong type. Sounds good to me.

cspiel1:
This is ready for merge. Any further suggestions?

alfredh:
looks good to me.. please feel free to merge it ...

cspiel1:
Thanks! Will proceed with further, hopefully smaller PRs regarding this.

PR #3095, opened 2024-07-31
sip/transp: add client certificate to all TLS transports

maximilianfridrich:
test/call: add account certificate selection test

Currently, the test fails (as expected). With the [fix in libre](https://github.com/baresip/re/pull/1173) the test is green.

Fix:

- https://github.com/baresip/re/pull/1173

### DIFF
--- a/test/call.c
+++ b/test/call.c
@@ -3385,7 +3385,7 @@ int test_call_sni(void)
 	ASSERT_EQ(1, fix.b.n_incoming);
 	ASSERT_EQ(1, fix.b.n_established);
 	ASSERT_EQ(1, fix.b.n_closed);
-	ASSERT_EQ(0, fix.a.close_scode);
+	ASSERT_EQ(0, fix.b.close_scode);
 
 	ASSERT_EQ(0, fix.c.n_incoming);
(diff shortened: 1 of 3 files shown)


test/call.c
```diff
@@ -3402,4 +3402,130 @@ int test_call_sni(void)
...
+
+	ASSERT_EQ(0, fix.a.n_incoming);
+	ASSERT_EQ(0, fix.a.n_established);
+	ASSERT_EQ(1, fix.a.n_closed);
+	ASSERT_EQ(0, fix.a.close_scode);
+
+	ASSERT_EQ(0, fix.b.n_incoming);
+	ASSERT_EQ(0, fix.b.n_established);
+	ASSERT_EQ(1, fix.b.n_closed);
+	ASSERT_EQ(0, fix.a.close_scode);
```
alfredh, line 3463: Is it correct to use `fix.a.close_code` here ? should it be `b` ?

maximilianfridrich: Thank you! Fixed.

alfredh:
Should this one be merged first, or the libre one first, or both at the same time ?

maximilianfridrich:
At the same time I'd say @alfredh.

PR #3096, opened 2024-07-31
uag: match transport and af when finding ua for msg

maximilianfridrich:
Currently, when finding a UA for an incoming msg, the transport protocol and address family only had to match for registrarless (regint=0) accounts.

Consider the case where `sip_transports  udp,tls` and there is a server account configured with `<sip:vuln@sip.example.com;transport=TLS>;regint=60`. Now BareSIP receives an incoming INVITE via UDP addressed to the account `vuln`. The account would be matched even though it is configured to use TLS.

Now, the transport protocol of the account (if configured) and the address family has to match, even if it is a server account (regint>0), which is the same behavior as it has always been for registrarless accounts.

### DIFF
--- a/src/uag.c
+++ b/src/uag.c
@@ -919,21 +919,19 @@ struct ua *uag_find_msg(const struct sip_msg *msg)
 		struct ua *ua = le->data;
 		struct account *acc = ua_account(ua);
 
-		if (!acc->regint) {
-			if (!uri_match_transport(&acc->luri, NULL, msg->tp))
-				continue;
+		if (!uri_match_transport(&acc->luri, NULL, msg->tp))
+			continue;
 
-			if (!uri_match_af(&acc->luri, &msg->uri))
-				continue;
-
-			if (!uaf && ua_catchall(ua))
-				uaf = ua;
-		}
+		if (!uri_match_af(&acc->luri, &msg->uri))
+			continue;
 
 		if (0 == pl_casecmp(cuser, &acc->luri.user)) {
 			ua_printf(ua, "account match for %r\n", cuser);
 			return ua;
 		}
+		if (!acc->regint && !uaf && ua_catchall(ua))
+			uaf = ua;
+
 	}
 
 	if (uaf)


cspiel1:
This is the match for registered accounts:
```c
	cuser = &msg->uri.user;
		struct ua *ua = le->data;

		if (0 == pl_strcasecmp(cuser, ua_local_cuser(ua))) {
			ua_printf(ua, "selected for %r\n", cuser);
			return ua;
		}
	}
```

This should always work because of the unique contact user. See ua.c:1165 !

maximilianfridrich:
> This is the match for registered accounts:

If the unique `cuser` is not matched, the loop below still runs and matches registered accounts by username.

juha-h:
If there is no `;transport` parameter in outbound proxy definition, and DNS query has resulted in, say, TLS transport connection to the outbound proxy, then that transport protocol should be the one that is required to match.

maximilianfridrich:
> If there is no ;transport parameter in outbound proxy definition

Then that check is ignored, no change of behavior in BareSIP.

> and DNS query has resulted in, say, TLS transport connection to the outbound proxy, then that transport protocol should be the one that is required to match.

That definitely makes sense, but it is out of scope for this PR. And it would have to be considered what happens if there are multiple outbound proxies (which could have different transports). It would be difficult to determine a transport protocol to match if it is not specifically configured in the account, so in that case the transport protocol match is ignored (no change of behavior in BareSIP).

cspiel1:
So this PR is a fallback UA match for registered accounts. Here is the related PR: https://github.com/baresip/baresip/pull/1206

cspiel1:
Reviewed and discussed with @maximilianfridrich. We found that currently the match in the second loop for registered accounts is to less restrictive.

```c
		if (0 == pl_casecmp(cuser, &acc->luri.user)) {
			ua_printf(ua, "account match for %r\n", cuser);
			return ua;
		}
```

This PR avoids that the wrong registered UA is selected. It's ready for the merge.

PR #3103, opened 2024-08-09
ausrc, aufile, gst: replace duration by callback handler

cspiel1:
Retrieving the duration of an audio file may lead to long blocking
delay. E.g. gst_element_query_duration()

The proposed solution meant for discussion.

A callback handler added to `struct ausrc_prm` can be called by the application only if needed.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -565,6 +565,7 @@ int message_encode_dict(struct odict *od, struct account *acc,
 
 struct ausrc;
 struct ausrc_st;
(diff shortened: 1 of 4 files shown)


cspiel1:
I'll try a different solution with API:

```c
int ausrc_info(struct list *ausrcl, const char *name, struct ausrc_prm *prm, const char *device);
```
- Sets the fields of given `ausrc_prm`.
- No need to alloc an `ausrc`.

cspiel1:
I would prefer this https://github.com/baresip/baresip/pull/3105 solution. But I am open for discussions.

PR #3104, opened 2024-08-09
account specific video source and display

paresy:
This PR is an extension of this PR: https://github.com/baresip/baresip/pull/796

It adds support for specifying video source and display per account.
The global config will be used by default as fallback.

### DIFF
--- a/src/video.c
+++ b/src/video.c
@@ -1054,6 +1054,7 @@ static int vrx_print_pipeline(struct re_printf *pf, const struct vrx *vrx)
  * @param streaml    List of streams
  * @param stream_prm Stream parameters
  * @param cfg        Global configuration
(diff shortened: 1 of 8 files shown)


src/video.c
```diff
@@ -1067,6 +1071,7 @@ static int vrx_print_pipeline(struct re_printf *pf, const struct vrx *vrx)
 int video_alloc(struct video **vp, struct list *streaml,
 		const struct stream_param *stream_prm,
 		const struct config *cfg,
+		const struct account *acc,
```
sreimers, line 1074: Doxygen `@param acc  Account (optional)` is missing

paresy: Thanks. Fixed.

sreimers:
Thanks!

PR #3105, opened 2024-08-12
ausrc, aufile, gst, debug_cmd: rework on ausrc duration retrieval

cspiel1:
replaces: https://github.com/baresip/baresip/pull/3103

Another solution for duration retrieval.

Retrieving the duration of an audio file may lead to long blocking
delay. E.g. gst_element_query_duration()
Now a callback handler can be called by the application only if
needed.

```c
int ausrc_info(struct list *ausrcl, const char *name, const char *device, struct ausrc_prm *prm);
```

- Sets the fields of given ausrc_prm.
- No need to alloc an ausrc_st.

### DIFF
--- a/modules/aufile/aufile_src.c
+++ b/modules/aufile/aufile_src.c
@@ -231,7 +231,6 @@ int aufile_src_alloc(struct ausrc_st **stp, const struct ausrc *as,
 	/* return wav format to caller */
 	prm->srate = fprm.srate;
 	prm->ch    = fprm.channels;
-	prm->duration = aufile_get_length(st->aufile, &fprm);
 
 	if (!rh)
 		goto out;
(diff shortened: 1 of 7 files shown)


alfredh:
Something like this approach is good ...

modules/aufile/aufile_src.c
```diff
@@ -231,7 +231,6 @@ int aufile_src_alloc(struct ausrc_st **stp, const struct ausrc *as,
 	/* return wav format to caller */
 	prm->srate = fprm.srate;
 	prm->ch    = fprm.channels;
-	prm->duration = aufile_get_length(st->aufile, &fprm);
```
maximilianfridrich, line 234: Is there an advantage to not setting this here? It is very cheap to compute for an opened aufile.

alfredh:
is this one ready to merge ?

Could you please rebase on top of git main HEAD ?

cspiel1:
Yes, its ready for the merge.

PR #3123, opened 2024-08-19
ua,menu: do not accept call by default

cspiel1:
Replaces https://github.com/baresip/baresip/pull/3088 "Moved 180 Ringing response from core to app (menu module)"

- The application decides when to send a 180 Ringing.
- A new config setting `call_accept` can be set to `yes` in order to let the baresip core reply with 180 Ringing, like before.

Note: As requested by @juha-h this changes the default behavior of the baresip core, because `call_accept` is `off` by default.

### Steps for bevent
- [x] add core bevent
- [x] ~~add event class filter mask to `bevent_register()`~~
- [x] add backwards wrapper
- [x] update function calls
  - [x] use bevent API in core
  - [x] use bevent API in test
  - [x] use bevent API in module menu
  - [x] module ctrl_dbus
  - [x] module ctrl_tcp
  - [x] module echo
  - [x] module gtk
  - [x] module mqtt
  - [x] module mwi
  - [x] module presence (forgot `publisher.c`)
  - [x] module rtcpsummary
  - [x] module serreg
  - [x ] module ebuacip
- [x] provide alternative to https://github.com/baresip/baresip/pull/3088
- [ ] move DnD feature from core to menu
- [ ] let `menu` decide if 180 Ringing is sent or 183 Session Progress. Not both like it is done currently if `answermode` is `early`.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -34,6 +34,7 @@ sip_tos			160 # See TOS fields!
 call_local_timeout	120
 call_max_calls		4
 call_hold_other_calls	yes
+call_accept		no
 
 # Audio
 #audio_path		/usr/local/share/baresip
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -372,6 +372,7 @@ struct config_call {
 	uint32_t local_timeout; /**< Incoming call timeout [sec] 0=off    */
 	uint32_t max_calls;     /**< Maximum number of calls, 0=unlimited */
 	bool hold_other_calls;  /**< Hold other calls */
+	bool accept;            /**< Accept call by baresip core          */
 };
 
 /** Audio */
@@ -909,6 +910,7 @@ int  ua_connect_dir(struct ua *ua, struct call **callp,
 		    enum vidmode vmode, enum sdp_dir adir, enum sdp_dir vdir);
 void ua_hangup(struct ua *ua, struct call *call,
 	       uint16_t scode, const char *reason);
+int  ua_accept(struct ua *ua, const struct sip_msg *msg);
 int  ua_answer(struct ua *ua, struct call *call, enum vidmode vmode);
 int  ua_hold_answer(struct ua *ua, struct call *call, enum vidmode vmode);
 int  ua_options_send(struct ua *ua, const char *uri,
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -664,6 +664,7 @@ static void event_handler(enum ua_event ev, struct bevent *event, void *arg)
 	const char           *prm  = bevent_get_text(event);
 	struct call          *call = bevent_get_call(event);
 	struct ua            *ua   = bevent_get_ua(event);
+	const struct sip_msg *msg  = bevent_get_msg(event);
 	struct account       *acc  = ua_account(bevent_get_ua(event));
 	int err;
 	(void)arg;
@@ -679,6 +680,17 @@ static void event_handler(enum ua_event ev, struct bevent *event, void *arg)
 	count = uag_call_count();
 
 	switch (ev) {
+
+	case UA_EVENT_SIPSESS_CONN:
+		ua = uag_find_msg(msg);
+		err = ua_accept(ua, msg);
+		if (err) {
+			warning("menu: could not accept incoming call (%m)\n",
+				err);
+			return;
+		}
+
+		break;
 
 	case UA_EVENT_CALL_INCOMING:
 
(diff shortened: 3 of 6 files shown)


cspiel1:
@juha-h please could you review and test this with baresip-studio?

juha-h:
Will do in the evening.

juha-h:
Why not use the ua value that has already been assigned in the beginning of `event_handler`?
```
case UA_EVENT_SIPSESS_CONN:
		ua = uag_find_msg(msg);
```

juha-h:
In order to know if incoming call can be accepted, at least caller's URI would need to be available.  How to get that from msg?

Also, why `ua_accept` needs msg parameter?  It means that I would need to pass the whole SIP message from native code (event_handler) to application code.

juha-h:
After `ua_accept`, what API calls are needed to generate 180 response and what API calls are needed to reject the call with 4xx response?

cspiel1:
> After `ua_accept`, what API calls are needed to generate 180 response and what API calls are needed to reject the call with 4xx response?

In the `event_handler`
```c
	case UA_EVENT_SIPSESS_CONN:
		ua = uag_find_msg(msg);
		if (dnd(ua)) {
			err = ua_accept(ua, msg);
			(void)sip_treply(NULL, uag_sip(), msg,
						    480,"Temporarily Unavailable");
		}
		else {
			err = ua_accept(ua, msg);
		}

		if (err) {
			warning("menu: could not accept incoming call (%m)\n",
				err);
			return;
		}

		break;
```

juha-h:
Tried to get From URI from msg->from.auri:
```
case UA_EVENT_SIPSESS_CONN:
  char from_uri[128];
  ua = uag_find_msg(msg);
  pl_strcpy(msg->from.auri, from_uri, 128);
  LOGD("****** from URI = %s\n", from_uri);
```
and got to log:
```
08-22 18:25:49.741  7820  7901 D Baresip Lib: ****** from URI = [ 	
08-22 18:25:49.741  7820  7901 D Baresip Lib: ]*[[ 	
08-22 18:25:49.741  7820  7901 D Baresip Lib: ]*[~ 
```
foll

juha-h:
I made more tests by setting from_uri to constant string "sip:test@test.tutpro.com".  Then both accept with `ua_accept` and reject with `sip_treply` worked OK.

So my question remains: how to get From URI from msg?  My code above gives garbage.

cspiel1:
Wrong:
```c
pl_strcpy(msg->from.auri, from_uri, 128);
```
Correct:
```c
pl_strcpy(&msg->from.auri, from_uri, 128);
```
The compiler normally will stop with an error.

Better use: `pl_strdup()`

juha-h:
@cspiel1 Based on the above suggestion, tried like this:
```
           char *from_uri;
            pl_strdup(&from_uri, &msg->from.auri);
            LOGD("****** from URI = %s\n", from_uri);
            mem_deref(from_uri);
```
and got to log:
```
08-23 16:46:08.725 30909 30965 D Baresip Lib: ****** from URI = SIP/2.0
08-23 16:46:08.725 30909 30965 D Baresip Lib: Record-Route: <sip:192.26.111.38:8003;transport=tls;sn=ext_tls;lr;n2;pm=0;ice>
...
```
i.e. the whole SIP message was printed.

Also, Android Studio does not like this:
```
        pl_strdup(&from_uri, &msg->from.auri);
```
It complains about the second parameter:
```
        Incompatible pointer types passing 'struct pl *const *' to parameter of type 'const struct pl *' passing argument to parameter 'src' here
```

juha-h:
Problem solved.  My re and/or baresip libs were not up to date.

Unrelated to from uri, I got this kind of mwi related warning:
```
08-23 19:31:00.870  6756  6818 D Baresip Lib: Used deprecated ua_event() for MWI_NOTIFY. Use one of event_xxx_emit() instead!
```
Haven't yet checked if the deprecated call is in my code or in baresip lib.

juha-h:
Module mwi is checked in the update list, but this PR has not updated mwi module yet.  It still has in `notify_handler`:
```
	if (mbuf_get_left(msg->mb)) {
		ua_event(mwi->ua, UA_EVENT_MWI_NOTIFY, NULL, "%b",
			  mbuf_buf(msg->mb), mbuf_get_left(msg->mb));
	}
```

cspiel1:
Thanks! PR: https://github.com/baresip/baresip/pull/3131

maximilianfridrich:
Looks good to me.

juha-h:
Can this be included in 3.15.0?  It is fixing a bug.

PR #3134, opened 2024-08-27
version 3.15.0

alfredh:


cspiel1:
Should we add this fix to 3.15.0?
https://github.com/baresip/baresip/pull/3133

alfredh:
I am very sceptical about adding last minute changes to a release.

The code has been stable now for a month. If we add this PR we will reset that stability.
Is the PR super critical ? Can we release without it ?

cspiel1:
I only can evaluate based on app module `menu`:
- The missing `UA_EVENT_CALL_LOCAL_SDP` for the answer on a re-INVITE is not relevant for `menu`.
- The missing `UA_EVENT_CALL_REMOTE_SDP` for the initial SDP offer (while in CALL_STATE_IDLE) is also not relevant for `menu`

The problem occurred in our test automation where the baresip events are processed. We can postpone this to 3.16.0.

maximilianfridrich:
Agreed, I don't see #3133 as critical.

PR #3138, opened 2024-09-05
audio: initialize ausrc_prm

cspiel1:
.duration was not initialized


src/audio.c
```diff
@@ -1059,6 +1059,7 @@ static int start_source(struct autx *tx, struct audio *a, struct list *ausrcl)
 		size_t sz;
 		size_t psize_alloc;
 
+		memset(&prm, 0, sizeof(prm));
```
sreimers, line 1062: Maybe:

```c
struct ausrc_prm prm = {.srate = srate_dsp, ...};
```

PR #3140, opened 2024-09-05
bevent: move dnd to menu

cspiel1:
- ua,menu: move DnD to menu
- uag: add uag_reject() with log info

### Steps for bevent
- [x] add core bevent
- [x] ~~add event class filter mask to `bevent_register()`~~
- [x] add backwards wrapper
- [x] update function calls
  - [x] use bevent API in core
  - [x] use bevent API in test
  - [x] use bevent API in module menu
  - [x] module ctrl_dbus
  - [x] module ctrl_tcp
  - [x] module echo
  - [x] module gtk
  - [x] module mqtt
  - [x] module mwi
  - [x] module presence (forgot `publisher.c`)
  - [x] module rtcpsummary
  - [x] module serreg
  - [x] module ebuacip
- [x] provide alternative to https://github.com/baresip/baresip/pull/3088
- [x] move DnD feature from core to menu
- [ ] let `menu` decide if 180 Ringing is sent or 183 Session Progress. Not both like it is done currently if `answermode` is `early`.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -972,6 +972,7 @@ void ua_close(void);
 void ua_stop_all(bool forced);
 int  uag_hold_resume(struct call *call);
 int  uag_hold_others(struct call *call);
+int  uag_reject(const struct sip_msg *msg, uint16_t scode, const char *reason);
 void uag_set_nodial(bool nodial);
 bool uag_nodial(void);
 void uag_set_exit_handler(ua_exit_h *exith, void *arg);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -972,7 +972,6 @@ void ua_close(void);
 void ua_stop_all(bool forced);
 int  uag_hold_resume(struct call *call);
 int  uag_hold_others(struct call *call);
-int  uag_reject(const struct sip_msg *msg, uint16_t scode, const char *reason);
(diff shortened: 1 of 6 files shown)


cspiel1:
I tested. It's ready for merge from my side. Any suggestions?

juha-h:
Why the new API call `uag_reject`?  My tests have worked fine with `sip_treply`.

juha-h:
If app rejects the call for whatever reason, it should be app's (not core's) responsibility to tell the user that the call was rejected.  In my app I have for example this:
```
val toastMsg = String.format(getString(R.string.call_auto_rejected),
                                Utils.friendlyUri(this, peerUri, ua.account))
if (toastMsg != "") {
                            Log.d(TAG, "Auto-rejecting incoming call $uap/$peerUri")
                            Api.sip_treply(callp, 486, "Busy Here")
                            toast(toastMsg)
}

**Comment by juha-h** [comment]
It would be useful for the user to see in the log message where the rejected call came from.  The user could then, for example, choose to call back later.

**Comment by cspiel1** [comment]
Log entry looks like this now:
```
menu: incoming call from ID5 <sip:id5@10.1.0.198> rejected: 480 Temporarily Unavailable
```

Issue #3135, opened 2024-09-01
re_unlock error when using sndfile module to record call


juha-h:
A user of my Android app noticed that the app hangs when call is recorded.  The app uses `sndfile` module for call recording.  The call starts fine, but then produces `re_unlock` error on the same debug line where sndfile module tells that it is dumping encode audio to file:
```
2024-09-01 08:44:51.031 25797-25850 Baresip Lib             com.tutpro.baresip                   D   started
                                                                                                    stream: Enable RTP timeout (60000 milliseconds)
                                                                                                    test@test.tutpro.com: Call established: sip:thetestcall@opensips.org
2024-09-01 08:44:51.041 25797-25850 Baresip Lib             com.tutpro.baresip                   D  rtp_receiver: RTP RX thread started
2024-09-01 08:44:51.141 25797-25972 Baresip Lib             com.tutpro.baresip                   D  ua event (MODULE) sndfile,dump,/data/user/0/com.tutpro.baresip/files/recordings/dump-sip:test@test.tutpro.com=>sip:thetestcall@opensips.org-2024-09-01-08-44-51-enc.wav
2024-09-01 08:44:51.143 25797-25850 Baresip Lib             com.tutpro.baresip                   D  sndfile: dumping encode audio to /data/user/0/com.tutpro.baresip/files/recordings/dump-sip:test@test.tutpro.com=>sip:thetestcall@opensips.org-2024-09-01-08-44-51-enc.wav
                                                                                                    [31mmain: re_unlock error
                                                                                                    [;m
2024-09-01 08:44:51.245 25797-25850 Baresip Lib             com.tutpro.baresip                   D  stream: incoming rtp for 'audio' established, receiving from 192.98.100.38:60336
...
```
After this the app stops responding and Android kills it.  

If call is not recorded, the is no errors and everything works fine. Calls with recording used to work OK, but I don't know when the error started to occur.  The user reports that call recording was broken already at least on August 1 whatever were then the master versions baresip and re.

Any hints on how to find out where and why the `re_unlock` error happens?  My app does not make any `re_unlock` calls, but does make `re_thread_enter` and `re_thread_leave` calls.

juha-h:
Tested with older versions of the app and found that the last version that worked OK was build using baresip/re master on 2024-03-24 and the next version that was build on 2024-03-30 had this issue.

juha-h:
[...]

PR #3141, opened 2024-09-05
sndfile: module_event from main thread

cspiel1:
fixes: https://github.com/baresip/baresip/issues/3135

The main thread allocates and sets the filename and emits the
module event.

### DIFF
--- a/modules/sndfile/sndfile.c
+++ b/modules/sndfile/sndfile.c
@@ -32,7 +32,7 @@ struct sndfile_enc {
 	SNDFILE *encf;
 	int err;
 	const struct audio *audio;
-	char filename[SNDFILE_NAME_LEN];
+	char *filename;
 };
 
 struct sndfile_dec {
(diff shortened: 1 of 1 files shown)


cspiel1:
Need to test, then will set it to Ready.

cspiel1:
Works correct now.
@juha-h please could you test this in your Android App?

juha-h:
Christian Spielberger writes:

> Works correct now.
> @juha-h please could you test this in your Android App?

Works fine if I don't leave/re-enter re_thread when forwarding module
event `dump`.

Is there a way to lock the re thread so that when I leave the re-thread
while forwarding the event, all other re-thread related actions are
queued until I re-enter the thread?

I re-tested and if I use this sndfile.c

https://github.com/baresip/baresip/blob/023ada990de350fab0d0a3599fead92b56208ef4/modules/sndfile/sndfile.c 

everything works fine even when I leave/re-enter re_thread while
forwarding module event `dump`.  How can this be explained?

alfredh:
this fix looks good.

There should be no need to change any application code.

PR #3163, opened 2024-10-15
bevent: avoid deprecated warnings in selftest

cspiel1:
- Remove tests for deprecated functions.


sreimers:
> * Remove tests for deprecated functions.

I think I would prefer this, to keep tests and API simple/clean.

PR #3165, opened 2024-10-17
wasapi: Add WASAPI Audio module

sreimers:
This is a replacement for the winwave module, since Windows 7 there are newer improved Core Audio APIs.

### DIFF
--- a/modules/wasapi/play.c
+++ b/modules/wasapi/play.c
@@ -1,0 +1,187 @@
+/**
+ * @file wasapi/play.c Windows Audio Session API (WASAPI)
+ *
+ * Copyright (C) 2024 Sebastian Reimers
+ * Copyright (C) 2024 AGFEO GmbH & Co. KG
+ */
+
+#include <re.h>
+#include <rem.h>
+#include <baresip.h>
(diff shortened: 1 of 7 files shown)


modules/wasapi/play.c
```diff
@@ -0,0 +1,52 @@
...
+	(void)device;
+
+	if (!stp || !ap || !prm)
+		return EINVAL;
+
+	st = mem_zalloc(sizeof(*st), auplay_destructor);
+	if (!st)
+		return ENOMEM;
+
+	st->wh	= wh;
```
alfredh, line 42: Perhaps rename to `writeh` as it is more readable ?

sreimers: Since `st->wh` is defined on all audio player modules, would it be better to be consistent?

modules/wasapi/wasapi.h
```diff
@@ -0,0 +1,70 @@
+/**
+ * @file wasapi.c Windows Audio Session API (WASAPI)
```
alfredh, line 2: please use the correct filename here ...

Issue #3164, opened 2024-10-17
conf_get_float is not in baresip.h


juha-h:
Function `conf_get_float` is in `core.h`, not in `baresip.h`.  Can it be moved so that modules can use it?

alfredh:
Yes, that function can be moved to the public api.

Some of the generic functions in conf.c can also be moved to libre conf module.

alfredh:
thanks for adding it:

https://github.com/baresip/re/commit/b9d11ef065a44e856c60709aad2cf0aae2496bf2

PR #3166, opened 2024-10-17
augain: add audio volume filter module

juha-h:
This module can be used to increase volume of audio source, for example, microphone.


modules/augain/augain.c
```diff
@@ -0,0 +1,165 @@
+/**
+ * @file augain.c  Audio gain
+ *
+ * Copyright (C) 2024 Juha Heinanen
+ */
+
+#include <re.h>
+#include <rem.h>
+#include <baresip.h>
+#include <stdlib.h>
```
sreimers, line 10: System headers like `stdlib.h` should be included before others:

https://github.com/baresip/re/wiki/Development-(re-usage)#include-headers

modules/augain/augain.c
```diff
@@ -0,0 +1,165 @@
...
+
+struct augain_enc {
+	struct aufilt_enc_st af;  /* base class */
+};
+
+
+static double gain = 1.0;
+
+
+static int conf_get_float(const struct conf *conf, const char *name,
```
sreimers, line 35: Should be removed after this is resolved: https://github.com/baresip/baresip/issues/3164

modules/augain/augain.c
```diff
@@ -0,0 +1,165 @@
...
+	size_t i;
+	unsigned int abs_sample, highest_abs_sample = 0;
+	double encode_gain = gain;
+	double highest_possible_gain;
+	int16_t sample, gained_sample;
+
+	if (!st || !af || !af->sampv || !af->sampc)
+		return EINVAL;
+
+	if (af->fmt != AUFMT_S16LE) {
```
sreimers, line 97: This check should be moved to `encode_update` (`prm->fmt`), not needed to execute for every frame.

modules/augain/augain.c
```diff
@@ -0,0 +1,165 @@
...
+	highest_possible_gain = 32767.0 / highest_abs_sample;
+	if (encode_gain > highest_possible_gain)
+		encode_gain = highest_possible_gain;
+
+	/*
+	info("augain: sample count=%d/highest abs=%d/encode_gain=%.2f\n",
+	     af->sampc, highest_abs_sample, encode_gain);
+	*/
+
+	for (i=0; i<af->sampc; i++) {
```
sreimers, line 118:
```suggestion
	for (size_t i=0; i<af->sampc; i++) {
```

modules/augain/augain.c
```diff
@@ -0,0 +1,165 @@
...
+		sample = ((int16_t *)(af->sampv))[i];
+		gained_sample = (int16_t)(sample * encode_gain);
+		/*
+		info("augain: sample=%d/gained=%d\n", sample, gained_sample);
+		*/
```
sreimers, line 123: Please remove debug code

modules/augain/augain.c
```diff
@@ -0,0 +1,165 @@
...
+			highest_abs_sample = abs_sample;
+	}
+
+	highest_possible_gain = 32767.0 / highest_abs_sample;
+	if (encode_gain > highest_possible_gain)
+		encode_gain = highest_possible_gain;
+
+	/*
+	info("augain: sample count=%d/highest abs=%d/encode_gain=%.2f\n",
+	     af->sampc, highest_abs_sample, encode_gain);
```
sreimers, line 115: Please remove debug code

modules/augain/augain.c
juha-h: Yes.

modules/augain/augain.c
juha-h: I moved `af->fmt` check but was not sure if also `if (!st || !af || !af->sampv || !af->sampc)` could have been moved.

juha-h:
Does it make any sense to have call
```
	conf_get_float(conf_cur(), "augain", &gain);
```
in `encode_update`?  Could it be removed?

modules/augain/augain.c
```diff
@@ -0,0 +1,156 @@
...
+		warning("augain: format not supported (%s)\n",
+			aufmt_name(prm->fmt));
+		return ENOTSUP;
+	}
+
+	st = mem_zalloc(sizeof(*st), enc_destructor);
+	if (!st)
+		return EINVAL;
+
+	conf_get_float(conf_cur(), "augain", &gain);
```
alfredh, line 84: can be done in init function

juha-h: Done.  It would have made sense in encode_update if there would be possibility to set the config var on the fly during call.  But I didn't find any conf_set functions.

modules/augain/augain.c
```diff
@@ -0,0 +1,161 @@
...
+
+	for (size_t i=0; i<af->sampc; i++) {
+		abs_sample = abs(((int16_t *)(af->sampv))[i]);
+		if (abs_sample > highest_abs_sample)
+			highest_abs_sample = abs_sample;
+	}
+
+	highest_possible_gain = 32767.0 / highest_abs_sample;
+	if (encode_gain > highest_possible_gain)
+		encode_gain = highest_possible_gain;
```
alfredh, line 89: It looks like the real gain is different for each frame, depending on the frame energy.
Does this algorithm work, is there any audible noise or artefacts ?

alfredh:
Here is another implementation, for inspiration:

https://github.com/FFmpeg/FFmpeg/blob/master/libavfilter/af_volume.c

modules/augain/augain.c
juha-h: Yes.  I thought that it is better than cutting those sample values that that would not fit 32768 - 32767 range.

I have made tests and haven't noticed anything very bad.  It is trade of between not hearing anything and hearing, but with some artifacts.

juha-h:
@alfredh If I interpreted the quoted  ffmpeg code correctly, it is after applying the gain cutting the resulting value to  -32768 - 32767 range, whereas the code in this PR lowers the gain so that the the resulting value stays within the range.  To me the ffmpeg version (if I understood it correctly), could result to lots of min or max sample values if gain is large.

juha-h:
One possibility might be to introduce two algorithms, say `adaptive` (current one) and `fixed` where out-of-range sample values are simply cut into the range.

modules/augain/augain.c
```diff
@@ -0,0 +1,161 @@
...
+
+static int module_init(void)
+{
+	aufilt_register(baresip_aufiltl(), &augain);
+
+	conf_get_float(conf_cur(), "augain", &gain);
+
+	info("augain: gaining by at most %.2f\n", gain);
+
+	return cmd_register(baresip_commands(), cmdv, RE_ARRAY_SIZE(cmdv));
```
alfredh, line 144: Please also add a matching call to `cmd_unregister()` in `module_close`

modules/augain/augain.c
juha-h: I made several tests using both adaptive and fixed algorithm and noticed less artifacts with the adaptive one.  Thus I don't think  that it is worth to add fixed algorithm to this PR.  It can be added later if someone needs it.

From my point of view, this PR is done.

alfredh:
thanks, can you please also update this page with the new module and your name ?

juha-h:
Thanks for the merge.  Added the module to Wiki.

PR #3169, opened 2024-10-18
conf: remove conf_get_float

juha-h:


sreimers:
Needs removed from `src/core.h` too.

Issue #3171, opened 2024-10-19
test: suppress warnings from test_call_sni


alfredh:
```
alfredh@debian:~/git/baresip$ ./test/selftest -r main test_call_sni
running baresip selftest version 3.16.0 with 1 tests
[ RUN      ] test_call_sni (rx main)
tls: tls_verify_handler: err          = 20
tls: tls_verify_handler: error_string = unable to get local issuer certificate
tls: tls_verify_handler: depth        = 1
tls: connect: error (r=-1, ssl_err=1)
tls: 40A70F17917F0000:error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:../ssl/statem/statem_clnt.c:1889:
tls: accept error: (r=-1, ssl_err=1)
tls: 40A70F17917F0000:error:0A000418:SSL routines:ssl3_read_bytes:tlsv1 alert unknown ca:../ssl/record/rec_layer_s3.c:1605:SSL alert number 48
tls: tls_verify_handler: err          = 20
tls: tls_verify_handler: error_string = unable to get local issuer certificate
tls: tls_verify_handler: depth        = 0
tls: connect: error (r=-1, ssl_err=1)
tls: 40A70F17917F0000:error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:../ssl/statem/statem_clnt.c:1889:
tls: accept error: (r=-1, ssl_err=1)
tls: 40A70F17917F0000:error:0A000418:SSL routines:ssl3_read_bytes:tlsv1 alert unknown ca:../ssl/record/rec_layer_s3.c:1605:SSL alert number 48
tcp: send: write(): Connection reset by peer [104] (fdc=21)
[       OK ]
OK. 1 tests passed successfully
```

Would it be possible to silence these warnings ?

maximilianfridrich:
I will take a look.

maximilianfridrich:
@alfredh Unfortunately, I don't see a simple way of silencing the warnings. The `DEBUG_LEVEL` is hardcoded at the top of `re/src/tls/openssl/tls.c` and the whole point of the SNI test is to have some negative test cases as well as positive ones.

If you have an idea of how to solve it, please share it and I can look into implementing it.

alfredh:
This can be done by either setting a dummy debug handler or setting the debug level.

This can be done temporarily while the test is running.

Here is one example from retest:
https://github.com/baresip/re/blob/main/test/uri.c#L422

[...]

PR #3173, opened 2024-10-21
test/call.c: test_call_sni: silence warnings

maximilianfridrich:
The warnings are expected because some TLS handshakes in the test are supposed to fail.

Fixes: #3171

### DIFF
--- a/test/call.c
+++ b/test/call.c
@@ -12,6 +12,11 @@
 #include "../src/core.h"  /* NOTE: temp */
 
 
+#define DEBUG_MODULE "testcall"
+#define DEBUG_LEVEL 5
+#include <re_dbg.h>
+
+
 #define MAGIC 0x7004ca11
 
 enum {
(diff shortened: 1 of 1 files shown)


test/call.c
```diff
@@ -3307,6 +3312,9 @@ int test_call_sni(void)
 	const char *dp = test_datapath();
 	char s[256];
 
+	/* warnings are expected for negative test cases, so silence them */
+	dbg_init(DBG_ERR, DBG_ANSI);
+
```
alfredh, line 3317: Please also restore the debug level at the end of the test ...

alfredh:
this looks good.

Could you also please do the same for `test_call_cert_select` ?

test/call.c
maximilianfridrich: Done.

maximilianfridrich:
> Could you also please do the same for test_call_cert_select ?

Done.

PR #3177, opened 2024-10-26
webrtc_aecm: added -Wno-missing-field-initializers compile option

juha-h:


alfredh:
I am not sure if this is needed.

With C99 you can init one struct member by name, and the rest of the members should be zero.

Please post the warning and the compiler details.

juha-h:
Below.  The warning started to appear when I upgrade NDK.  Also, see `modules/webrtc_aec/aec.cpp`.  Since it is there, why not in  `modules/webrtc_aecm/aec.cpp`?
```
/usr/src/libbaresip-android/baresip/modules/webrtc_aecm/aec.cpp:132:37: warning: missing field 'le' initializer [-Wmissing-field-initializers]
  132 |                                    .dech    = webrtc_aecm_decode};
      |                                                                 ^
```

juha-h:
Got the same warning about .enabled field:
```
/usr/src/libbaresip-android/baresip/modules/webrtc_aecm/aec.cpp:135:1: warning: missing field 'enabled' initializer [-Wmissing-field-initializers]
  135 | };
      | ^
```

juha-h:
These warnings appeared with clang version 18.0.3.

juha-h:
If "the rest of the members should be zero", should `.enabled` be intialized to false for backwards compatibility?

juha-h:
Could it be that the warnings only appear with cpp source code?

alfredh:
https://reviews.llvm.org/D157879

alfredh:
I guess there is a difference between C and CPP.

That warning is disabled for the other module:

```
$ git grep missing-field
modules/gzrtp/CMakeLists.txt:    -Wno-missing-field-initializers
modules/webrtc_aec/CMakeLists.txt:  -Wno-missing-field-initializers
```


https://github.com/baresip/baresip/blob/main/modules/webrtc_aec/CMakeLists.txt#L27

juha-h:
OK, but would it be better to avoid the warnings in the code rather than disable them?

alfredh:
It should be almost the same if we update the makefile or the code.

But I am slightly in favour of updating the makefile, since that is done for the other modules.

juha-h:
OK, I can modify CMakeLists.txt instead and remove .le init from webrtc_aec module so that it is compatible with webrtc_aecm module.

PR #3186, opened 2024-11-04
menu dnd event

cspiel1:
This adds an enum `bevent_class` field to `struct bevent` because it should not be derived from the `enum ua_event`. This does not always give wanted results. The second commit gives an example how the fixed API can be used.

- **bevent: add enum bevent_class to struct**
- **menu: emit bevent for dnd reject**

### DIFF
--- a/src/bevent.c
+++ b/src/bevent.c
@@ -482,20 +482,6 @@ int event_encode_dict(struct odict *od, struct ua *ua, enum ua_event ev,
 }
 
 
-static int odict_pl_add(struct odict *od, const char *key,
-			const struct pl *val)
-{
-	char *str;
-	int err = pl_strdup(&str, val);
-	if (err)
-		return err;
-
-	err = odict_entry_add(od, key, ODICT_STRING, str);
-	mem_deref(str);
-	return err;
-}
-
-
 int odict_encode_bevent(struct odict *od, struct bevent *event)
 {
 	struct ua *ua     = bevent_get_ua(event);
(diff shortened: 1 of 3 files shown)


maximilianfridrich:
Looks good to me.

src/bevent.c
```diff
@@ -531,6 +482,20 @@ int event_encode_dict(struct odict *od, struct ua *ua, enum ua_event ev,
 }
 
 
+static int odict_pl_add(struct odict *od, const char *key,
+			const struct pl *val)
```
cspiel1, line 486: Should we add this to `re`?

cspiel1: https://github.com/baresip/re/pull/1208

juha-h:
Does this PR mean that API has changed and existing apps will need to be modified?  If so, the PR should document the steps needed.

cspiel1:
No, the API didn't change.

But now it allows to use the "emit" functions in a more flexible way. E.g. in main only this ev is allowed in first parameter:

```c
bevent_sip_msg_emit(UA_EVENT_SIPSESS_CONN, msg, ...);
```

Now this is also possible:
```c
bevent_sip_msg_emit(UA_EVENT_MODULE, msg, ...);
```

juha-h:
@cspiel1 Yes, but before this commit, there was no `bevent_sip_msg_emit` call in `menu.c`.  Why is it now needed or is it there just for fun?

cspiel1:
We have an application on top of menu which collects rejected calls in a list. The user may use it to call back.

Issue #3199, opened 2024-11-11
Missing: Extra headers from the start ...


flash7777:
I missed the option to set extra SIP headers, ideally right at the start, on the command line—and even better, not just one, but several. That would be great.

usage:  
`baresip  -x <key>:<val> -x <key>:<val>`

Attached is a forward patch of an older 1.0.0 hotfix.  So i kept this uag fragment.

[baresip-3.16.0_mod.patch.txt](https://github.com/user-attachments/files/17708524/baresip-3.16.0_mod.patch.txt)

alfredh:
Could you please create a Pull Request for this patch ?

flash7777:
i have no branch here. how to proceed?

sreimers:
> i have no branch here. how to proceed?

You can fork the project and submit the PR from your fork.

alfredh:
we can track the progress in the PR

PR #3202, opened 2024-11-13
xtra headers

flash7777:
xtra headers fresh from the start

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -294,6 +294,12 @@ int custom_hdrs_add(struct list *hdrs, const char *name, const char *fmt, ...);
 int custom_hdrs_apply(const struct list *hdrs, custom_hdrs_h *h, void *arg);
 
 
+void debug_list_hdrs( struct list *custom_headers );
+int uag_set_custom_hdrs( struct list *custom_headers);
+struct list *uag_get_custom_hdrs(void);
+void uag_flush_hdrs(void);
+
+
 /*
  * Conf (utils)
  */
--- a/modules/contact/contact.c
+++ b/modules/contact/contact.c
@@ -94,6 +94,7 @@ static int cmd_dial_contact(struct re_printf *pf, void *arg)
 
 	uri = contact_uri(cnt);
 
+
 	err = ua_connect(uag_find_requri(uri), NULL, NULL, uri, VIDMODE_ON);
 	if (err) {
 		warning("contact: ua_connect(%s) failed: %m\n",
--- a/src/main.c
+++ b/src/main.c
@@ -107,7 +107,7 @@ int main(int argc, char *argv[])
 	const char *ch_par = NULL;
 	char *ch_val = NULL;
 	char *x_par,*x_val = NULL;
-	int l_ch = 0;
+	unsigned int l_ch = 0;
 
 	struct list x_custom_hdrs;
 
--- a/src/main.c
+++ b/src/main.c
@@ -75,6 +75,7 @@ static void usage(void)
 			 "\t-n <net_if>      Specify network interface\n"
 			 "\t-u <parameters>  Extra UA parameters\n"
 			 "\t-v               Verbose debug\n"
+			 "\t-x <key>:<val>   extra header\n"
 			 "\t-T               Enable timestamps log\n"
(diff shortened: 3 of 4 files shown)


flash7777:
It’s not implemented very cleanly, as this is a forward patch from 1.0.0 to 3.x.y. Please take a look at the UAG structure; I’ve taken it over in a rudimentary way, as it’s no longer present in 3.x.y.

flash7777:
isnt there any auto-linter in git? has trailing tab(s)...   
and there's no testcase for this implementation. sorry.

sreimers:
Not sure if this is needed, you can maybe use `usage: /uaaddheader <key>=<value> <ua-idx>`

```bash
baresip -e "/uaaddheader x-custom=test 0" -e "/uaaddheader x-custom2=test 0"
```

flash7777:
works on register too?

include/baresip.h
```diff
@@ -294,6 +294,12 @@ int custom_hdrs_add(struct list *hdrs, const char *name, const char *fmt, ...);
 int custom_hdrs_apply(const struct list *hdrs, custom_hdrs_h *h, void *arg);
 
 
+void debug_list_hdrs( struct list *custom_headers );
```
alfredh, line 297: please add a prefix to the function name ...

if the parameter `custom_headers` is not modified, it can be `const`

modules/contact/contact.c
```diff
@@ -94,6 +94,7 @@ static int cmd_dial_contact(struct re_printf *pf, void *arg)
 
 	uri = contact_uri(cnt);
 
+
```
alfredh, line 97: unrelated change ..

src/main.c
```diff
@@ -75,6 +75,7 @@ static void usage(void)
 			 "\t-n <net_if>      Specify network interface\n"
 			 "\t-u <parameters>  Extra UA parameters\n"
 			 "\t-v               Verbose debug\n"
+			 "\t-x <key>:<val>   extra header\n"
```
alfredh, line 78: Please use "Extra" like the other lines. Also please specify "SIP" header.

src/main.c
```diff
@@ -200,6 +210,36 @@ int main(int argc, char *argv[])
 			ua_eprm = optarg;
 			break;
 
+		case 'x':
+			l_ch  = strlen(optarg);
```
alfredh, line 214: here you can also use `str_len()` which is portable ...

src/ua.c
```diff
@@ -37,6 +37,12 @@ struct ua_xhdr_filter {
 	char *hdr_name;
 };
 
+static struct {
+	struct list custom_hdrs;
+} uag = {
+	LIST_INIT,
+};
```
alfredh, line 44: please move this block to `uag.c`

cspiel1:
Solves issue: https://github.com/baresip/baresip/issues/3199

TODOs
- please squash the commits and set a meaningful commit message
- solve lint checks
- please add a test case

alfredh:
I agree with the comment from @sreimers about the command `/uaaddheader`

what do you think @cspiel1

flash7777:
My time budget is a bit tight at the moment, so I wouldn't be able to get around to implementing the changes anytime soon anyway. What's not necessary doesn't need to be included. I must have been quite naïve to add something like that without consulting the detailed documentation or studying the code, especially since it was for version 1.0.0 and I only patched it forward. Software is only as good as its documentation .... and even more so, its users. ;)

cspiel1:
> I agree with the comment from @sreimers about the command `/uaaddheader`
> 
> what do you think @cspiel1

I realized the comment of @sreimers. But then I made a mistake and thought that the `-e` does not work for SIP REGISTER. Now I checked again that the modules like account are loaded first, this creates the `ua` objects, then the "Execute commands" from command line options are executed. Thus, the solution @sreimers mentioned should work. (I didn't test it.)

So the only benefit might be that it is a little bit more comfortable. I'd agree with you if you think this is to less benefit and we could close this.

alfredh:
What is the next step here -- is it good enough to use the command solution ?

alfredh:
Lets try to use the existing commands.

if you have time later and want to add your patch, please create a new PR later ...

PR #3208, opened 2024-11-17
Aaudio module

juha-h:
I created PR event when the module is still work in progress so that others can comment easier.

Here is an example on how recording sounds at my Linux baresip when I call it using aaudio module on my Android device: https://box.tutpro.com/tmp/recorder.wav.  Voice is encoded far too fast and there are glitches. Playing works at correct speed, but there a lots of glitches.

There must be something wrong in dataCallback function in both recorder.c and player.c, but I don't have a glue.

### DIFF
--- a/modules/aaudio/aaudio.c
+++ b/modules/aaudio/aaudio.c
@@ -1,0 +1,60 @@
+/**
+ * @file aaudio.c  AAudio audio driver
+ *
+ * Copyright (C) 2024 Juha Heinanen
+ */
+
+#include <re.h>
+#include <rem.h>
+#include <baresip.h>
+
+#include <aaudio/AAudio.h>
+
+#include "aaudio.h"
+
+
+/**
+ * @defgroup aaudio aaudio
+ *
+ * Aaudio audio driver module for Android
+ */
+
+
+static struct auplay *auplay;
+static struct ausrc *ausrc;
+
+
+static int module_init(void)
+{
+	int err;
+
+	err  = auplay_register(&auplay, baresip_auplayl(),
+			       "aaudio", aaudio_player_alloc);
+	err |= ausrc_register(&ausrc, baresip_ausrcl(),
+			      "aaudio", aaudio_recorder_alloc);
+
+	return err;
+}
+
+
+static int module_close(void)
+{
+
+	if (playerStream)
+		AAudioStream_close(playerStream);
+	if (recorderStream)
+		AAudioStream_close(recorderStream);
+
+	auplay = (struct auplay *)mem_deref((void *)auplay);
+	ausrc = (struct ausrc *)mem_deref((void *)ausrc);
+
+	return 0;
+}
+
+
+EXPORT_SYM const struct mod_export DECL_EXPORTS(aaudio) = {
+	"aaudio",
+	"audio",
+	module_init,
+	module_close,
+};
--- a/modules/aaudio/aaudio.h
+++ b/modules/aaudio/aaudio.h
@@ -1,0 +1,18 @@
+/**
+ * @file aaudio.h  Aaudio sound driver
+ *
+ */
+
+
+#include <aaudio/AAudio.h>
+
+extern AAudioStream *playerStream;
+extern AAudioStream *recorderStream;
+
+int aaudio_player_alloc(struct auplay_st **stp, const struct auplay *ap,
+	struct auplay_prm *prm, const char *device,
+	auplay_write_h *wh, void *arg);
+
+int aaudio_recorder_alloc(struct ausrc_st **stp, const struct ausrc *as,
+			  struct ausrc_prm *prm, const char *device,
+			  ausrc_read_h *rh, ausrc_error_h *errh, void *arg);
--- a/modules/aaudio/player.c
+++ b/modules/aaudio/player.c
@@ -39,7 +39,7 @@ static void auplay_destructor(void *arg)
  * the audioData buffer.
  */
  static int dataCallback(AAudioStream *stream, void *userData,
-			      void *audioData, int32_t numFrames) {
+			 void *audioData, int32_t numFrames) {
 	(void)stream;
 	struct auplay_st *st = userData;
 	struct auframe af;
@@ -60,14 +60,14 @@ static int open_player_stream(struct auplay_st *st) {
 
 	AAudioStreamBuilder *builder;
 	aaudio_result_t result;
-	
+
 	result = AAudio_createStreamBuilder(&builder);
 	if (result != AAUDIO_OK) {
 		warning("oboe: failed to create stream builder: error %s\n",
 			AAudio_convertResultToText(result));
 		return result;
 	}
-	
+
 	AAudioStreamBuilder_setDirection(builder, AAUDIO_DIRECTION_OUTPUT);
 	AAudioStreamBuilder_setSharingMode(builder,
 		AAUDIO_SHARING_MODE_SHARED);
@@ -99,9 +99,9 @@ static int open_player_stream(struct auplay_st *st) {
 	     AAudioStream_getUsage(playerStream));
 
 	AAudioStreamBuilder_delete(builder);
-	
+
 	AAudioStream_setBufferSizeInFrames(playerStream,
-                AAudioStream_getFramesPerBurst(playerStream) * 2);
+		AAudioStream_getFramesPerBurst(playerStream) * 2);
 
 	return AAUDIO_OK;
 }
--- a/modules/aaudio/player.c
+++ b/modules/aaudio/player.c
@@ -14,7 +14,6 @@
 #include "aaudio.h"
 
 struct auplay_st {
-	AAudioStream *stream;
 	struct auplay_prm play_prm;
 	auplay_write_h *wh;
 	size_t sampsz;
@@ -28,8 +27,8 @@ static void auplay_destructor(void *arg)
 {
 	struct auplay_st *st = arg;
 
-	if (st->stream)
-		AAudioStream_close(st->stream);
+	if (playerStream)
+		AAudioStream_close(playerStream);
 
 	st->wh = NULL;
 }
@@ -71,7 +70,7 @@ static int open_player_stream(struct auplay_st *st) {
 	
 	AAudioStreamBuilder_setDirection(builder, AAUDIO_DIRECTION_OUTPUT);
 	AAudioStreamBuilder_setSharingMode(builder,
-		AAUDIO_SHARING_MODE_EXCLUSIVE);
+		AAUDIO_SHARING_MODE_SHARED);
 	AAudioStreamBuilder_setSampleRate(builder, st->play_prm.srate);
 	AAudioStreamBuilder_setChannelCount(builder, 1);
 	AAudioStreamBuilder_setFormat(builder, AAUDIO_FORMAT_PCM_I16);
@@ -82,7 +81,7 @@ static int open_player_stream(struct auplay_st *st) {
                AAUDIO_USAGE_VOICE_COMMUNICATION),
 	AAudioStreamBuilder_setDataCallback(builder, &dataCallback, st);
 
-	result = AAudioStreamBuilder_openStream(builder, &st->stream);
+	result = AAudioStreamBuilder_openStream(builder, &playerStream);
 	if (result != AAUDIO_OK) {
 		warning("aaudio: failed to open player stream: error %s\n",
 			AAudio_convertResultToText(result));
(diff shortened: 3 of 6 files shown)


juha-h:
Commit https://github.com/baresip/baresip/pull/3208/commits/73850c30e21875fd0dc2a6602d856b01532b8694 fixed aaudio recorder.  numFrames parameter to dataCallback contains number of frames, not bytes.  Now recorder seems to work flawlessly, right speed no glitches!  Player still needs to get fixed.

juha-h:
Same think with player, fixed by https://github.com/baresip/baresip/pull/3208/commits/1e701d37750fa6abd0bd1c9c7329b8aaf2e1a669.  Now aaudio module works OK in my tests.  But I would still appreciate a proper review, since I'm not an audio expert and some other issues may still go unnoticed.

modules/aaudio/player.c
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	st = mem_zalloc(sizeof(*st), auplay_destructor);
+	if (!st)
+		return ENOMEM;
+
+	st->play_prm.srate = prm->srate;
+	st->play_prm.ch    = prm->ch;
+	st->play_prm.ptime = prm->ptime;
+	st->play_prm.fmt   = prm->fmt;
```
sreimers, line 139: Can be replaced with:
```c
st->play_prm = *prm;
```

modules/aaudio/player.c
```diff
@@ -0,0 +1,173 @@
...
+	struct auplay_st *st = userData;
+	struct auframe af;
+
+	auframe_init(&af, st->play_prm.fmt, audioData, numFrames,
+		     st->play_prm.srate, st->play_prm.ch);
+
+	st->wh(&af, st->arg);
+
+	for (size_t i = 0; i < af.sampc; ++i)
+		((int16_t *)audioData)[i] = ((int16_t *)(af.sampv))[i];
```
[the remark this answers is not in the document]
sreimers: But wait, this should not needed, since `af.sampv` and `audioData` is the same pointer.

```c
st->wh(&af, st->arg);
```

Already writes to the buffer. So you can simply drop the `for` loop

modules/aaudio/player.c
```diff
@@ -0,0 +1,173 @@
...
+	st = mem_zalloc(sizeof(*st), auplay_destructor);
+	if (!st)
+		return ENOMEM;
+
+	st->play_prm.srate = prm->srate;
+	st->play_prm.ch    = prm->ch;
+	st->play_prm.ptime = prm->ptime;
+	st->play_prm.fmt   = prm->fmt;
+
+	st->sampsz = aufmt_sample_size(prm->fmt);
```
sreimers, line 141: Looks unused for player.

modules/aaudio/recorder.c
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	st = mem_zalloc(sizeof(*st), ausrc_destructor);
+	if (!st)
+		return ENOMEM;
+
+	st->src_prm.srate = prm->srate;
+	st->src_prm.ch    = prm->ch;
+	st->src_prm.ptime = prm->ptime;
+	st->src_prm.fmt   = prm->fmt;
```
sreimers, line 162: Use:

```c
st->src_prm = *prm;
```

modules/aaudio/recorder.c
```diff
@@ -0,0 +1,205 @@
...
+		warning("oboe: failed to create stream builder: error %s\n",
+			AAudio_convertResultToText(result));
+		return result;
+	}
+
+	AAudioStreamBuilder_setDirection(builder, AAUDIO_DIRECTION_INPUT);
+	AAudioStreamBuilder_setSharingMode(builder,
+		AAUDIO_SHARING_MODE_EXCLUSIVE);
+	AAudioStreamBuilder_setSampleRate(builder, st->src_prm.srate);
+	AAudioStreamBuilder_setChannelCount(builder, 1);
```
sreimers, line 97:
```suggestion
	AAudioStreamBuilder_setChannelCount(builder, st->src_prm.ch);
```

modules/aaudio/recorder.c
```diff
@@ -0,0 +1,205 @@
...
+	sampc = numFrames;
+	if (sampc > st->sampc) {
+		st->sampv = mem_realloc(st->sampv, st->sampsz * sampc);
+		st->sampc = sampc;
+	}
+
+	if (!st->sampv)
+		return ENOMEM;
+
+	memset((uint8_t *)st->sampv, 0, numFrames);
```
sreimers, line 67:
```suggestion
	memcpy(af.sampv, audioData, auframe_size(af) * af.ch);
```

modules/aaudio/recorder.c
```diff
@@ -0,0 +1,205 @@
...
+
+	auframe_init(&af, st->src_prm.fmt, audioData, sampc,
```
sreimers, line 69:
```suggestion
	auframe_init(&af, st->src_prm.fmt, st->sampv, sampc,
```

sreimers: `auframe_init` needs to be moved before `memcpy`

modules/aaudio/player.c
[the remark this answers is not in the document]
juha-h: Is it so that also stereo (2 channels) should be supported and other formats than S16LE should be supported?  I have never needed those in my Android app.

Regarding `memcpy`, the change didn't compile until I changed it to this:
```
	memcpy(audioData, af.sampv, auframe_size(&af) * af.ch);
```

juha-h:
@sreimers Thank you very much for your fixes/suggestions.  I committed them and tested that module worked with them.

modules/aaudio/player.c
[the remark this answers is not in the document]
sreimers: As said, you can remove the memcpy. Since `st->wh` writes directly to `audioData` here, since `af.sampc` and `audioData` are the same pointer, so it's currently copied twice.

sreimers:
A `cmake/modules.cmake` entry is missing.

juha-h:
I made some cleanups and did more testing.  Replacing opensles module and webrtc_aecm with this module greatly improves sound quality when  aec, agc, and noise suppression can now be done in the app using device's hardware.

I don't currently have anything in mind to add or change.  The PR could thus be merged unless more issues show up in reviews.

modules/aaudio/recorder.c
```diff
@@ -0,0 +1,200 @@
...
+	result = AAudio_createStreamBuilder(&builder);
+	if (result != AAUDIO_OK) {
+		warning("aaudio: recorder: failed to create stream builder: "
+			"error %s\n", AAudio_convertResultToText(result));
+		return result;
+	}
+
+	AAudioStreamBuilder_setDirection(builder, AAUDIO_DIRECTION_INPUT);
+	AAudioStreamBuilder_setSharingMode(builder,
+		AAUDIO_SHARING_MODE_EXCLUSIVE);
```
sreimers, line 96: Maybe better use `AAUDIO_SHARING_MODE_SHARED`?

juha-h: You mean allow the microphone to be used at the same time by many apps?  Anyway, I checked and the exclusive mode was not accepted and the result is shared.

modules/aaudio/aaudio.c
```diff
@@ -0,0 +1,45 @@
...
+
+	return err;
+}
+
+
+static int module_close(void)
+{
+
+	auplay = (struct auplay *)mem_deref((void *)auplay);
+	ausrc = (struct ausrc *)mem_deref((void *)ausrc);
```
alfredh, line 34: for these two lines, the casting is not needed

modules/aaudio/aaudio.h
```diff
@@ -0,0 +1,20 @@
+/**
+ * @file aaudio/aaudio.h AAudio audio driver for Android
+ *
+ * Copyright (C) 2024 Juha Heinanen
+ */
+
+#include <aaudio/AAudio.h>
+
+extern AAudioStream *playerStream;
+extern AAudioStream *recorderStream;
```
alfredh, line 10: is there a need for global variables ?

modules/aaudio/player.c
```diff
@@ -0,0 +1,163 @@
...
+#include <baresip.h>
+
+#include "aaudio.h"
+
+
+struct auplay_st {
+	struct auplay_prm play_prm;
+	auplay_write_h *wh;
+	size_t sampsz;
+	void *arg;
```
alfredh, line 19: I would suggest to group the write-handler and arg together, they usually go in a pair

modules/aaudio/player.c
```diff
@@ -0,0 +1,163 @@
...
+
+	module_event("aaudio", "player: sessionid", NULL, NULL, "%d",
+		     AAudioStream_getSessionId(playerStream));
+
+	info ("aaudio: player: stream started\n");
+
+  out:
+	if (result != AAUDIO_OK) {
+		close_stream(playerStream);
+		close_stream(recorderStream);
```
alfredh, line 156: is it needed to mix player and recorder instance ?

normally the audio recorder and player are orthogonal to each other

juha-h:
Sebastian Reimers writes:

> I think this works only best for dedicated audio devices (OTG
> USB). And if you have multiple calls, multiple streams are opened. So
> better use default shared mode here.

Done SHARED for both streams.

juha-h:
Alfred E. Heggestad writes:

> is it needed to mix player and recorder instance ?
> normally the audio recorder and player are orthogonal to each other

Yes, if one stream does not start the other (if still running) needs to
be closed.

juha-h:
Alfred E. Heggestad writes:
> @alfredh commented on this pull request.

> +extern AAudioStream *playerStream;
> +extern AAudioStream *recorderStream;
> 
> is there a need for global variables ?

That is because in case of an error in one stream, the other needs to be
closed.  Perhaps the destructors would handle it, but I have not taken
the risk that one stream stays hanging in the app.

juha-h:
Found a new issue.  If audio device (recorder or player) gets disconnected for example when headset is connected, error callback is called and the stream should be closed and then re-opened.  The problem here is that the docs tell that closing/re-opening must be done on a separate thread.  So far I have only found C++ examples on how to do it.

I'll add error callback to recorder and player and include comment in each on what should be called in a separate thread.  Any suggestions are welcome on how to make the calls on separate threads.  I can try with pthread tomorrow.

juha-h:
Latest commit added restarting of streams from errorCallbacks. The current code works in my tests, for example, when I connect/disconnect headset, but the code does not follow the rules.  Restarting should be done from another thread  (see comment in errorCallbacks) and I have not been able to do it.  Any help would be appreciated. Lint warnings are about too long comment lines.,

juha-h:
Here is C++ example: https://github.com/AliAsadi/aaudio-api-sample-app/blob/master/app/src/main/cpp/StreamEngine.cpp.  Should I try to use that?

juha-h:
With commit https://github.com/baresip/baresip/commit/d8c2558f0dfe607193e959b974bba688b4f70f6a  disconnected streams are started from a new thread.

modules/aaudio/player.c
```diff
@@ -0,0 +1,211 @@
...
+
+struct auplay_st {
+	auplay_write_h *wh;
+	void *arg;
+	struct auplay_prm play_prm;
+	size_t sampsz;
+};
+
+
+AAudioStream *playerStream = NULL;
```
sreimers, line 23: This should be part of the `auplay_st`, since with multiple calls or if a sound is played you have multiple streams to handle, same for recording. These have to be independent.

juha-h: How can I then close playerStream if starting of recorderStream fails and vice versa?

juha-h: Done by commit https://github.com/baresip/baresip/commit/8bce6730165288be5c5df5967e521a8f4bf670cc .

modules/aaudio/aaudio.h
```diff
@@ -0,0 +1,17 @@
+/**
+ * @file aaudio/aaudio.h AAudio audio driver for Android
+ *
+ * Copyright (C) 2024 Juha Heinanen
+ */
+
+#include <aaudio/AAudio.h>
+
+void close_stream(AAudioStream *stream);
```
alfredh, line 9: Please use the name of the module as prefix for global symbols ...

juha-h: Done by https://github.com/baresip/baresip/commit/f5c2e3338770422c2421782aeb12391392f9b391

modules/aaudio/utils.c
```diff
@@ -0,0 +1,20 @@
+/**
+ * @file aaudio/util.c  AAudio audio driver for Android
```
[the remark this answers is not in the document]
juha-h: Done by https://github.com/baresip/baresip/commit/f5c2e3338770422c2421782aeb12391392f9b391

alfredh:
This looks good now -- we can merge it to main

Adding Android CI can be done in a separate task, please see libre CI for working Android build.

juha-h:
@alfredh @sreimers  Can I do the merge or should you do it?

PR #3224, opened 2024-11-26
Use real path instead of CMAKE_C_IMPLICIT_LINK_DIRECTORIES

U2FsdGVkX1:
On RISC-V 64, glibc will prefer to be installed in /lib64/lp64d, however on some distributions (such as fedora), /lib64/lp64d will be symlinked to /lib64 itself for compatibility. That is

```
lib64
├── bin
└── lp64d -> .
```

Therefore, the actual value of the CMAKE_C_IMPLICIT_LINK_DIRECTORIES variable will be
`/lib64/lp64d;/usr/lib64/lp64d`

Under the previous method of determination, the `IS_SYSTEM_DIR` variable would always be equal to -1, causing the `CMAKE_INSTALL_RPATH` variable to be set, which can lead to issues on some distributions (such as Fedora).

This change ensures that the correct directories are identified as link directories, preventing unnecessary setting of `CMAKE_INSTALL_RPATH` and avoiding compatibility problems on these distributions.


robert-scheck:
I'm the package maintainer for baresip at Fedora. And maybe I'm missing something, but I don't think Fedora made an actual decision so far, there just have been IRC and [mailing list discussions](https://lists.pagure.io/archives/list/devel@lists.fedoraproject.org/thread/XFCBU3VCEGBE246VNZN7UV4X2RZG2PEN/), because Fedora itself is not yet building for RISC-V.

U2FsdGVkX1:
> I'm the package maintainer for baresip at Fedora. And maybe I'm missing something, but I don't think Fedora made an actual decision so far, there just have been IRC and [mailing list discussions](https://lists.pagure.io/archives/list/devel@lists.fedoraproject.org/thread/XFCBU3VCEGBE246VNZN7UV4X2RZG2PEN/), because Fedora itself is not yet building for RISC-V.

As of now, glibc in Fedora will be installed to the directory I mentioned above, which actually blocks the package check
https://src.fedoraproject.org/rpms/glibc/blob/rawhide/f/glibc.spec#_1433
https://openkoji.iscas.ac.cn/kojifiles/work/tasks/7347/6197347/build.log

alfredh:
It is difficult for me to understand what is the problem, and what the solution is doing

Is this a work around for RISC-V only, should it be encapsulated in ifdef RISC-V ?

Is there someone who can review and test this patch ?

sreimers:
I think we can simply replace this with:

```cmake
if(APPLE)
  set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
  set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
endif()
```

This was the original intention of the RPATH usage. 

https://github.com/baresip/baresip/pull/2839
https://github.com/baresip/baresip/pull/2831

U2FsdGVkX1:
> It is difficult for me to understand what is the problem, and what the solution is doing
> 
> Is this a work around for RISC-V only, should it be encapsulated in ifdef RISC-V ?
> 
> Is there someone who can review and test this patch ?

As I said above, under RISC-V, the CMAKE_C_IMPLICIT_LINK_DIRECTORIES variable for some distributions will be /lib64/lp64d;/usr/lib64/lp64d
This will cause the RPATH to always be set.
This patch is not just for RISC-V.

sreimers:
I think we can close this and simplify RPATH usage: https://github.com/baresip/baresip/pull/3232

PR #3227, opened 2024-11-28
video: add picture update bool to vidpacket

maximilianfridrich:
Currently, video encoders only receive picture update information (i.e. new keyframe request) when the encoder's frame handler is called. When the encoder's packetize handler is called, this information is not passed on. Some videosrc/vidcodec combinations only work in this "pass-through" mode where *only* the packetize handler is called, and the encoder will never receive picture update information. However, some encoders can still act on picture update requests even when operated in "pass-through" mode, so now the `update` bool is also passed to the packetize handler.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1096,6 +1096,7 @@ struct vidpacket {
 	size_t size;         /**< Size of buffer                    */
 	uint64_t timestamp;  /**< Timestamp in VIDEO_TIMEBASE units */
 	bool keyframe;       /**< True=keyframe, False=deltaframe   */
+	bool picup;          /**< Picture update requested          */
 };
 
 /* Declare function pointer */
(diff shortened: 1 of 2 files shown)


alfredh:
is it possible to reuse this struct instead of adding new parameters:

```c
struct vidpacket {
	uint8_t *buf;        /**< Buffer memory                     */
	size_t size;         /**< Size of buffer                    */
	uint64_t timestamp;  /**< Timestamp in VIDEO_TIMEBASE units */
	bool keyframe;       /**< True=keyframe, False=deltaframe   */
};
```

can the member `keyframe` be set instead ?

maximilianfridrich:
I would certainly like that solution, but I didn't want to incorrectly set the `keyframe` flag in the baresip core and then break some other encoder.

I see only the `av1` module which evaluates this bool right now (`av1/encode.c:240`) and I fear that if I (mis)use this bool now to indicate a picture update, we might break the `av1` encoder.

Please let me know which way you'd prefer. For us, setting the `keyframe` bool would work nicely, but I don't want to break other things.

alfredh:
can you consider to add the bool
to the struct ?

maximilianfridrich:
Done. Thanks for the review! @alfredh

PR #3228, opened 2024-11-28
ua,call: add API for rejecting incoming call

cspiel1:
This adds a more general version of `ua_hangup()` to the API which allows to add custom headers. E.g. `Contact:` and `Diversion:` for a 302 Moved Temporarily.

```c
void ua_hangupf(struct ua *ua, struct call *call,
                uint16_t scode, const char *reason, const char *fmt, ...);
```

related application: https://github.com/baresip/baresip-apps/pull/57
related discussion: https://github.com/baresip/baresip/discussions/3207

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -222,6 +222,8 @@ int  call_progress_dir(struct call *call,
 		       enum sdp_dir adir, enum sdp_dir vdir);
 int  call_progress(struct call *call);
 void call_hangup(struct call *call, uint16_t scode, const char *reason);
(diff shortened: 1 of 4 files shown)


cspiel1:
Maybe should check if the call is in incoming state

juha-h:
Good, would simplify code in apps.

zucher:
@cspiel1 , Thank you that's working very fine !

cspiel1:
I'll add a test case with custom headers.

maximilianfridrich:
Looks good to me.

PR #3230, opened 2024-12-02
video: allow RTCP FIR packets as payload-specific feedback

maximilianfridrich:
The RTCP FIR messages of [RFC 2032](https://www.rfc-editor.org/rfc/rfc2032) have been obsoleted in [RFC 4587](https://www.rfc-editor.org/rfc/rfc4587) (see section 5) and instead the feedback mechanism of [RFC 4585](https://www.rfc-editor.org/rfc/rfc4585) should be used which has been extended to include FIR messages in [RFC 5104](https://www.rfc-editor.org/rfc/rfc5104).

With this change, the FIR messages of RFC 5104, which are payload-specific feedback messages (PSFB), are also interpreted as picture update requests.

### DIFF
--- a/src/video.c
+++ b/src/video.c
@@ -968,13 +968,18 @@ static void rtcp_handler(struct stream *strm, struct rtcp_msg *msg, void *arg)
 
 	case RTCP_FIR:
 		mtx_lock(vtx->lock_enc);
+		debug("video: recv Full Intra Request (FIR)\n");
 		vtx->picup = true;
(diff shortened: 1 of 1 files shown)


maximilianfridrich:
I think it could make sense to keep them for interoperability purposes (e.g. older UAs). Asterisk for example processes the obsolete `RTCP_FIR` packets exactly like the PSFB messages.

PR #3234, opened 2024-12-04
ua,menu: move max calls check to menu

cspiel1:
Mentioned here https://github.com/baresip/baresip/pull/3018#issuecomment-2094645141
by @juha-h


sreimers:
Not sure about this. I mostly not use the menu module, and max calls is a good DDoS prevention. What's the reason for this change?

PR #3236, opened 2024-12-04
ua prevent double call accept

cspiel1:
- **ua,call: prevent double ua_accept**
  - API function ua_accept() can be called by multiple modules which handle
    the shortly added UA_EVENT_SIPSESS_CONN. This commit prevents that
    multiple call/sipsess/sip_dialog objects are created for a single SIP
    INVITE message.
- **menu,test: bevent_stop after ua_accept**
  -  After accepting a call the bevent should not be handled by following
     event handlers.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -2294,6 +2294,12 @@ int call_accept(struct call *call, struct sipsess_sock *sess_sock,
 }
 
 
+bool call_sess_cmp(const struct call *call, const struct sip_msg *msg)
+{
+	return sipsess_msg_equal(call->sess, msg);
+}
+
+
 static void delayed_answer_handler(void *arg)
 {
 	struct call *call = arg;
--- a/src/call.c
+++ b/src/call.c
@@ -2296,6 +2296,9 @@ int call_accept(struct call *call, struct sipsess_sock *sess_sock,
 
 bool call_sess_cmp(const struct call *call, const struct sip_msg *msg)
 {
+	if (!call || !msg)
+		return false;
+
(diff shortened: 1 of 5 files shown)


maximilianfridrich:
LGTM.

src/call.c
```diff
@@ -2294,6 +2294,12 @@ int call_accept(struct call *call, struct sipsess_sock *sess_sock,
 }
 
 
+bool call_sess_cmp(const struct call *call, const struct sip_msg *msg)
+{
```
alfredh, line 2298: please check the `call` pointer here ..

src/ua.c
```diff
@@ -786,6 +802,11 @@ int ua_accept(struct ua *ua, const struct sip_msg *msg)
 	if (!ua || !msg)
 		return EINVAL;
 
+	if (ua_find_call_msg(ua, msg)) {
+		warning("ua: call was already accepted\n");
+		return EINVAL;
+	}
+
```
alfredh, line 809: I would try to find a solution that does not involve a raw pointer.

You can add some logic in the application that looks up a call in **INCOMING** state.
This call is then passed to `ua_accept()` which will set the state to something else.
When the next module tries to call `ua_accept()` it will also lookup incoming calls,
and it will find none because the previous call was already accepted.

cspiel1: Sorry, had to look again.

This is not possible, because the purpose of `ua_accept()` is to allocate the one and only call for the given `sip_msg`.

PR #3238, opened 2024-12-05
menu: add scode and reason arguments to hangup command

maximilianfridrich:
This can be useful for replying for example with "603 Decline" which indicates that the user does not want to answer the call at all. From [RFC 3261 section 21.6.2](https://www.rfc-editor.org/rfc/rfc3261#section-21.6.2):

```
21.6.2 603 Decline

   The callee's machine was successfully contacted but the user
   explicitly does not wish to or cannot participate.  The response MAY
   indicate a better time to call in the Retry-After header field.  This
   status response is returned only if the client knows that no other
   end point will answer the request.
```

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -1159,6 +1159,76 @@ int menu_param_decode(const char *prm, const char *name, struct pl *val)
 }
 
 
+/**
+ * Find ua and call from command arguments.
+ *
+ * Assumes that the first argument passed in carg->prm is a valid call-id
+ * (if passed at all). If carg->data is set, it MUST be a pointer to a
+ * struct ua. If no call-id is passed in carg->prm, the currently active call
+ * is returned, if there is an active call.
+ *
+ * @param pf     Print backend
+ * @param carg   Command arguments
+ * @param uap    Pointer-pointer to ua. Set on successful return.
+ * @param callp  Pointer-pointer to call. Set on successful return.
+ *
+ * @return 0 for success, otherwise errorcode
+ */
+int menu_get_call_ua(struct re_printf *pf, const struct cmd_arg *carg,
+		     struct ua **uap, struct call **callp)
+{
+	int err = 0;
+	struct ua *ua;
+	struct call *call;
+	const char *eq;
+	char *cid = NULL;
+	struct pl pl = PL_INIT;
+
+	if (!carg || !uap || !callp)
+		return EINVAL;
+
+	/* fallback */
+	ua = carg->data ? carg->data : menu_uacur();
+	call = ua_call(ua);
+
+	if (re_regex(carg->prm, str_len(carg->prm), "[^ ]+", &pl))
+		goto out;
+
+	/* A call-id MUST NOT contain an '='. See RFC 3261 section 25.1. */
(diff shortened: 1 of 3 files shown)


maximilianfridrich:
Will update this PR and incorporate features from

- https://github.com/baresip/baresip-apps/pull/57

modules/menu/static_menu.c
```diff
@@ -773,40 +773,86 @@ static int cmd_enable_transp(struct re_printf *pf, void *arg)
...
 	const struct cmd_arg *carg = arg;
 	struct ua *ua = carg->data ? carg->data : menu_uacur();
 	struct call *call = ua_call(ua);
+	const char *rest = carg->prm;
 
-	(void)pf;
+	if (!str_isset(carg->prm)) {
+		if (!ua) {
+			err = ENOENT;
+			goto out;
```
alfredh, line 808: looks like a double `goto out` ?

maximilianfridrich: Thank you. Fixed.

modules/menu/menu.c
```diff
@@ -1159,6 +1159,68 @@ int menu_param_decode(const char *prm, const char *name, struct pl *val)
...
+	int err = 0;
+	struct ua *ua;
+	struct call *call;
+	const char *eq;
+	char *cid = NULL;
+	struct pl pl = PL_INIT;
+
+	if (!carg || !uap || !callp)
+		return EINVAL;
+
```
cspiel1, line 1186: Add comment:
/* fallback */

modules/menu/menu.c
```diff
@@ -1159,6 +1159,68 @@ int menu_param_decode(const char *prm, const char *name, struct pl *val)
...
+	ua = carg->data ? carg->data : menu_uacur();
+	call = ua_call(ua);
+
+	(void)re_regex(carg->prm, str_len(carg->prm), "[^ ]+", &pl);
```
cspiel1, line 1190: Is it possible to combine the two checks with regexp `"[^ =]+"`?

modules/menu/menu.c
```diff
@@ -1159,6 +1159,68 @@ int menu_param_decode(const char *prm, const char *name, struct pl *val)
...
+
+	/* A call-id MUST NOT contain an '='. See RFC 3261 section 25.1. */
+	eq = pl_strchr(&pl, '=');
+	if (!eq)
+		(void)pl_strdup(&cid, &pl);
```
cspiel1, line 1195: check err

modules/menu/menu.c
maximilianfridrich: Done.

modules/menu/menu.c
maximilianfridrich: I tried, but was not successful. With your mentioned regex, the Call-Id `abcd=ef` would be matched as `abcd` and we would look for the Call-Id `abcd`. Instead we want to detect that this is not a valid Call-Id.

modules/menu/menu.c
maximilianfridrich: Thank you. Done.

cspiel1 (approved):
Looks good now

PR #3244, opened 2024-12-11
bevent: fix encode bevent without display name

cspiel1:
This fixes serialization of `BEVENT_CLASS_SIP` events. If display name is not set `odict_encode_bevent()` failed and the event was not sent e.g. via dbus.

### DIFF
--- a/src/bevent.c
+++ b/src/bevent.c
@@ -507,7 +507,7 @@ int odict_encode_bevent(struct odict *od, struct bevent *event)
 			err = odict_pl_add(od, "contact", &hdr->val);
 
 		if (pl_isset(&msg->from.dname))
-			err = odict_pl_add(od, "display", &msg->from.dname);
+			err |= odict_pl_add(od, "display", &msg->from.dname);
 
 		err |= re_sdprintf(&buf, "%H", uri_encode, &msg->from.uri);
 		err |= odict_entry_add(od, "from", ODICT_STRING, buf);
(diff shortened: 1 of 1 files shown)


src/bevent.c
```diff
@@ -506,8 +506,10 @@ int odict_encode_bevent(struct odict *od, struct bevent *event)
 		if (hdr)
 			err = odict_pl_add(od, "contact", &hdr->val);
 
-		err |= odict_pl_add(od, "display", &msg->from.dname);
-		err |= re_sdprintf(&buf, "%H", uri_encode, &msg->from.uri);
+		if (pl_isset(&msg->from.dname))
+			odict_pl_add(od, "display", &msg->from.dname);
```
alfredh, line 510: perhaps check return value?

cspiel1: yes, with the check if `dname` is set the only err that might occur is `ENOMEM`. Commit follows

src/bevent.c
```diff
@@ -506,7 +506,9 @@ int odict_encode_bevent(struct odict *od, struct bevent *event)
 		if (hdr)
 			err = odict_pl_add(od, "contact", &hdr->val);
 
-		err |= odict_pl_add(od, "display", &msg->from.dname);
+		if (pl_isset(&msg->from.dname))
+			err = odict_pl_add(od, "display", &msg->from.dname);
+
```
alfredh, line 511: this will overwrite `err` from previous line where it is set.

the correct thing to do is to check `err` or to use `err |=`

PR #3270, opened 2025-01-10
vidinfo: avoid use of floating numbers

alfredh:


sreimers:
The coverage fail is unrelated, fixed here: https://github.com/baresip/baresip/pull/3271

PR #3271, opened 2025-01-10
ci/coverage: use ubuntu-22.04

sreimers:
I need to investigate why ubuntu 24.04 fails, for now let's stick with 22.04


alfredh:
thanks for the quick fix!

PR #3275, opened 2025-01-13
Fixed FreeBSD builds

rhaberkorn:
Fixes FreeBSD builds - it has already been in the FreeBSD port in the form of patches.

### DIFF
--- a/modules/alsa/alsa.c
+++ b/modules/alsa/alsa.c
@@ -4,7 +4,7 @@
  * Copyright (C) 2010 Alfred E. Heggestad
  */
 #define _DEFAULT_SOURCE 1
-#define _POSIX_SOURCE 1
+#define _POSIX_C_SOURCE 200112L
 #include <sys/types.h>
 #include <sys/time.h>
 #include <stdlib.h>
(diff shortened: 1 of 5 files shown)


sreimers:
Please make a separate PR for the FreeBSD macro changes.

rhaberkorn:
I cannot even test one without the other. But if you insist.

sreimers:
I think we should use `_GNU_SOURCE` like we do for libre, so we can remove the individual posix related define statements completely, like here: https://github.com/baresip/re/commit/14ad287966304b6eec37603236b6d6b7651611cf

rhaberkorn:
> I think we should use `_GNU_SOURCE` like we do for libre, so we can remove the individual posix related define statements completely, like here: [baresip/re@14ad287](https://github.com/baresip/re/commit/14ad287966304b6eec37603236b6d6b7651611cf)

Yes, this would also fix FreeBSD builds. If you like, I could prepare such a patch and send another PR. This PR (#3275) could be closed in this case.

sreimers:
Can you try: https://github.com/baresip/baresip/pull/3282

rhaberkorn:
works

PR #3276, opened 2025-01-14
portaudio: fixed inconsistency between the device specification in audio_source and via /ausrc

rhaberkorn:
* audio_source actually required an device index integer as reported by `baresip -v` and silently fell back to the default device for everything else. /ausrc however checked against the media registry, ie. accepted only long names based on the host backend and Portaudio device name. Called with such a name, it would still fall back to the default device. In other words `/ausrc portaudio` was totally broken.
* src_alloc() and play_alloc() should accept a name as stored in the media registry, ie. as reported by `baresip -v`. Additionally, it accepts "default" as is apparently supported by all the other audio modules. All other strings result in an error. This is good - you don't want the system to silently fall back to the default device after requesting a specific device.
* For instance, to open source device /dev/dsp1 (e.g. on FreeBSD), you could write: `audio_source portaudio,"OSS: /dev/dsp1"`

### DIFF
--- a/modules/portaudio/portaudio.c
+++ b/modules/portaudio/portaudio.c
@@ -208,13 +208,33 @@ static void auplay_destructor(void *arg)
 	}
 }
 
+static int find_device(const char *device)
+{
+	struct mediadev *dev;
+	char *endp = NULL;
+	int dev_index;
+
+	dev = str_isset(device) && 0 != str_casecmp(device, "default")
+		? mediadev_find(&ausrc->dev_list, device)
+		: mediadev_get_default(&ausrc->dev_list);
+	if (dev)
+		return dev->device_index;
+
+	/*
+	 * Accept a numeric index as well for backwards compatibility.
+	 * This will only be supported for the audio_source and audio_player
+	 * commands and is not supported by any other audio backend.
+	 */
+	dev_index = (int)strtol(device, &endp, 10);
+	return *endp == '\0' ? dev_index : -1;
+}
 
 static int src_alloc(struct ausrc_st **stp, const struct ausrc *as,
 		     struct ausrc_prm *prm, const char *device,
 		     ausrc_read_h *rh, ausrc_error_h *errh, void *arg)
 {
 	struct ausrc_st *st;
-	PaDeviceIndex dev_index;
+	int dev_index;
 	int err;
 
 	(void)device;
@@ -223,10 +243,9 @@ static int src_alloc(struct ausrc_st **stp, const struct ausrc *as,
 	if (!stp || !as || !prm)
 		return EINVAL;
 
-	if (str_isset(device))
-		dev_index = atoi(device);
-	else
-		dev_index = Pa_GetDefaultInputDevice();
+	dev_index = find_device(device);
+	if (dev_index < 0)
+		return ENODEV;
 
 	st = mem_zalloc(sizeof(*st), ausrc_destructor);
 	if (!st)
@@ -257,7 +276,7 @@ static int play_alloc(struct auplay_st **stp, const struct auplay *ap,
 		      auplay_write_h *wh, void *arg)
 {
 	struct auplay_st *st;
-	PaDeviceIndex dev_index;
+	int dev_index;
 	int err;
 
 	(void)device;
@@ -265,10 +284,9 @@ static int play_alloc(struct auplay_st **stp, const struct auplay *ap,
 	if (!stp || !ap || !prm)
 		return EINVAL;
 
-	if (str_isset(device))
-		dev_index = atoi(device);
-	else
-		dev_index = Pa_GetDefaultOutputDevice();
+	dev_index = find_device(device);
+	if (dev_index < 0)
+		return ENODEV;
 
 	st = mem_zalloc(sizeof(*st), auplay_destructor);
 	if (!st)


modules/portaudio/portaudio.c
```diff
@@ -257,18 +258,19 @@ static int play_alloc(struct auplay_st **stp, const struct auplay *ap,
...
 	(void)device;
 
 	if (!stp || !ap || !prm)
 		return EINVAL;
 
-	if (str_isset(device))
-		dev_index = atoi(device);
-	else
-		dev_index = Pa_GetDefaultOutputDevice();
+	dev = !str_cmp(device, "default")
```
cspiel1, line 269: Should we also allow the empty string for the default device?

cspiel1:
Today is release day. Since this seems to solve completely broken config setting for portaudio and only touches this module I suggest to merge this.

Who is able to review this?

modules/portaudio/portaudio.c
```diff
@@ -223,10 +223,11 @@ static int src_alloc(struct ausrc_st **stp, const struct ausrc *as,
 	if (!stp || !as || !prm)
 		return EINVAL;
 
-	if (str_isset(device))
-		dev_index = atoi(device);
-	else
-		dev_index = Pa_GetDefaultInputDevice();
+	dev = !str_cmp(device, "default")
+		? mediadev_get_default(&ausrc->dev_list)
+		: mediadev_find(&ausrc->dev_list, device);
```
cspiel1, line 228: This breaks the current behavior because `mediadev_find()` compares only the device name and ignores the index. There might be already applications that rely on the index.

sreimers:
For the next release I think we should refactor this: https://github.com/baresip/baresip/discussions/3277
Since string comparison is not very reliable with all available drivers.

cspiel1:
My understanding is that if someone provides the correct index it works. I'll move this to next milestone.

modules/portaudio/portaudio.c
rhaberkorn: Looking through the other audio modules, it seems that the idea is indeed that the string can be empty or "default" (matched case-insensitively). In case of errors, we should perhaps return ENODEV instead of ENOENT (cf. coreaudio).

modules/portaudio/portaudio.c
rhaberkorn: The current behavior is undocumented and inconsistent. No other audio module appears to accept indexes. I only realized that `audio_source` expects an index after loading it in the debugger. What's the point in allowing it in `audio_source portaudio`, but not in `/ausrc`? But the `/ausrc` device names appear to be checked centrally. To allow indexes in `/ausrc` it seems you'd have to allow indexes in all audio modules.

rhaberkorn:
> For the next release I think we should refactor this: #3277
> Since string comparison is not very reliable with all available drivers.

If you would refactor this, it would of course make this patch superfluous. Any selection of the default device would have to happen centrally and behavior would be consistent across all audio modules.

alfredh:
Are we planning to merge this PR ?

cspiel1:
I suggest:
- empty device should fall back to the default device
- **maybe** the index could be the second try in order to stay backward compatible

or @sreimers has a better idea like mentioned above.

rhaberkorn:
> Are we planning to merge this PR ?

Of course if this would be my project, I would have merged it. You can still refactor it later on. But there is no reason to not immediately benefit from the fix, especially if the patch is so simple.

alfredh:
if you run baresip in verbose mode, you will see that portaudio dumps all audio devices detected:


```
portaudio: device count is 7
portaudio: device 0: ALSA: HD-Audio Generic: HDMI 0 (hw:0,3)
portaudio: device 1: ALSA: HD-Audio Generic: HDMI 1 (hw:0,7)
portaudio: device 2: ALSA: HD-Audio Generic: HDMI 2 (hw:0,8)
portaudio: device 3: ALSA: HD-Audio Generic: HDMI 3 (hw:0,9)
portaudio: device 4: ALSA: HD-Audio Generic: ALC256 Analog (hw:1,0)
portaudio: device 5: ALSA: hdmi
portaudio: device 6: OSS: /dev/dsp1
```


The idea was that you can use a number for the index, instead of the long string.
This was portaudio specific.

Regarding semantics for audio device name:

1. "default" -- Default device
2. "integer" -- Index for audio device
3. "string" -- Name of audio device
4. "(empty)" -- Perhaps default device

modules/portaudio/portaudio.c
```diff
@@ -223,10 +223,11 @@ static int src_alloc(struct ausrc_st **stp, const struct ausrc *as,
 	if (!stp || !as || !prm)
 		return EINVAL;
 
-	if (str_isset(device))
-		dev_index = atoi(device);
-	else
-		dev_index = Pa_GetDefaultInputDevice();
+	dev = !str_cmp(device, "default")
```
alfredh, line 226: Could you please use `str_casecmp` here to make it more robust ?

alfredh: You can update the PR if you want.

The important part is that the device string should handle values like integer,string,default and empty.

rhaberkorn:
I made all of the requested changes and rebased the patch.

alfredh:
This patch looks good now.

Please note, it should be possible to merge this patch AND refactor the media-device api.

PR #3278, opened 2025-01-15
release v3.19.0

cspiel1:


cspiel1:
`clang` macos test_call_bundle had a timeout in line time/call.c:2886

cspiel1:
The `test_call_bundle` seems to have some instability. I'll have a look on this tomorrow.

alfredh:
you can ignore the bundle test for now and go ahead with the release

The bundle test is a bit unstable and can be fixed later ...

cspiel1:
In `include/baresip.h` two functions where added. So I'll keep the ABI version. Okay?

Issue #3281, opened 2025-01-17
unstable test_call_bundle


alfredh:
This test is unstable when running on macOS:

```
[ RUN      ] test_call_bundle (rx main)
selftest: re_main() loop timed out -- test hung..
TEST_ERR: /Users/runner/work/baresip/baresip/test/call.c:2887: (Operation timed out [60])
test: call bundle test failed with mnat=off menc=off (Operation timed out [60])
srtp: failed to decrypt RTCP packet with 52 bytes (Authentication error [80])
srtp: failed to decrypt RTCP packet with 52 bytes (Authentication error [80])
srtp: failed to decrypt RTP packet with 34 bytes (Authentication error [80])
test_call_bundle (rx main): test failed (Operation timed out [60])
test failed (Operation timed out [60])
re main loop:
  maxfds:       1024
  nfds:         3
  method:       kqueue
  polling:      0
  sig:          0
  timers:       1
  mutex:        0x6000025e00c0
  tid:          0x1efad4f80
  thread_enter: 0
  async:        0x137e05020
```

alfredh:
Here is a log from a failing case:

```
Alfreds-MBP-M1:baresip alfredh$ ./test/selftest -v test_call_bundle
running baresip selftest version 3.19.0 with 1 tests
Local network addresses:
        lo0:  fe80::1
      anpi1:  fe80::1038:7bff:fe88:cc17
      anpi2:  fe80::1038:7bff:fe88:cc18
      anpi0:  fe80::1038:7bff:fe88:cc16
        ap1:  fe80::bcd0:74ff:fe9f:c963
        en0:  fe80::453:e185:24e2:6e7a
        en0:  10.0.1.136
      awdl0:  fe80::dcf1:c1ff:fed9:c060
      utun0:  fe80::c218:1168:da6b:8580
      utun1:  fe80::3178:c03f:b2d9:e075
      utun2:  fe80::ce81:b1c:bd2c:69e
      utun3:  fe80::9266:c575:cbc3:61c8
[ RUN      ] test_call_bundle (rx main)
vidcodec: H266
vidisp: mock-vidisp
vidsrc: fakevideo
vidisp: fakevideo
aucodec: PCMU/8000/1
aucodec: PCMA/8000/1
call: connecting to 'sip:b@127.0.0.1:50168'..
test: [ sip:a@127.0.0.1 ] event: CALL_OUTGOING (sip:b@127.0.0.1:50168)
bundle: alloc
[...]

PR #3287, opened 2025-01-21
test: make test_call_bundle stable

cspiel1:
fix for: https://github.com/baresip/baresip/issues/3281

For discussion: Should we fix the test or do we have a potential problem that RTP packets are missed when the call becomes established?

### DIFF
--- a/test/call.c
+++ b/test/call.c
@@ -2872,8 +2872,10 @@ static int test_call_bundle_base(bool use_mnat, bool use_menc)
 		fixture_init_prm(f, "");
 	}
 
-	cancel_rule_new(UA_EVENT_CALL_RTPESTAB, f->b.ua, 1, 0, 1);
-	cancel_rule_and(UA_EVENT_CALL_RTPESTAB, f->a.ua, 0, 0, 1);
+	cancel_rule_new(UA_EVENT_CALL_RTPESTAB, f->b.ua, 1, 0, -1);
+	cancel_rule_and(UA_EVENT_CALL_RTPESTAB, f->a.ua, 0, 0, -1);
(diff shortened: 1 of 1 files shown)


cspiel1:
This fixes the test. But not sure if it is good to ignore the order of first SIP 200 Ok vs first RTP packet.
The first RTP video packet normally contains the first key frame. In the detected case, the key frame would be missed by the decoder.

cspiel1:
Checked in stream.c and rtprecv.c that:
- 200 Ok enables the audio/video decoder

alfredh:
We could also add a keyframe flag to the mock video codec in test, or use a real video codec such as H.264

```c
static int mock_encode(struct videnc_state *ves, bool update,
		       const struct vidframe *frame, uint64_t timestamp)
{
	struct mbuf *hdr;
	uint8_t payload[2] = {0,0};
	uint64_t rtp_ts;
	int err;
	(void)update;

	if (!ves || !frame)
		return EINVAL;

	hdr = mbuf_alloc(16);

	err  = mbuf_write_u32(hdr, htonl(frame->fmt));
	err |= mbuf_write_u32(hdr, htonl(frame->size.w));
	err |= mbuf_write_u32(hdr, htonl(frame->size.h));
	if (err)
		goto out;

	rtp_ts = video_calc_rtp_timestamp_fix(timestamp);

	err = ves->pkth(true, rtp_ts, hdr->buf, hdr->end,
			payload, sizeof(payload), ves->vid);
	if (err)
		goto out;

 out:
	mem_deref(hdr);

	return err;
}
```

cspiel1:
Adding H.264 to the test or as separate test sound good to me. Maybe also with a simple image input/output comparison.

PR #3288, opened 2025-01-21
avacapture: fix vidframe init

sreimers:


modules/avcapture/avcapture.m
```diff
@@ -257,7 +257,7 @@ - (void)captureOutput:(AVCaptureOutput *)captureOutput
 {
 	const CVImageBufferRef b = CMSampleBufferGetImageBuffer(sampleBuffer);
 	CMTime ts = CMSampleBufferGetOutputPresentationTimeStamp(sampleBuffer);
-	struct vidframe vf;
+	struct vidframe vf = {.data[0] = NULL};
```
chucongqing, line 260: struct vidframe vf = {0};
should let all member be 0;

PR #3295, opened 2025-01-23
contact commands

cspiel1:
- **contact: add/rm contacts via commands**
- **contact: add audio/video local direction**
- **contact: avoid duplicate contacts**
- **menu: apply contact media direction if specified**

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -512,6 +512,8 @@ struct list     *contact_list(const struct contacts *contacts);
 const char      *contact_str(const struct contact *c);
 const char      *contact_uri(const struct contact *c);
 const char      *contact_presence_str(enum presence_status status);
+void             contact_get_ldir(const struct contact *c,
+				  enum sdp_dir *aup, enum sdp_dir *vidp);
 struct le       *contact_le(struct contact *cnt);
 void contacts_set_current(struct contacts *contacts, struct contact *cnt);
 struct contact *contacts_current(const struct contacts *contacts);
--- a/modules/contact/contact.c
+++ b/modules/contact/contact.c
@@ -237,12 +237,58 @@ static int cmd_current_next(struct re_printf *pf, void *arg)
 }
 
 
+static int cmd_add_contact(struct re_printf *pf, void *arg)
+{
+	const struct cmd_arg *carg = arg;
+	struct pl pl;
+
+	pl_set_str(&pl, carg->prm);
+	int err = contact_add(baresip_contacts(), NULL, &pl);
+	if (err)
+		re_hprintf(pf, "contact: could not add '%s'\n", carg->prm);
+
+	return err;
+}
+
+
+static int cmd_rm_contact(struct re_printf *pf, void *arg)
+{
+	const struct cmd_arg *carg = arg;
+	struct pl pl;
+	struct sip_addr addr;
+	char *uri;
+	struct contacts *contacts = baresip_contacts();
+	struct contact *c;
+	int err;
+	(void)pf;
+
+	pl_set_str(&pl, carg->prm);
+	err = sip_addr_decode(&addr, &pl);
+	if (err) {
+		warning("contact: could not decode '%r'\n", &pl);
+		return err;
+	}
+
+	err = pl_strdup(&uri, &addr.auri);
+	if (err)
+		return err;
+
+	while ((c = contact_find(contacts, uri)))
+		contact_remove(baresip_contacts(), c);
+
+	mem_deref(uri);
+	return 0;
+}
+
+
 static const struct cmd cmdv[] = {
 {"contacts",     'C',        0, "List contacts",          print_contacts    },
 {"dialcontact",  'D',        0, "Dial current contact",   cmd_dial_contact  },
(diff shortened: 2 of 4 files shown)


src/contact.c
```diff
@@ -269,6 +280,24 @@ enum presence_status contact_presence(const struct contact *c)
 }
 
 
+enum sdp_dir contact_adir(const struct contact *c)
+{
+	if (!c)
+		return SDP_SENDRECV;
```
sreimers, line 286: Maybe return `SDP_INACTIVE`?

include/baresip.h
```diff
@@ -512,6 +512,8 @@ struct list     *contact_list(const struct contacts *contacts);
 const char      *contact_str(const struct contact *c);
 const char      *contact_uri(const struct contact *c);
 const char      *contact_presence_str(enum presence_status status);
+enum sdp_dir     contact_adir(const struct contact *c);
+enum sdp_dir     contact_vdir(const struct contact *c);
```
alfredh, line 516: Perhaps these two functions can be combined into one function
that takes one extra argument `bool audio` ?

Also `au` or `audio` is more descriptive than just `a` ...

src/contact.c
cspiel1: For `sdp_dir` there is often the default value `SDP_SENDRECV`.

sreimers: But this is a error case not default.

include/baresip.h
cspiel1:
```
void contact_get_ldir(struct call *call, enum sdp_dir *ap, enum sdp_dir *vp);
```
similar to existing `call_get_mdir()`.

We don't need return int err here.

cspiel1:
Maybe will add some application code. Then switch to "Ready for review".

modules/contact/contact.c
```diff
@@ -237,12 +237,58 @@ static int cmd_current_next(struct re_printf *pf, void *arg)
...
+
+static int cmd_rm_contact(struct re_printf *pf, void *arg)
+{
+	const struct cmd_arg *carg = arg;
+	struct pl pl;
+	struct sip_addr addr;
+	char *uri;
+	struct contacts *contacts = baresip_contacts();
+	struct contact *c;
+	int err;
```
alfredh, line 262: perhaps move err to first init ...

modules/contact/contact.c
```diff
@@ -237,12 +237,58 @@ static int cmd_current_next(struct re_printf *pf, void *arg)
...
+
+
 static const struct cmd cmdv[] = {
 {"contacts",     'C',        0, "List contacts",          print_contacts    },
 {"dialcontact",  'D',        0, "Dial current contact",   cmd_dial_contact  },
 {"message",      'M',  CMD_PRM, "Message current contact",cmd_message       },
 {"contact_prev", '<',        0, "Set previous contact",   cmd_current_prev  },
 {"contact_next", '>',        0, "Set next contact",       cmd_current_next  },
+{"addcontact",     0,        0, "Add a contact",          cmd_add_contact   },
+{"rmcontact",      0,        0, "Remove a contact",       cmd_rm_contact   },
```
alfredh, line 291: Does these commands take parameters ?

src/contact.c
```diff
@@ -22,6 +22,9 @@ struct contact {
 	char *uri;
 	enum presence_status status;
 	enum access access;
+	enum sdp_dir aldir;   /* audio local direction */
+	enum sdp_dir vldir;   /* video local direction */
```
alfredh, line 26: please use "au" for audio and "vid" for video ...

src/contact.c
```diff
@@ -98,6 +101,7 @@ int contact_add(struct contacts *contacts,
 	if (err)
 		goto out;
 
+	contact_remove(contacts, contact_find(contacts, c->uri));
```
alfredh, line 104: perhaps the contact handling could be moved to dictionary
and allowed for multiple entries ?

src/contact.c
```diff
@@ -269,6 +280,20 @@ enum presence_status contact_presence(const struct contact *c)
 }
 
 
+void contact_get_ldir(const struct contact *c,
+		      enum sdp_dir *ap, enum sdp_dir *vp)
```
alfredh, line 284: please use "au" for audio, and "vid" for video ...

modules/contact/contact.c
cspiel1: Yes. I'll add a usage print.

src/contact.c
cspiel1: I can't follow. This line comes from commit 010c2294fb27b78d3444137f6db6592eeb747760.

alfredh: if the number of contacts is very large, i.e. 1000 entries, this code might be slow.
If you are adding a new entry, it means that you have to search the whole list first.

In another follow-up PR we can consider to switch from `struct list` to `struct odict` which
is an ordered dictionary.

cspiel1: Ok, I undestand. It's a hash table but for low bucket size it's like a list.

I can't see that odict is ordered or has any performance benefits compared to a hash map. But let's discuss this in a separate PR!

alfredh:
this looks good now

PR #3306, opened 2025-02-17
Add OCSP Stapling

maximilianfridrich:
config,uag: add `sip_ocsp_stapling` setting

See [RFC 6066](https://datatracker.ietf.org/doc/html/rfc6066) (for the `status_request` extension) and [RFC 2560](https://datatracker.ietf.org/doc/html/rfc2560) (for the Online Certificate Status Protocol - OCSP definition).

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -14,6 +14,7 @@ sip_capath		/etc/ssl/certs
 sip_verify_server	yes
 #sip_verify_client	no
 #sip_tls_resumption	all		# none, ids, tickets
+#sip_ocsp_stapling  no		# yes, required
 sip_tos			160 # See TOS fields!
 #filter_registrar   udp,tcp
 
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -357,18 +357,19 @@ const char *rtp_receive_mode_str(enum rtp_receive_mode rxmode);
 
 /** SIP User-Agent */
 struct config_sip {
-	char uuid[64];          /**< Universally Unique Identifier  */
-	char local[64];         /**< Local SIP Address              */
-	char cert[256];         /**< SIP Certificate                */
-	char cafile[256];       /**< SIP CA-file                    */
-	char capath[256];       /**< SIP CA-path                    */
-	uint32_t transports;    /**< Supported transports mask      */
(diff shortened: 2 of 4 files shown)


alfredh:
we should find out if we want this feature or not ...

Is this popular in commercial deployment ? There might be some government or military deployment
that is still using this.

Why have you decided to use OCSP Stapling and not something else ?

maximilianfridrich:
I intend to add multiple features for certificate revocation checking to BareSIP, this is just the first one. I was planning on adding in this order:

1. OCSP stapling ([RFC 6066](https://datatracker.ietf.org/doc/html/rfc6066) - because it is proabably simpler and does not add any extra requests to the TLS setup - no added latency).
2. CRLDP checking ([RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.13) and [RFC 2585](https://datatracker.ietf.org/doc/html/rfc2585)).
3. (Maybe) "Normal" [OCSP](https://datatracker.ietf.org/doc/html/rfc6960) requests through "Authority Information Access" extension.

In the related `re` PR I mentioned:

> Certificate revocation is still a difficult topic with no clear cut solution.
> 
> [RFC 9325](https://www.rfc-editor.org/rfc/rfc9325) ([section 7.5](https://www.rfc-editor.org/rfc/rfc9325#section-7.5)) provides some guidance and does suggest OCSP, particularly OCSP stapling since it solves the privacy issues around OCSP.

We have customers who require us to do certificate revocation checking and we want to be flexible and support multiple standards. I can check back with my team though and see if only CRLDP would be enough. I can make this a draft until I have an answer.

alfredh:
some other people using OCSP:

https://docs.pexip.com/admin/verify_tls.htm

https://docs.paloaltonetworks.com/pan-os/11-1/pan-os-admin/certificate-management/certificate-revocation/online-certificate-status-protocol-ocsp


It would be an interesting experiment to see if this feature can be implemented as a module.
Then the new module called e.g. `ocsp.so` could be added and loaded without any changes
to the core. The module would have to read the account configuration and enable/disable the
global TLS context accordingly.

alfredh:
did you get any feedback from your customers ?

It would probably make sense to add support for the method that is most commonly used...

PR #3307, opened 2025-02-17
call: use call_find_replaces for finding call to replace

### LINKED ISSUE #3304: baresip crashes when receiving invite with replaces header
this [call_find_id](https://github.com/baresip/baresip/blob/2e6f009e274658e8defb439073fff1ff94ba5674/src/call.c#L2247) does not find a call as a replaces header looks like this according to [RFC3891](https://datatracker.ietf.org/doc/html/rfc3891)
```
425928@bobster.example.org;to-tag=7743;from-tag=6472
```
Maybe the call id should be extracted from the header or the comparison should be changed to `str_str`

maximilianfridrich:
and return 481 if no matching dialog is found as required by [RFC 3891 section 3](https://datatracker.ietf.org/doc/html/rfc3891#section-3).

Fixes #3304.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -2165,6 +2165,63 @@ static bool valid_addressfamily(struct call *call, const struct stream *strm)
 }
 
 
+/**
+ * Find a call given the value of a Replaces header
+ *
+ * @param calls   List of calls
+ * @param hdr     Value of Replaces header
+ *
+ * @return Call object if found, NULL if not found
+ */
+static struct call *call_find_replaces(const struct list *calls,
+				       const struct pl *hdr)
+{
+	struct le *le;
+	struct sip_msg msg = { 0 };
+	struct call *ret = NULL;
+
+	if (!calls || !pl_isset(hdr))
+		return NULL;
+
+	msg.req = true;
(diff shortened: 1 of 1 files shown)


src/call.c
```diff
@@ -2165,6 +2165,87 @@ static bool valid_addressfamily(struct call *call, const struct stream *strm)
...
+			p += 7;  /* Skip "to-tag=" */
+			p2 = strchr(p, ';');
+			msg.to.tag.p = p;
+			msg.to.tag.l = p2 ? (size_t)(p2 - p) : str_len(p);
+		}
+	}
+
+	/* Find from-tag */
+	if ((p = strchr(hdr_val, ';'))) {
+		if ((p = str_str(p, "from-tag="))) {
```
sreimers, line 2214: can re_regex used for `to-tag` and `from-tag` value detection?

maximilianfridrich: Oh, right. That's much simpler. Done.

maximilianfridrich:
Tested locally with multiple variations of just callid, or call-id with tag(s), and matching/non-matching Replaces header value. Looks good.

PR #3312, opened 2025-02-20
video: do not request keyframe immediately if a packet is lost

maximilianfridrich:
The video decoder is in a much better position to decide if a new keyframe is needed, and decoders already indicate that by returning an error in the decode handler.

For example see `avcodec/decode.c`. A lost fragment does not immediately mean that a new keyframe is needed, while other errors can not be recovered and a new keyframe must be requested.

### DIFF
--- a/src/video.c
+++ b/src/video.c
@@ -889,12 +889,9 @@ static void stream_recv_handler(const struct rtp_header *hdr,
 	(void)extv;
 	(void)extc;
 	(void)ignore;
+	(void)lostc;
 
 	MAGIC_CHECK(v);
-
-	/* in case of packet loss, we need to receive a new keyframe */
-	if (lostc)
-		request_picture_update(&v->vrx);
 
 	(void)video_stream_decode(&v->vrx, hdr, mb);
 }


sreimers:
I'm not sure if hardware accelerated codecs behave like this, too.

maximilianfridrich:
The BareSIP `vidcodec` layer should still be able to indicate an error condition, even if hardware acceleration is used under the hood. I don't think it would be much different for a hardware accelerated codec, and all `vidcodec`s in BareSIP behave like that too.

E.g. in `avcodec/decode.c`, `avcodec_send_packet`/`avcodec_receive_frame` should return an error no matter if hardware acceleration is configured or not.

sreimers:
If I remember correctly hardware codecs don't return a error here, since it's only a direct memory copy and no decode state is shared. Maybe @alfredh can remember why this was added. It would take me some time to setup a hardware accelerated setup, to verify this.

alfredh:
I tested with H.263 a long time ago, and in case of packet loss it was concealing the picture and returning 0.

In general for any video codec it is a stream so if you loose one packet, you have to generate a new keyframe.
There are more advanced codecs and options to request slices, but I dont have knowledge about these ...

Each video codec decoder also checks the RTP sequence number btw ...

maximilianfridrich:
I can do more testing with other codecs and also with hardware acceleration to get a more definitive answer if the decoders always correctly indicate that a new keyframe is needed or if this line in the BareSIP core is needed.

I can re-open this PR once I have an answer.

PR #3313, opened 2025-02-20
menu: return on UA_EVENT_SIPSESS_CONN

maximilianfridrich:
Otherwise, when `UA_EVENT_SIPSESS_CONN` is done processing in the menu event_handler, `menu_set_incall` and `menu_update_callstatus` will be incorrect, leading to the `dynamic_menu` being broken.

### DIFF
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -731,7 +731,7 @@ static void event_handler(enum ua_event ev, struct bevent *event, void *arg)
 		}
 
 		bevent_stop(event);
-		break;
+		return;
 
 	case UA_EVENT_CALL_INCOMING:
 


cspiel1:
I think the problem is that `UA_EVENT_SIPSESS_CONN` directly calls `ua_accept()` -> `call_accept()` -> emits `CALL_EVENT_INCOMING`
- The `CALL_EVENT_INCOMING` activates the dynamic menu.
- But then the processing of `UA_EVENT_SIPSESS_CONN` proceeds at line 733, reaches `break` and then the dynamic menu is disabled again.

The `break` should be replaced by a `return`.

modules/menu/menu.c
```diff
@@ -730,6 +730,7 @@ static void event_handler(enum ua_event ev, struct bevent *event, void *arg)
 			return;
 		}
 
+		++count;
 		bevent_stop(event);
 		break;
```
cspiel1, line 735: Please could you try to replace the break by a return?

Increment count is not needed. It counts the numer of "real" calls. With SIPSESS_CONN the call is not yet allocated.

maximilianfridrich: In line `704` we count the number of calls. Right above this in line `726` we called `ua_accept` **successfully**. So we know our call count is no longer correct and must be incremented. That was my reasoning for incrementing the call count here.

Every other event handled here (except for **one** - `UA_EVENT_UNREGISTERING`) does not exit early either. So I thought the handling of `UA_EVENT_SIPSESS_CONN` should be fixed to not break the general `event_handler` functionality either, just like every other event. Further, `bevent_stop` is called, so this fix seemed pretty safe.

If you want me to put a `return` there instead, I will.

cspiel1: Processing of `UA_EVENT_SIPSESS_CONN` should end after the call of `ua_accept()`.
Yes, `return` is correct.

cspiel1:
Looks good now.

PR #3317, opened 2025-02-24
pulse: add NULL checks for paconn_get

maximilianfridrich:
This fixes SEGVs when pulse is restarted during a call.


cspiel1:
Thanks! An important fix.

PR #3327, opened 2025-03-19
misc: Transport-wide Congestion Control (TWCC) Receiver Status

sreimers:
https://datatracker.ietf.org/doc/html/draft-holmer-rmcat-transport-wide-cc-extensions-01

Todo:

- [ ] Offer SDP extension
- [ ] Configurable
- [ ] Testing
- [ ] Bundle support

### DIFF
--- a/src/twcc.c
+++ b/src/twcc.c
@@ -1,0 +1,412 @@
+/**
+ * @file twcc.c Transport-wide Congestion Control (TWCC) Receiver Status
+ *
+ * https://tools.ietf.org/
+ * draft-holmer-rmcat-transport-wide-cc-extensions-01
+ *
+ * Copyright (C) 2025 Sebastian Reimers
+ */
+
+#include <re.h>
+#include <baresip.h>
+#include "core.h"
+
+
+enum {
+	TWCC_INTERVAL = 100,  /**< ms                                */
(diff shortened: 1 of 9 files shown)


src/twcc.c
```diff
@@ -0,0 +1,434 @@
...
+ */
+
+#include <re.h>
+#include <baresip.h>
+#include "core.h"
+
+
+enum {
+	TWCC_INTERVAL	= 100,	/**< ms                                */
+	TWCC_MAX_PAKETS = 100,	/**< Limit pakets per Feedback message */
```
alfredh, line 17: MAX_PACKETS ?

src/twcc.c
```diff
@@ -0,0 +1,434 @@
...
+static const char uri[] = "http://www.ietf.org/id/"
+			  "draft-holmer-rmcat-transport-wide-cc-extensions-01";
+
+struct twcc_packet {
+	struct le le;
+	uint16_t tseq;
+	int32_t delta;
+	enum packet_state state;
+	uint64_t ts;
+};
```
alfredh, line 36: would it be possible to move this type and `enum packet_state` to libre ?

There might be useful to have functions to encode/decode to and from mbuf chunks/deltas ...

sreimers: Sounds good, will refactor this.

Issue #3330, opened 2025-03-25
portaudio find_device function fails to correctly identify Audio Playback Devices


darioxtx:
Hello @rhaberkorn, 

The latest changes introduced in the main branch for PortAudio have resulted in incorrect behavior within the `find_device` function. Specifically, this function successfully identifies audio source devices but fails to correctly search and locate audio playback (output) devices.

**Expected Behavior:**
The function should accurately locate and return the index of available audio playback devices.

**Actual Behavior:**
The function returns incorrect results or fails entirely when searching for audio playback devices. However, it operates as expected when used with audio source devices.

**Environment:**
baresip main branch
Windows 11

rhaberkorn:
Hello!

I am not a member of the project. But I will help to clean up the mess I created.

Would you please post your `baresip -v` (esp. the part following `portaudio: device count is`) and the relevant parts of your `.baresip/config`, i.e. the `audio_player` and `audio_source` commands?

I suppose that you are using `audio_player portaudio,default` and `audio_source portaudio,default` and the default playback and source devices differ on your system.
After my change, the same device is chosen as default input and output device - it's always the first one from the list. This is what `mediadev_get_default()` does. Previously, it used `Pa_GetDefaultInputDevice()` and `Pa_GetDefaultOutputDevice()`.

darioxtx:
The problem is with this part of code
```
dev = str_isset(device) && 0 != str_casecmp(device, "default")
		? mediadev_find(&ausrc->dev_list, device)
		: mediadev_get_default(&ausrc->dev_list);
```

this checks only audio source list `&ausrc->dev_list`. If you try to find player you should look devices in `&auplay->dev_list`

rhaberkorn:
> this checks only audio source list `&ausrc->dev_list`. If you try to find player you should look devices in `&auplay->dev_list`

You're right. I will fix this and create another merge request. I wasn't noticing this since it does work for devices that have both input and output channels (playback and recording). If you have a playback-only device, you won't be able to select it properly via `audio_player portaudio,...`.

As a workaround, you can use the deprecated numeric addressing mode, e.g. `audio_player portaudio,3`. Start with `baresip -v` to find out which index to use.

rhaberkorn:
Should be fixed now in the main branch. @darioxtx, can you confirm and close the ticket, please?

rhaberkorn:
[...]

PR #3331, opened 2025-03-26
portaudio: fixed selecting playback-only devices by name (refs #3330)

rhaberkorn:
find_device() was looking only in the ausrc source device list. If a playback device does not have input channels, it won't be in the ausrc device list and cannot be found, which broke the `audio_player` command for such devices.

We now pass the exact device list to find_device() as well.

### DIFF
--- a/modules/portaudio/portaudio.c
+++ b/modules/portaudio/portaudio.c
@@ -208,15 +208,15 @@ static void auplay_destructor(void *arg)
 	}
 }
 
-static int find_device(const char *device)
+static int find_device(const struct list *dev_list, const char *device)
 {
 	struct mediadev *dev;
 	char *endp = NULL;
 	int dev_index;
 
 	dev = str_isset(device) && 0 != str_casecmp(device, "default")
-		? mediadev_find(&ausrc->dev_list, device)
-		: mediadev_get_default(&ausrc->dev_list);
+		? mediadev_find(dev_list, device)
+		: mediadev_get_default(dev_list);
 	if (dev)
 		return dev->device_index;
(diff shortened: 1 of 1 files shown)


rhaberkorn:
I believe there might still be a bug related to selecting default devices. This is currently done via mediadev_get_default(), which is simply returning the first device from the given list. In portaudio, this will be the first device with playback or record capabilities in the portaudio device list. I am not sure whether Portaudio guarantees that its "default" devices always come first in their device iteration.

The default-flag is stored in the mediadev entries, so we could find the "real" default devices (as reported by Portaudio via Pa_GetDefaultInputDevice() and Pa_GetDefaultOutputDevice()). mediadev has "src" and "play" members, each with "is_default" flags, but we know that in this particular case only one of the two members will be in use, so we could simply pick the first device that has the flag set in either "src" or "play". Or we could call these functions directly as was done before 9ffabc53160a7539b23c3b67ac7ef5cc5f071187 and pass the default device index as a parameter to find_device().

If you'd like me to do this, you will probably want me to open a separate PR?

alfredh:
Many thanks for your fix, I will merge it to main ...

rhaberkorn:
> Many thanks for your fix, I will merge it to main ...

Can you comment on the possible bug for selecting the default devices? IMHO we'd be on the safe side if we respected what Portaudio reports as default devices for playback and recording devices respectively.

sreimers:
> Can you comment on the possible bug for selecting the default devices?

Should be fixed by: https://github.com/baresip/baresip/pull/3332

PR #3354, opened 2025-04-13
evdev: allowing evdev module compilation on FreeBSD

### LINKED ISSUE #3352: evdev: please do not restrict module compilation to Linux OS only
Greetings,

first of all, thanks for supporting `baresip` development.
I would like to ask you to consider changing this line of code:

https://github.com/baresip/baresip/blob/91645c3de43cd774ca14d94dd252625a9dd73509/modules/evdev/CMakeLists.txt#L3

to include FreeBSD into the list of OSes on which evdev module can be built.
The `evdev` driver available on FreeBSD:
https://man.freebsd.org/cgi/man.cgi?evdev(4)
and `baresip/evdev` module works for me if the line above changed as follows:
```
if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Linux")
  if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
    return()
  endif()
endif()
```
Note that I am not sure the exact code above is the most suitable from the `baresip` development perspective because I am not familiar enough with `cmake` magic.

Best Regards,
Denis.

DenisPolygalov:
resolves #3352

### DIFF
--- a/modules/evdev/CMakeLists.txt
+++ b/modules/evdev/CMakeLists.txt
@@ -3,7 +3,6 @@ project(evdev)
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
 elseif(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
 else()
-  message(STATUS "Skipping ${PROJECT_NAME} module compilation due to unsupported platform ${CMAKE_SYSTEM_NAME}")
   return()
 endif()
 
--- a/modules/evdev/CMakeLists.txt
+++ b/modules/evdev/CMakeLists.txt
@@ -1,6 +1,9 @@
 project(evdev)
 
-if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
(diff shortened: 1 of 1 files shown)


modules/evdev/CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 project(evdev)
 
-if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+elseif(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
+else()
+  message(STATUS "Skipping ${PROJECT_NAME} module compilation due to unsupported platform ${CMAKE_SYSTEM_NAME}")
```
alfredh, line 6: Can you please remove the message() line, I dont think that it is needed ...

alfredh:
also one cmake lint error:

```
./modules/evdev/CMakeLists.txt:6: Lines should be <= 80 characters long [linelength]
```

DenisPolygalov:
Done. Sorry for the extra troubles.

PR #3375, opened 2025-05-01
ctrl_tcp preload: allow preloading of ctrl_tcp module on baresip command line

IkerAriz:
The `ctrl_tcp` module fails to initialize when attempting to preload it with `baresip -m`. This MR addresses this issue with the following changes:

1. Update main to initialize the UA before preloading modules - makes the UA available to the initialization routine of  `ctrl_tcp` (and any other modules like it).
2. Update `load_module` to skip dynamic loading for static modules. If a static module is found an attempt is made to load/initialize it, and the result returned to the caller immediately. Previously, a static module that failed to initialize would fall through to the dynamic case which leads to misleading errors (in the `ctrl_tcp` case) or confusing behavior.
3. Update `module_preload` to allow pre-loading modules from the current working directory or from the configured module path. This allows preloading standard modules in the same way whether they're compiled as separate libs or statically.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1686,7 +1686,7 @@ int sip_req_send(struct ua *ua, const char *method, const char *uri,
 #endif
 
 
-int  module_preload(const char *module, const struct conf *cfg);
+int  module_preload(const char *module);
 int  module_load(const char *path, const char *name);
 void module_unload(const char *name);
 void module_app_unload(void);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1686,7 +1686,7 @@ int sip_req_send(struct ua *ua, const char *method, const char *uri,
 #endif
 
 
-int  module_preload(const char *module);
+int  module_preload(const char *module, const struct conf *cfg);
 int  module_load(const char *path, const char *name);
 void module_unload(const char *name);
 void module_app_unload(void);
--- a/src/main.c
+++ b/src/main.c
@@ -281,7 +281,7 @@ int main(int argc, char *argv[])
 
 		for (i=0; i<modc; i++) {
 
-			err = module_preload(modv[i]);
+			err = module_preload(modv[i], conf_cur());
 			if (err) {
 				re_fprintf(stderr,
 					   "could not pre-load module"
--- a/src/main.c
+++ b/src/main.c
@@ -267,7 +267,14 @@ int main(int argc, char *argv[])
 			      conf_config()->audio.audio_path);
 	}
 
-	/* NOTE: must be done after all arguments are processed */
+    /* Initialise User Agents */
+    err = ua_init(software, true, true, true);
+    if (err)
+        goto out;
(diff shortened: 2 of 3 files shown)


include/baresip.h
```diff
@@ -1686,7 +1686,7 @@ int sip_req_send(struct ua *ua, const char *method, const char *uri,
 #endif
 
 
-int  module_preload(const char *module);
+int  module_preload(const char *module, const struct conf *cfg);
```
alfredh, line 1689: perhaps the config can be fetched inside this function, so that we dont have to break the API ?

src/module.c
```diff
@@ -190,6 +197,15 @@ int module_preload(const char *module)
 	pl_set_str(&path, ".");
 	pl_set_str(&name, module);
 
+	char file[FS_PATH_MAX];
+	if (re_snprintf(file, sizeof(file), "%r/%r", &path, &name) < 0) {
```
alfredh, line 201: Perhaps better to use re_sdprint to avoid a large buffer on the stack ?

include/baresip.h
IkerAriz: Ah, sorry - didn't catch that `module_preload` was part of the public API. Will follow up. 

BTW, any interest in patches to add a C API that avoids global variables? (ie, by explicitly passing a "context" object around)

src/main.c
```diff
@@ -267,14 +267,21 @@ int main(int argc, char *argv[])
 			      conf_config()->audio.audio_path);
 	}
 
-	/* NOTE: must be done after all arguments are processed */
+	/* Initialise User Agents */
+	err = ua_init(software, true, true, true);
+	if (err)
+		goto out;
```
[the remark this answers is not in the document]
IkerAriz: I tested preloading a few others while making calls and found no issues. Eg, `cons`, `httpd`, `srtp`, `presence`, `l16`, `sndfile`.

src/module.c
IkerAriz: Will do.

IkerAriz: Addressed with commit `8e5b0bdf`.

include/baresip.h
IkerAriz: Addressed with commit `bbcb50cd`.

src/module.c
```diff
@@ -174,22 +179,37 @@ void module_app_unload(void)
 
 
 /**
- * Pre-load a module from the current working directory
+ * Pre-load a module. First check the current working directory
+ * then fall back to the configured module path
  *
  * @param module Module name including extension
+ * @param conf Config from which to obtain module path
```
alfredh, line 186: Please remove this line ...

alfredh:
I think this looks good now, and we can merge it to main ...

src/module.c
IkerAriz: Done

PR #3409, opened 2025-06-03
call: Get cancel reason from close message

KillingSpark:
Lib-re now passes a close message when the call has been canceled which means we can pass the reason line via the reason string with the `CALL_EVENT_CLOSED`

This goes towards the goal of providing better UI when a call gets cancelled

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -2104,19 +2104,17 @@ static void sipsess_close_handler(int err, const struct sip_msg *msg,
 
 	call_stream_stop(call);
 
-	if (msg) {
+	if (msg)
 		reason_hdr = sip_msg_hdr(msg, SIP_HDR_REASON);
-	}
 
 	if (reason_hdr) {
 		info("Cancel reason: %r\n", &reason_hdr->val);
-		call_event_handler(call, CALL_EVENT_CLOSED, "%s | %r",
+		call_event_handler(call, CALL_EVENT_CLOSED, "%s,%r",
 						   reason, &reason_hdr->val);
 	}
 	else {
 		call_event_handler(call, CALL_EVENT_CLOSED, "%s", reason);
 	}
-
 }
 
 


alfredh:
Many thanks

A new release is planned for tomorrow, so we will merge it after that

alfredh:
I think this PR looks fine. The build is failing due to windows-2019 deprecation.

@cspiel1 could you please also take a look ?

src/call.c
```diff
@@ -2102,7 +2103,20 @@ static void sipsess_close_handler(int err, const struct sip_msg *msg,
 		xfer_cleanup(call, reason);
 
 	call_stream_stop(call);
-	call_event_handler(call, CALL_EVENT_CLOSED, "%s", reason);
+
+	if (msg) {
+		reason_hdr = sip_msg_hdr(msg, SIP_HDR_REASON);
+	}
```
cspiel1: Often we omit the brackets if the body is a one liner.

src/call.c
```diff
@@ -2102,7 +2103,20 @@ static void sipsess_close_handler(int err, const struct sip_msg *msg,
...
+
+	if (reason_hdr) {
+		info("Cancel reason: %r\n", &reason_hdr->val);
+		call_event_handler(call, CALL_EVENT_CLOSED, "%s | %r",
+						   reason, &reason_hdr->val);
```
cspiel1, line 2113: Maybe we should use `"%s,%r"`. In `module_event()` we already use a comma as separater.

KillingSpark: I wanted that string to be parseable again in the upper layers. One could still look for `SIP ;cause=` and start parsing there if a comma is used as a separator. It just felt better to have the `|` in there so that upper layers know that there is a reason to be parsed in that string.

If you prefer the comma I'll change it, just let me know :)

cspiel1: Of course it should be parseable. I looked again for other use cases with separators. And found only `menu_event()` where we have the comma as separator. So, yes `,` is better.

bevents are serialized by API modules like `ctrl_tcp`, `ctrl_dbus` to JSON format. The `fmt` string will result in `"param":"<reason>,<reason_hdr>"`. So, if you like you could also make it a JSON object. E.g.
`{"local-reason":"%s","remote-reason":"%s"}`

KillingSpark: I'll change it to use the comma. Using json seems complicated, since I'd need to escape any quotes in the enclosed strings. Thanks for the quick feedback :)

PR #3425, opened 2025-06-16
menu: Add hangup sound on closed calls

dddaniel:
Gives the user the option to play an audio file when the call closes. This is a common way in many phones to provide the user an additional audio feedback when a call is terminated.

The new config option is called 'hangup_aufile' and is set to 'none' per default. Meaning there is no playback on hangup unless the user specifies a file.

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -248,6 +248,7 @@ video_selfview		window # {window,pip}
 #ring_aufile		ring.wav
 #callwaiting_aufile	callwaiting.wav
 #ringback_aufile	ringback.wav
+#hangup_aufile		none
 #notfound_aufile	notfound.wav
 #busy_aufile		busy.wav
 #error_aufile		error.wav
--- a/modules/menu/menu.c
+++ b/modules/menu/menu.c
@@ -551,6 +551,9 @@ static void menu_play_closed(struct call *call)
 
 		menu_play(call, key, fb, 1, DEVICE_ALERT);
 	}
+	else {
+		menu_play(call, "hangup_aufile", "none", 0, DEVICE_PLAYER);
+	}
 }
 
 
--- a/src/config.c
+++ b/src/config.c
@@ -1306,6 +1306,7 @@ int config_write_template(const char *file, const struct config *cfg)
 			"#sip_autoanswer_method\trfc5373 "
 			"# {rfc5373,call-info,alert-info}\n"
 			"#ring_aufile\t\tring.wav\n"
+			"#hangup_aufile\t\tnone\n"
 			"#callwaiting_aufile\tcallwaiting.wav\n"
 			"#ringback_aufile\tringback.wav\n"
 			"#notfound_aufile\tnotfound.wav\n"


docs/examples/config
```diff
@@ -248,6 +248,7 @@ video_selfview		window # {window,pip}
 #ring_aufile		ring.wav
 #callwaiting_aufile	callwaiting.wav
 #ringback_aufile	ringback.wav
+#hangup_aufile		none
```
cspiel1, line 251: Good!
But, for the module menu we also print the defaults to the template. Please see config.c how we did it for e.g. `ring_aufile`!

modules/menu/menu.c
```diff
@@ -881,6 +885,7 @@ static void event_handler(enum ua_event ev, struct bevent *event, void *arg)
 		if (call_is_outgoing(call))
 			--menu.outcnt;
 
+		play_hangup(call);
```
cspiel1: There is a collision with `menu_play_closed()`. We should discuss
- Currently if the current call terminates and there is an `scode`, then an error is played. Then the `hangup_aufile` should not be played in addition.
- What should be the expected behavior if multiple calls are handled?

cspiel1: Let me make a suggestion!
Would it be okay, to play the `hangup_aufile` only if the current call terminates and it is the one and only call?

If this would be acceptable for you, then try to solve this in function `menu_play_closed()` or/and in `errorcode_key_aufile()`!

The reason for this suggestion is that it could be annoying if a tone is played while talking to someone else.

dddaniel: > The reason for this suggestion is that it could be annoying if a tone is played while talking to someone else.

Makes sense, I did not think about multiple calls before

>If this would be acceptable for you, then try to solve this in function menu_play_closed() or/and in errorcode_key_aufile()!


Will try that.

dddaniel: >Currently if the current call terminates and there is an scode, then an error is played. Then the hangup_aufile should not be played in addition.

The hangup file is now only played when there is no other sip error

modules/menu/menu.c
```diff
@@ -550,6 +550,8 @@ static void menu_play_closed(struct call *call)
 		fb = errorcode_fb_aufile(scode);
 
 		menu_play(call, key, fb, 1, DEVICE_ALERT);
```
cspiel1, line 552: For `scode==487` there is still no tone. Is this expected?

modules/menu/menu.c
```diff
@@ -550,6 +550,8 @@ static void menu_play_closed(struct call *call)
...
+	} else {
```
cspiel1: See lint error!

modules/menu/menu.c
dddaniel: No, there should be a hangup tone for 487. I just did not recognize it due to my testing phone number always forwarded me directly to the mailbox when the call was rejected

modules/menu/menu.c
dddaniel: I don't really see the issue here. But the code might change anyway when I add support for the scode==487

modules/menu/menu.c
dddaniel: How about that @cspiel1 ? Now the hangup call is played back for 487 or when no sip code is availble.
Otherweise the error codes are played

cspiel1: Due to https://www.rfc-editor.org/rfc/rfc3261.html#page-91 Section 15.1.2 UAS behavior for incoming BYE, I think that it is best to hide a SIP 487 "Request Terminated" from the user.

Any SIP response after the BYE is not interesting for the user.

edit: Sorry, I was a few minutes too late. Please could you revert the last change?

cspiel1:
Please see: https://github.com/baresip/baresip/pull/3425#discussion_r2161080765 !

modules/menu/menu.c
cspiel1: An "else" has to be in the next line. "lint" means code formatting checker. See `test/ccheck.py`!

dddaniel: got it. Is changed now

modules/menu/menu.c
```diff
@@ -551,6 +551,9 @@ static void menu_play_closed(struct call *call)
 
 		menu_play(call, key, fb, 1, DEVICE_ALERT);
 	}
+	else {
+		menu_play(call, "hangup_aufile", "none", 0, DEVICE_PLAYER);
```
cspiel1, line 555: Are you/we sure that it should be played on the same device where the speech can be heard?

- DEVICE_PLAYER: speech, call-waiting, ringback (for outgoing calls)
- DEVICE_ALERT: incoming ringtone, error tones

E.g. when a headset is used for the call.

I would also suggest to play it on `DEVICE_PLAYER`.

This PR looks good now for me.

cspiel1:
@dddaniel thank you for the contribution!

PR #3426, opened 2025-06-17
uuid: check that uuid file contains valid uuid

maximilianfridrich:
Currently, if the uuid file gets corrupted, BareSIP gets stuck in a loop where it keeps trying to use an invalid uuid instead of generating a new one. Valid UUIDs are needed for example when SIP outbound is used with an account, so in this scenario, registrations to a SIP server are broken and stay broken until someone manually deletes the uuid file.

This is fixed now by checking that the uuid file contains a valid uuid.

### DIFF
--- a/modules/uuid/uuid.c
+++ b/modules/uuid/uuid.c
@@ -3,12 +3,27 @@
  *
  * Copyright (C) 2010 Alfred E. Heggestad
  */
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#ifdef HAVE_UNISTD_H
+#include <unistd.h>
+#endif
+#ifdef HAVE_IO_H
+#include <io.h>
+#endif
 #include <stdlib.h>
 #include <string.h>
 #include <stdio.h>
 #include <re.h>
 #include <baresip.h>
 
+
+#ifdef WIN32
+#define open _open
+#define read _read
+#define close _close
+#endif
 
 /**
  * @defgroup uuid uuid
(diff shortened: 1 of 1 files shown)


modules/uuid/uuid.c
```diff
@@ -20,6 +35,69 @@
...
+}
+
+
+/*
+ * Returns zero if provided file is present and contains a valid UUID.
+ * Returns non-zero in all other cases.
+ */
+static int check_uuid_file(const char *file)
+{
+	int fd = open(file, O_RDONLY);
```
sreimers, line 57: Maybe easier to use `fs_fread` for this.

maximilianfridrich: I wanted to make this as robust as possible. `fs_fread` for example does not have an upper bound on the file size. If somehow an attacker got control over this size, the attacker could DOS us by handing us a very large file. Further, the `fs_fopen` function also has some peculiarities that make it not quite as robust when an attacker might have control over that file.

sreimers: If a attacker has write access to the config directory I think you have more problems ;-)

maximilianfridrich: Okay, maybe I got a bit carried away on this one. I can probably simplify it.

maximilianfridrich: It's a bit simpler now. Is that okay?

alfredh:
The UUID module is currently unmaintained:

@maximilianfridrich it would be great if you could step up and volunteer to maintain this module :)

maximilianfridrich:
@alfredh No problem, I can do that :)

I cannot edit the baresip wiki pages, could you please put my name down there as maintainer of the UUID module?

cspiel1:
> @alfredh No problem, I can do that :)
> 
> I cannot edit the baresip wiki pages, could you please put my name down there as maintainer of the UUID module?

Done.

PR #3437, opened 2025-07-14
bevent: remove deprecated ua_event API

cspiel1:
improves also bevent tests

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -281,8 +281,6 @@ void call_start_answtmr(struct call *call, uint32_t ms);
 bool          call_supported(struct call *call, uint16_t tags);
 const char   *call_user_data(const struct call *call);
 int call_set_user_data(struct call *call, const char *user_data);
-void call_set_evstop(struct call *call, bool stop);
-bool call_is_evstop(struct call *call);
 int call_msg_src(const struct call *call, struct sa *sa);
 enum sip_transp call_transp(const struct call *call);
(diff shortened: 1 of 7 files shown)


juha-h:
`ua_event` API function is gone, but not `enum ua_event`.  Does not feel logical to me.

cspiel1:
> `ua_event` API function is gone, but not `enum ua_event`. Does not feel logical to me.

The enum has a central role for baresip events. It was requested by @alfredh 
 to keep it as first parameter for the event handler functions `bevent_h`: https://github.com/baresip/baresip/pull/3018#discussion_r1604499979

alfredh:
It might be possible to rename it to something with `bevent` or add a wrapper for it.

juha-h:
Alfred E. Heggestad writes:

> It might be possible to rename it to something with `bevent` or add a
> wrapper for it.

I thought that if `enum ua_event` could be renamed to `enum b_event`.

cspiel1:
I'll try to make a suggestion next week after my vacation.

alfredh:
it should be fine to rename the enum if that is wanted, please keep in mind that it will break
the public API. I am not sure if it is possible to add a wrapper to an enum...

cspiel1:
Maybe rename and backward compatibility is possible with an "implicit conversation" warning of applications that use the old `enum ua_event`. I am on it.

Should we merge this first in order to keep the PR and final commit small?

alfredh:
Sounds good to merge it now ...

PR #3441, opened 2025-07-22
srtp: fix size err printf

sreimers:
`Format: "srtp: srtp_tx not ready (%m<-- SIZE ERROR`


alfredh:
Thanks for the quick fix!

PR #3445, opened 2025-07-24
enum ua event rename

cspiel1:
- **bevent: rename enum ua_event without breaking API**

It would be really nice if we could rename the enum. But we should not break the API.

The backwards compatibility code is somehow complicated, but can be removed after a while (again one year?).

Is there a better solution? Or should we do without it?

### DIFF
--- a/src/bevent.c
+++ b/src/bevent.c
@@ -179,7 +179,7 @@ void *bevent_get_apparg(const struct bevent *event)
  *
  * @return the event type
  */
-enum bevent_id bevent_get_type(const struct bevent *event)
+enum bevent_id bevent_get_id(const struct bevent *event)
 {
 	return event ? event->ev : BEVENT_MAX;
 }
@@ -382,7 +382,7 @@ int odict_encode_bevent(struct odict *od, struct bevent *event)
 		mem_deref(buf);
 	}
 
-	enum bevent_id ev = bevent_get_type(event);
+	enum bevent_id ev = bevent_get_id(event);
 	err |= odict_entry_add(od, "type", ODICT_STRING, bevent_id_str(ev));
 	if (ua) {
 		err |= odict_entry_add(od, "accountaor",
--- a/src/bevent.c
+++ b/src/bevent.c
@@ -383,7 +383,7 @@ int odict_encode_bevent(struct odict *od, struct bevent *event)
 	}
 
 	enum bevent_id ev = bevent_get_type(event);
-	err |= odict_entry_add(od, "type", ODICT_STRING, uag_event_str(ev));
+	err |= odict_entry_add(od, "type", ODICT_STRING, bevent_id_str(ev));
 	if (ua) {
 		err |= odict_entry_add(od, "accountaor",
 				       ODICT_STRING,
@@ -720,7 +720,7 @@ int bevent_sip_msg_emit(enum bevent_id ev, const struct sip_msg *msg,
  *
  * @return Name of the event
  */
-const char *uag_event_str(enum bevent_id ev)
+const char *bevent_id_str(enum bevent_id ev)
 {
 	switch (ev) {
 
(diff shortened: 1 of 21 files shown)


cspiel1:
Todo: 
- Fix warnings in modules. -> hopefully will fix ci
- Test warnings with baresip-apps.

sreimers:
I'm not sure if maintaining backwards compatibility is worthwhile. There are also many API breakages within this PR. Renaming the enum might be less disruptive and cleaner. Maybe this a good time to increase the major version for the next release?

cspiel1:
For me it would be okay to simply rename the enum in the whole repository. A baresip-apps PR will follow. Then @juha-h needs to update also the Android app.

@alfredh and @juha-h do you agree with @sreimers suggestion? If so, I'll remove the compatibility code.

cspiel1:
The last commit removes the backward compatibility and fixes all compiler warnings.

src/bevent.c
```diff
@@ -382,7 +382,7 @@ int odict_encode_bevent(struct odict *od, struct bevent *event)
 		mem_deref(buf);
 	}
 
-	enum ua_event ev = bevent_get_type(event);
+	enum bevent_id ev = bevent_get_type(event);
```
cspiel1: Should be fitted. Which one should we rename?
- `enum bevent_id` -> `enum bevent_type` or
-  `bevent_get_type(event)` -> `bevent_get_id(event)`

Other suggestions?

juha-h:
> For me it would be okay to simply rename the enum in the whole repository. A baresip-apps PR will follow. Then @juha-h needs to update also the Android app.
> 
> @alfredh and @juha-h do you agree with @sreimers suggestion? If so, I'll remove the compatibility code.

FIne with me.

src/bevent.c
cspiel1: Some suggestions:
- `bevent_id`, like now in this PR.
- `bevent_cat` for category
- `bevent_type`
- `bevent_ev` which matches `ev` in `struct  bevent`. `ev` is used very often, but we could change also this.

alfredh:
Bumping the major version number on the next release is a good idea ...

cspiel1:
> Bumping the major version number on the next release is a good idea ...

Okay, the August release is mine I think. Will do this.

@alfredh should I change the name of the enum to one of https://github.com/baresip/baresip/pull/3445#discussion_r2230489791?

juha-h:
If possible, could you wait until tomorrow with merging of this commit, since F-Droid is about to build new version of my baresip app and the build would fail if the PR is merged.

cspiel1:
There is still no decision for the enum name. I am not so happy with the name inconsistency `enum bevent_id ev`. And `ev` is used very often. So, let's rename the enum to
`enum bevent_ev` with doxygen
`@param ev Baresip event value` and
```
struct bevent {
    enum bevent_ev ev;
    ...
};
```
and the getter
```
enum bevent_ev bevent_get_value(const struct bevent *event);
```

cspiel1:
Are there any wishes for the naming or other suggestions? If not, this PR is ready for the merge.

alfredh:
I dont have any strong opinions about the naming, the PR looks fine to me.

For reference other `enum` definitions in the codebase are using `type` as a suffix:

```
include/baresip.h:76:enum jbuf_type {
include/baresip.h:1763:enum sdp_type {
src/core.h:298:enum media_type {
```

cspiel1:
Then let's keep the shorter `ev`! E.g. one line
```c
static void event_handler(enum bevent_ev ev, struct bevent *event, void *arg)
```
vs two line
```c
static void event_handler(enum bevent_type type, struct bevent *event,
			  void *arg)
```

cspiel1:
Now I proceed with a PR for baresip-apps. Applications can be ported with sed:
```sh
sed -i -e "s/enum ua_event/enum bevent_ev/g" modules/*/*{c,h}
sed -i -e "s/UA_EVENT_/BEVENT_/g" modules/*/*c
sed -i -e "s/uag_event_str/bevent_str/g" modules/*/*c
```

PR #3450, opened 2025-07-30
in_band_dtmf: fix cross compiler alignment warning

cspiel1:
With arm gcc cross compiler there was a warning
```
src/baresip/modules/in_band_dtmf/in_band_dtmf.c:173:44: warning: cast increases required alignment of target type [-Wcast-align]
  173 |         struct in_band_dtmf_filt_dec *sf = (struct in_band_dtmf_filt_dec *)st;
      |
```


cspiel1:
Can we merge this? I am not sure if a cast to `void *` is the best.

PR #3455, opened 2025-08-06
package: Fix not working libdir

danielroedl:
Because of a typo in the `packaging/libbaresip.pc.in` file the libdir for baresip package is not set correctly. This brings issues using baresip in CMake when it's not installed to system paths but to local paths.


sreimers:
Thanks! You can ignore the failed Windows checks. It will take some time for the new OpenSSL release to be approved.

cspiel1:
Do we need a hotfix release with this fix?

sreimers:
I think, next regular release is fine.

PR #3460, opened 2025-08-12
audio: a config parameter for tx audio buffer size

cspiel1:
- use tx audio_buffer_tx range setting instead of hard coded 30
- the default max now can be reduced

### DIFF
--- a/docs/examples/config
+++ b/docs/examples/config
@@ -52,6 +52,7 @@ auplay_format		s16		# s16, float, ..
 auenc_format		s16		# s16, float, ..
 audec_format		s16		# s16, float, ..
 audio_buffer		20-160		# ms
+#audio_buffer_tx	20-120		# ms
 audio_buffer_mode	fixed		# fixed, adaptive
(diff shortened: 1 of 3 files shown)


alfredh:
this buffer is for packetizing outgoing audio packets that has to be e.g. 20ms
Some audio drivers gives audio frames that are too small, e.g. 5ms, so we have to buffer them.
Some audio drivers gives audio frames that are too large, e.g. 50ms

The code should send out the prepared audio packets as quickly as possible,
perhaps add a short delay of 1-2 ms to avoid network spikes.

The transmit buffer should not add any delay, nor should it drop any packets.
If the maximum size is e.g. 1 second and the audio driver exceeds that, there is something
wrong with the audio driver or device.

It would be more user-friendly if the user does not have to worry about this buffer size,
and that the buffer size is automatically adjusted.

cspiel1:
The reason was to avoid oom issues on very limited devices. e.g. 16kHz 30 packets with 20ms needs ~19kB.

I close this because this should not be a problem nowadays.

PR #3461, opened 2025-08-12
module: add warning if static module could not be loaded

cspiel1:


src/module.c
```diff
@@ -81,6 +81,9 @@ static int load_module(struct mod **modp, const struct pl *modpath,
 		err = mod_add(&m, mex);
 		goto out;
 	}
+	else {
+		warning("could not load static module %r\n", name);
```
sreimers: Not sure, you can have a mixed setup with static and dynamic modules.

cspiel1:
clang analyzer tells "Value stored to 'err' is never read". And it's true. ;-)

PR #3462, opened 2025-08-12
A flag for unique contact user

cspiel1:
Discussion: https://github.com/baresip/baresip/discussions/3149

- **account,ua: add flag cuser_ua for unique contact user**

    A new accounts flag `cuser_ua` (by default true) can be used to disable
    the UA pointer suffix for the contact user. The suffix makes the contact
    user unique and thus the UA selection for incoming SIP requests stable
    against chosing the wrong, if there are multiple accounts with equal AOR
    users at different registrars.

```
alice@iptel.org
alice@sipgate.com
``` 
 This setting fixes registration issues on a Siemens HIpath and 1und1.de (see https://github.com/baresip/baresip/discussions/3149)

```
<sip:alice@1und1.de>;cuser_ua=no
```

- **uag: rework for cuser_ua is false**

    The first loop in `uag_find_msg()` compares only the contact user which
    works only reliable if the contact user is unique. For `cuser_ua==false` the
    second loop uses better checks.

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -1206,7 +1206,7 @@ int ua_alloc(struct ua **uap, const char *aor)
 		goto out;
 
 
-	if (ua->acc->cuser_ua && list_count(uag_list()) > 0) {
+	if (ua->acc->cuser_ua) {
 		/* generate a unique contact-user, this is needed to route
 		   incoming requests when using multiple useragents */
 		err = re_sdprintf(&ua->cuser, "%r-%z", &ua->acc->luri.user,
--- a/src/ua.c
+++ b/src/ua.c
@@ -1206,11 +1206,11 @@ int ua_alloc(struct ua **uap, const char *aor)
 		goto out;
 
 
-	if (ua->acc->cuser_ua) {
+	if (ua->acc->cuser_ua && list_count(uag_list()) > 0) {
 		/* generate a unique contact-user, this is needed to route
 		   incoming requests when using multiple useragents */
-		err = re_sdprintf(&ua->cuser, "%r-%p", &ua->acc->luri.user,
-				  ua);
+		err = re_sdprintf(&ua->cuser, "%r-%z", &ua->acc->luri.user,
+				  list_count(uag_list()) + 1);
 	}
 	else {
 		err = re_sdprintf(&ua->cuser, "%r", &ua->acc->luri.user);
--- a/src/ua.c
+++ b/src/ua.c
@@ -1206,9 +1206,16 @@ int ua_alloc(struct ua **uap, const char *aor)
 		goto out;
 
 
-	/* generate a unique contact-user, this is needed to route
-	   incoming requests when using multiple useragents */
-	err = re_sdprintf(&ua->cuser, "%r-%p", &ua->acc->luri.user, ua);
+	if (ua->acc->cuser_ua) {
+		/* generate a unique contact-user, this is needed to route
+		   incoming requests when using multiple useragents */
+		err = re_sdprintf(&ua->cuser, "%r-%p", &ua->acc->luri.user,
+				  ua);
+	}
+	else {
+		err = re_sdprintf(&ua->cuser, "%r", &ua->acc->luri.user);
+	}
+
 	if (err)
 		goto out;
 
--- a/docs/examples/accounts
+++ b/docs/examples/accounts
@@ -22,6 +22,7 @@
 #    ;call_transfer=no
 #    ;catchall={yes, no}  # default: no
 #    ;cert=cert.pem
+#    ;cuser_ua={yes, no}  # default: yes
 #    ;mediaenc={srtp,srtp-mand,srtp-mandf,dtls_srtp,zrtp}
 #    ;medianat={stun,turn,ice}
 #    ;natpinhole={yes, no}  # default: no
--- a/modules/account/account.c
+++ b/modules/account/account.c
@@ -63,6 +63,7 @@ static int account_write_template(const char *file)
 			 "#    ;auth_pass=password\n"
 			 "#    ;call_transfer=no\n"
 			 "#    ;cert=cert.pem\n"
+			 "#    ;cuser_ua={yes, no}  # default: yes\n"
 			 "#    ;mediaenc={srtp,srtp-mand,srtp-mandf"
 			 ",dtls_srtp,zrtp}\n"
 			 "#    ;medianat={stun,turn,ice}\n"
--- a/src/account.c
+++ b/src/account.c
@@ -512,6 +512,9 @@ static int sip_params_decode(struct account *acc, const struct sip_addr *aor)
 
 	err |= param_dstr(&acc->regq, &aor->params, "regq");
 
+	acc->cuser_ua = true;
+	err |= param_bool(&acc->cuser_ua, &aor->params, "cuser_ua");
+
 	for (i=0; i<RE_ARRAY_SIZE(acc->outboundv); i++) {
 
 		char expr[16] = "outbound";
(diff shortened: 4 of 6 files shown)


maximilianfridrich:
The `cuser` is sent over the network in SIP messages (e.g. in the Contact header) and it contains a memory address of the device. From a security perspective, this is not ideal - information about internal memory addresses should not be leaked (e.g. it would render ASLR pointless and could aid other attacks). Maybe this could be a good time to change this and make the suffix random for example.

cspiel1:
> The `cuser` is sent over the network in SIP messages (e.g. in the Contact header) and it contains a memory address of the device. From a security perspective, this is not ideal - information about internal memory addresses should not be leaked (e.g. it would render ASLR pointless and could aid other attacks). Maybe this could be a good time to change this and make the suffix random for example.

I'll add a commit.

src/ua.c
```diff
@@ -1206,9 +1206,16 @@ int ua_alloc(struct ua **uap, const char *aor)
...
 
 
-	/* generate a unique contact-user, this is needed to route
-	   incoming requests when using multiple useragents */
-	err = re_sdprintf(&ua->cuser, "%r-%p", &ua->acc->luri.user, ua);
+	if (ua->acc->cuser_ua) {
+		/* generate a unique contact-user, this is needed to route
+		   incoming requests when using multiple useragents */
+		err = re_sdprintf(&ua->cuser, "%r-%p", &ua->acc->luri.user,
+				  ua);
```
alfredh: perhaps this pointer could be obfuscated or use a random number / index instead ?

alfredh:
I agree that sending the pointer on the network is bad.

We could make a different approach, where the contact user is different depending on how many
accounts are configured. This contact username is used to map incoming SIP requests to the
correct account.

If the number of accounts is one, which is 99%, you can use the default username

If the number of accounts is more than one, we can add a account number as a trailing suffix.

For example:

```
1: username@domain
2. username-2@domain
3. username-3@domain
4. username-4@domain
5. etc...
```

Please be aware that two or more accounts will not work with 1und1

cspiel1:
> For example:
> 
> ```
> 1: username@domain
> 2. username-2@domain
> 3. username-3@domain
> 4. username-4@domain
> 5. etc...
> ```
> 
> Please be aware that two or more accounts will not work with 1und1

Changed to a counter. But let's keep the new flag for users who want more than one accounts, maybe on different servers!

alfredh:
Lets consider this example accounts file:

```
<sip:1234@iptel.org>
<sip:1234@iptel.org>
```

For incoming calls, both line 1 and line 2 will ring.
If line 1 is busy, line 2 will ring.

Why does a user want to configure two accounts for the same SIP account ?

juha-h:
I'm wondering the same.

cspiel1:
> Lets consider this example accounts file:
> 
> ```
> <sip:1234@iptel.org>
> <sip:1234@iptel.org>
> ```
> 
> For incoming calls, both line 1 and line 2 will ring. If line 1 is busy, line 2 will ring.
> 
> Why does a user want to configure two accounts for the same SIP account ?

You are right that this example doesn't make sense. But this has to work:

```
<sip:alice@iptel.org>
<sip:alice@1und1.de>
```

@maximilianfridrich and me studied again RFC-3261. Section "8.2.2.1 To and Request-URI" tells that the To header should not be used to chose the right UA. Only the Request-URI should be used for that. But the host part of the Request-URI is the local IP address and thus only the contact user remains for the match. Thus, the contact user has to be unique.

We further found that in order to make this
```
<sip:alice@iptel.org>
<sip:alice@1und1.de>;cuser_ua=no
```
 example work, the first UA also needs a suffix. I'll change this.

juha-h:
I still don't understand.   In the example above, iptel.org or 1und1.org in Request-URI chooses for which UA the request is for.

cspiel1:
With last push force I only updated the commit message of the first commit.

Are there any wishes to change the name of the setting?
- `cuser_ua`
- `cuser_unique`
- `unique_cuser`
- `unique_cu`

cspiel1:
> I still don't understand. In the example above, iptel.org or 1und1.org in Request-URI chooses for which UA the request is for.

The point is that `iptel.org` or any domain will never be in the Request-URI. There is often or always an IP address. Our local IP address, which doesn't carry any information to decide anything.

Edit: If I understand @alfredh and @juha-h questions correclty, we are discussing now if we can drop the suffix completely. We gave you arguments, that we should keep it in general. But this PR adds a setting to omit the suffix to make it compatible to some special SIP registrars.

sreimers:
Maybe it's option to use a suffix only if the `user` part is not unique (to avoid a new option):

```
<sip:alice@iptel.org> -> alice
<sip:alice@example.net> -> alice-01
<sip:49xxxxxxxxxx@1und1.net> -> 49xxxxxxxxxx
```

Since `ua_alloc` is only called once per account a simple loop `uag_list` check to identify should be fine (with a global suffix counter):

```
<sip:alice@iptel.org> -> alice
<sip:alice@example.net> -> alice-01
<sip:bob@iptel.org> -> bob
<sip:bob@example.org> -> bob-02
<sip:49xxxxxxxxxx@1und1.net> -> 49xxxxxxxxxx
```

juha-h:
> The point is that iptel.org or any domain will never be in the Request-URI. There is often or always an IP address. Our local IP address, which doesn't carry any information to decide anything.

This doesn't make any sense.  If the server serves more than one domain, Request-URI must include a domain of the user.

maximilianfridrich:
>  If the server serves more than one domain, Request-URI must include a domain of the user.

That would make sense, but we're thinking of the case where we are registered at two different servers with (two different domains).

In BareSIP, the matching of an incoming INVITE to a UA account [only looks at the user part](https://github.com/baresip/baresip/blob/main/src/uag.c#L897) of the Request-URI and for a good reason. We cannot know if the peer (i.e. server) will have the host part of the URI set to a FQDN or IP address or something else - see [RFC 3261 section 8.2.2.1](https://datatracker.ietf.org/doc/html/rfc3261#section-8.2.2.1) and the examples (esp. message F5) in [section 24.2](https://datatracker.ietf.org/doc/html/rfc3261#section-24.2):

```
8.2.2.1 To and Request-URI
...
   However, the Request-URI identifies the UAS that is to process the
   request.
   ...
   If the Request-URI does not identify an address that the
   UAS is willing to accept requests for, it SHOULD reject the request
   with a 404 (Not Found) response.  Typically, a UA that uses the
   REGISTER method to bind its address-of-record to a specific contact
   address will see requests whose Request-URI equals that contact
   address.

24.2 Session Setup
...
F5 INVITE biloxi.com proxy -> Bob

INVITE sip:bob@192.0.2.4 SIP/2.0
Via: SIP/2.0/UDP server10.biloxi.com;branch=z9hG4bK4b43c2ff8.1
Via: SIP/2.0/UDP bigbox3.site3.atlanta.com;branch=z9hG4bK77ef4c2312983.1
 ;received=192.0.2.2
Via: SIP/2.0/UDP pc33.atlanta.com;branch=z9hG4bKnashds8
 ;received=192.0.2.1
Max-Forwards: 68
To: Bob <sip:bob@biloxi.com>
From: Alice <sip:alice@atlanta.com>;tag=1928301774
Call-ID: a84b4c76e66710
CSeq: 314159 INVITE
Contact: <sip:alice@pc33.atlanta.com>
Content-Type: application/sdp
Content-Length: 142
```

Ideally, the incoming INVITE will match what we sent in the Contact header of our REGISTER, but this is not required by the RFC.

juha-h:
When baresip UA registers with a SIP Proxy/Registrar, it uses Contact URI that is unique for that UA.  If baresip has two UAs with equal URI userparts, but different domains, Contact URIs will differ and based on that baresip knows for which UA the incoming request is for.

cspiel1:
> Maybe it's option to use a suffix only if the `user` part is not unique (to avoid a new option):
> 
> ```
> <sip:alice@iptel.org> -> alice
> <sip:alice@example.net> -> alice-01
> <sip:49xxxxxxxxxx@1und1.net> -> 49xxxxxxxxxx
> ```
> 
> Since `ua_alloc` is only called once per account a simple loop `uag_list` check to identify should be fine (with a global suffix counter):
> 
> ```
> <sip:alice@iptel.org> -> alice
> <sip:alice@example.net> -> alice-01
> <sip:bob@iptel.org> -> bob
> <sip:bob@example.org> -> bob-02
> <sip:49xxxxxxxxxx@1und1.net> -> 49xxxxxxxxxx
> ```

@maximilianfridrich and me discussed this proposition and found that with an entry in the user manual of our devices the result will be better than with the added config option. I'll start with a fresh PR.

Issue #350, opened 2018-01-24
g722 codec implementation


sreimers:
Hi Alfred, at the moment the g722 module needs the external spandsp library. But the spandsp author has released the g722 code under a separate license, so maybe we can import them like asterisk directly in baresip or rem:

https://github.com/asterisk/asterisk/tree/master/codecs/g722

```
/*
 * SpanDSP - a series of DSP components for telephony
 *
 * g722_decode.c - The ITU G.722 codec, decode part.
 *
 * Written by Steve Underwood <steveu@coppice.org>
 *
 * Copyright (C) 2005 Steve Underwood
 *
 *  Despite my general liking of the GPL, I place my own contributions
 *  to this code in the public domain for the benefit of all mankind -
 *  even the slimy ones who might try to proprietize my work and use it
 *  to my detriment.
 *
 * Based in part on a single channel G.722 codec which is:
 *
 * Copyright (c) CMU 1993
 * Computer Science, Speech Group
 * Chengxiang Lu and Alex Hauptmann
 *
 * $Id$
 */
```

alfredh:
I think we should not import source code project other projects, unless no other
option exist.

What happens after the code is imported/copied, is in practice that the code is forked.
From this moment we have to maintain 2 projects and sync features/bugfixes between
them. After some years the codes are completely out of sync, and merging between
them becomes impossible.

For example distributions like Debian and others have the `spandsp` package which
is properly maintained. For the baresip project it is very easy and convenient to
link to `-lspandsp` and have a package dependency to spandsp.

We could of course make an exception for g722, do you have specific usecase for that?

We should also ask Steve Underwood about his opinion of this.


Alfred

sreimers:
My problem with spandsp is the license, the GPL forbids static linking with licenses that are not compatible  (like BSD/MIT) with my understanding of the law restrictions. Dynamic linking and distribution is no option for my use case (easy execution, without a installer on all platforms). I understand your argumentation and this is why I ask before and put work into a pull request. So I can fork the g722 by myself and implement it this way.

alfredh:
https://www.soft-switch.org/installing-spandsp.html

```
Note than recent versions of SpanDSP are licenced under LGPL 2.1. Earlier versions were licenced under GPL 2.0.
```

oej:
Even if you fork, the license does not change. In fact, running Baresip with old versions of SpanDSP makes your baresip GPL licensed. LGPL is a better choice.

[...]

PR #3467, opened 2025-08-13
g722: add libg722 as alternative to avoid spandsp dependency

fosple:
This PR adds support for using **libg722** as an alternative G.722 codec backend in addition to the existing g722 implementation which depends on spandsp.

### Key changes

- Added `FindLIBG722.cmake` to locate the library and headers.
- Updated `modules/g722/CMakeLists.txt` to:
    - Prefer SPANDSP if available, but fall back to libg722 if not.
    - Allow building the G.722 module without a SPANDSP dependency.
- Modified `modules/g722/g722.c` to:
    - Support both SPANDSP and libg722 APIs.
    - Add runtime and compile-time selection of the codec backend.
    - Implement proper resource cleanup for libg722.

### Motivation

SPANDSP is licensed under LGPL, which complicates static linking (e.g., for iOS App Store submissions).  
libg722 is in the public domain, making it a lightweight and license-friendly alternative.

This change enables developers to:
- Avoid LGPL restrictions where necessary.
- Build baresip with G.722 support even without SPANDSP installed.

### References
- Discussion: https://github.com/baresip/baresip/discussions/3465
- Related past issue: https://github.com/baresip/baresip/issues/350
- libg722 repository: https://github.com/sippy/libg722

### DIFF
--- a/cmake/FindLIBG722.cmake
+++ b/cmake/FindLIBG722.cmake
@@ -1,0 +1,40 @@
+# First try to use the CMake variables set by contrib.mk
+if(DEFINED G722_INCLUDE_DIR AND DEFINED G722_LIBRARY)
+  set(LIBG722_INCLUDE_DIR ${G722_INCLUDE_DIR})
+  set(LIBG722_LIBRARY ${G722_LIBRARY})
+  set(LIBG722_FOUND TRUE)
+else()
+  # Fallback to searching in standard locations
+  find_path(LIBG722_INCLUDE_DIR
+    NAME g722_codec.h
+    HINTS
+      "${LIBG722_INCLUDE_DIRS}"
+      "${LIBG722_HINTS}/include"
+    PATHS /usr/local/include /usr/include
+  )
+
+  find_library(LIBG722_LIBRARY
+    NAME g722
+    HINTS
+      "${LIBG722_LIBRARY_DIRS}"
+      "${LIBG722_HINTS}/lib"
+    PATHS /usr/local/lib /usr/lib
+  )
+endif()
+
+# Only use find_package_handle_standard_args if we haven't already found the library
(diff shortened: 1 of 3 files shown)


cspiel1:
Should we drop spandsp completely, or mark it as deprecated for a while?

sreimers:
> Should we drop spandsp completely, or mark it as deprecated for a while?

As far as I know, libg722 has no packages for most linux distributions, so keeping spandsp support is necessary I think.

maximilianfridrich:
[Alfred said the same in 2018](https://github.com/baresip/baresip/issues/350#issuecomment-361692263):

> For example distributions like Debian and others have the spandsp package which
is properly maintained. For the baresip project it is very easy and convenient to
link to -lspandsp and have a package dependency to spandsp.

Dynamic linking is okay with LGPL libraries. Users who want to statically link just have to be careful to not violate the LGPL or use libg722.

alfredh:
Would it be an option to create a new module calleg e.g. `libg722` instead ?

fosple:
I'm happy to change it to a separate module `libg722` - I would also say this is the best solution. If you agree @sreimers and @cspiel1 I can make a suggestion?

cspiel1:
> I'm happy to change it to a separate module `libg722` - I would also say this is the best solution. If you agree @sreimers and @cspiel1 I can make a suggestion?

Yes, please. Sorry for the late answer! Maybe its best to close this PR and open a new one. But its your choice.

juha-h:
@fosple Are you still planning to create PR for new g722 that does not depend on spandsp?  Such a module would be also good for Android, since spandsp depends on libtiff and now also on libjpeg.

fosple:
@juha-h Thanks for reminding me. Added the PR: https://github.com/baresip/baresip/pull/3545

PR #3470, opened 2025-08-19
ua: avoid suffix for contact user if possible

cspiel1:
This fixes issues with some SIP registrars like Siemens HIpath or
1und1.de. Furhter there had been problems with some Session Border
Controller (SBC) if the contact user and the registration user differ.

Proposed by @sreimers: https://github.com/baresip/baresip/pull/3462#issuecomment-3196765546

replaces: https://github.com/baresip/baresip/pull/3462

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -1189,6 +1189,7 @@ int ua_alloc(struct ua **uap, const char *aor)
 	struct uri *luri;
 	char *buf = NULL;
 	char *host = NULL;
+	static uint32_t ua_suffix = 0;
 	int err;
 
 	if (!aor)
@@ -1220,7 +1221,7 @@ int ua_alloc(struct ua **uap, const char *aor)
 		/* generate a unique contact-user, this is needed to route
 		 * incoming requests when using multiple useragents */
 		err = re_sdprintf(&ua->cuser, "%r-%u", &ua->acc->luri.user,
-				  list_count(uag_list()));
+				  ++ua_suffix);
 	}
 	else {
 		err = pl_strdup(&ua->cuser, &ua->acc->luri.user);
--- a/src/ua.c
+++ b/src/ua.c
@@ -1166,7 +1166,7 @@ static const char *autoans_header_name(enum answer_method met)
 }
 
 
-static bool find_user(struct le *le, void *arg)
+static bool user_cmp_handler(struct le *le, void *arg)
 {
 	struct ua *ua = le->data;
 	struct pl *user = arg;
@@ -1215,7 +1215,8 @@ int ua_alloc(struct ua **uap, const char *aor)
 		goto out;
 
 
-	if (list_apply(uag_list(), true, find_user, &ua->acc->luri.user)) {
+	if (list_apply(uag_list(), true, user_cmp_handler,
+		       &ua->acc->luri.user)) {
 		/* generate a unique contact-user, this is needed to route
 		 * incoming requests when using multiple useragents */
 		err = re_sdprintf(&ua->cuser, "%r-%u", &ua->acc->luri.user,
--- a/src/ua.c
+++ b/src/ua.c
@@ -1189,7 +1189,6 @@ int ua_alloc(struct ua **uap, const char *aor)
 	struct uri *luri;
 	char *buf = NULL;
 	char *host = NULL;
-	static uint32_t ua_suffix = 0;
 	int err;
 
 	if (!aor)
@@ -1221,7 +1220,7 @@ int ua_alloc(struct ua **uap, const char *aor)
 		/* generate a unique contact-user, this is needed to route
 		 * incoming requests when using multiple useragents */
 		err = re_sdprintf(&ua->cuser, "%r-%u", &ua->acc->luri.user,
-				  ++ua_suffix);
+				  uag_unique());
 	}
 	else {
 		err = pl_strdup(&ua->cuser, &ua->acc->luri.user);
--- a/src/ua.c
+++ b/src/ua.c
@@ -1175,6 +1175,32 @@ static bool user_cmp_handler(struct le *le, void *arg)
 }
 
 
+static int ua_cuser_gen(struct ua *ua)
+{
+	bool suffix = false;
+	int err;
+
+	conf_get_bool(conf_cur(), "cuser_random", &suffix);
+	if (suffix) {
+		err = re_sdprintf(&ua->cuser, "%r-%u", &ua->acc->luri.user,
+				  rand_u32());
+	}
+	else if (list_apply(uag_list(), true, user_cmp_handler,
+		       &ua->acc->luri.user)) {
+		/* generate a unique contact-user, this is needed to route
+		 * incoming requests when using multiple useragents */
+		err = re_sdprintf(&ua->cuser, "%r-%u", &ua->acc->luri.user,
+				  uag_unique());
+	}
+	else {
+		err = pl_strdup(&ua->cuser, &ua->acc->luri.user);
+	}
+
+	debug("ua: contact user %s\n", ua->cuser);
+	return err;
+}
+
+
 /**
  * Allocate a SIP User-Agent
  *
@@ -1214,18 +1240,7 @@ int ua_alloc(struct ua **uap, const char *aor)
 	if (err)
 		goto out;
 
-
-	if (list_apply(uag_list(), true, user_cmp_handler,
-		       &ua->acc->luri.user)) {
-		/* generate a unique contact-user, this is needed to route
-		 * incoming requests when using multiple useragents */
-		err = re_sdprintf(&ua->cuser, "%r-%u", &ua->acc->luri.user,
-				  uag_unique());
-	}
-	else {
-		err = pl_strdup(&ua->cuser, &ua->acc->luri.user);
-	}
-
+	err = ua_cuser_gen(ua);
 	if (err)
 		goto out;
 
--- a/src/ua.c
+++ b/src/ua.c
@@ -1166,6 +1166,15 @@ static const char *autoans_header_name(enum answer_method met)
 }
 
 
+static bool find_user(struct le *le, void *arg)
+{
+	struct ua *ua = le->data;
+	struct pl *user = arg;
+
+	return pl_cmp(&ua->acc->luri.user, user) == 0;
+}
+
+
 /**
  * Allocate a SIP User-Agent
  *
@@ -1206,9 +1215,16 @@ int ua_alloc(struct ua **uap, const char *aor)
 		goto out;
 
 
-	/* generate a unique contact-user, this is needed to route
-	   incoming requests when using multiple useragents */
-	err = re_sdprintf(&ua->cuser, "%r-%p", &ua->acc->luri.user, ua);
+	if (list_apply(uag_list(), true, find_user, &ua->acc->luri.user)) {
+		/* generate a unique contact-user, this is needed to route
+		 * incoming requests when using multiple useragents */
+		err = re_sdprintf(&ua->cuser, "%r-%u", &ua->acc->luri.user,
+				  list_count(uag_list()));
+	}
+	else {
+		err = pl_strdup(&ua->cuser, &ua->acc->luri.user);
+	}
+
 	if (err)
 		goto out;
 
--- a/src/ua.c
+++ b/src/ua.c
@@ -1181,16 +1181,14 @@ static int ua_cuser_gen(struct ua *ua)
 	int err;
 
 	conf_get_bool(conf_cur(), "cuser_random", &suffix);
+
+	suffix |= list_apply(uag_list(), true, user_cmp_handler,
+			     &ua->acc->luri.user) != NULL;
 	if (suffix) {
-		err = re_sdprintf(&ua->cuser, "%r-%u", &ua->acc->luri.user,
-				  rand_u32());
-	}
-	else if (list_apply(uag_list(), true, user_cmp_handler,
-		       &ua->acc->luri.user)) {
-		/* generate a unique contact-user, this is needed to route
-		 * incoming requests when using multiple useragents */
-		err = re_sdprintf(&ua->cuser, "%r-%u", &ua->acc->luri.user,
-				  uag_unique());
+		char buf[16];
+		rand_str(buf, sizeof(buf));
+		err = re_sdprintf(&ua->cuser, "%r-%s", &ua->acc->luri.user,
+				  buf);
 	}
 	else {
 		err = pl_strdup(&ua->cuser, &ua->acc->luri.user);
--- a/src/uag.c
+++ b/src/uag.c
@@ -1270,3 +1270,14 @@ sip_msg_h *uag_subh(void)
 {
 	return uag.subh;
 }
+
+
+/**
+ * Increments the unique counter uag.unique
+ *
+ * @return Unique counter
+ */
+uint32_t uag_unique(void)
+{
+	return ++uag.unique;
+}
--- a/src/uag.c
+++ b/src/uag.c
@@ -906,6 +906,10 @@ struct ua *uag_find_msg(const struct sip_msg *msg)
 	cuser = &msg->uri.user;
 	for (le = uag.ual.head; le; le = le->next) {
 		struct ua *ua = le->data;
+		struct account *acc = ua_account(ua);
+
+		if (!acc || !acc->regint)
+			continue;
 
 		if (0 == pl_strcasecmp(cuser, ua_local_cuser(ua))) {
 			ua_printf(ua, "selected for %r\n", cuser);
--- a/src/uag.c
+++ b/src/uag.c
@@ -1270,14 +1270,3 @@ sip_msg_h *uag_subh(void)
 {
 	return uag.subh;
 }
-
-
-/**
- * Increments the unique counter uag.unique
- *
- * @return Unique counter
- */
-uint32_t uag_unique(void)
-{
-	return ++uag.unique;
-}
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1007,6 +1007,7 @@ struct sip  *uag_sip(void);
 struct list *uag_list(void);
 uint32_t     uag_call_count(void);
 struct tls  *uag_tls(void);
+uint32_t     uag_unique(void);
 struct sipsess_sock  *uag_sipsess_sock(void);
 struct sipevent_sock *uag_sipevent_sock(void);
 struct call *uag_call_find(const char *id);
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1007,7 +1007,6 @@ struct sip  *uag_sip(void);
 struct list *uag_list(void);
 uint32_t     uag_call_count(void);
 struct tls  *uag_tls(void);
-uint32_t     uag_unique(void);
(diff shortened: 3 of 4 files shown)


juha-h:
Why should baresip be complicated by issues in other software?  Why don't you report the issues with them instead?

cspiel1:
> Why should baresip be complicated by issues in other software? Why don't you report the issues with them instead?

It is not so clear that this is an issue at "other software". And we don't think that this patch is so complicated.

sreimers:
And because they are won't change anything for years now, but we got issue reports multiple times to handle this. From a user perspective, this is very frustrating, since there is no workaround, other than patching baresip. ([1und1 is a big provider in germany and the "issue" exist over decade now](https://lists.freeswitch.org/pipermail/freeswitch-users/2012-July/086093.html)).

cspiel1:
I have to find now why `test_call_deny_udp()` fails. The `transport` parameter is twice:
```
test: [ sip:a@127.0.0.1 ] event: CALL_ESTABLISHED (sip:b@127.0.0.1:54940;transport=udp;transport=tcp)
```
But the same in `main` where the test is green. -> So, another issue that maybe is not a problem.

The `fix.a.n_established` is `1` but has to be zero.

juha-h:
See this re patch:
```
diff --git a/src/sip/auth.c b/src/sip/auth.c
index 70681e4..f341312 100644
--- a/src/sip/auth.c
+++ b/src/sip/auth.c
@@ -166,7 +166,9 @@ static bool auth_handler(const struct sip_hdr *hdr, const struct si
p_msg *msg,
 			goto out;
 	}
 	else {
-		if (!pl_isset(&ch.stale) || pl_strcasecmp(&ch.stale, "true")) {
+		/* error if first auth attempt fails */
+		if ((!pl_isset(&ch.stale) ||
+		     pl_strcasecmp(&ch.stale, "true")) && (realm->nc == 2)) {
 			err = EAUTH;
 			goto out;
 		}
```
WIthout it, authentication fails with many servers, but the patch has not been accepted and I have had to patch re myself.

juha-h:
What is the use of contact header if server doesn't need to care about it?  I have always thought that with contact header UA tells how it can be reached.

cspiel1:
> What is the use of contact header if server doesn't need to care about it? I have always thought that with contact header UA tells how it can be reached.

We think that some SIP servers or SBC check if the registration user matches the contact user. If they do not match, the registration is declined.

juha-h:
Have I understood correctly that the PR makes baresip accept incoming requests even when itheir Request-URI does not match Contact URI of the REGISTER request?  If so, it sounds like serious security issue.

sreimers:
> Have I understood correctly that the PR makes baresip accept incoming requests even when itheir Request-URI does not match Contact URI of the REGISTER request?

No, its all about how baresip sets the contact user:
Before:
```
<sip:alice@iptel.org> -> alice-0x61fa7c1a46d0
<sip:alice@example.net> -> alice-0x71fa7c1a46d0
....
```

After
```
<sip:alice@iptel.org> -> alice
<sip:alice@example.net> -> alice-2
<sip:bob@iptel.org> -> bob
<sip:bob@example.org> -> bob-4
<sip:49xxxxxxxxxx@1und1.net> -> 49xxxxxxxxxx
```

juha-h:
That too creates security issue.  Now it would be easy for spammers to guess the Request-URI.

juha-h:
Christian Spielberger writes:

> > Why should baresip be complicated by issues in other software? Why don't you report the issues with them instead?
> 
> It is not so clear that this is an issue at "other software".

For sure it is.  There is no requirement about how userpart of Contact
URI should look like.

cspiel1:
> I have to find now why `test_call_deny_udp()` fails. The `transport` parameter is twice:
> 
> ```
> test: [ sip:a@127.0.0.1 ] event: CALL_ESTABLISHED (sip:b@127.0.0.1:54940;transport=udp;transport=tcp)
> ```

Fix for the double `transport` parameter: https://github.com/baresip/baresip/pull/3471

> But the same in `main` where the test is green. -> So, another issue that maybe is not a problem.
> 
> The `fix.a.n_established` is `1` but has to be zero.

The function `uag_find_msg()` needs an update now.

cspiel1:
This PR is ready now. The other one https://github.com/baresip/baresip/pull/3471 also.

src/ua.c
```diff
@@ -1166,6 +1166,15 @@ static const char *autoans_header_name(enum answer_method met)
 }
 
 
+static bool find_user(struct le *le, void *arg)
```
alfredh: nit: add suffix `_handler` as is done for many other callback handlers ...

alfredh:
if you have time, it would be nice to create a new testcase that covers 2 useragents
and one SIP registrar, and checks that incoming requests are routed correctly ...

cspiel1:
> if you have time, it would be nice to create a new testcase that covers 2 useragents and one SIP registrar, and checks that incoming requests are routed correctly ...

I'll give it a try

cspiel1:
> if you have time, it would be nice to create a new testcase that covers 2 useragents and one SIP registrar, and checks that incoming requests are routed correctly ...

For the test we need two registrars, because like you already mentioned it makes no sense and will not work to have two "alice" on one registrar. I'll try this with `sip_server_alloc()` in `test/sip/sipsrv.h`.

src/uag.c
```diff
@@ -906,6 +906,10 @@ struct ua *uag_find_msg(const struct sip_msg *msg)
 	cuser = &msg->uri.user;
 	for (le = uag.ual.head; le; le = le->next) {
 		struct ua *ua = le->data;
+		struct account *acc = ua_account(ua);
+
+		if (!acc->regint)
```
alfredh: perhaps also check that `acc` is not NULL, in order to avoid a potential null deref crash ...

alfredh:
The menu module has commands to add and remove useragents.
I tried to add couple of useragents, remove one, add one again and
it ends up that some cuser fields are not unique:


```
 cuser:     aeh-1
 cuser:     aeh-2
 cuser:     aeh-2
```

cspiel1:
I have a unit test with a stateless SIP proxy. But it's still red. Currently working on it.

src/ua.c
```diff
@@ -1180,6 +1189,7 @@ int ua_alloc(struct ua **uap, const char *aor)
 	struct uri *luri;
 	char *buf = NULL;
 	char *host = NULL;
+	static uint32_t ua_suffix = 0;
```
alfredh: This should work just fine, but I think it is better to move this to the UAG state.
Personally I would prefer that we try to avoid global/static variables.

juha-h:
I don't know what this PR is doing, but this must be the default like it has been before:
```
err = re_sdprintf(&ua->cuser, "%r-%p", &ua->acc->luri.user, ua);
```
It is not acceptable that baresip is opened to spamming without explicit account configuration.

Edit: Please add a config variable such as `please_spam_me` with default value `no` that retains the current behavior.

sreimers:
Baresip supports peer-to-peer calls and so the normal sip user is/was always used for matching too. So nothing is changed here:

https://github.com/baresip/baresip/blob/ee25400f0fc8a6771d08dab22c6f6347ad8599ea/src/uag.c#L897-L941

juha-h:
Peer-To-Peer fine, but if UA registers, it should not accept calls with other Request-URIs except the one it used in the Contact,

juha-h:
For improved security, If UA registers, it should be possible to reject requests with other Request-URIs than the one it uses in the Contact.

juha-h:
In any case, it must be possible to keep the existing behavior with a configuration variable.

sreimers:
I just tested other sip user agents like linphone and pjsip and both send `Contact` user (also if registered) without any random user suffix string:

```
Contact: <sip:android@10.0.0.1:24890;transport=tcp>;expires=600;received="sip:10.0.0.1:24890;transport=tcp";+sip.instance="<urn:uuid:e6bc1234-1234-1234-1234-cbeb9acf1234>";+org.linphone.specs="conference/2.0,ephemeral/1.1,groupchat/1.2"
```

juha-h:
> I just tested other sip user agents like linphone and pjsip and both send `Contact` user (also if registered) without any random user suffix string: ...

Fine with me as long there is a means to keep the current behavior.  I want to add a check to my app that rejects incoming request if its Request-URI does not match the Contact URI with random string.

It is really weird that this kind of major security issue has not been noticed before.

cspiel1:
I am nearly there, but the diff is big:
```sh
$ git diff unique_contact_user --stat
 test/call.c       |  97 +++++++++++++++++++++++++++++++++++++
 test/main.c       |   1 +
 test/sip/sipsrv.c | 311 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--
 test/test.h       |   1 +
 4 files changed, 406 insertions(+), 4 deletions(-)
```
Should we merge this PR first?

cspiel1:
> For improved security, If UA registers, it should be possible to reject requests with other Request-URIs than the one it uses in the Contact.

This PR doesn't change that. Incoming requests where the request URI doesn't match the contact user are still rejected with 404 "Not Found".

cspiel1:
I deleted the branch by accident. Let's see if I can restore the PR somehow!

cspiel1:
From my side now it's ready for the merge. We discussed already that it makes sense to remove the random suffix as long as the contact user is unique for all accounts. This doesn't introduce any security leak or false UA selection.

I have a unit test, but the diff is large:
```sh
$ git diff unique_contact_user --stat
 test/call.c       |  99 +++++++++++++++++++++++++++++++++++++++++++++++++++
 test/main.c       |   1 +
 test/sip/sipsrv.c | 309 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--
 test/test.h       |   1 +
 4 files changed, 407 insertions(+), 3 deletions(-)
```
I suggest to open another PR for that. @alfredh or should I add it here?

juha-h:
> From my side now it's ready for the merge. We discussed already that it makes sense to remove the random suffix as long as the contact user is unique for all accounts. This doesn't introduce any security leak or false UA selection.

I want to be able to keep the random suffix even if there is only one account in order to be able to make baresip more secure , for example, like this:

case BEVENT_SIPSESS_CONN:
            ua = uag_find_msg(msg);
            call = (struct call *)msg;
            struct uri ruri;
            uri_decode(&ruri, &msg->ruri);
            if (pl_strcmp(&ruri.user, ua_local_cuser(ua)) == 0) {
                LOGW("******* request uri match\n");
            } else {
                LOGD(request uri mismatch\n");
               // reject the call
            }

You cannot go and just remove that possibility and break backwards compatibility.

juha-h:
Usually there has been high bar to break backwards compatibility.  At least you should add a configuration variable that allows to retain it.

cspiel1:
Isn't TLS a better solution, with verify server? A CA-file or path can be set.

edit:
Okay, some users might find TLS to complicated, use UDP with listen port 5060 and a very simple account user name. In such cases an unwanted peer to peer call from a hacker would be possible.
I'll add a config setting (without change in struct config). Another commit for discussion with the maintainers.

juha-h:
I have account `sip:test@test.tutpro.com` with`outbound="sip:test.tutpro.com:5061;transport=tls"` and `sip_verify_server	yes`.  Still this comes in:
```
UDP 192.168.221.215:5060 -> 192.168.221.215:5022
INVITE sip:test@192.168.221.215:5022 SIP/2.0
```
This is not my personal baresip use issue.  As baresip app developer, I try to protect users of the app from spam calls even when they do not use TLS.

I don't understand why backwards compatibility cannot be retained here.  Instead, securing baresip is made harder if not even impossible to achieve.

juha-h:
But since I only have one vote here, do your thing.  I'll start using my own fork of baresip.

alfredh:
> From my side now it's ready for the merge. We discussed already that it makes sense to remove the random suffix as long as the contact user is unique for all accounts. This doesn't introduce any security leak or false UA selection.
> 
> I have a unit test, but the diff is large:
> 
> ```shell
> $ git diff unique_contact_user --stat
>  test/call.c       |  99 +++++++++++++++++++++++++++++++++++++++++++++++++++
>  test/main.c       |   1 +
>  test/sip/sipsrv.c | 309 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--
>  test/test.h       |   1 +
>  4 files changed, 407 insertions(+), 3 deletions(-)
> ```
> 
> I suggest to open another PR for that. @alfredh or should I add it here?

hi it should be fine to create a new PR for the unit test. Ideally it should be orthogonal to this PR ?

cspiel1:
The new setting is called `cuser_random`. In order to keep it simple, it is a gobal config setting (not account param) and I didn't add it to the `struct config_sip`. We want the default value to be `false`.

alfredh:
The reason why we have a unique suffix for the contact-user, is that baresip has a shared SIP-stack
and a shared transport. Perhaps other implementations have one SIP-stack per account.

The suffix can be the address of the UA, a random string, a counter etc.
It was pointed out that sending a memory address of the UA out is bad practice.
This can be solved by adding a global mask which the address is XOR'ed with.

Or a random string per UA that is long enough:

```
sip:aeh@iptel.org  -->  cuser=aeh-2PeWy27cFcg
sip:aeh@iptel.org  -->  cuser=aeh-ggFBc9H4yBA
```

juha-h:
Thanks for the `cuser_random` setting.  I can make it `true`.  Here is the initial code from my app that protects the UA from spam calls:
```
case BEVENT_SIPSESS_CONN:
            ua = uag_find_msg(msg);
            // There is no call yet and call is thus used to hold SIP message
            call = (struct call *)msg;
            if (account_regint(ua_account(ua)) > 0) {
                struct uri ruri;
                uri_decode(&ruri, &msg->ruri);
                if (pl_strcmp(&ruri.user, ua_local_cuser(ua)) != 0) {
                    char user_buf[256];
                    len = re_snprintf(user_buf, sizeof user_buf, "%r", &ruri.user);
                    LOGW("ruri user mismatch: %s != %s\n", user_buf, ua_local_cuser(ua));
                    sip_treply(NULL, uag_sip(), msg, 404, "Not Found");
                    bevent_stop(event);
                    return;
                }
            }
```

cspiel1:
@juha-h the code looks okay but this is already done here 
https://github.com/cspiel1/baresip/blob/507e9cfa294641254e9dd629740b14fb9dccfe48/src/ua.c#L712-L718

Please could you try this with `cuser_random   true`? Then you code in the app should not be reached.

cspiel1:
> Or a random string per UA that is long enough:
> 
> ```
> sip:aeh@iptel.org  -->  cuser=aeh-2PeWy27cFcg
> sip:aeh@iptel.org  -->  cuser=aeh-ggFBc9H4yBA
> ```

@alfredh If you agree with the `cuser_random` setting, then I'll replace the `rand_u32()` with such random string. 

edit: And then I am able to simplify it again, because then we need only the random suffix. It should be stable enough against wrong UA selection.

cspiel1:
If everybody can live with the last 3 commits and there are no further suggestions, it's ready for the merge.
- ua: add config cuser_random
- ua: simplify cuser generation
- uag: drop obsolete unique counter

Of course we can squash the commits.

juha-h:
@cspiel1 I tried with `cuser_random yes` and my check was still executed.  If I un-commented my check, incoming call with wrong cuser was coming in.

juha-h:
Looks like  `user_cmp_handler` is executed only at baresip start, not when call comes in.

cspiel1:
> ruri

Thanks for the test! You can use `msg->uri` which contains already the decoded request URI.

But that's not the point. We should fix this in `uag_find_msg()`. The second loop should only return a UA for non-registered accounts. For our applications and products, this would be okay, because we also register separate non-registration accounts for peer-to-peer calls. After a fix your check will no longer be needed.

juha-h:
> Thanks for the test! You can use `msg->uri` which contains already the decoded request URI.

Thanks for the tip.  Made the code simpler.

> But that's not the point. We should fix this in `uag_find_msg()`. The second loop should only return a UA for non-registered accounts. For our applications and products, this would be okay, because we also register separate non-registration accounts for peer-to-peer calls. After a fix your check will no longer be needed.

I can test again after your fix.

cspiel1:
So, the last commit changes the behavior such that
- peer-to-peer calls are no longer possible to a registered account.
- Thus, this avoids what @juha-h calls "spam calls".

Are there any objections against it?

juha-h:
I tried with  https://github.com/baresip/baresip/pull/3470/commits/a7b124d4ca37d918cdad61498291ebeb956ffd4b, but the test is not executed and spam calls get through to registered account.  Also my test fails, because now `account_regint(ua_account(ua))` is 0 even when the account is registering.

sreimers:
We should not mix suffix change and uag_find_msg changes within one PR here, better separate.

juha-h:
Re-tested and the result is the same as in above.  With the new ua.c and uag.c my also test fails, because account_regint(ua_account(ua)) is 0.

cspiel1:
I remove the last uag commit and keep it in another branch for the next PR. @juha-h what does `uag_find_msg()` print out when this "spam call" happens?

I suggest to keep this commit which is essential if we remove the suffix for the first account with a specific cuser: 243435e201e37f8d09fb671d93daba87a89ee7a1
```
    uag: cuser match only for registered accounts
```

cspiel1:
> I suggest to keep this commit which is essential if we remove the suffix for the first account with a specific cuser: [243435e](https://github.com/baresip/baresip/commit/243435e201e37f8d09fb671d93daba87a89ee7a1)
> 
> ```
>     uag: cuser match only for registered accounts
> ```

Okay, I changed opinion. Will also drop, revert this change. Let's fix it in the next PR!

juha-h:
Here is the log:
```
2025-08-28 08:50:12.854  9921-10178 Baresip Lib             com.tutpro.baresip                   D  ua: sipsess connect via UDP 192.168.221.215:5060 --> 192.168.221.125:60012
2025-08-28 08:50:12.854  9921-10178 Baresip Lib             com.tutpro.baresip                   D  foo@test.tutpro.com: account match for foo
2025-08-28 08:50:12.854  9921-10177 Baresip Lib             com.tutpro.baresip                   D  ua event (SIPSESS_CONN) incoming call
2025-08-28 08:50:12.855  9921-10178 Baresip Lib             com.tutpro.baresip                   D  foo@test.tutpro.com: account match for foo
2025-08-28 08:50:12.855  9921-10177 Baresip Lib             com.tutpro.baresip                   D  ruri user mismatch: foo != foo-6f3deu3rETUO7ff
```
My code detects the spam which was missed by uag.c.

cspiel1:
We need at least this change 2a56ad39130c04214b04967a76cd85037cf3858b in `uag_find_msg()` in order to pass `test_call_deny_udp`.

juha-h:
I have to go now.  Can do more tests in the afternoon or evening.

cspiel1:
For further analysis of spam calls I opened this issue: https://github.com/baresip/baresip/issues/3479

Let's merge this one first!

PR #3473, opened 2025-08-19
jbuf: wish size is given in #packets

cspiel1:
The wish size for the adaptive mode depends on how much the packets are re-
ordered. It can only be measured by looking on the sequence number of the RTP
header and how much it jumps around because of re-ordered packets. Counting
video frames is not needed and causes failures when the wish size is computed.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1575,7 +1575,6 @@ int  jbuf_drain(struct jbuf *jb, struct rtp_header *hdr, void **mem);
 void jbuf_flush(struct jbuf *jb);
 int  jbuf_stats(const struct jbuf *jb, struct jbuf_stat *jstat);
 int  jbuf_debug(struct re_printf *pf, const struct jbuf *jb);
-uint32_t jbuf_frames(const struct jbuf *jb);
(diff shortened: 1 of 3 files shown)


sreimers:
PR #2757 is ready and works fine for years now in my production with different WebRTC browsers (audio and video).

cspiel1:
> PR #2757 is ready and works fine for years now in my production with different WebRTC browsers (audio and video).

This sounds good! Did you test it with jitter simulation?

sreimers:
> > PR #2757 is ready and works fine for years now in my production with different WebRTC browsers (audio and video).
> 
> 
> 
> This sounds good! Did you test it with jitter simulation?

Sure, you can find many different Experiments and Stats in the PR comments.

cspiel1:
Can we close this?
I change it to draft for now

alfredh:
The meaning of "packet" should be one RTP packet

The meaning of "frame" is a complete video-frame that consist of many RTP packets.

For audio the definition of "frame" is outside the scope of the jitter buffer.

To me is seems logical that the jitter buffer handles "packets".

I would suggest that we merge this PR first:
https://github.com/baresip/baresip/pull/2757

PR #3475, opened 2025-08-19
test: ua - fix index for domain_add

cspiel1:
@alfredh please have a look! I am unsure.


alfredh:
this looks correct. it has been working because the `server_count` is 1 ...

PR #3481, opened 2025-08-29
doc: update jbuf settings in examples

cspiel1:


cspiel1:
The deprecated "_delay" settings have to be fixed also. Note that there is `jbtype` and no evaluation!
```c
if (0 == conf_get(conf, "audio_jitter_buffer_delay", &jbtype)) {
...
```

cspiel1:
Thanks for review!



> The deprecated "_delay" settings have to be fixed also. Note that there is `jbtype` and no evaluation!
> 
> ```c
> if (0 == conf_get(conf, "audio_jitter_buffer_delay", &jbtype)) {
> ...
> ```

Should we handle it like here for multicast? https://github.com/baresip/baresip-apps/pull/79

cspiel1:
> Thanks for review!
> 
> > The deprecated "_delay" settings have to be fixed also. Note that there is `jbtype` and no evaluation!
> > ```c
> > if (0 == conf_get(conf, "audio_jitter_buffer_delay", &jbtype)) {
> > ...
> > ```
> 
> Should we handle it like here for multicast? [baresip/baresip-apps#79](https://github.com/baresip/baresip-apps/pull/79)

Answer myself. I realized that this was wanted. The old settings will no longer work and if still used, there is a fallback to the default values. This okay for us.

PR #3484, opened 2025-09-01
test: add test_call_uag_find_msg

cspiel1:
- **test: sipsrv - SIP message forwarding**
- **test: call - add test_call_uag_find_msg()**

### DIFF
--- a/test/call.c
+++ b/test/call.c
@@ -3858,12 +3858,12 @@ int test_call_uag_find_msg(void)
 	TEST_ERR(err);
 	err = ua_alloc(&f->a.ua, aor);
 	TEST_ERR(err);
-	mem_deref(aor);
+	aor = mem_deref(aor);
 	err = re_sdprintf(&aor, "B <sip:alice@%J>;regint=60", &sa2);
 	TEST_ERR(err);
 	err = ua_alloc(&f->b.ua, aor);
 	TEST_ERR(err);
-	mem_deref(aor);
+	aor = mem_deref(aor);
 	err = re_sdprintf(&aor, "C <sip:bob@%J>;regint=60", &sa2);
 	TEST_ERR(err);
 	err = ua_alloc(&f->c.ua, aor);
--- a/test/call.c
+++ b/test/call.c
@@ -3833,6 +3833,7 @@ int test_call_uag_find_msg(void)
 	struct sa sa1;
 	struct sa sa2;
 	char *aor=NULL;
+	char *curi=NULL;
 	struct cancel_rule *cr;
 	int err = 0;
 
@@ -3891,8 +3892,8 @@ int test_call_uag_find_msg(void)
 	cancel_rule_new(BEVENT_CALL_ESTABLISHED, f->c.ua, 0, 0, 1);
 	cancel_rule_and(BEVENT_CALL_ESTABLISHED, f->b.ua, 1, 0, 1);
 
-	char curi[32];
-	re_snprintf(curi, sizeof(curi), "sip:alice@%J", &sa2);
+	err = re_sdprintf(&curi, "sip:alice@%J", &sa2);
+	TEST_ERR(err);
 	err = ua_connect(f->c.ua, 0, NULL, curi, VIDMODE_OFF);
 	TEST_ERR(err);
 
@@ -3911,6 +3912,7 @@ int test_call_uag_find_msg(void)
 	mem_deref(srv1);
 	mem_deref(srv2);
 	fixture_close(f);
+	mem_deref(curi);
 
 	return err;
 }
--- a/test/call.c
+++ b/test/call.c
@@ -3894,7 +3894,7 @@ int test_call_uag_find_msg(void)
 
 	err = re_sdprintf(&curi, "sip:alice@%J", &sa2);
 	TEST_ERR(err);
-	err = ua_connect(f->c.ua, 0, NULL, curi, VIDMODE_OFF);
+	err = ua_connect(f->c.ua, NULL, NULL, curi, VIDMODE_OFF);
 	TEST_ERR(err);
 
 	err = re_main_timeout(5000);
@@ -3906,6 +3906,24 @@ int test_call_uag_find_msg(void)
 	ASSERT_EQ(0, fix.a.n_established);
 	ASSERT_EQ(1, fix.b.n_incoming);
 	ASSERT_EQ(1, fix.b.n_established);
+
+	/* 2nd test: peer-to-peer call to registered UAs should be rejected */
+	cancel_rule_pop();
+	curi = mem_deref(curi);
+	err = re_sdprintf(&curi, "sip:alice@127.0.0.1");
+	TEST_ERR(err);
+
+	f->a.n_closed = 0;
+	f->exp_closed = 1;
+	err = ua_connect(f->c.ua, NULL, NULL, curi, VIDMODE_OFF);
+	TEST_ERR(err);
+	err = re_main_timeout(5000);
+	TEST_ERR(err);
+	TEST_ERR(fix.err);
+
+	ASSERT_EQ(0, fix.a.n_incoming);
+	ASSERT_EQ(0, fix.b.n_incoming);
+	ASSERT_EQ(0, fix.c.n_incoming);
 
  out:
 	mem_deref(aor);
--- a/test/call.c
+++ b/test/call.c
@@ -9,6 +9,7 @@
 #include <rem.h>
 #include <baresip.h>
 #include "test.h"
+#include "sip/sipsrv.h"
 #include "../src/core.h"  /* NOTE: temp */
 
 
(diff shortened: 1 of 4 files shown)


test/call.c
```diff
@@ -3815,3 +3816,101 @@ int test_call_cert_select(void)
...
+
+	err = re_main_timeout(5000);
+	TEST_ERR(err);
+	TEST_ERR(fix.err);
+
+	/* very that correct UA was selected and got established call */
+	ASSERT_EQ(0, fix.a.n_incoming);
+	ASSERT_EQ(0, fix.a.n_established);
+	ASSERT_EQ(1, fix.b.n_incoming);
+	ASSERT_EQ(1, fix.b.n_established);
```
alfredh, line 3908: I am wondering if it is enough to check that the incoming call was working.

The other testcases are covering calls that are established ...

test/call.c
```diff
@@ -3815,3 +3816,101 @@ int test_call_cert_select(void)
...
+	char curi[32];
+	re_snprintf(curi, sizeof(curi), "sip:alice@%J", &sa2);
+	err = ua_connect(f->c.ua, 0, NULL, curi, VIDMODE_OFF);
+	TEST_ERR(err);
+
+	err = re_main_timeout(5000);
+	TEST_ERR(err);
+	TEST_ERR(fix.err);
+
+	/* very that correct UA was selected and got established call */
```
alfredh: there is a small typo in the comment ...

test/call.c
```diff
@@ -3815,3 +3816,101 @@ int test_call_cert_select(void)
...
+	err = re_sdprintf(&aor, "C <sip:bob@%J>;regint=60", &sa2);
+	TEST_ERR(err);
+	err = ua_alloc(&f->c.ua, aor);
+	TEST_ERR(err);
+
+	err = ua_register(f->a.ua);
+	TEST_ERR(err);
+	err = ua_register(f->b.ua);
+	TEST_ERR(err);
+	err = ua_register(f->c.ua);
```
alfredh, line 3877: account C is used for outgoing call, is it required to register it ?

test/call.c
cspiel1: Test agent `a` represents the first "alice" account, while `b` the second. The checks ensure that the second UA is selected.

The other call tests all cover peer-to-peer calls. There is no other `sip_server_alloc()` in test/call.c.

test/call.c
cspiel1: Yes, it is a SIP call handled by the second `sip_server`, `srv2`. Agent `a` registers at `srv1`, while agent `b` and `c` at `srv2`. The call is handled by SIP server `srv2`.

The test ensures that `b` is selected for the incoming call, not `a` which would be selected if there were no cuser suffix for both "alice" accounts.

alfredh:
there is currently a new CI build warning/error:

```
/Users/runner/work/baresip/baresip/modules/avfilter/filter.c:71:8: error: 'av_int_list_length_for_size' is deprecated [-Werror,-Wdeprecated-declarations]
```

sreimers:
> there is currently a new CI build warning/error:

Looks like macOS homebrew is upgraded to ffmpeg 8.0, maybe downgrade until we can fix this, will make a PR.

alfredh:
thanks, it is building now ...

@cspiel1 could you please rebase this PR on top of main HEAD ?

test/call.c
alfredh: a SIP account does not have to be registered in order to make outgoing call...

The selection of transport option for INVITE/200-OK/ACK does not depend on
if the endpoint is registered or not.

test/call.c
```diff
@@ -3815,3 +3816,101 @@ int test_call_cert_select(void)
...
+
+	mem_deref(f->a.ua);
+	mem_deref(f->b.ua);
+	mem_deref(f->c.ua);
+
+	err = re_sdprintf(&aor, "A <sip:alice@%J>;regint=60", &sa1);
+	TEST_ERR(err);
+	err = ua_alloc(&f->a.ua, aor);
+	TEST_ERR(err);
+	mem_deref(aor);
```
alfredh: if the next line fails, it will try to free `aor` at the end of the function, but `aor` will point
to invalid memory. It is probably safer to set aor to NULL:

`aor = mem_deref(aor)`

test/call.c
```diff
@@ -3815,3 +3816,101 @@ int test_call_cert_select(void)
...
+	cancel_rule_pop();
+
+	f->b.peer = &f->c;
+	f->c.peer = &f->b;
+
+	f->behaviour = BEHAVIOUR_ANSWER;
+	cancel_rule_new(BEVENT_CALL_ESTABLISHED, f->c.ua, 0, 0, 1);
+	cancel_rule_and(BEVENT_CALL_ESTABLISHED, f->b.ua, 1, 0, 1);
+
+	char curi[32];
```
alfredh: is the buffer big enough for an IPv6 address ?

test/sip/sipsrv.c
```diff
@@ -183,14 +184,310 @@ static bool handle_register(struct sip_server *srv, const struct sip_msg *msg)
...
+	if (!hdr)
+		return 0;
+
+	values = mbuf_alloc(16);
+	if (!values)
+		return ENOMEM;
+
+	sip_msg_hdr_apply(msg, true, id, enc_list_handler, values);
+
+	values->pos = 0;
```
alfredh, line 218: if possible please use `mbuf_` functions for setting position, e.g. `mbuf_set_pos()`

test/sip/sipsrv.c
```diff
@@ -183,14 +184,310 @@ static bool handle_register(struct sip_server *srv, const struct sip_msg *msg)
...
+	int err;
+};
+
+
+static bool enc_handler(const struct sip_hdr *hdr, const struct sip_msg *msg,
+			void *arg)
+{
+	struct hdr_handler_arg *harg = arg;
+	struct mbuf *mb = harg->mb;
+	(void)msg;
```
alfredh: looks like `msg` is used in this function ..

test/sip/sipsrv.c
```diff
@@ -183,14 +184,310 @@ static bool handle_register(struct sip_server *srv, const struct sip_msg *msg)
...
+
+
+static bool reply_via_handler(const struct sip_hdr *hdr,
+			      const struct sip_msg *msg, void *arg)
+{
+	struct via_handler_arg *vharg = arg;
+	struct mbuf *mb = vharg->mb;
+	struct sa *dst = vharg->dst;
+	int err = 0;
+	(void)msg;
```
alfredh: `msg` is used ..

test/sip/sipsrv.c
```diff
@@ -183,14 +184,310 @@ static bool handle_register(struct sip_server *srv, const struct sip_msg *msg)
...
+	sip_msg_hdr_apply(msg, true, SIP_HDR_VIA, enc_handler, &harg);
+	if (harg.err) {
+		err |= harg.err;
+		return err;
+	}
+
+	const struct sip_hdr *maxfwd = sip_msg_hdr(msg, SIP_HDR_MAX_FORWARDS);
+	if (maxfwd) {
+		uint32_t mf = pl_u32(&maxfwd->val);
+		err |= mbuf_printf(mb, "%r: %u\r\n", &maxfwd->name, mf - 1);
```
alfredh, line 352: if Max-Forwards goes down to zero, the message should be dropped in the SIP server

test/sip/sipsrv.c
```diff
@@ -183,14 +184,310 @@ static bool handle_register(struct sip_server *srv, const struct sip_msg *msg)
...
+	mb->pos = 0;
+	err = sip_send_conn(srv->sip, NULL, msg->tp, &dst, NULL, mb, NULL,
+			    NULL);
+	if (err)
+		goto out;
+
+	DEBUG_INFO("successfully forwarded SIP message\n");
+out:
+	mem_deref(mb);
+	return err ? false : true;
```
alfredh: can be simplified:

`return err == 0;`

test/sip/sipsrv.c
cspiel1: Which error code fits best? `EADDRNOTAVAIL` was suggested by copilot.

alfredh: any error code can be used, for example `EPROTO` ...

cspiel1: Changed to EPROTO.

test/call.c
cspiel1: But we want to test a SIP call `C -> srv2 -> B` and not directly `C -> B`.

So, I don't understand what you mean.

I could add another check for `C -> B` or `A -> B` which now should fail, which is wanted. It would check, that "spam-calls" are declined.

cspiel1: I have to postpone this. Hopefully can proceed next week here. The "reject spam calls" check is not ready. I'll push it anyway as start point when I have more time.

alfredh:
I will close it for now... Please open a new PR when you have more time ...

PR #3486, opened 2025-09-02
message: log when UA is not found for incoming MESSAGE

juha-h:


src/message.c
```diff
@@ -92,6 +92,8 @@ static bool request_handler(const struct sip_msg *msg, void *arg)
 
 	ua = uag_find_msg(msg);
 	if (!ua) {
+		info("ua: %r: UA not found: %H\n",
+		     &msg->from.auri, uri_encode, &msg->uri);
```
alfredh, line 96: Perhaps it is better if the log prefix matches the name of the .c file, i.e. `message: xxx` ?

juha-h: Replaced "ua" prefix with "message" prefix

PR #3488, opened 2025-09-07
readme: update list of platforms and compilers

alfredh:


juha-h:
Android (8.0 or later) does not support aaudio set/get sessionid, which is required for hw aec and is available in Android 9 (API level 28).  Of course one can build baresip on earlier versions, but calls are pretty much useless without hw aec.

alfredh:
Are you still using `webrtc_aec` modules ?

juha-h:
Alfred E. Heggestad writes:

> alfredh left a comment (baresip/baresip#3488)
> 
> Are you still using `webrtc_aec` modules ?

No, webrtc_aecm turned out to be useless. See
https://github.com/baresip/baresip/discussions/3477.

PR #3492, opened 2025-09-08
misc: check return value of str_dup()

alfredh:


alfredh:
I can split this up into smaller PRs if that is wanted ...

PR #3499, opened 2025-09-18
mpa: move MPA audio codec to baresip-apps

alfredh:
This is a proposal to remove the old audio codec MPA. This codec is very old and not sure if anyone is using it at all.

Today there are much better alternatives for
audio codecs such as Opus and AAC.


sreimers:
I think MPA was developed because https://tech.ebu.ch/docs/tech/tech3326.pdf requires this as mandatory codec for compatibility. Maybe we should move this to baresip-apps.

alfredh:
https://github.com/baresip/baresip-apps/pull/80

PR #3509, opened 2025-09-22
webrtc_aec: update module to Debian Trixie compatibility

juha-h:
With patch provided by @alfredh


modules/webrtc_aec/encode.cpp
```diff
@@ -82,11 +83,13 @@ static AudioProcessing::ChannelLayout get_layout(uint8_t ch)
 	default: return (AudioProcessing::ChannelLayout)-1;
 	}
 }
+#endif
```
alfredh: I think this if 0 block can be removed ...

alfredh:
this PR looks fine to me ...

@sreimers please review and merge if you agree ...

PR #3511, opened 2025-09-22
test/ua: add test_ua_cuser

maximilianfridrich:
Add a simple test for the recently merged https://github.com/baresip/baresip/pull/3470.

### DIFF
--- a/test/ua.c
+++ b/test/ua.c
@@ -969,3 +969,47 @@ int test_ua_refer(void)
  out:
 	return err;
 }
+
+
+int test_ua_cuser(void)
+{
+	int err;
+	struct ua *ua1, *ua2, *ua3, *ua4;
+	uint32_t n_uas = list_count(uag_list());
+
+	/* make sure we dont have that UA already */
+	ASSERT_TRUE(NULL == uag_find_aor("sip:alice@test.invalid"));
+	ASSERT_TRUE(NULL == uag_find_aor("sip:bob@test.invalid"));
+
+	err = ua_alloc(&ua1, "Alice <sip:alice@test.invalid>;regint=0");
+	TEST_ERR(err);
+
+	err = ua_alloc(&ua2, "Alice <sip:alice@test.invalid>;regint=0");
+	TEST_ERR(err);
+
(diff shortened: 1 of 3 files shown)


maximilianfridrich:
Could someone of the maintainers also please add documentation to the BareSIP wiki about the new `cuser_random` config settting?

https://github.com/baresip/baresip/wiki/Configuration#sip

cspiel1:
> Could someone of the maintainers also please add documentation to the BareSIP wiki about the new `cuser_random` config settting?
> 
> https://github.com/baresip/baresip/wiki/Configuration#sip

done. https://github.com/baresip/baresip/wiki/Configuration#cuser_random

@juha-h Should we rename the setting to `sip_cuser_random`?

test/ua.c
```diff
@@ -969,3 +969,47 @@ int test_ua_refer(void)
...
+	/* verify global UA keeper */
+	ASSERT_EQ((n_uas + 4), list_count(uag_list()));
+	ASSERT_TRUE(ua1 == uag_find_aor("sip:alice@test.invalid"));
+
+	mem_deref(ua1);
+	mem_deref(ua2);
+	mem_deref(ua3);
+	mem_deref(ua4);
+
+	ASSERT_EQ((n_uas), list_count(uag_list()));
```
alfredh: There is an extra parenthesis around `n_uas` not sure if it is needed ?

juha-h:
> @juha-h Should we rename the setting to `sip_cuser_random`?

All other sip settings start with sip_ prefix.  But I have lost interest, because its default value breaks backwards compatibility.

test/ua.c
maximilianfridrich: Certainly not. Thanks. Removed.

maximilianfridrich:
I like the renaming idea because it would make the naming more consistent (and this setting is very new - renaming shouldn't be too problematic). However, I would suggest doing it in a separate PR. This PR shouldn't do too many things at once.

cspiel1:
rename PR: https://github.com/baresip/baresip/pull/3515

PR #3513, opened 2025-09-23
call: add missing input argument checking (struct call pointer)

alfredh:


alfredh:
hi @cspiel1 could you please review and merge ?

PR #3518, opened 2025-09-25
test: add test_uag_find_msg()

cspiel1:
This PR adds stateless proxy to test/sip/sipsrv that up to now only had
registrar features.

Futher adds positive and negative test for UA selection. Thus it is also tested
if - so called - spam calls are rejected.

- **test: sipsrv - SIP message forwarding**
- **test: call - add test_call_uag_find_msg()**
- **test: registered UA rejects peer-to-peer calls**

### DIFF
--- a/test/call.c
+++ b/test/call.c
@@ -3851,9 +3851,9 @@ int test_call_uag_find_msg(void)
 	err = sip_transp_laddr(srv2->sip, &sa2, SIP_TRANSP_UDP, NULL);
 	TEST_ERR(err);
 
-	mem_deref(f->a.ua);
-	mem_deref(f->b.ua);
-	mem_deref(f->c.ua);
+	f->a.ua = mem_deref(f->a.ua);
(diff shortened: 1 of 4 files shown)


test/call.c
```diff
@@ -3815,3 +3816,138 @@ int test_call_cert_select(void)
...
+	err = ua_register(f->a.ua);
+	TEST_ERR(err);
+	cancel_rule_new(BEVENT_REGISTER_OK, f->a.ua, 0, 0, 0);
+	err = re_main_timeout(5000);
+	cancel_rule_pop();
+	TEST_ERR(err);
+
+	curi = mem_deref(curi);
+	/* alice --> rejected. alice-<suffix> would be correct */
+	/* note: now both alice UAs have a suffix */
```
cspiel1: I replace this comment by an assertion.

cspiel1:
This PR is ready.

PR #3531, opened 2025-10-08
cmake: fix usage of SPANDSP_HINTS

cspiel1:
if SPANDSP_HINTS is not set, then /usr/local should be preferred over /usr

discussion: #3530


cspiel1:
Thanks! I think that there are more cmake files like this.

PR #3532, opened 2025-10-11
video: find new encoder if not available

xiaokuang95:
Fix an issue where video_encoder_set may encounter an unimplemented vc->encupdh.
A new vidcodec is enumerated, and if none is found, ENOENT is returned.

### DIFF
--- a/src/video.c
+++ b/src/video.c
@@ -1585,12 +1585,12 @@ int video_encoder_set(struct video *v, struct vidcodec *vc,
 	vtx = &v->vtx;
 
 	if (!vc->encupdh) {
-		struct list *vidcodecl = vc->le.list;
-		struct vidcodec *vcd;
-
 		info("video: vidcodec '%s' has no encoder\n", vc->name);
 
-		vcd = (struct vidcodec *)vidcodec_find_encoder(vidcodecl,
+		struct list *vidcodecl = vc->le.list;
+
+		struct vidcodec *vcd =
+			(struct vidcodec *)vidcodec_find_encoder(vidcodecl,
 							       vc->name);
 		if (!vcd) {
 			warning("video: could not find encoder (%s)\n",


sreimers:
Not sure about this, why has `struct vidcodec *vc` no encoder at this stage? Which encoder do you use?

xiaokuang95:
> Not sure about this, why has `struct vidcodec *vc` no encoder at this stage? Which encoder do you use?

In the Baresip Studio project, I implemented MediaCodec. If only the decoder is implemented without the encoder, there may be no video stream sent. I found that in video_encoder_set, when vc->encupdh is not implemented, it doesn’t check for other available encoders. I believe the encoder should behave like the decoder: if no updh is found, it should select another encoder that has updh implemented. This is important because some Android devices may support only encoding but not decoding, or only decoding but not encoding.

Reference: [video.c#L1636-L1651](https://github.com/baresip/baresip/blob/main/src/video.c#L1636-L1651)

alfredh:
The next release is planned for Wed 15. Oct and if you dont mind I will suggest to merge this
PR after the release is out.

xiaokuang95:
That works too, but I’d suggest merging it before the release so it can be part of the upcoming release package.

sreimers:
We have not a good experience with late fixes and a freeze window currently, I would also prefer a merge after release.

xiaokuang95:
Okay, then let’s merge it after version 4.2.0 is released.

src/video.c
```diff
@@ -1585,8 +1585,20 @@ int video_encoder_set(struct video *v, struct vidcodec *vc,
 	vtx = &v->vtx;
 
 	if (!vc->encupdh) {
+		struct list *vidcodecl = vc->le.list;
+		struct vidcodec *vcd;
```
alfredh: the declaration and setting of `vcd` can be done in one operation

alfredh:
is there any updates on this PR ?

xiaokuang95:
> is there any updates on this PR ?

Sorry, I forgot that this PR still needed some adjustments. I’ve pushed the updated code.

alfredh:
many thanks.

This PR looks good to me. Should we merge it ?
@sreimers

PR #3533, opened 2025-10-13
rtprecv, aureceiver: fix ssrc re-invite

Turkish:
# Baresip SSRC Handling Regression (v4.1.0)

## Overview

Some SIP endpoints (e.g., conferencing bridges, SBCs) renegotiate media mid-call
by sending a re-INVITE that introduces new SSRC values and, occasionally, a new
dynamic payload type. On baresip v4.1.0 the receive pipeline does not reset its
internal state when this happens. The jitter buffer continues to hold packets,
but the decoder is never called again, so inbound audio mutes silently.

This note documents the failure and the proposed fix.
## Steps to Reproduce

1. Build baresip v4.1.0.
2. Register a user agent and call a SIP device that upgrades to audio+video.
   A simple reproducer is a conferencing service that starts audio-only and, a
   few seconds in, sends a re-INVITE to add video. During that step many devices
   switch SSRC and payload type for the audio stream.
3. Observe that:
   * `rtprecv` logs “SSRC changed …”,
   * `audio_recv` emits “Set audio decoder …”, but
   * no further `audio_recv` decode/push events appear  any consumer of the
     decoded audio sees silence.

The remote side keeps transmitting, yet baresip produces only silence.

## Cause

* `rtprecv` only does a `jbuf_flush` on SSRC changes. This removes queued packets
  but keeps the jitter buffer in the previous epoch (sequence numbers, timing).
* The payload-type guard in `aurecv_receive` still expects the old PT until the
  decoder updates. Packets with the new PT are quietly ignored, so no data is
  passed to the decoder.
* Together this leaves the receive pipeline stuck: the jitter buffer retains
  packets but never marks them as ready, and the decoder is never called.

## Fix

We introduce a helper, `rtprecv_resync`, that performs a hard reset when the
SSRC changes:

* Flush the jitter buffer.
* Reset the SSRC/sequence baselines (`ssrc`, `pseq`, `pseq_set`).
* Clear the cached payload-type fields so the next packet rebinds them (`pt` and
  `pt_tel`).

In `aurecv_receive` we add a throttled warning when a packet is ignored because
its PT doesn’t match the current decoder, making the problem visible.

After these changes the very next packet from the new SSRC is decoded normally,
and audio continues across the re-INVITE without manual intervention.

## Patch Summary

* `src/rtprecv.c`
  * Add `rtprecv_resync` and call it when the SSRC changes before inserting a
    packet into the jitter buffer.
  * Reset SSRC, `pseq`, and payload-type fields inside that helper.
* `src/aureceiver.c`
  * Log SSRC changes and flush the aubuf to drop stale data.
  * Surface payload-type mismatches so they can be addressed quickly.

The modifications touch only the core receive path; no downstream integration
code is required. With this patch applied, baresip handles mid-call SSRC changes
from the far end as expected.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -638,11 +638,11 @@ static int stream_pt_handler(uint8_t pt, struct mbuf *mb, void *arg)
  * Stream receive handler for audio is called from RX thread if enabled
  */
 static void stream_recv_handler(const struct rtp_header *hdr,
-					struct rtpext *extv, size_t extc,
-					struct mbuf *mb, unsigned lostc,
-					bool new_source,
-					bool *ignore,
-					void *arg)
+				struct rtpext *extv, size_t extc,
+				struct mbuf *mb, unsigned lostc,
+				bool new_source,
+				bool *ignore,
+				void *arg)
 {
 	struct audio *a = arg;
 


alfredh:
Many thanks for your patch.

We are planning a new release this Wed 15. Oct
so we can try to merge this after the release is out.

sreimers:
The PR contains unrelated debug code, it would be nice if you could split up this PR (or remove the debug code), to keep the related fix small.

src/rtprecv.c
```diff
@@ -322,6 +322,26 @@ static void decode_tmr(void *arg)
...
 
 
+static void rtprecv_resync(struct rtp_receiver *rx,
+			 const struct rtp_header *hdr)
+{
+	if (!rx || !hdr)
+		return;
+
+	if (rx->jbuf)
+		jbuf_flush(rx->jbuf);
```
sreimers: `jbuf_flush` has a `*jb` NULL check. So you can remove the`if`.

Turkish: @sreimers  removed

alfredh:
In a follow-up PR, would it be possible to add a new testcase for this scenario ?
The new testcase would help to avoid regressions next time ...

Turkish:
> In a follow-up PR, would it be possible to add a new testcase for this scenario ? The new testcase would help to avoid regressions next time ...

Yeah sure, I'll do that in a separate PR

src/aureceiver.c
```diff
@@ -291,6 +285,13 @@ void aurecv_receive(struct audio_recv *ar, const struct rtp_header *hdr,
 		return;
 
 	mtx_lock(ar->mtx);
+	if (hdr->ssrc != ar->ssrc) {
+		ar->ssrc = hdr->ssrc;
```
alfredh: It is quite unlikely, but SSRC=0 is also a valid SSRC.
For the first packet arriving, it is not really necessary to flush the aubuf.

src/aureceiver.c
```diff
@@ -220,14 +220,11 @@ static int aurecv_stream_decode(struct audio_recv *ar,
...
 	int err = 0;
 	const struct aucodec *ac = ar->ac;
-	bool flush = ar->ssrc != hdr->ssrc;
 
 	/* No decoder set */
 	if (!ac)
 		return 0;
 
-	ar->ssrc = hdr->ssrc;
-
```
alfredh, line 230: The SSRC for incoming packets is tracked in two places: rtprecv and aureceiver.

Perhaps it would be better if the recv SSRC was only tracked in one place,
to avoid double book-keeping and potential errors? Since `rtprecv` is used by both
audio and video it seems like a good place.

This can also be done in a separate follow-up PR.

alfredh:
hello, is there any updates regarding this PR ?

src/aureceiver.c
Turkish: I centralized it in rtprecv and aureceiver now just gets its SSRC info from there

Turkish:
@alfredh  just updated the PR with the requested refactor

alfredh:
hi @cspiel1 could you please also take a look at this PR ?

src/aureceiver.c
```diff
@@ -42,7 +42,8 @@ struct audio_recv {
 	const struct aucodec *ac;     /**< Current audio decoder             */
 	struct aubuf *aubuf;          /**< Audio buffer before auplay        */
 	mtx_t *aubuf_mtx;             /**< Mutex for aubuf allocation        */
-	uint32_t ssrc;                /**< Incoming synchronization source   */
+	uint32_t ssrc_remote;         /**< Last observed remote SSRC         */
+	bool ssrc_remote_valid;       /**< True if remote SSRC is known      */
```
alfredh: it looks like the members `ssrc_remote` and `ssrc_remote_valid` is
only set and never read.

is it possible to remove them to simplify the code ?

Turkish: Cleaned up dangling logic from previous implementation

src/rtprecv.c
```diff
@@ -627,6 +649,28 @@ int rtprecv_get_ssrc(struct rtp_receiver *rx, uint32_t *ssrc)
 }
 
 
+bool rtprecv_consume_ssrc_change(struct rtp_receiver *rx, uint32_t *ssrc)
```
cspiel1: Should we name it `rtprecv_ssrc_changed()` and `stream_ssrc_changed()`?

cspiel1: Okay, I see that it really "consumes" the change. So, its your choice.

src/rtprecv.c
```diff
@@ -322,6 +323,26 @@ static void decode_tmr(void *arg)
...
+	rx->ssrc_set = true;
+	rx->pseq = (uint16_t)(hdr->seq - 1);
+	rx->pseq_set = true;
+	rx->pt = -1;
+	rx->pt_tel = 0;
+	rx->ssrc_changed = true;
+	mtx_unlock(rx->mtx);
+}
+
+
```
cspiel1, line 349: Now there is some redundant code in https://github.com/Turkish/baresip/blob/099d60b06ffa562005c9c595b689c2495d374d88/src/rtprecv.c#L463-L465

src/rtprecv.c
```diff
@@ -451,11 +473,11 @@ void rtprecv_decode(const struct sa *src, const struct rtp_header *hdr,
 			return;
 	}
 
+	if (flush)
+		rtprecv_resync(rx, hdr);
```
cspiel1: I think now it should be okay if we call `rtprecv_resync()` directly in line 466 and drop the `bool flush`. Note, that the payload handler needs `pt` and `mb` not more. And `mb` is not flushed. So it should be safe. Anyway, the payload handler in audio.c runs in another thread.
Be careful with the lock!

src/aureceiver.c
```diff
@@ -278,19 +271,32 @@ static int aurecv_stream_decode(struct audio_recv *ar,
 
 
 /* Handle incoming stream data from the network */
-void aurecv_receive(struct audio_recv *ar, const struct rtp_header *hdr,
+void aurecv_receive(struct audio_recv *ar, struct stream *strm,
+		    const struct rtp_header *hdr,
```
alfredh: another solution would be to add a new argument to this function
that is called e.g. `bool new_source` that would be set to `true` when
the incoming SSRC changed.

cspiel1: Good idea. This would be similar to
- `lostc` and `drop` parameters of function `handle_rtp()`,
- `lostc` and `ignore` parameters of the `stream_rtp_h()`
Adding a parameter would avoid accessing the `stream` object from the `audio_recv`.

src/rtprecv.c
```diff
@@ -322,6 +323,26 @@ static void decode_tmr(void *arg)
...
+			 const struct rtp_header *hdr)
+{
+	if (!rx || !hdr)
+		return;
+
+	jbuf_flush(rx->jbuf);
+
+	mtx_lock(rx->mtx);
+	rx->ssrc = hdr->ssrc;
+	rx->ssrc_set = true;
```
cspiel1, line 340: Is `rx->ssrc_set` set to `false` somewhere?

alfredh:
hello, is there any updates regarding this patch ?

If you think there are too many changes inside this PR, it is possible to split it up
into multiple PRs if that makes it easier.

src/aureceiver.c
cspiel1: > another solution would be to add a new argument to this function that is called e.g. `bool new_source` that would be set to `true` when the incoming SSRC changed.

Please could you try this suggestion of @alfredh?!

Turkish: @cspiel1  Yes, I just pushed an implementation of that, please review and let me know if I got the idea right

src/audio.c
```diff
@@ -649,6 +651,9 @@ static void stream_recv_handler(const struct rtp_header *hdr,
 	if (!a->aur)
 		return;
 
+	if (new_source)
+		aurecv_reset_on_source_change(a->aur);
```
cspiel1: Looks better now. I suggest to call it simply `aurecv_reset(a->aur)`.

Turkish: renamed :heavy_check_mark:

src/audio.c
```diff
@@ -638,9 +638,11 @@ static int stream_pt_handler(uint8_t pt, struct mbuf *mb, void *arg)
...
  */
 static void stream_recv_handler(const struct rtp_header *hdr,
-				struct rtpext *extv, size_t extc,
-				struct mbuf *mb, unsigned lostc, bool *ignore,
-				void *arg)
+					struct rtpext *extv, size_t extc,
+					struct mbuf *mb, unsigned lostc,
+					bool new_source,
+					bool *ignore,
+					void *arg)
```
cspiel1: Indentation changed. Please use "tabstop=8"!

alfredh:
In general this PR now looks quite good and I think we can merge it

Thanks for your good work and patience!

PR #3543, opened 2025-10-25
call: remove SDP label attribute (not used)

alfredh:


alfredh:
hi @cspiel1 could you please review and merge this one ?

alfredh:
Ping

cspiel1:
Yes, sorry! I missed that.

cspiel1:
I could see that `label` is from the beginning there. We don't use it.

PR #3545, opened 2025-10-26
g722: add libg722 module as alternative to avoid spandsp dependency

fosple:
This PR adds support for using **libg722** as an alternative G.722 codec backend in addition to the existing g722 implementation which depends on spandsp.

This is a continuation of https://github.com/baresip/baresip/pull/3467 , where the suggestion was made to include `libg722` as a separate module. 

### Motivation

SPANDSP is licensed under LGPL, which complicates static linking (e.g., for iOS App Store submissions).  
libg722 is in the public domain, making it a lightweight and license-friendly alternative.

This change enables developers to:
- Avoid LGPL restrictions where necessary.
- Build baresip with G.722 support even without SPANDSP installed.

### References
- Discussion: https://github.com/baresip/baresip/discussions/3465
- Related past issue: https://github.com/baresip/baresip/issues/350
- libg722 repository: https://github.com/sippy/libg722
- https://github.com/baresip/baresip/pull/3467

### DIFF
--- a/modules/libg722/CMakeLists.txt
+++ b/modules/libg722/CMakeLists.txt
@@ -10,8 +10,6 @@ endif()
 find_package(LIBG722)
 
 if(NOT LIBG722_FOUND)
-  message(WARNING
-      "libg722: libg722 not found - module will not be built")
   return()
 endif()
 


modules/libg722/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+project(libg722)
+
+find_package(LIBG722)
+
+if(NOT LIBG722_FOUND)
+  message(WARNING "libg722: libg722 not found - module will not be built")
```
cspiel1: No warning! Simply returns, like the other modules do so.

modules/libg722/libg722.c
```diff
@@ -0,0 +1,225 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <stdint.h>
+#include <re.h>
+#include <rem_au.h>
+#include <baresip.h>
+
+#ifndef USE_LIBG722
+#error "The libg722 module requires libg722"
+#endif
```
cspiel1: This seems to be also debugging code.

`.github/workflows/build.yml` needs an update to solve the github automatic check. If everything is green, you could drop the #error.

But it depends if we want to build the new module in build.yml or not.

src/account.c
```diff
@@ -379,16 +379,21 @@ static int audio_codecs_decode(struct account *acc, const struct pl *prm)
 		while (0 == csl_parse(&acs, cname, sizeof(cname))) {
 			struct aucodec *ac;
 			struct pl pl_cname, pl_srate, pl_ch = PL_INIT;
-			uint32_t srate = 8000;
-			uint8_t ch = 1;
+		uint32_t srate = 8000;
+		uint8_t ch = 1;
```
cspiel1: Unrelated change

src/account.c
```diff
@@ -379,16 +379,21 @@ static int audio_codecs_decode(struct account *acc, const struct pl *prm)
...
 
-			/* Format: "codec/srate/ch" */
+		if (0 == str_casecmp(cname, "libg722"))
+			str_ncpy(cname, "G722", sizeof(cname));
+
+		/* Format: "codec/srate/ch" */
```
cspiel1: Wrong indentation.

Do we really need the change in account.c?
An alternative behavior could be to prefer libg722 over spandsp if it is available.

juha-h:
Thanks, I built libg722 module for Android and G722 test call with Linux baresip using g722 module worked fine.

Also, I noticed that in order to get rid of spandsp dependency, I needed to remove g726 module from the build.

src/account.c
sreimers: yes let's introduce a preferred build order within cmake, so only one g722 module is build, no need for runtime configuration, I think.

alfredh:
> Thanks, I built libg722 module for Android and G722 test call with Linux baresip using g722 module worked fine.
> 
> Also, I noticed that in order to get rid of spandsp dependency, I needed to remove g726 module from the build.

is anyone using G.726 ? it is quite an old audio codec and maybe not so much used in the wild.
should we deprecate it or remove it ?

Or it can also be disabled by default.

juha-h:
> is anyone using G.726 ? it is quite an old audio codec and maybe not so much used in the wild. should we deprecate it or remove it ?
> 
> Or it can also be disabled by default.

My plan is to remove support of G726 codec from my baresip Android app once this PR is merged.  That will greatly simplify dependencies.

fosple:
Added the changes as proposed. Simplified the build so it's the same build code for Linux & MacOS now.

Feel free to make changes, if you don't feel this is right :)

juha-h:
I made more tests with the latest version of the PR and everything worked fine.  So merge would be OK to me.

PR #3553, opened 2025-11-03
play: warnings for failed audio devices

cspiel1:


src/play.c
```diff
@@ -302,8 +306,12 @@ int play_tone(struct play **playp, struct player *player,
 	err = auplay_alloc(&play->auplay, baresip_auplayl(),
 			   play_mod, &wprm,
 			   play_dev, write_handler, play);
-	if (err)
+       if (err) {
```
sreimers: the indentation looks wrong (spaces, instead tab), not sure why linting is not complaining.

cspiel1: Thanks! I am struggling around with my neovim config. :-(

cspiel1:
Any further comments?

alfredh:
Should we merge this one after 4.3.0 is released (this Wednesday) ?

cspiel1:
This are only warnings, including NULL pointer checks. Double checked again. Should be fine. You both may decide.

alfredh:
The release is planned for tomorrow, and you are the release manager so you can decide.

cspiel1:
Thanks Alfred!

PR #3556, opened 2025-11-05
menu: cleanup dialdir params

cspiel1:
A lot of command parameter parsing in menu looks similar. This PR moves parsing to the core and adds tests.

Reason: In a next PR we want to add another parameter `cuser=` to e.g. dialdir that allows to reliably choose a UA. This is needed for calling back a peer of a previous incoming INVITE.

- Coverage: 54.3% (min 54.0%)

### DIFF
--- a/test/cparam.c
+++ b/test/cparam.c
@@ -1,5 +1,5 @@
 /**
- * @file test/parse.c  Baresip selftest -- parse
+ * @file test/cparam.c  Baresip selftest -- Command parameter parse helpers
  *
  * Copyright (C) 2025 Christian Spielberger - c.spielberger@commend.com
  */


cspiel1:
Very strange that this violates the coverage boundary. Should we add a unit test for the menu module?
Unit test PR: https://github.com/baresip/baresip/pull/3557

Edit: modules are also counted for the coverage, right?

cspiel1:
TODO: unit test

cspiel1:
Removed the include to menu.h: https://github.com/baresip/baresip/pull/3558

Changes something. The parsing functions should be moved to the core.

modules/menu/static_menu.c
```diff
@@ -104,50 +104,29 @@ static int cmd_answer(struct re_printf *pf, void *arg)
...
 		return EINVAL;
 	}
 
 	call = ua_call(ua);
 
-	(void)pl_strdup(&cid, &callid);
+	(void)pl_strdup(&cid, &cp->callid);
 	if (str_isset(cid)) {
 		call = uag_call_find(cid);
 		cid = mem_deref(cid);
```
cspiel1: In a follow-up PR I'd like to add `uag_call_find_pl(const struct pl callid)` in order to simplify such code parts.

include/baresip.h
```diff
@@ -282,12 +282,49 @@ void call_get_media_estdir(struct call *call,
...
 
+/**
+ * Command parse helpers
+ */
+
+int cmd_prm_decode(const char *prm, const char *name, struct pl *val);
+
+struct media_dirs {
+	enum sdp_dir a;
+	enum sdp_dir v;
```
alfredh: Perhaps a bit longer struct name such as `au` or `audio` ?

cspiel1: Sorry, this is not used. I forgot to drop it.

src/call.c
```diff
@@ -3351,19 +3351,20 @@ const char *call_user_data(const struct call *call)
  * Set the user data of the call
  *
  * @param call Call object
- * @param user_data User data to be set
+ * @param pl User data to be set
```
alfredh: the parameter name can remain unchanged, even if the parameter type change ...

src/call.c
```diff
@@ -3351,19 +3351,20 @@ const char *call_user_data(const struct call *call)
...
  * @return int
  */
 
-int call_set_user_data(struct call *call, const char *user_data)
+int call_set_user_data(struct call *call, const struct pl *pl)
```
alfredh: now *set* and *get* user data is using different types (pl vs C-string)
and is no longer symmetrical.

it would be more elegant if both *set* and *get* used the same type ...

src/parse.c
```diff
@@ -0,0 +1,223 @@
...
+ * @file parse.c  Parsing helpers
+ *
+ * Copyright (C) 2025 Christian Spielberger - c.spielberger@commend.com
+ */
+
+#include <re.h>
+#include <baresip.h>
+
+
+static bool mdir_isvalid(struct pl *pl)
```
alfredh: this input parameter can be made `const`

alfredh:
is there a need to add generic support for command parameter parsing ?
for example:

```
/command param1=FOO param2 param3 param4
```

cspiel1:
> is there a need to add generic support for command parameter parsing ? for example:
> 
> ```
> /command param1=FOO param2 param3 param4
> ```

There are some special cases like "short form", "long form", "only one argument" that makes it difficult to apply a general parsing to e.g. odict. The output could be an odict. I am not sure if this would make it more readable or simpler to use.

cspiel1:
`pl_dup()` seems to miss to set the destructor like `pl_alloc_str()` does.

edit: Okay, does something different. What would be nice is a function like `struct pl * pl_copy(const struct pl *pl)`

edit: See: https://github.com/baresip/baresip/pull/3556#discussion_r2545057771

src/cparam.c
```diff
@@ -0,0 +1,223 @@
...
+ * Copyright (C) 2025 Christian Spielberger - c.spielberger@commend.com
+ */
+
+#include <re.h>
+#include <baresip.h>
+
+
+static bool mdir_isvalid(const struct pl *pl)
+{
+	if (!pl_strcmp(pl, "sendrecv"))
```
alfredh, line 13: perhaps more robust to use case-insensitive compare here, maybe the
user typed `Sendrecv` ... ?

src/parse.c
```diff
@@ -0,0 +1,223 @@
...
+ *
+ * @param prm  Command arguments parameter string
+ * @param name Parameter name
+ * @param val  Returned parameter value
+ *
+ * @return 0 for success, otherwise errorcode
+ */
+int cmd_prm_decode(const char *prm, const char *name, struct pl *val)
+{
+	char expr[128];
```
alfredh: it would be more robust to init expr to `""` -- in case the printf function below failed...

src/cparam.c
cspiel1: `sdp_dir_decode()` doesn't support case-insensitive checks.

include/baresip.h
```diff
@@ -281,13 +281,42 @@ void call_get_media_estdir(struct call *call,
...
 int call_msg_src(const struct call *call, struct sa *sa);
 enum sip_transp call_transp(const struct call *call);
 enum sdp_neg_state call_sdp_neg_state(const struct call *call);
 bool call_sdp_change_allowed(const struct call *call);
 
+/**
+ * Command parse helpers
+ */
+
+int cmd_prm_decode(const char *prm, const char *name, struct pl *val);
```
alfredh: I think the prefix of the functions should match the name of the file, i.e. `parse_` ...?

modules/menu/static_menu.c
```diff
@@ -572,15 +551,10 @@ static int dial_handler(struct re_printf *pf, void *arg)
...
-	char *ud_pos = NULL;
-	if (carg->prm != NULL)
-		ud_pos = strstr(carg->prm, ud_sentinel);
-	char *user_data = NULL;
-	if (ud_pos != NULL) {
-		user_data = ud_pos + strlen(ud_sentinel);
-		call_set_user_data(call, user_data);
-	}
+	struct pl pl = PL_INIT;
+	cmd_prm_decode(carg->prm, "userdata", &pl);
```
alfredh: check return value

src/call.c
```diff
@@ -3355,19 +3357,23 @@ const char *call_user_data(const struct call *call)
  * @return int
  */
```
maximilianfridrich: Style: Remove empty line?

src/call.c
```diff
@@ -83,7 +83,7 @@ struct call {
 	enum sdp_dir estvdir;      /**< Established video direction         */
 	bool use_video;
 	bool use_rtp;
-	char *user_data;           /**< User data related to the call       */
+	struct pl *user_data;      /**< User data related to the call       */
```
maximilianfridrich, line 86: Could it make sense to define this as `struct pl user_data` (no pointer)?

src/call.c
```diff
@@ -3355,19 +3357,23 @@ const char *call_user_data(const struct call *call)
...
 		return EINVAL;
 
 	call->user_data = mem_deref(call->user_data);
+	if (!pl_isset(user_data))
+		return 0;
 
-	int err = str_dup(&call->user_data, user_data);
-
+	char *str;
+	int err = pl_strdup(&str, user_data);
```
maximilianfridrich: Could this be simplified by just using `pl_dup` and making `call->user_data` a `pl` and not `pl*` (and `mem_deref(call->user_data.p)` in the `call_destructor`)? Avoids unnecessary memcpys.

src/parse.c
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * Decode call command parameters
+ *
+ * @param cp   Call command parameters
+ * @param prm  Command arguments parameter string
+ * @param pf   Print handler for debug output
+ * @return 0 for success, otherwise errorcode
+ */
+int call_cmd_prm_decode(struct call_cmd_prm **prmp, const char *prm,
```
maximilianfridrich: Would it make sense to incorporate the finding of the actual `call*` into this function if a `callid` was provided? I.e. call `uag_call_find` in this function and add a `call*` to the `call_cmd_prm` struct? That would potentially reduce code duplication and centralize the handling of this logic (which is the point of this PR).

src/parse.c
```diff
@@ -0,0 +1,223 @@
...
+	if (err)
+		mem_deref(cp);
+	else
+		*prmp = cp;
+
+	return err;
+}
+
+
+int ua_cmd_prm_decode(struct ua_cmd_prm **prmp,
```
maximilianfridrich: Would it make sense to incorporate the finding of the actual `ua*` into this function based on the `cp->uri`? I.e. call `uag_find_requri_pl` in this function and add `ua*` to the `ua_cmd_prm` struct? That would potentially reduce code duplication and centralize the handling of this logic (which is the point of this PR).

src/parse.c
```diff
@@ -0,0 +1,223 @@
...
+/**
+ * Decode a command parameter
+ *
+ * @param prm  Command arguments parameter string
+ * @param name Parameter name
+ * @param val  Returned parameter value
+ *
+ * @return 0 for success, otherwise errorcode
+ */
+int cmd_prm_decode(const char *prm, const char *name, struct pl *val)
```
maximilianfridrich: This function is basically identical to `fmt_param_sep_get`. Are both needed or could you just use `fmt_param_sep_get` (or `fmt_param_get`) in this file?

src/cparam.c
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	if (!ok) {
+		re_hprintf(pf, "dial URI missing\n");
+		err = EINVAL;
+		goto out;
+	}
+
+	prm = cp->uri.p + cp->uri.l;
+	struct pl pla = PL("sendrecv");
```
maximilianfridrich, line 185: Style: In the above function you don't mix code and declarations, but here you do. Maybe keeping it consistent could be nice.

src/call.c
cspiel1: I tried with both. And decided to use `pl_alloc_str()` which uses a destructor internally.

src/call.c
cspiel1: Yes, but then we would have to `mem_deref(call->user_data.p)`. Compared with `mem_deref(call->user_data)`.

src/cparam.c
cspiel1: In the above its not really possible / doesn't make it more readable.

edit: ok, there is a tiny improvement

src/parse.c
cspiel1: Yes, maybe. Will try this.

src/call.c
maximilianfridrich: So a double memcpy of the same data is better?

src/parse.c
cspiel1: I tried it. It would make parse.c and menu code a little bit simpler.
But it makes the `test/parse.c` very complicated because `struct ua` and `struct call` is not available. A lot of code will have to be duplicated from `test/call.c`.

src/parse.c
cspiel1: Same here.

src/parse.c
cspiel1: Instead I propose functions:
```c
struct call *call_find_id_pl(const struct list *calls, const struct pl *id);
struct call *uag_call_find_pl(const struct pl *id);
```

src/parse.c
cspiel1: Let's use it and add a wrapper! Thanks!

include/baresip.h
cspiel1: Then I suggest to rename the file to `cparam.c` and use the prefix `cparam_` for structs and functions.

modules/menu/static_menu.c
cspiel1: done

src/call.c
cspiel1: Okay, then let's add a new function to `re`!
```c
struct pl *pl_alloc_dup(const struct pl *src);
```
It really simplifies pl usage for this case. https://github.com/baresip/baresip/pull/3556/commits/03809b6c08cf33da7831e548316f5356dcbeefdb

re PR: https://github.com/baresip/re/pull/1469

maximilianfridrich: Thanks!

cspiel1:
Any further comments?

CMakeLists.txt
```diff
@@ -179,6 +179,7 @@ set(SRCS
   src/mnat.c
   src/module.c
   src/net.c
+  src/cparam.c
```
alfredh: it would be nice if source files are in alphabetical order ...

src/call.c
alfredh, line 86: `struct pl` is a very small struct so it could be better to just declare it as a compound
member and not a pointer:

```c
struct pl user_data
```

test/cparam.c
```diff
@@ -0,0 +1,270 @@
+/**
+ * @file test/parse.c  Baresip selftest -- parse
```
alfredh: comment does not match the filename ...

src/call.c
cspiel1: We found in combination with this `re` extension: https://github.com/baresip/re/pull/1469 it is slightly cleaner to use an alloc/destructor pattern like suggested. You may decide between:
```c
call->user_data = pl_alloc_dup(user_data);
...
call->user_data = mem_deref(call->user_data);
```
and
```c
err = pl_dup(&call->user_data, user_data);
...
mem_deref(call->user_data.p);
call->user_data = pl_null;
```

alfredh:
This one looks ready to merge ...

cspiel1:
Thanks!

PR #3557, opened 2025-11-05
test: a test for menu dialdir command

cspiel1:
menu.so is a central module, the example application and can be used also as API for higher level applications, like we do.

Should we add it to ci and start with a first test?


cspiel1:
I vaguely remember that we decided against it before to add tests for module menu.
Instead I am thinking about putting a dial parameter parsing function into the core with a unit test.

PR #3560, opened 2025-11-16
account: added account_set_pubint API function

juha-h:


juha-h:
I have no clue what `[Coverage / coverage (pull_request)](https://github.com/baresip/baresip/actions/runs/19402329219/job/55511618553?pr=3560)Failing after 55s` means.

sreimers:
Please rebase against main, should be fixed.

juha-h:
I did some tests with `account_set_pubint`.  When baresip starts with `;pubint=900`, PUBLISH is sent out.  When I start baresip with `;pubint=0` and then call `account_set_pubint(acc, 900)` followed by `ua_update_account(ua)`, no PUBLISH is sent out.  

Is this a bug in presence module or somewhere else?

alfredh:
Perhaps you can try to reload the `presence` module ?

juha-h:
> Perhaps you can try to reload the `presence` module ?

I'll try tomorrow, but it would be better if `pubint` change in one account had no effect on other accounts.

I suspect that there is some issue with `ua_update_account` when it does not do that the name implicates.

juha-h:
I did some tests with presence module unload/load.  If there is two accounts with presence state open and I set set pubint of one to 0 and then call presence module load/unload, presence state of the one with pubint 900 becomes closed.

So presence module unload/load after pubint change is not a solution.

alfredh:
I am not sure if changes in the accounts parameters will be picked up by modules.
At least there should be some kind of timer started in the module.

juha-h:
OK, so this PR is pretty much useless before `ua.update_account` also updates presence `pubint`, but it does not do any harm either.

cspiel1:
Right, a setter that is missing anyway, will not hurt. I'll merge this and prepare the release v4.3.0.

PR #3564, opened 2025-11-25
menu: add cuser parameter to dialdir command

cspiel1:
- **menu: add parameter cuser to dialdir**
- **bevent: add cuser to encoded JSON string**
- **uag: use pub-gruu to select UA by cuser**

This can be used in order to implement a call-back feature. In order to call back the peer of a missed incoming call it is important to select the right local UA.


cspiel1:
baresip does not append `;pub-gruu` as Contact parameter in the REGISTER. Should we add this?

PR #3565, opened 2025-11-25
uag: use GRUU if available for selecting UA

cspiel1:
If the registrar generated a GRUU which is returned in the response to the SIP REGISTER, then following requests will use the GRUU in the Request-URI. Thus baresip needs to compare the gruu if it is set to select the right UA.

Note: We try to find relevant sections in [RFC 5627](https://www.rfc-editor.org/rfc/rfc5627) and find a test setup.

### DIFF
--- a/src/uag.c
+++ b/src/uag.c
@@ -941,7 +941,7 @@ struct ua *uag_find_msg(const struct sip_msg *msg)
 		if (!acc || !acc->regint)
 			continue;
 
-		if (0 == pl_strcasecmp(cuser, ua_local_cuser(ua))) {
+		if (0 == pl_strcasecmp(cuser, ua_cuser(ua))) {
 			ua_printf(ua, "selected for %r\n", cuser);
 			return ua;
 		}


cspiel1:
Maybe we misinterpreted the use of the GRUU in the request line. Is it true that it is removed from the request line by the final proxy?
```
6.1.  Request Targeting
...
   Note that the "gr" URI parameter is retained just for the purposes of
   finding the GRUU in the location service; if a match is found, the
   Request-URI will be rewritten with the registered contacts, replacing
   the GRUU and its "gr" URI parameter.  The "gr" URI parameter is not
   carried forward into the rewritten Request-URI.
```
Currently we are not able to test it.

juha-h:
Christian Spielberger writes:

> Maybe we misinterpreted the use of the GRUU in the request line. Is it true that it is removed from the request line by the final proxy?
> ```
> 6.1.  Request Targeting
> ...
>    Note that the "gr" URI parameter is retained just for the purposes of
>    finding the GRUU in the location service; if a match is found, the
>    Request-URI will be rewritten with the registered contacts, replacing
>    the GRUU and its "gr" URI parameter.  The "gr" URI parameter is not
>    carried forward into the rewritten Request-URI.
> ```
> Currently we are not able to test it.

Try with my SIP proxy (Kamailio).  I'm sure that it works according to
the specs.

juha-h:
Check if the trace here, that includes invite and re-invite, has the answer:  https://box.tutpro.com/tmp/trace.txt.

cspiel1:
Thanks! At least we can see in your trace, that the request line does not contain the gruu.

Currently we are not sure if gruu can help to improve the UA selection for incoming SIP requests. This PR makes no sense, this is sure.

PR #3586, opened 2025-12-21
account: added ;check_origin parameter and API functions to get and set it

juha-h:
API function `ua_req_check_origin` now checks origin only if UA's account has `;check_origin=yes`. Default is "yes" for backwards compatibility.

### DIFF
--- a/src/account.c
+++ b/src/account.c
@@ -531,7 +531,7 @@ static int sip_params_decode(struct account *acc, const struct sip_addr *aor)
 	if (0 != msg_param_decode(&aor->params, "check_origin", &tmp))
 		acc->check_origin = true;
 	else
-		acc->check_origin = pl_strcasecmp(&tmp, "no") != 0;
+		pl_bool(&acc->check_origin, &tmp);
 
 	value = NULL;
 	err |= param_dstr(&value, &aor->params, "sipnat");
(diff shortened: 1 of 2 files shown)


src/account.c
```diff
@@ -528,6 +528,11 @@ static int sip_params_decode(struct account *acc, const struct sip_addr *aor)
 				  "outbound");
 	}
 
+	if (0 != msg_param_decode(&aor->params, "check_origin", &tmp))
+		acc->check_origin = true;
+	else
+		acc->check_origin = pl_strcasecmp(&tmp, "no") != 0;
```
sreimers: Please use `pl_bool`

src/account.c
```diff
@@ -750,6 +756,25 @@ int account_set_outbound(struct account *acc, const char *ob, unsigned ix)
...
 
+/**
+ * Sets check origin capability on (value true) or off (value false)
+ *
+ * @param acc      User-Agent account
+ * @param value    true or false
+ *
+ * @return 0 if success, otherwise errorcode
+ */
+int account_set_check_origin(struct account *acc, bool value)
```
sreimers: no error code needed:

```c
void account_set_check_origin(struct account *acc, bool value)
```

src/account.c
juha-h: Should be done now

src/account.c
juha-h: Should be done now

juha-h:
I don't know how to resolve the conflict,  I'll start from the beginning.

juha-h:
Due to the conflict, I'll start from scratch.

PR #3587, opened 2025-12-21
account: added ;check_origin parameter and API functions to get and set it

juha-h:
API function `ua_req_check_origin` now checks origin only if UA's account has `;check_origin=yes`. Default is "yes" for backwards compatibility.

### DIFF
--- a/src/account.c
+++ b/src/account.c
@@ -528,10 +528,7 @@ static int sip_params_decode(struct account *acc, const struct sip_addr *aor)
 				  "outbound");
 	}
 
-	if (0 != msg_param_decode(&aor->params, "check_origin", &tmp))
-		acc->check_origin = true;
-	else
-		pl_bool(&acc->check_origin, &tmp);
+	param_bool(&acc->check_origin, &aor->params, "check_origin");
 
 	value = NULL;
 	err |= param_dstr(&value, &aor->params, "sipnat");


alfredh:
I am now working to increase the testing coverage.

if you dont mind it would be nice if you could also add a small test for this new code.
There is a test in test/account.c that could be expanded to also include
decoding if these new parameters.

maximilianfridrich:
So this new option only works UDP, right? Because the `check_origin` setting is only checked if `filter_registrar` is set in which case only for UDP a listening socket is opened. Seems very specific. In the `filter_registrar` PR we first suggested to make it an account option, which was decided against.

- https://github.com/baresip/baresip/pull/3080

juha-h:
Function `ua_req_check_origin` is defined like this:
```
bool ua_req_check_origin(const struct ua *ua, const struct sip_msg *msg)
{
	struct le *le;

	if (!ua || !msg)
		return false;

	if (!u32mask_enabled(uag_cfg()->reg_filt, msg->tp))
		return true;

	if (!ua->acc->check_origin)
		return true;

	for (le = ua->regl.head; le; le = le->next) {
		struct reg *reg = le->data;
		if (sa_cmp(reg_paddr(reg), &msg->src, SA_ADDR))
			return true;
	}

	return false;
}
```
With  `filter_registrar udp,tcp,tls,ws,wss`, `if (!u32mask_enabled(uag_cfg()->reg_filt, msg->tp))` always returns false.

juha-h:
If reg_filter would be per account, this PR won't be needed.

juha-h:
`filter_registrar` makes only sense with accounts that register.  If there is also an account that does not register, then the global filter would always make the check fail.  Thus in my opinion, global filter makes no general sense and I don't have no idea why it was added to baresip.

src/account.c
```diff
@@ -528,6 +528,11 @@ static int sip_params_decode(struct account *acc, const struct sip_addr *aor)
 				  "outbound");
 	}
 
+	if (0 != msg_param_decode(&aor->params, "check_origin", &tmp))
+		acc->check_origin = true;
+	else
+		pl_bool(&acc->check_origin, &tmp);
+
```
alfredh, line 532: You can also use `param_bool()` instead of this block

src/ua.c
```diff
@@ -2343,6 +2344,9 @@ bool ua_req_check_origin(const struct ua *ua, const struct sip_msg *msg)
 	if (!u32mask_enabled(uag_cfg()->reg_filt, msg->tp))
 		return true;
 
+	if (!ua->acc->check_origin)
+		return true;
```
alfredh, line 2348: is this config option also valid for connection oriented protocols such as TLS
that has an active connection from client to server ?

juha-h: Yes, the check is made each time an initial request arrives.

The whole thing is much bigger mess than it could be.The best account configuration parameter for clients using connection oriented protocols (tcp, tls, ws. wss) for registrations would be  "do not accept any connection attempts from anywhere".  The reason is that all well designed SIP clients keep the connection(s) open to the registrar(s) and registrars/proxies use the existing connection(s) from client to the server to send all in-dialog and new out-of-dialog request to the client.  Thus the server never has any need to setup a connection to the client.

src/account.c
juha-h: Done

src/ua.c
cspiel1: This is not always the case for TCP. E.g.
- The qualifying feature of Asterisk periodically sends SIP OPTIONS to registered clients. A separate TCP connection is used by default.
- Also SIP INVITE may be sent via a separate TCP connection to the client.

cspiel1: The global setting should be a simple option to switch security checks globally on. We need this setting for our customers.

If a per account exceptions from that are needed, then this PR would be okay, if like you suggest to keep the check active by default.

cspiel1: In other words, this PR is okay for us like it is.

juha-h: It is OK to keep the global setting and this PR does not make any changes to its behavior. 

Regarding establishing a new TCP connection to the client,  it would not work in my tests when the client is connected to Internet using mobile network.  Just tested with netcat and got
```
$ nc -v 103.165.148.241 5060
nc: connect to 103.165.148.241 port 5060 (tcp) failed: Connection refused
```
@cspiel1 In your Asterisk setup, is Asterisk able to send OPTIONS requests over new TCP connections to mobile network connected devices?

juha-h: If there is no issues remaining, then let's merge the PR.

cspiel1: No, I don't think so. I meant in local networks. So the SIP server and the clients are in the same local network. If somebody wants to keep this working he needs to disable the `filter_registrar` global setting or use the new `check_origin=off` for the account.

juha-h: Yes, the reason I wrote this PR was that I want have `check_origin=yes` in register accounts and `no` in peer-to-peer accounts.

juha-h:
So, could I get merge as Christmas present? :)

juha-h:
Thanks!

PR #3596, opened 2025-12-27
audio: remove audio_txtelev_empty() -- unused

alfredh:


cspiel1:
Now baresip-apps build is broken:
```
/home/cspiel/src/baresip-apps/modules/intercom/ichidden.c:104:8: error: call to undeclared function 'audio_txtelev_empty'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  104 |                         if (audio_txtelev_empty(call_audio(call))) {
      |                             ^
```

alfredh:
Sorry about that -- I have reverted the change in main... Please try again ...

Are you able to add usage of this function in the selftest ?
This serves as documentation that the function is used.

cspiel1:
Thanks! I'll try to add a selftest.

Explanation
----------------
The hidden call in the intercom module can be used to send a DTMF (e.g. open a door) without having a real call with speech. The removed function is used to check that the DTMF characters have been sent out before terminating the SIP call.

PR #3599, opened 2025-12-30
mixausrc: rework

cspiel1:
- There are two threads involved. One for the stream and one for the
  ausrc. This commit should avoid race conditions.
- Use auframe functions of aubuf.
- General code cleanup.


alfredh:
Would it be possible to include this module in one of the existing testcases ?

cspiel1:
The file test/call.c already has nearly 4k lines. Should we separate the utility functions first? Here is a PR: https://github.com/baresip/baresip/pull/3610 ?

PR #3606, opened 2026-01-02
test: cast expected and actual to int

cspiel1:
I got this runtime error output from the warning in `ASSERT_EQ()` if `expected != actual`

https://github.com/baresip/baresip/pull/3607#issuecomment-3705311492

```
Format: "selftest: ASSERT_EQ: %s:%u: %s(): expected=%d<-- SIZE ERROR
selftest: /home/cspiel/src/re/src/fmt/print.c:501: int vhprintf(const char *, struct __va_list_tag *, re_vprintf_h *, void *, _Bool): Assertion `0 && "RE_VA_ARG: arg is not compatible"' failed.
```
@sreimers Note, the call stack printed by the `vhprintf()` Assertion seems to be wrong. See https://github.com/baresip/baresip/pull/3607#issuecomment-3705311492!

### DIFF
--- a/test/test.h
+++ b/test/test.h
@@ -15,11 +15,11 @@
 
 #define ASSERT_EQ(expected, actual)				\
 	if ((expected) != (actual)) {				\
-		warning("selftest: ASSERT_EQ: %s:%u: %s():"	\
-			" expected=%d(0x%x), actual=%d(0x%x)\n",\
-			__FILE__, __LINE__, __func__,		\
-			(expected), (expected),			\
-			(actual), (actual));			\
+		warning("selftest: ASSERT_EQ: %s:%u: %s(): "	\
+			"expected=%ld(0x%lx), actual=%ld(0x%lx)\n",	\
+			__FILE__, __LINE__, __func__,			\
+			(long int)(expected), (long int)(expected),	\
+			(long int)(actual),   (long int)(actual));	\
 		err = EINVAL;					\
 		goto out;					\
 	}


sreimers:
> Note, the call stack printed by the `vhprintf()` Assertion seems to be wrong

The stack belongs to the bad signal not the ASSERT, you have to use a debugger and backtrace, to find the right ASSERT_EQ.

cspiel1:
Sorry, the description is confusing. But please look at the diff!
If other types than `int` are passed to the macro, and the `if` condition is false, then the warning `fmt` does not fit to the arguments.

Let me reopen for now!

cspiel1:
A cast to `long int` works also for pointers which is used at some places (e.g. test/jbuf.c:108)

sreimers:
I don't think a generic cast will work here for all cases. It has to be fixed/cast within the test code itself. And it's not a big problem, since ASSERT_EQ is only for testing and if it fails something is already wrong and breaks the test.

cspiel1:
Not sure if we are talking about the same problem. I updated my `re` and have your newest commit b3a26b570a818f8adb5a08e3f8f4cdc8caf7daba "fmt/print: add backtrace for incompatible format arguments".

When somebody is developing a unit test, the warning is important. Currently when I work on `test_call_dtmf()` then you're right that the "SIZE ERROR" with
```c
ASSERT_EQ(n, fix.b.n_dtmf_recv);
```
can be avoided. But I got confused because the backtrace starts with:
```
test_call_ipv6ll at /home/cspiel/src/baresip/test/call.c:3076
```

You're right that with
```c
ASSERT_EQ((unsigned) n, fix.b.n_dtmf_recv);
```
the warning is printed correctly. It might be okay if an unsigned int is passed to `%d`.

Another example is test/jbuf.c line 108. As an experiment I changed the line to
```c
ASSERT_EQ(mem, frv[0]);
```
If `i==1` the assertion fails and the warning produces a SIZE ERROR. It is not okay to pass a pointer to `%d` nor to `%x`. And the backtrace is confusing.

Note, that debugging of real-time threads is not always possible.

sreimers:
I think we can keep it simple and use the insecure version for test cases:

https://github.com/baresip/baresip/pull/3608

I'm not a big fan of casting everything it hides problems and should avoided as much as possible.

cspiel1:
Okay, I close this.

Then one thing for me is open. I want to understand what `btrace` prints in this case. When I have more time I'll try to find the explanation myself.

cspiel1:
I found the backtrace problem: https://github.com/baresip/re/pull/1500

PR #3610, opened 2026-01-07
test: split test/call.c

cspiel1:
Separates test/call utility functions from the test functions.

Now test/call.c contains only the test functions.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -11,8 +11,8 @@ include_directories(${RE_INCLUDE_DIRS} ${BARESIP_INCLUDE_DIRS})
 add_executable(${PROJECT_NAME}
   account.c
   call.c
-  call_fixture.c
   call_cancelrule.c
+  call_fixture.c
   cmd.c
   contact.c
   bevent.c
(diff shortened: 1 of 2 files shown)


test/call_cancelrule.c
```diff
@@ -0,0 +1,402 @@
+/**
+ * @file call_fixture.c  Call tests -- Cancel rules
```
sreimers: Wrong filename

test/CMakeLists.txt
```diff
@@ -11,6 +11,8 @@ include_directories(${RE_INCLUDE_DIRS} ${BARESIP_INCLUDE_DIRS})
 add_executable(${PROJECT_NAME}
   account.c
   call.c
+  call_fixture.c
+  call_cancelrule.c
```
sreimers, line 14: Keep alphabetic ordering

test/call_cancelrule.c
```diff
@@ -0,0 +1,402 @@
...
+	struct le *le;
+
+	LIST_FOREACH(&f->rules, le) {
+		struct cancel_rule *cr = le->data;
+
+		cancel_rule_reset(cr);
+	}
+}
+
+#define UINTSET(u) (u) != (unsigned) -1
```
sreimers: Bette move all definitions to the top, the name is a bit misleading btw and defined two times already

PR #3612, opened 2026-01-07
Add bidirectional rtsp module.

jabjoe:
Add bidirectional audio rtsp module. Based of gst module, but changed as needed for bidirectional RTSP audio.

Tested on a Wyze Doorbell V1 flashed with Thingino.

### DIFF
--- a/modules/rtsp/rtsp.c
+++ b/modules/rtsp/rtsp.c
@@ -25,7 +25,8 @@
  *
  * Audio source/player module using gstreamer 1.0
  *
- * The module 'rtsp' is using the Gstreamer framework to do audio bi-directionally to rtsp
+ * The module 'rtsp' is using the Gstreamer framework to do audio
+ * bi-directionally to rtsp
  *
  * Example config:
  \verbatim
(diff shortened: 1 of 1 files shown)


cspiel1:
We could add this to https://github.com/baresip/baresip-apps. For the baresip repository it might be a use-case that is too specific.

Edit: I discussed with my colleagues. And I think that @alfredh will also agree that this should be moved to baresip-apps. Please open a new PR there and close this! Note also the following comments! Thanks!

modules/rtsp/rtsp.c
```diff
@@ -0,0 +1,740 @@
...
+    GstElement * src = NULL, * sink = NULL, *pipeline = NULL;
+
+    GError *error = NULL;
+    const GstStructure *s;
+
+    const gchar *encoding;
+
+    mtx_lock(&backchannel.lock);
+    printf("Trying to setup backchannel.\n");
+    if (!backchannel.options_num || backchannel.option < 0 || (unsigned)backchannel.option >= backchannel.options_num) {
```
cspiel1: Line length is 80 characters. Use `test/ccheck.py` to check code formatting!

modules/rtsp/rtsp.c
```diff
@@ -0,0 +1,740 @@
...
+        sys_msleep(dt);
+    }
+    info("rtsp: Stopping write thread.\n");
+    return 0;
+}
+
+
+static void backchannel_init(void)
+{
+    const gchar * outfmt;
```
cspiel1: format like this:
```c
const gchar *outfmt;
```

modules/rtsp/rtsp.c
```diff
@@ -0,0 +1,740 @@
...
+    /* Empty buffer now */
+    while (st->run) {
+        const struct timespec delay = {0, st->prm.ptime*1000000/2};
+
+        play_packet(st);
+
+        if (aubuf_cur_size(st->aubuf) < st->psize)
+            break;
+
+        (void)nanosleep(&delay, NULL);
```
cspiel1: use `sys_usleep()` or `sys_msleep()` from `re`

PR #3614, opened 2026-01-08
uag: uag_filter_calls() support unlink in listh

cspiel1:
If in the given list handler the call is unlinked (terminated) then the `le` got invalid.

### DIFF
--- a/src/uag.c
+++ b/src/uag.c
@@ -185,12 +185,11 @@ void uag_filter_calls(call_list_h *listh, call_match_h *matchh, void *arg)
 
 	for (leu = list_head(uag_list()); leu; leu = leu->next) {
 		struct ua *ua = leu->data;
-		struct le *lec;
-		struct le *leprev;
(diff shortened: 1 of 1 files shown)


src/uag.c
```diff
@@ -186,9 +186,11 @@ void uag_filter_calls(call_list_h *listh, call_match_h *matchh, void *arg)
 	for (leu = list_head(uag_list()); leu; leu = leu->next) {
 		struct ua *ua = leu->data;
 		struct le *lec;
+		struct le *leprev;
 
-		for (lec = list_tail(ua_calls(ua)); lec; lec = lec->prev) {
+		for (lec = list_tail(ua_calls(ua)); lec; lec = leprev) {
 			struct call *call = lec->data;
+			leprev = lec->prev;
```
sreimers: I like to use `while` for these cases:

```c
struct le *lec = list_tail(ua_calls(ua));
while(lec) {
  struct call *call = lec->data;
  lec = lec->prev;
...
}

PR #3615, opened 2026-01-12
test mixausrc

cspiel1:
- **test: call - add test for mixausrc**
- **mixausrc: add command parameter cname**
- **mixausrc: add command parameter fadetime**

Requested by @alfredh : https://github.com/baresip/baresip/pull/3599#issuecomment-3704798765

### DIFF
--- a/test/call.c
+++ b/test/call.c
@@ -472,7 +472,7 @@ static int test_call_mixausrc_priv(bool dec)
 	struct ausrc *ausrc = NULL;
 	struct auplay *auplay = NULL;
 	const char *dp = test_datapath();
-	char *prm;
+	char *prm = NULL;
 	int err = 0;
 
 	err = re_sdprintf(&prm, ";ptime=10"
@@ -480,7 +480,7 @@ static int test_call_mixausrc_priv(bool dec)
 			  ";audio_player=mock-auplay,a", dp);
 	TEST_ERR(err);
 	fixture_init_prm(f, prm);
-	mem_deref(prm);
+	prm = mem_deref(prm);
 	mem_deref(f->b.ua);
 
 	err = re_sdprintf(&prm, "B <sip:b@127.0.0.1>;regint=0;ptime=10"
@@ -488,7 +488,7 @@ static int test_call_mixausrc_priv(bool dec)
 			  ";audio_player=mock-auplay,b", dp);
 	TEST_ERR(err);
 	err = ua_alloc(&f->b.ua, prm);
-	mem_deref(prm);
+	prm = mem_deref(prm);
 	TEST_ERR(err);
 
 	conf_config()->avt.rtp_stats = true;
@@ -511,7 +511,6 @@ static int test_call_mixausrc_priv(bool dec)
 	f->estab_action = ACTION_NOTHING;
 
 	/* Make a call from A to B */
-	TEST_ERR(err);
 	err = ua_connect(f->a.ua, 0, NULL,f->buri, VIDMODE_OFF);
 	TEST_ERR(err);
 
@@ -530,7 +529,7 @@ static int test_call_mixausrc_priv(bool dec)
 				dec ? "dec" : "enc", dp, dec ? "b@" : "a@");
 	TEST_ERR(err);
 	err = fixture_delayed_command(f, 0, prm);
-	mem_deref(prm);
+	prm = mem_deref(prm);
 	TEST_ERR(err);
 	err = re_main_timeout(1000);
 	TEST_ERR(err);
@@ -549,6 +548,7 @@ static int test_call_mixausrc_priv(bool dec)
 	TEST_ERR(fix.err);
 
  out:
+	mem_deref(prm);
 	conf_config()->avt.audio.jbuf_del.min = 100;
 	fixture_close(f);
 	mem_deref(ausrc);
--- a/test/test.h
+++ b/test/test.h
@@ -154,7 +154,6 @@ int dns_server_add_srv(struct dns_server *srv, const char *name,
  */
 
 struct auplay;
-struct ausrc;
 
 typedef void (mock_sample_h)(struct auframe *af, const char *dev, void *arg);
 


cspiel1:
TODO: test `mixausrc_dec_start` and `mixausrc_dec_stop`.

test/CMakeLists.txt
```diff
@@ -41,6 +41,7 @@ add_executable(${PROJECT_NAME}
 
   mock/mock_aucodec.c
   mock/mock_auplay.c
+  mock/mock_ausrc.c
```
alfredh: is it possible to reuse an existing audio-source module (e.g. aufile, ausine)
instead of adding a new mock source ?

By reusing a module we can get more testing coverage of that module ...

cspiel1: Yes, with aufile. I'll try it.

cspiel1:
TODO: replace mock_ausrc.c by using aufile

edit: done

cspiel1:
There is a "lock-order inversion" detected at rtprecv.c:731.
Run job: https://github.com/baresip/baresip/actions/runs/21023271875/job/60442070623?pr=3615 

Also mentioned here: https://github.com/baresip/baresip/pull/3616#issuecomment-3743642510

cspiel1:
|acquire|holding|thread|code|lock|description|
|---|---|--|--|--|--|
| M1 | M0 | main | rtprecv.c:731 --> async.c:357 | async_work->mtx | During stream shutdown the async thread is canceled with `re_thread_async_main_cancel()`. |
| M2 | M1 | main | rtprecv.c:539 | rx->mtx | When mnat connected, the streams starts RTCP. |
| M0 | M2 | RX | rtprecv.c:148 --> async.c:288 | async->mtx | The RX thread pushes an async work to the main thread with `re_thread_async_main_id()`. |

alfredh:
Here is a patch related to the "lock order inversion":


```patch
diff --git a/src/async/async.c b/src/async/async.c
index 9d17876..cd532ee 100644
--- a/src/async/async.c
+++ b/src/async/async.c
@@ -152,12 +152,17 @@ static void queueh(int id, void *data, void *arg)
        (void)id;
 
        mtx_lock(work->mtx);
-       if (work->cb) {
-               work->cb(work->err, work->arg);
-               work->cb =NULL;
-       }
+       re_async_h *cb = work->cb;
+       void *cb_arg   = work->arg;
+       int err        = work->err;
+       work->cb = NULL;
        mtx_unlock(work->mtx);
 
+       /* NOTE: calling callback handler with mutex unlocked */
+       if (cb) {
+               cb(err, cb_arg);
+       }
+
        mtx_lock(&async->mtx);
        list_move(&work->le, &async->freel);
        mtx_unlock(&async->mtx);
```


You are welcome to test it :)

cspiel1:
> Here is a patch related to the "lock order inversion":
> ...
> You are welcome to test it :)

Thanks! I will have a look. I guess that the lock-order inversion happens only very sporadic.

In the meanwhile I came to the assumption that the last row in the above table is the problem and tried to fix it like that in `rtprecv.c` function `pass_rtpestab_work()`. What do you think?

```c
	mtx_unlock(rx->mtx);
	re_thread_async_main_id((intptr_t)rx, NULL, async_work_main, w);
	mtx_lock(rx->mtx);
```

Edit: I understand your patch now. It is the second row in my table and makes more sense to fix it there.

cspiel1:
@alfredh The patch looks good. I ran the selftest several times with clang sanitizer=thread and it didn't detect the lock order inversion. Do you want to open a `re` PR?

alfredh:
> @alfredh The patch looks good. I ran the selftest several times with clang sanitizer=thread and it didn't detect the lock order inversion. Do you want to open a `re` PR?

I think Sebastian has to review the patch first, you can create a new re PR if you want ...


In general a mutex should protect the data members of a struct, and not function calls.
In my experience the mutex should be unlocked while calling functions, because you
cannot know what is that function doing, or what other functions is is calling.

```c
static void pass_rtpestab_work(struct rtp_receiver *rx)
{
	struct work *w;

	if (!re_atomic_rlx(&rx->run)) {
		rx->rtpestabh(rx->strm, rx->sessarg);
		return;
	}

	w = mem_zalloc(sizeof(*w), work_destructor);
	w->type = WORK_RTPESTAB;
	w->rx   = rx;

	re_thread_async_main_id((intptr_t)rx, NULL, async_work_main, w);
}
```


in order to avoid deadlock it would be good if mutex is unlocked
while calling `rx->rtpestabh()`

cspiel1:
> in order to avoid deadlock it would be good if mutex is unlocked while calling `rx->rtpestabh()`

This is called from `async_work_main()` the main thread. So I guess you mean the mutex of `struct async_work` in `re/src/async/async.c`, right?

sreimers:
> Here is a patch related to the "lock order inversion":
> 
> ```diff
> diff --git a/src/async/async.c b/src/async/async.c
> index 9d17876..cd532ee 100644
> --- a/src/async/async.c
> +++ b/src/async/async.c
> @@ -152,12 +152,17 @@ static void queueh(int id, void *data, void *arg)
>         (void)id;
>  
>         mtx_lock(work->mtx);
> -       if (work->cb) {
> -               work->cb(work->err, work->arg);
> -               work->cb =NULL;
> -       }
> +       re_async_h *cb = work->cb;
> +       void *cb_arg   = work->arg;
> +       int err        = work->err;
> +       work->cb = NULL;
>         mtx_unlock(work->mtx);
>  
> +       /* NOTE: calling callback handler with mutex unlocked */
> +       if (cb) {
> +               cb(err, cb_arg);
> +       }
> +
>         mtx_lock(&async->mtx);
>         list_move(&work->le, &async->freel);
>         mtx_unlock(&async->mtx);
> ```
> 
> You are welcome to test it :)

The patch looks good to me :) We should not lock the callback function.

cspiel1:
For this PR I'll extract the mixausrc fixes to a separate smaller PR, and then rebase this one.

cspiel1:
> The patch looks good to me :) We should not lock the callback function.

Here is the PR: https://github.com/baresip/re/pull/1503. Thanks @alfredh !

alfredh:
Here is an example of a robust code that has been working well for many years, in the coreaudio module:

```c
static void record_handler(void *userData, AudioQueueRef inQ,
			   AudioQueueBufferRef inQB,
			   const AudioTimeStamp *inStartTime,
			   UInt32 inNumPackets,
			   const AudioStreamPacketDescription *inPacketDesc)
{
	struct ausrc_st *st = userData;
	struct auframe af;
	ausrc_read_h *rh;
	void *arg;
	(void)inStartTime;
	(void)inNumPackets;
	(void)inPacketDesc;

	mtx_lock(&st->mutex);
	rh  = st->rh;
	arg = st->arg;
	mtx_unlock(&st->mutex);

	if (!rh)
		return;

	auframe_init(&af, st->prm.fmt, inQB->mAudioData,
		     inQB->mAudioDataByteSize / st->sampsz, st->prm.srate,
		     st->prm.ch);

	af.timestamp = AUDIO_TIMEBASE*inStartTime->mSampleTime / st->prm.srate;

	rh(&af, arg);

	AudioQueueEnqueueBuffer(inQ, inQB, 0, NULL);
}
```

This code is also so robust that the callback handler can delete the ausrc state object.

The similar pattern can be applied also to other parts of the code.

1. Use the lock/unlock to copy the callback handler and argument
2. Call the callback handler with the mutex unlocked
3. Dont access the state object after calling the callback handler

test/call.c
```diff
@@ -413,6 +414,167 @@ static bool linenum_are_sequential(const struct ua *ua)
...
+	struct cancel_rule *cr;
+	struct ausrc *ausrc = NULL;
+	struct auplay *auplay = NULL;
+	const char *dp = test_datapath();
+	char *prm;
+	int err = 0;
+
+	err = re_sdprintf(&prm, ";ptime=10"
+			  ";audio_source=aufile,%s/wav/square_500Hz_0.8.wav"
+			  ";audio_player=mock-auplay,a", dp);
```
alfredh, line 480: is `prm` also freed at the end of the function ?

test/call.c
```diff
@@ -413,6 +414,167 @@ static bool linenum_are_sequential(const struct ua *ua)
...
+
+	err = mock_auplay_register(&auplay, baresip_auplayl(),
+				   mixdetect_handler, f);
+	TEST_ERR(err);
+
+	f->behaviour = BEHAVIOUR_ANSWER;
+	f->estab_action = ACTION_NOTHING;
+
+	/* Make a call from A to B */
+	TEST_ERR(err);
```
alfredh: is this `TEST_ERR` needed ?

test/test.h
```diff
@@ -154,6 +154,7 @@ int dns_server_add_srv(struct dns_server *srv, const char *name,
  */
 
 struct auplay;
+struct ausrc;
```
alfredh: is this line needed ?

test/call.c
cspiel1: No, always immediately after use. I'll add a final `mem_deref()` at the end in order to be save for future changes.

test/call.c
cspiel1: No

test/test.h
cspiel1: No.

cspiel1:
Coverage is now at 67%. The boundary can be increased.

PR #3618, opened 2026-01-14
test: add memory cleanup check

sreimers:
each test should keep a clean memory state after execution

### DIFF
--- a/test/main.c
+++ b/test/main.c
@@ -182,6 +182,7 @@ static int run_tests(void)
 
 		/* Flush mainloop after each test */
 		re_fhs_flush();
+		dnsc_cache_flush(net_dnsc(baresip_network()));
 
 		mem_get_stat(&mstat_after);
 		if (mstat_after.bytes_cur != mstat_before.bytes_cur) {


test/main.c
```diff
@@ -169,6 +171,7 @@ static int run_tests(void)
 		re_printf("[ RUN      ] %s (rx %s)\n",
 			  tests[i].name, rtp_receive_mode_str(rxmode));
 
+		mem_get_stat(&mstat_before);
```
alfredh, line 174: for Release builds this function will return `ENOSYS`

test/main.c
```diff
@@ -177,6 +180,20 @@ static int run_tests(void)
 			return err;
 		}
 
+		/* Flush mainloop after each test */
+		re_fhs_flush();
```
alfredh, line 184: should we also flush the DNS client cache and the async module here ?

test/main.c
```diff
@@ -177,6 +180,21 @@ static int run_tests(void)
...
+		dnsc_cache_flush(net_dnsc(baresip_network()));
```
alfredh, line 185: should we also flush FHS where we run a single testcase ?

sreimers:
Will open a new PR and align this like in libre

PR #3646, opened 2026-02-07
fedora: add valgrind to selftest

alfredh:


alfredh:
during one of the jobs this warning appeared:

`async: all async workers are busy`

Could this lead to an error, should we increase number of workers ?

PR #3653, opened 2026-02-17
audio: remove audio_mode

alfredh:
This is a proposal to remove audio txmode config and only use a thread.

The chosen mode could be "polling" or "thread" but I think thread is slightly better
as it gives better packet pacing and also aligns with the "receive thread" mode.


cspiel1:
We use tx-mode **thread** because we also observed that we obtain lower TX jitter on our embedded devices.

alfredh:
It looks like there is consensus for this change, thanks for the feedback.

PR #3655, opened 2026-02-18
mixausrc: do not exit if cname is not set

maximilianfridrich:
cname is not strictly required and if no audio struct is passed, mixausrc should still work.

### DIFF
--- a/modules/mixausrc/mixausrc.c
+++ b/modules/mixausrc/mixausrc.c
@@ -417,12 +417,10 @@ static int encode_update(struct aufilt_enc_st **stp, void **ctx,
 			 const struct audio *au)
 {
 	struct mixausrc_enc *enc;
-	int err;
-	(void)au;
-	(void)af;
 	(void)ctx;
+	(void)af;
 
-	if (!stp || !ctx || !prm)
+	if (!stp || !prm)
 		return EINVAL;
 
 	if (*stp)
@@ -433,9 +431,11 @@ static int encode_update(struct aufilt_enc_st **stp, void **ctx,
 		return ENOMEM;
 
 	enc->st.fadetime = DEFAULT_FADE_TIME;
-	err = str_dup(&enc->cname, stream_cname(audio_strm(au)));
-	if (err)
-		return err;
+	if (au) {
+		int err = str_dup(&enc->cname, stream_cname(audio_strm(au)));
+		if (err)
+			return err;
+	}
 
 	list_append(&encs, &enc->le_priv, enc);
 	*stp = (struct aufilt_enc_st *) enc;
@@ -449,10 +449,8 @@ static int decode_update(struct aufilt_dec_st **stp, void **ctx,
 			 const struct audio *au)
 {
 	struct mixausrc_dec *dec;
-	int err;
-	(void)au;
-	(void)af;
 	(void)ctx;
+	(void)af;
 
 	if (!stp || !prm)
 		return EINVAL;
@@ -465,9 +463,11 @@ static int decode_update(struct aufilt_dec_st **stp, void **ctx,
 		return ENOMEM;
 
 	dec->st.fadetime = DEFAULT_FADE_TIME;
-	err = str_dup(&dec->cname, stream_cname(audio_strm(au)));
-	if (err)
-		return err;
+	if (au) {
+		int err = str_dup(&dec->cname, stream_cname(audio_strm(au)));
+		if (err)
+			return err;
+	}
 
 	list_append(&decs, &dec->le_priv, dec);
 	*stp = (struct aufilt_dec_st *) dec;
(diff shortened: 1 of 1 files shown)


modules/mixausrc/mixausrc.c
```diff
@@ -433,7 +432,14 @@ static int encode_update(struct aufilt_enc_st **stp, void **ctx,
 		return ENOMEM;
 
 	enc->st.fadetime = DEFAULT_FADE_TIME;
-	err = str_dup(&enc->cname, stream_cname(audio_strm(au)));
+
+	/* Get cname from audio stream if possible, fallback to ctx */
+	cname = stream_cname(audio_strm(au));
+	if (!cname && ctx && *ctx) {
+		cname = (const char *)*ctx;
+	}
```
cspiel1, line 438: Background information that `*ctx` is a `const char *` is used here.

@maximilianfridrich please replace this by a logic like before 764ac1c63cc12e3b79d71780e592be2c7e2489e9.
If the user specifies a `cname=` parameter and there is no audio object, or cname, then simply return from the command functions!
If the user does not specify the `cname=` paramter, then take the head of the list!

Note that normally only the mixausrc test in `test/call.c` uses this parameter.

modules/mixausrc/mixausrc.c
```diff
@@ -465,9 +459,7 @@ static int decode_update(struct aufilt_dec_st **stp, void **ctx,
 		return ENOMEM;
 
 	dec->st.fadetime = DEFAULT_FADE_TIME;
-	err = str_dup(&dec->cname, stream_cname(audio_strm(au)));
-	if (err)
-		return err;
+	(void)str_dup(&dec->cname, stream_cname(audio_strm(au)));
```
cspiel1: Better do not ignore ENOMEM!

```c
if (au)
	err = str_dup(&dec->cname, stream_cname(audio_strm(au)));
if (err)
	return err;
```

modules/mixausrc/mixausrc.c
```diff
@@ -847,9 +839,11 @@ static struct mixausrc_enc *enc_find(const struct cmd_arg *carg)
 
 	struct mixausrc_enc *enc = NULL;
 	for (struct le *le = encs.head; le; le = le->next) {
-		enc = le->data;
-		if (!re_regex(enc->cname, str_len(enc->cname), cname_str))
+		struct mixausrc_enc *tmp_enc = le->data;
+		if (!re_regex(tmp_enc->cname, str_len(tmp_enc->cname), cname_str)) {
+			enc = tmp_enc;
 			goto out;
+		}
```
cspiel1, line 854: Is the change needed?

modules/mixausrc/mixausrc.c
```diff
@@ -874,9 +868,11 @@ static struct mixausrc_dec *dec_find(const struct cmd_arg *carg)
 
 	struct mixausrc_dec *dec = NULL;
 	for (struct le *le = decs.head; le; le = le->next) {
-		dec = le->data;
-		if (!re_regex(dec->cname, str_len(dec->cname), cname_str))
+		struct mixausrc_dec *tmp_dec = le->data;
+		if (!re_regex(tmp_dec->cname, str_len(tmp_dec->cname), cname_str)) {
+			dec = tmp_dec;
 			goto out;
+		}
```
cspiel1, line 883: Is this change needed?

modules/mixausrc/mixausrc.c
```diff
@@ -433,9 +431,7 @@ static int encode_update(struct aufilt_enc_st **stp, void **ctx,
 		return ENOMEM;
 
 	enc->st.fadetime = DEFAULT_FADE_TIME;
-	err = str_dup(&enc->cname, stream_cname(audio_strm(au)));
-	if (err)
-		return err;
+	(void)str_dup(&enc->cname, stream_cname(audio_strm(au)));
```
cspiel1: See https://github.com/baresip/baresip/pull/3655/changes#r2822289423

modules/mixausrc/mixausrc.c
maximilianfridrich: The find logic seemed a bit off to me. Before this change, even if there was no `cname` match found, simply the last element of the encoders would be returned. That logic seemed wrong to me, so now if there is no `cname` match, NULL is returned. Seemed more logical to me.

modules/mixausrc/mixausrc.c
maximilianfridrich: The find logic seemed a bit off to me. Before this change, even if there was no `cname` match found, simply the last element of the decoders would be returned. That logic seemed wrong to me, so now if there is no `cname` match, NULL is returned. Seemed more logical to me.

modules/mixausrc/mixausrc.c
cspiel1: Your right. Thanks!

cspiel1:
Looks good to me.

PR #3660, opened 2026-03-02
uag: allow registered ua catchall

sreimers:
A registered user agent should also be possible to used as catchall

### DIFF
--- a/src/uag.c
+++ b/src/uag.c
@@ -946,6 +946,9 @@ struct ua *uag_find_msg(const struct sip_msg *msg)
 			ua_printf(ua, "selected for %r\n", cuser);
 			return ua;
 		}
+
+		if (!uaf && ua_catchall(ua))
+			uaf = ua;
 	}
 
 	/* match for peer-to-peer calls (only un-registered accounts) */


cspiel1:
We observe now the following problem.
- For each local IP address baresip at device A has an un-registered account (`regint=0`).
- Device A has also one registered account at SIP server X.
- X is currently offline and thus the application at the peer B decides to make a peer-to-peer call to A. It uses the IPv6ll of A.
- All accounts have the `catchall=yes` parameter set. I know that this is not ideal.
- At A the wrong UA for X is selected.

```
"Alice" <sip:********@[fe80::be6a:44ff:fe00:85bf]>;regint=0;100rel=yes;catchall=yes
"Alice" <sip:********@X;transport=tls>;auth_user=***;auth_pass="********";regint=3600;catchall=yes
```

Should we apply this patch?
```patch
diff --git a/src/uag.c b/src/uag.c
index 8f7f6e704..92b15ed07 100644
--- a/src/uag.c
+++ b/src/uag.c
@@ -947,7 +947,7 @@ struct ua *uag_find_msg(const struct sip_msg *msg)
                        return ua;
                }
 
-               if (!uaf && ua_catchall(ua))
+               if (!uaf && ua_catchall(ua) && ua_isregistered(ua))
                        uaf = ua;
        }
```

sreimers:
> Should we apply this patch?

Looks good to me

cspiel1:
PR: https://github.com/baresip/baresip/pull/3752

Issue #3659, opened 2026-02-28
Incoming OPTIONS messages for keepalive causes repeated log entries


petercolberg:
My SIP provider sends an OPTIONS message to the client every 15 seconds to keep the connection alive behind NAT, which causes repeated entries at the default log level INFO in the form of
```
USER@DOMAIN: selected for USER
USER@DOMAIN: selected for USER
USER@DOMAIN: selected for USER
...
```
Only when setting the log level to DEBUG does the reason become apparent:
```
ua: incoming OPTIONS message from sip:KEEPALIVE@DOMAIN (1.2.3.4:5060)
call: alloc with params laddr=192.168.1.1, af=AF_INET, use_rtp=0
call: use_video=0
stream: disable audio RTP sender
stream: disable audio RTP receiver
audio: destroyed (started=0)
stream: disable audio RTP sender
stream: disable audio RTP receiver
USER@DOMAIN: selected for USER
```

Since baresip is otherwise perfectly quiet at log level INFO unless there is, e.g., call activity, and log level INFO is generally very useful, would it be reasonable to downgrade the `ua_printf()` calls in `uag_find_msg()` to `debug()` calls? Or should a solution be more complex, i.e., take into account the context in which `uag_find_msg()` is being called?

alfredh:
I think it should be fine to remove that logline or change it to `debug()` ...

Would you be so kind to provide a Pull Request for this change ?

petercolberg:
Thank you for the quick response and to all contributors for a great project.

PR #3661, opened 2026-03-03
uag: log selected account in uag_find_msg() at level DEBUG

petercolberg:
This avoids repeated log entries at level INFO when a SIP registrar frequently sends OPTIONS messages to the user agent to keep alive, e.g., a NAT UDP mapping. Log selected account at level DEBUG instead, which aligns with other log entries triggered upon such a message.

Closes: https://github.com/baresip/baresip/issues/3659
Fixes: 917326002203 ("ua: improve UA selection for incoming calls")

### DIFF
--- a/src/uag.c
+++ b/src/uag.c
@@ -943,7 +943,7 @@ struct ua *uag_find_msg(const struct sip_msg *msg)
 			continue;
 
 		if (0 == pl_strcasecmp(cuser, ua_local_cuser(ua))) {
-			info("%H: selected for %r\n", ua_printf, ua, cuser);
+			debug("%H: selected for %r\n", ua_printf, ua, cuser);
 			return ua;
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/core.h
```diff
@@ -364,8 +364,8 @@ void stream_mnat_connected(struct stream *strm, const struct sa *raddr1,
 
 struct ua;
 
-void         ua_printf(const struct ua *ua, const char *fmt, ...);
-
+void ua_printf(const struct ua *ua, const char *fmt, ...);
+int ua_print(struct re_printf *pf, const struct ua *ua);
```
cspiel1: Since `ua_printf()` is an internal function we simply could replace the implementation and how it is used. Finally there would be only one function:

```c
int ua_printf(struct re_printf *pf, const struct ua *ua);
```

petercolberg: Thank you for the quick review and suggestion.

cspiel1:
Look good for me.

PR #3666, opened 2026-03-05
modules/selfview: added support for screen rotation

juha-h:
I noticed that when video screen is rotated from landscape to portrait or vice versa, selfview pip window was not rotating. This PR tries to fix that.  Tested with my Android baresip app.


alfredh:
This looks fine to me.

Have you tested when the picture is a square, for example 400x400 ?

juha-h:
Just tested with 720x720 that was supported by my device's camera and also with that rotation of selfview worked OK.

PR #3667, opened 2026-03-07
Revert "call: do not go back to RINGING state if already in EARLY"

dddaniel:
Causes to never get into CALL_STATE_RINGING on some SIP providers. Therefore the ringback_aufile is never played.

This reverts commit 5699d42e508ca3a222c4550e42c114dbf08aaba3.

-----

Hi,

since that commit 5699d42e508ca3a222c4550e42c114dbf08aaba3, I noticed all my outgoing SIP calls do not play back the ringback_aufile audio. The PR might not be a final solution for you, but it will shine some light on the issue.

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -2493,8 +2493,7 @@ static void sipsess_progr_handler(const struct sip_msg *msg, void *arg)
 	switch (msg->scode) {
 
 	case 180:
-		if (call_state(call) != CALL_STATE_EARLY)
-			set_state(call, CALL_STATE_RINGING);
+		set_state(call, CALL_STATE_RINGING);
 		break;
 
 	case 183:
@@ -2509,7 +2508,7 @@ static void sipsess_progr_handler(const struct sip_msg *msg, void *arg)
                                    call->peer_uri);
 		mem_deref(call);
 	}
-	else if (call_state(call) != CALL_STATE_EARLY) {
+	else {
 		call_stream_stop(call);
 		call_event_handler(call, CALL_EVENT_RINGING, "%s",
                                    call->peer_uri);


cspiel1:
I'm not sure if a revert is the right option now. @maximilianfridrich and me have to look closer and discuss this on Monday.

juha-h:
I fully support reverting https://github.com/baresip/baresip/commit/5699d42e508ca3a222c4550e42c114dbf08aaba3.  I just made test by calling from baresip test@test.tutpro.com account to foo@test.tutpro.com on another device via my early announce app.  The app responded with 183 to test and played the announcement, then the app forwarded the invite to foo@test.tutpro.com, which responded with 180 ringing.  Then the 180 ringing was received by test, but baresip never started to ring.  Of course it has to start ringing so that the called knows what is going on.

juha-h:
Just to add, I don't care in which call state baresip lib is as long as the app gets  ua event (CALL_RINGING).

maximilianfridrich:
We see the following issue: In BareSIP, we always send a 180 Ringing response immediately and in the early media case a 183 immediately follows. We cannot usually (in UDP case) guarantee the order of these two, which is why early media was broken for us sometimes. For now, we can revert this and maybe find another solution for our issue.

juha-h:
Maximilian Fridrich writes:

> We see the following issue: In BareSIP, we always send a 180 Ringing
> response immediately and in the early media case a 183 immediately
> follows. We cannot usually (in UDP case) guarantee the order of these
> two, which is why early media was broken for us sometimes. For now, we
> can revert this and maybe find another solution for our issue.

Please do and stop using UDP.

maximilianfridrich:
> Please do and stop using UDP.

Agreed, but we can't force our customers to do so and we can't take a transport protocol away from them and break their systems. That would ruffle some feathers.

juha-h:
Thanks, my early announce test now worked fine.

Issue #3645, opened 2026-02-06
Baresip 4.5.0 testsuite fails: mem: mem_deref: magic check failed 0xb5b5b5b5b5b5b5b5 (0x55b706ae65d0)


robert-scheck:
Trying to build baresip 4.5.0 on Fedora 45 (Rawide) succeeds, but testsuite fails on all hardware architectures (i686, x86_64, aarch64, ppc64le, s390x) like this:

```
[ RUN      ] test_cparam_ua_decode (rx main)
test 0: "display name" <sip:user@domain> audio=sendonly video=inactive userdata=mydata
test 1: displayname <sip:user@domain> userdata=mydata
test 2: dn <user> userdata=mydata2 audio=recvonly
test 3: dn <user> userdata=mydata2 video=fail4
test 4: sip:user2@domain2 audio=recvonly video=sendonly
test 5: sip:user3@domain3 audio=fail video=sendonly
test 6: sip:user4@domain4 video=inactive
test 7: sip:user5@domain5 video=fail2
test 8: user6 sendonly
test 9: audio=sendonly
test 10: sip:user4@domain4;transport=tcp video=recvonly
[       OK ]
[ RUN      ] test_contact (rx main)
[       OK ]
[ RUN      ] test_bevent_register (rx main)
[       OK ]
[ RUN      ] test_jbuf (rx main)
[       OK ]
[ RUN      ] test_jbuf_adaptive (rx main)
[       OK ]
[ RUN      ] test_jbuf_video (rx main)
[       OK ]
[ RUN      ] test_jbuf_gnack (rx main)
[       OK ]
[ RUN      ] test_message (rx main)
b@127.0.0.1: account match for b
[ sip:b@127.0.0.1:49989;tra
mem: mem_deref: magic check failed 0xb5b5b5b5b5b5b5b5 (0x55b706ae65d0)
[;m/var/tmp/rpm-tmp.SkxfZ6: line 50:  8461 Trace/breakpoint trap      (core dumped) ./test/selftest -d ../test/data/ -v
```

Full [x86_64 build log](https://kojipkgs.fedoraproject.org//work/tasks/2230/142012230/build.log) and [x86_64 root log](https://kojipkgs.fedoraproject.org//work/tasks/2230/142012230/root.log).

robert-scheck:
This is maybe a duplicate of #3623 - and maybe the following builds help to track the issue down:
 
[...]

PR #3668, opened 2026-03-09
test: check for real IP-address in test_peerconn

alfredh:
The ICE module needs a real IP-address to work. This test is failing on Red Hat systems due to
missing real IP-address.

https://github.com/baresip/baresip/issues/3645


alfredh:
hi @cspiel1 could you please review the network logic in this patch ?

cspiel1:
Looks good.

PR #3688, opened 2026-04-03
stream,rtprecv: harden RTP receiver teardown against races

KORuL:
Add RX teardown guards in stream/rtprecv shutdown paths.

This change cancels pending RX-related timers, marks the receiver as closed
before dereferencing it, and avoids taking a mutex after it has been cleared
during receiver destruction.

The goal is to prevent use-after-free / teardown races during stream shutdown
without changing normal RTP/RTCP/media behavior.


sreimers:
Please avoid AI PRs with such low quality and ignoring our coding standards.

PR #3689, opened 2026-04-03
net: guard net_laddr_apply() against NULL local address entries

KORuL:
This change adds a defensive guard in net_laddr_apply().

The current implementation passes laddr->ifname to the callback without
validating the local-address entry first. If a malformed entry reaches
the local address list, this can lead to a NULL dereference during
address iteration.

With this change, entries with a NULL element or NULL ifname are skipped,
while the behavior for valid entries remains unchanged.

This is a small robustness fix and does not change the normal code path.


sreimers:
> With this change, entries with a NULL element or NULL ifname are skipped,
> while the behavior for valid entries remains unchanged.

This should never happen. Otherwise `laddrs` list is already broken.

PR #3690, opened 2026-04-03
Android: serialize AAudio stream open/close and harden media teardown…

KORuL:
Android: serialize AAudio stream open/close and harden media teardown paths

MR description

This MR addresses stability issues seen during Android AAudio stream teardown and restart, especially when player and recorder streams are closed and reopened from different threads.

Summary

The changes in this MR do two things:

Serialize critical AAudio stream lifecycle operations to reduce races between concurrent stream teardown and recreation.
Harden a few media teardown paths in the generic audio/video stack to avoid invalid access during shutdown.
AAudio changes
add a global recursive AAudio lock in aaudio/utils.c
serialize AAudioStreamBuilder_openStream()
serialize AAudioStream_close()
keep AAudioStream_requestStop() inside the serialized close path
restore stream state checks before closing
keep the existing player/recorder restart handling introduced in the previous changes

This is intended to reduce vendor/device-specific failures caused by overlapping stream close/open activity across threads.

Media teardown hardening
src/audio.c
make stop_tx() tolerate partially destroyed state
avoid returning early before TX thread shutdown and source cleanup
src/stream.c
add defensive checks in stream_enable_tx() for incomplete stream state
src/video.c
protect video_stop_display() with vrx.lock during display teardown

### DIFF
--- a/modules/aaudio/aaudio.h
+++ b/modules/aaudio/aaudio.h
@@ -7,6 +7,8 @@
 #include <aaudio/AAudio.h>
 
 void aaudio_close_stream(AAudioStream *stream);
+void aaudio_lock(void);
+void aaudio_unlock(void);
 
 int aaudio_player_alloc(struct auplay_st **stp, const struct auplay *ap,
 			struct auplay_prm *prm, const char *device,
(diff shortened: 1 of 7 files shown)


sreimers:
Please avoid AI PRs with such low quality and ignoring our coding standards.

PR #3696, opened 2026-04-07
Aaudio stream lifecycle hardening

KORuL:
Thanks for the feedback. I reworked the MR to better match the project coding standards and reduced the scope to the essential changes.

What I updated:

* cleaned up the implementation and removed unnecessary defensive changes
* aligned naming/style with the surrounding code
* kept the Android AAudio synchronization changes focused on the stream lifecycle race
* rechecked teardown paths to avoid unrelated changes and premature returns
* updated the MR description to better reflect the actual intent and scope

I believe the result is now much cleaner and easier to review. Please take another look when you have a moment.

### DIFF
--- a/modules/aaudio/player.c
+++ b/modules/aaudio/player.c
@@ -8,30 +8,161 @@
 #include <re.h>
 #include <rem.h>
 #include <baresip.h>
+#include <stdatomic.h>
+#include <string.h>
 
 #include "aaudio.h"
 
 
-struct auplay_st {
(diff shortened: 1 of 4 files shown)


juha-h:
I have not had any issues what so ever with current version of aaudio module and I have been using it on many different devices and Android versions starting from 10.  So I'm really afraid that if this large  PR is merged, something breaks.  It would be safer to create a new module for those who need these changes.

PR #3697, opened 2026-04-08
contact: avoid duplicate contacts

cspiel1:


src/contact.c
```diff
@@ -101,6 +101,7 @@ int contact_add(struct contacts *contacts,
 	if (err)
 		goto out;
 
+	contact_remove(contacts, contact_find(contacts, c->uri));
```
alfredh, line 104: it looks like the contact struct is stored in two containers, list and hash.
would it be enough with one container, or perhaps use `struct odict` ?

In case of duplicates, which entry should you keep, the old one or the new one ?

alfredh:
could you please rebase to main ?

src/contact.c
cspiel1: I'll check if both `list` and `hash` are needed.

The "duplicate" should replace the existing contact. More concretely if the URI part matches it should be replaced.

cspiel1: Yes, both are needed. The `hash` table makes the find fast for large contact sets. And the `list` is needed for the API function `contact_list()`. We could get rid of the `list` by changing this function to generate the returned list from the hash buckets. But it seems to be not worth.

PR #3699, opened 2026-04-08
contact: intelligent matching for incoming calls

cspiel1:
- **contact: intelligent matching for incoming calls**
- **menu: intelligent contact matching**

    
    Matching a contact with the From header of an incoming INVITE nearly always fails
    for registered User-Agents. This commit adds a new API function
    `contact_find_call()`.
    
    - first try the complete From header (like `contact_find()` does),
    - then try the AOR of the UA with the user-ID of the From header (for
      registered accountts),
    - as last fallback only for P2P calls try the From header without user-ID

### DIFF
--- a/src/contact.c
+++ b/src/contact.c
@@ -511,7 +511,7 @@ struct contact *contact_find_call(const struct contacts *contacts,
 
 	/* replace the user */
 	aor.user = from_addr.uri.user;
-	char *uri;
+	char *uri = NULL;
 	err = re_sdprintf(&uri, "%H", uri_encode, &aor);
 	if (err)
 		return NULL;
--- a/src/contact.c
+++ b/src/contact.c
@@ -129,6 +129,7 @@ int contact_add(struct contacts *contacts,
 	c->status = PRESENCE_UNKNOWN;
 
 	list_append(&contacts->cl, &c->le, c);
+	debug("contact: adding <%r>\n", &c->addr.auri);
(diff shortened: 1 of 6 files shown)


src/contact.c
```diff
@@ -440,6 +441,101 @@ struct contact *contact_find(const struct contacts *contacts, const char *uri)
...
+	int err;
+
+	pl_set_str(&pl, call_peeruri(call));
+	err = sip_addr_decode(&from_addr, &pl);
+	if (err)
+		return NULL;
+
+	/* replace the user */
+	aor.user = from_addr.uri.user;
+	char *uri;
```
alfredh: perhaps init `uri` to NULL for increased robustness

alfredh:
when adding new API functions, it would be nice if there is a matching new testcase for it.

please consider also making a test, either in this PR or a follow-up PR ...

cspiel1:
> when adding new API functions, it would be nice if there is a matching new testcase for it.
> 
> please consider also making a test, either in this PR or a follow-up PR ...

The PR adds a dependency from contact to call. And `test/call.c` contains a lot of call logic. I try to find a way to make a unit test for `contact_find_call()`.

cspiel1:
Added a minimal test for the new function. I'll add a commit to run through more code parts of `contact_find_call()`.

cspiel1:
The test runs now through the whole `contact_find_call()`.

PR #3702, opened 2026-04-14
call: support rejecting re-INVITE

cspiel1:
Now it is possible to reject a re-INVITE in the application in
BEVENT_CALL_REMOTE_SDP or BEVENT_CALL_LOCAL_SDP

### DIFF
--- a/src/call.c
+++ b/src/call.c
@@ -313,8 +313,10 @@ static int update_media(struct call *call)
 {
 	debug("call: update media\n");
 
-	bevent_call_emit(BEVENT_CALL_REMOTE_SDP, call,
-			 call->got_offer ? "offer" : "answer");
+	int err = bevent_call_emit(BEVENT_CALL_REMOTE_SDP, call,
+				   call->got_offer ? "offer" : "answer");
+	if (err)
+		return err;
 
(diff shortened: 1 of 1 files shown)


cspiel1:
TODO: Can the application distinguish between re-INVITE and 200 OK (call answer)?

juha-h:
`TODO: Can the application distinguish between re-INVITE and 200 OK (call answer)?

Yes, in
```
static void event_handler(enum bevent_ev ev, struct bevent *event, void *arg) {
   const char *prm  = bevent_get_text(event);
```
when `ev == BEVENT_CALL_REMOTE_SDP`, `prm` is either "offer" or "answer".

cspiel1:
> TODO: Can the application distinguish between re-INVITE and 200 OK (call answer)?

This is the wrong question. This PR has not directly something to do with re-INVITE. It allows a reaction on SDP offer/answer. If the application sets an error on the bevent, it depends if called from
- connect.c: terminates call,
- modify.c: suppress the SIP ACK,
- listen.c:
  - from PRACK, ACK ignores the error
  - re-INVITE, UPDATE rejects with a SIP 488

PR #3731, opened 2026-05-13
directfb: remove deprecated module

alfredh:
This is a proposal to remove a deprecated video display module. For video display the SDL module can be used instead.

If someone wants to keep this module, it can be
moved to baresip-apps.


cspiel1:
Should we merge this? It could be retrieved from git log later if needed and then moved to baresip-apps.

alfredh:
There are no objections so I think it should be fine to merge it ..

PR #3733, opened 2026-05-14
ua: emit event for failed incoming INVITE

SnowMaster12:
## Summary

This PR follows the discussion here: https://github.com/baresip/baresip/discussions/3727

The pr adds a new Baresip event:
```c
BEVENT_INCOMING_CALL_FAILED
```
The event is emitted when an incoming `INVITE` is rejected before a normal incoming call lifecycle is available to the application layer.

### Added a new bevent
```c
BEVENT_INCOMING_CALL_FAILED
```
and the corresponding string mapping:
```c
"INCOMING_CALL_FAILED"
```

### Emit the event for early incoming INVITE rejections
The new event is emitted in `sipsess_conn_handler()` before sending the SIP final response for these cases:

- `606 Not Acceptable`
- `404 Not Found`
- `403 Forbidden`
- `486 Max Calls`
- `420 Bad Extension`
- `421 Extension required`

These paths reject the incoming `INVITE` before the call is accepted and before a normal `struct call` lifecycle is available

## Fixed the 421 path
The `421 Extension required` path now returns immediately after sending the final response.
Previously, after sending `421`, the function could continue toward call acceptance.

Since `421` is a final response, the `INVITE` MUST NOT continue into `ua_accept()` after that response has been sent.

## Handle the existing 488 path correctly
`ua_call_alloc()` already sends:
```c
488 Not Acceptable Here
```
when the address family is not supported.

Before this change, that path returned `EINVAL`, so `ua_accept` treated it as a generic failure and could continue to the fallback `500 Call Error` path.

That could result in two final responses being attempted for the same incoming `INVITE`:
```c
488 Not Acceptable Here
500 Call Error
```
I've checked the lower SIP transaction path, `sip_treply(NULL,...)` doesn't provide a high-level guard, in fact, the second call can still go through the reply path and attempt to send another response.

To avoid this, the `488` path now returns `EAFNOSUPPORT`.

`ua_accept()` checks for `EAFNOSUPPORT` and exits without sending the generic `500 Call Error`.

The same guard was also added in `ua_handle_options()` to avoid falling back to `500 Call Error` after `ua_call_alloc()` has already sent `488`


src/ua.c
```diff
@@ -831,9 +848,15 @@ int ua_accept(struct ua *ua, const struct sip_msg *msg)
...
 
+ out:
+	mem_deref(to_uri);
+	return err;
+
  error:
 	mem_deref(call);
 	mem_deref(to_uri);
+	(void)bevent_sip_msg_emit(BEVENT_INCOMING_CALL_FAILED,
+			msg, "500 Call Error");
```
sreimers: Since call object is available `BEVENT_CALL_CLOSED` should be used here instead.

src/bevent.c
```diff
@@ -767,6 +767,7 @@ const char *bevent_str(enum bevent_ev ev)
 	case BEVENT_MODULE:               return "MODULE";
 	case BEVENT_END_OF_FILE:          return "END_OF_FILE";
 	case BEVENT_CUSTOM:               return "CUSTOM";
+	case BEVENT_INCOMING_CALL_FAILED: return "INCOMING_CALL_FAILED";
```
sreimers: Maybe `BEVENT_SIPSESS_FAILED` is a better name for this.

src/ua.c
```diff
@@ -69,7 +69,6 @@ static void ua_destructor(void *arg)
 	list_flush(&ua->hdr_filter);
 }
 
-
```
sreimers, line 72: Unrelated change

sreimers:
> The `421 Extension required` path now returns immediately after sending the final response. Previously, after sending `421`, the function could continue toward call acceptance.
> 
> Since `421` is a final response, the `INVITE` MUST NOT continue into `ua_accept()` after that response has been sent.

Please create a separate PR for this.

sreimers:
> Handle the existing 488 path correctly

This should be a separate PR too.

SnowMaster12:
I updated the PR to keep it focused only on the new SIP session failure event.

I removed the behavioral changes related to the 421 path and the existing 488 handling. 
I also removed the ua_accept() change, because that error path can be reached both before and after a call object is available. 
It probably needs a separate, more precise change so that failures with an existing call object can use the existing call event semantics, while pre-call failures can still be reported as SIP_Session failures.

This PR now only adds BEVENT_SIPSESS_FAILED and emits it in the early sipsess_conn_handler() rejection paths where no call object exists yet.

I will open two separate PRs for the other issues:
- making the 421 Extension required path return immediately after sending the final response
- fixing the existing 488 path to avoid falling back to 500 after a final response has already been sent

PR #3736, opened 2026-05-19
ua: stop processing INVITE after 421 response

SnowMaster12:
## Summary

This PR fixes the `421 Extension required` path in `sipsess_conn_handler()`.

When `rel100_mode` is set to `REL100_REQUIRED`, but the incoming `INVITE` does not advertise `100rel` through either `Supported` or `Require`, Baresip sends:

```c
sip_treplyf(..., 421, "Extension required", ...);
```
However, after sending the final `421` response, the function could continue and reach the normal incoming call handling path.
Since `421` should be a final response, the rejected INVITE must not continue toward ua_accept() after the response has been sent.

## What is changed
The `421 Extension required`path now returns immediately after sending the final response.

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -764,6 +764,7 @@ void sipsess_conn_handler(const struct sip_msg *msg, void *arg)
 				  421, "Extension required",
 				  "Require: 100rel\r\n"
 				  "Content-Length: 0\r\n\r\n");
+		return;
 	}
 
 	if (config->call.accept)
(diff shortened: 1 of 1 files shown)


alfredh:
Thank you, this looks very good.

A new testcase for this usecase is also welcome, either in this PR or a separate PR.

SnowMaster12:
@alfredh i'll make a new pr for the usecase

PR #3738, opened 2026-05-23
test: add 100rel required reject testcase

SnowMaster12:
Added a regression test for the 100rel=required reject path, requested in #3736.

The test configures only the callee with 100rel=required, sends an INVITE from a caller without 100rel support, and verifies that the caller receives 421 Extension required while no call is allocated on the callee side.

Tested with:
./build/test/selftest test_call_100rel_required_reject
./build/test/selftest

The new test passes with both rx main and rx thread.

### DIFF
--- a/test/call.c
+++ b/test/call.c
@@ -1074,6 +1074,57 @@ int test_call_100rel_video(void)
 }
 
 
+int test_call_100rel_required_reject(void)
+{
+	struct fixture fix, *f = &fix;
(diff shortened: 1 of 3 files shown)


src/ua.c
alfredh, line 767: I think this change is already in `main` ?

Could you please rebase to main ?

SnowMaster12: You're right, sorry.
Rebased.

PR #3750, opened 2026-06-10
Adding /dtmf <string> command to the static_menu (can then also be used by MQTT for sending DTMF signals in a call)

ehasting:
by using /dtmf <dtmf string> or over mqtt '{"command": "dtmf", "params": "*#535", "token": "123"}'
it will in-call trigger dtmf sequence.  If ua is not in call it will result in error.


cspiel1:
See `dynamic_menu.c` command "sndcode"!

Edit: This is a complete duplicate. Can we close it?

alfredh:
Many thanks for your PR... The new command is already implemented :)

PR #3753, opened 2026-06-17
release v4.9.0

cspiel1:


cspiel1:
`enum bevent_ev` changed --> ABI version increased

PR #3762, opened 2026-06-30
call: add callback for incoming SIP INFO

SnowMaster12:
## Summary

This PR adds a call-level callback for incoming SIP INFO requests.

The new callback allows applications embedding baresip to inspect and handle incoming SIP INFO messages directly from the call layer.

The callback exposes:

- the related `struct call`
- the SIP INFO `Content-Type`
- the raw body payload
- the body length
- the original `struct sip_msg`
- custom SIP headers through the original SIP message reference

## Changes

Added a new SIP INFO handler type:
```C
typedef bool (call_sip_info_h)(struct call *call,
			       const char *content_type,
			       const uint8_t *body,
			       size_t body_len,
			       const struct sip_msg *msg,
			       void *arg);
```

Added a public registration API:
```C
void call_set_sip_info_handler(call_sip_info_h *handler, void *arg);
```

Added internal SIP INFO event emission:
```c
call_emit_sip_info(call, msg);
```

Added a new internal helper:
```c
static void call_emit_sip_info(struct call *call, const struct sip_msg *msg);
```

The helper extracts the SIP INFO `Content-Type`, including optional content-type parameters, maps the SIP message body from the SIP message buffer, and invokes the registered application handler.

The callback return value is propagated as `app_handled`, allowing the call code to know whether the SIP INFO request was handled by the application layer.

## Motivation
Some applications use SIP INFO for custom in-call signaling, not only for standard DTMF-related use cases.

For example, applications may need to receive custom commands, metadata, or vendor-specific headers carried inside SIP INFO requests.

Before this change, there was no clean call-level API for applications or SDK wrappers to receive incoming SIP INFO messages while preserving access to the raw body and original SIP headers.

### DIFF
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -502,12 +502,12 @@ struct config *conf_config(void);
 
 /** Sip Info Configuration*/
 typedef bool (call_sip_info_h)(
-    struct call *call,
-    const char *content_type,
-    const uint8_t *body,
-    size_t body_len,
-    const struct sip_msg *msg,
-    void *arg
+	struct call *call,
+	const char *content_type,
+	const uint8_t *body,
+	size_t body_len,
+	const struct sip_msg *msg,
+	void *arg
 );
 void call_set_sip_info_handler(call_sip_info_h *handler, void *arg);
 
@@ -1347,8 +1347,9 @@ typedef int(viddec_update_h)(struct viddec_state **vdsp,
 			     const struct vidcodec *vc, const char *fmtp,
 			     const struct video *vid);
 
-typedef int (viddec_decode_h)(struct viddec_state *vds, struct vidframe *frame,
-                              struct viddec_packet *pkt);
+typedef int (viddec_decode_h)(struct viddec_state *vds,
+			      struct vidframe *frame,
+			      struct viddec_packet *pkt);
 
 struct vidcodec {
 	struct le le;
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -509,7 +509,11 @@ typedef bool (call_sip_info_h)(
 	const struct sip_msg *msg,
 	void *arg
 );
-void call_set_sip_info_handler(call_sip_info_h *handler, void *arg);
+void call_set_sip_info_handler(
+	struct call *call,
+	call_sip_info_h *handler,
+	void *arg
+);
 
 
 /*
@@ -1347,9 +1351,8 @@ typedef int(viddec_update_h)(struct viddec_state **vdsp,
 			     const struct vidcodec *vc, const char *fmtp,
 			     const struct video *vid);
 
-typedef int (viddec_decode_h)(struct viddec_state *vds,
-			      struct vidframe *frame,
-			      struct viddec_packet *pkt);
+typedef int (viddec_decode_h)(struct viddec_state *vds, struct vidframe *frame,
+                              struct viddec_packet *pkt);
 
 struct vidcodec {
 	struct le le;
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -509,6 +509,7 @@ typedef bool (call_sip_info_h)(
 	const struct sip_msg *msg,
 	void *arg
 );
+
 void call_set_sip_info_handler(
 	struct call *call,
 	call_sip_info_h *handler,
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -501,7 +501,7 @@ int config_write_template(const char *file, const struct config *cfg);
 struct config *conf_config(void);
 
 /** Sip Info Configuration*/
-typedef bool (call_sip_info_h)(
+typedef void (call_sip_info_h)(
 	struct call *call,
 	const char *content_type,
 	const uint8_t *body,
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -501,7 +501,7 @@ int config_write_template(const char *file, const struct config *cfg);
 struct config *conf_config(void);
 
 /** Sip Info Configuration*/
-typedef void (call_sip_info_h)(
+typedef bool (call_sip_info_h)(
 	struct call *call,
 	const char *content_type,
 	const uint8_t *body,
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -500,6 +500,17 @@ int config_print(struct re_printf *pf, const struct config *cfg);
 int config_write_template(const char *file, const struct config *cfg);
 struct config *conf_config(void);
 
+/** Sip Info Configuration*/
+typedef bool (call_sip_info_h)(
+    struct call *call,
+    const char *content_type,
+    const uint8_t *body,
+    size_t body_len,
+    const struct sip_msg *msg,
+    void *arg
+);
+void call_set_sip_info_handler(call_sip_info_h *handler, void *arg);
+
 
 /*
  * Contact
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -217,7 +217,14 @@ typedef void (call_event_h)(struct call *call, enum call_event ev,
 typedef void (call_dtmf_h)(struct call *call, char key, void *arg);
 typedef bool (call_match_h)(const struct call *call, void *arg);
 typedef void (call_list_h)(struct call *call, void *arg);
-
+typedef bool (call_sip_info_h)(
+	struct call *call,
+	const char *content_type,
+	const uint8_t *body,
+	size_t body_len,
+	const struct sip_msg *msg,
+	void *arg
+);
 
 int  call_connect(struct call *call, const struct pl *paddr);
 int  call_answer(struct call *call, uint16_t scode, enum vidmode vmode);
@@ -248,6 +255,9 @@ int  call_notify_sipfrag(struct call *call, uint16_t scode,
 			 const char *reason, ...);
 void call_set_handlers(struct call *call, call_event_h *eh,
 		       call_dtmf_h *dtmfh, void *arg);
+void call_set_sip_info_handler(struct call *call,
+				call_sip_info_h *handler,
+				void *arg);
 struct account *call_account(const struct call *call);
 uint16_t      call_scode(const struct call *call);
 enum call_state call_state(const struct call *call);
@@ -499,22 +509,6 @@ int config_parse_conf(struct config *cfg, const struct conf *conf);
 int config_print(struct re_printf *pf, const struct config *cfg);
 int config_write_template(const char *file, const struct config *cfg);
 struct config *conf_config(void);
-
-/** Sip Info Configuration*/
-typedef bool (call_sip_info_h)(
-	struct call *call,
-	const char *content_type,
-	const uint8_t *body,
-	size_t body_len,
-	const struct sip_msg *msg,
-	void *arg
-);
-
-void call_set_sip_info_handler(
-	struct call *call,
-	call_sip_info_h *handler,
-	void *arg
-);
 
 
 /*
--- a/src/call.c
+++ b/src/call.c
@@ -85,13 +85,13 @@ struct call {
 	bool use_video;
 	bool use_rtp;
 	struct pl *user_data;      /**< User data related to the call       */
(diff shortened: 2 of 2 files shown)


include/baresip.h
```diff
@@ -1336,8 +1347,9 @@ typedef int(viddec_update_h)(struct viddec_state **vdsp,
 			     const struct vidcodec *vc, const char *fmtp,
 			     const struct video *vid);
 
-typedef int (viddec_decode_h)(struct viddec_state *vds, struct vidframe *frame,
-                              struct viddec_packet *pkt);
+typedef int (viddec_decode_h)(struct viddec_state *vds,
+			      struct vidframe *frame,
+			      struct viddec_packet *pkt);
```
sreimers: Unrelated change

src/call.c
```diff
@@ -90,7 +90,8 @@ struct call {
 
 static int send_invite(struct call *call);
 static int send_dtmf_info(struct call *call, char key);
-
+static call_sip_info_h *sip_info_handler = NULL;
+static void *sip_info_arg = NULL;
```
sreimers: Better to bind these values to the `call` object itself.

src/call.c
```diff
@@ -2181,7 +2230,7 @@ static void xfer_cleanup(struct call *call, char *reason)
 	if (call->xcall->state == CALL_STATE_TRANSFER) {
 		set_state(call->xcall, CALL_STATE_ESTABLISHED);
 		call_event_handler(call->xcall, CALL_EVENT_TRANSFER_FAILED,
-                                   "%s", reason);
+						"%s", reason);
```
sreimers: More unrelated format changes

src/call.c
```diff
@@ -975,10 +975,14 @@ int call_alloc(struct call **callp, const struct config *cfg, struct list *lst,
 	return err;
 }
```
cspiel1, line 978: two empty lines between functions

include/baresip.h
```diff
@@ -509,7 +509,11 @@ typedef bool (call_sip_info_h)(
 	const struct sip_msg *msg,
 	void *arg
 );
```
cspiel1: insert an empty line here

src/call.c
```diff
@@ -2500,7 +2553,7 @@ int call_accept(struct call *call, struct sipsess_sock *sess_sock,
 	call->estvdir = stream_ldir(video_strm(call_video(call)));
 	if (!call->acc->mnat)
 		call_event_handler(call, CALL_EVENT_INCOMING, "%s",
-                                   call->peer_uri);
+						call->peer_uri);
```
cspiel1: unrelated change and not aligned now

src/call.c
```diff
@@ -2577,13 +2630,13 @@ static void sipsess_progr_handler(const struct sip_msg *msg, void *arg)
 	if (media) {
 		mem_ref(call);
 		call_event_handler(call, CALL_EVENT_PROGRESS, "%s",
-                                   call->peer_uri);
+						call->peer_uri);
```
cspiel1: unrelated change

src/call.c
```diff
@@ -2577,13 +2630,13 @@ static void sipsess_progr_handler(const struct sip_msg *msg, void *arg)
...
 		mem_deref(call);
 	}
 	else {
 		call_stream_stop(call);
 		call_event_handler(call, CALL_EVENT_RINGING, "%s",
-                                   call->peer_uri);
+						call->peer_uri);
```
cspiel1: unrelated change

src/call.c
```diff
@@ -2010,6 +2020,48 @@ static uint32_t randwait(uint32_t minwait, uint32_t maxwait)
...
 }
 
+static void call_emit_sip_info(struct call *call, const struct sip_msg *msg)
+{
+	struct pl body;
+	char content_type[256] = "";
+
+	if (!call || !msg || !call->sip_infoh) {
+		return;
+	}
```
cspiel1: brackets can be omitted here

src/call.c
```diff
@@ -2010,6 +2020,48 @@ static uint32_t randwait(uint32_t minwait, uint32_t maxwait)
...
+	pl_set_mbuf(&body, msg->mb);
+
+	call->sip_infoh(
+		call,
+		content_type[0] ? content_type : NULL,
+		(const uint8_t *)body.p,
+		body.l,
+		msg,
+		call->sip_info_arg
+	);
```
cspiel1, line 2056: maybe
```c
	call->sip_infoh(call,
			content_type[0] ? content_type : NULL,
			(const uint8_t *)body.p, body.l,
			msg,
			call->sip_info_arg);
```

Align lines 2-4 under the first parameter `call`! Use tab width 8!

src/call.c
```diff
@@ -2010,6 +2020,48 @@ static uint32_t randwait(uint32_t minwait, uint32_t maxwait)
...
+		}
+		else {
+			re_snprintf(
+				content_type,
+				sizeof(content_type),
+				"%r/%r",
+				&msg->ctyp.type,
+				&msg->ctyp.subtype
+			);
+		}
```
cspiel1: Did you try?

```c
		re_snprintf(content_type, sizeof(content_type),
			    "%r/%r%s%r",
			    &msg->ctyp.type, &msg->ctyp.subtype,
			    pl_isset(&msg->ctyp.params) ? ";" : "",
			    &msg->ctyp.params);
```

cspiel1: It should also work if there are no params.

src/call.c
SnowMaster12: done

cspiel1:
looks good now

include/baresip.h
```diff
@@ -500,6 +500,22 @@ int config_print(struct re_printf *pf, const struct config *cfg);
 int config_write_template(const char *file, const struct config *cfg);
 struct config *conf_config(void);
 
+/** Sip Info Configuration*/
+typedef bool (call_sip_info_h)(
```
alfredh: the handler type returns `bool` but the called does not check the return value.
is this intentional ?

Either change it to `void` or check the return value ?

chucongqing:
there is a situation that we receive  INFO requests with Content-Type: application/media_control+xml from the peer, e.g.:

   ```xml
     <media_control>
       <vc_primitive>
         <to_encoder>
           <picture_fast_update/>
         </to_encoder>
         <stream_id>2</stream_id>
       </vc_primitive>
     </media_control>
   ```

   This is a standard FIR (Full Intra Request) used to request an immediate video keyframe.

   With the current void return type, our app can parse the body, but baresip still replies 488 Not Acceptable Here afterwards because
   application/media_control+xml is not handled in sipsess_info_handler.

   Could the callback return bool instead, so the app can signal that it has handled the INFO and baresip should skip the default
   response (and reply 200 OK)?

SnowMaster12:
> there is a situation that we receive INFO requests with Content-Type: application/media_control+xml from the peer, e.g.:
> 
> ```
>   <media_control>
>     <vc_primitive>
>       <to_encoder>
>         <picture_fast_update/>
>       </to_encoder>
>       <stream_id>2</stream_id>
>     </vc_primitive>
>   </media_control>
> ```
> 
> This is a standard FIR (Full Intra Request) used to request an immediate video keyframe.
> 
> With the current void return type, our app can parse the body, but baresip still replies 488 Not Acceptable Here afterwards because application/media_control+xml is not handled in sipsess_info_handler.
> 
> Could the callback return bool instead, so the app can signal that it has handled the INFO and baresip should skip the default response (and reply 200 OK)?

Your observation is correct.

However, I would prefer to keep baresip’s native SIP INFO handling separate from the application level handling.

The intended flow should be:

- baresip first handles the Content-Types it supports natively, such as `application/dtmf-relay`
- INFO requests with unsupported or application-specific Content-Types are then forwarded to the application callback
- if the callback returns `true`, baresip considers the request handled and replies with 200 OK;
- if the callback returns `false`, baresip keeps the current fallback behavior and replies with 488 Not Acceptable Here

With this, we should be able to support any application specific sip info request without changing baresip's native behavior.
The application will be responsible for handling any content-type that is not supported by baresip.

alfredh:
it would be quite flexible if the new info handler can get a chance to inspect the incoming
INFO message first.

The incoming INFO message is first passed to the new handler. If the new handler understands
the message it can return true or reply 200 ok.

Otherwise the INFO message is handled by the current code, i.e. `dtmf-relay` etc.

SnowMaster12:
> it would be quite flexible if the new info handler can get a chance to inspect the incoming INFO message first.
> 
> The incoming INFO message is first passed to the new handler. If the new handler understands the message it can return true or reply 200 ok.
> 
> Otherwise the INFO message is handled by the current code, i.e. `dtmf-relay` etc.

At the moment, with the way the new application level handler is implemented, it forwards any incoming SIP INFO message to the application and reports it as handled.
Therefore, if the handler is invoked before the existing native processing, it would also intercept and consume `application/dtmf-relay `requests.

In that case, baresip would reply with `200 OK` before reaching the current DTMF branch, so this callback would no longer be invoked: 
```C
call->dtmfh(call, s, call->arg);
```
Resulting in breaking backward compatibility.

For this reason, i currently invoke the generic sip info only after the content-types already supported natively by baresip have been processed.

include/baresip.h
```diff
@@ -500,6 +500,22 @@ int config_print(struct re_printf *pf, const struct config *cfg);
...
+	const struct sip_msg *msg,
+	void *arg
+);
+
+void call_set_sip_info_handler(
+	struct call *call,
+	call_sip_info_h *handler,
+	void *arg
+);
+
```
alfredh: please move this block with the new API to the existing `struct call` block.

src/call.c
```diff
@@ -85,13 +85,14 @@ struct call {
...
 	struct pl *user_data;      /**< User data related to the call       */
+	call_sip_info_h *sip_infoh;/**< SIP INFO handler for this call      */
+	void *sip_info_arg;        /**< SIP INFO handler argument           */
 };
 
 
 static int send_invite(struct call *call);
 static int send_dtmf_info(struct call *call, char key);
 
-
```
alfredh, line 94: please revert unrelated changes (this and also check rest of patch)

src/call.c
```diff
@@ -2113,7 +2158,7 @@ static void sipsess_info_handler(struct sip *sip, const struct sip_msg *msg,
 			}
 		}
 	}
-	else if (!mbuf_get_left(msg->mb)) {
+	else if (!mbuf_get_left(msg->mb) || call_emit_sip_info(call, msg)) {
```
alfredh: Incoming INFO messages with no body will not be forwarded to the new handler.
Is this on purpose ?

alfredh:
hello @SnowMaster12 

are you planning to address the review comments in this PR ?

src/call.c
SnowMaster12: I reversed the condition so that the handler is now invoked first
```C
else if (call_emit_sip_info(call, msg) || !mbuf_get_left(msg->mb)) {
```

Now requests with an empty body will be inspected by the new handler

src/call.c
```diff
@@ -236,7 +238,7 @@ static void mnat_handler(int err, uint16_t scode, const char *reason,
 
 	case CALL_STATE_INCOMING:
 		call_event_handler(call, CALL_EVENT_INCOMING, "%s",
-                                   call->peer_uri);
+				   call->peer_uri);
```
alfredh: unrelated change ..

src/call.c
```diff
@@ -1963,7 +1976,7 @@ static int sipsess_answer_handler(const struct sip_msg *msg, void *arg)
 	    msg->scode >= 200 && msg->scode < 300 &&
 	    call_state(call) != CALL_STATE_ESTABLISHED)
 		call_event_handler(call, CALL_EVENT_ANSWERED, "%s",
-                                   call->peer_uri);
+				   call->peer_uri);
```
alfredh: unrelated change ..

alfredh:
this PR is almost ready to merge ...

just 2 small whitespace issues left to fix ...

PR #3764, opened 2026-07-06
rtprecv: fix use after free on ENODEV errors

cspiel1:
There was a use after free in the decode_frames() loop if CALL_CLOSED was emitted from
video_error_handler().

The wanted behavior is to hangup the call if there is no video display or in any other case where a CALL_CLOSED event is emitted. After emitting CALL_CLOSED it is not allowed to access any data that belongs to the call object.

This PR adds an early exit to the `decode_frames()` loop in `rtprecv.c`

```
Thread 1 "baresip" received signal SIGSEGV, Segmentation fault.
0x00007ffff7f7f528 in jbuf_get (jb=0xb5b5b5b5b5b5b5b5, hdr=0x7fffffffd788, mem=0x7fffffffd780) at /home/cspiel/src/baresip/src/jbuf.c:683
683		mtx_lock(jb->lock);
(gdb) bt
#0  0x00007ffff7f7f528 in jbuf_get (jb=0xb5b5b5b5b5b5b5b5, hdr=0x7fffffffd788, mem=0x7fffffffd780) at /home/cspiel/src/baresip/src/jbuf.c:683
#1  0x00007ffff7f8aa3b in decode_frames (rx=0x555555710720) at /home/cspiel/src/baresip/src/rtprecv.c:372
#2  0x00007ffff7f893dd in decode_tmr (arg=0x555555710720) at /home/cspiel/src/baresip/src/rtprecv.c:330
#3  0x00007ffff7ed6b88 in tmr_poll (tmrl=0x55555555b8e0) at /home/cspiel/src/re/src/tmr/tmr.c:166
#4  0x00007ffff7eb4961 in re_main (signalh=0x555555556c60 <signal_handler>) at /home/cspiel/src/re/src/main/main.c:1085
#5  0x0000555555556b16 in main (argc=1, argv=0x7fffffffdbd8) at /home/cspiel/src/baresip/src/main.c:329
```

### DIFF
--- a/src/video.c
+++ b/src/video.c
@@ -885,7 +885,7 @@ static int stream_pt_handler(uint8_t pt, struct mbuf *mb, void *arg)
 
 
 /* Handle incoming stream data from the network */
-static void stream_recv_handler(const struct rtp_header *hdr,
+static int stream_recv_handler(const struct rtp_header *hdr,
 				struct rtpext *extv, size_t extc,
 				struct mbuf *mb, unsigned lostc,
 				bool new_source,
@@ -902,7 +902,7 @@ static void stream_recv_handler(const struct rtp_header *hdr,
 	if (lostc)
 		request_picture_update(&v->vrx);
 
-	(void)video_stream_decode(&v->vrx, hdr, mb);
+	return video_stream_decode(&v->vrx, hdr, mb);
 }
 
 
--- a/src/audio.c
+++ b/src/audio.c
@@ -622,7 +622,7 @@ static int stream_pt_handler(uint8_t pt, struct mbuf *mb, void *arg)
 /**
  * Stream receive handler for audio is called from RX thread if enabled
  */
-static void stream_recv_handler(const struct rtp_header *hdr,
+static int stream_recv_handler(const struct rtp_header *hdr,
 				struct rtpext *extv, size_t extc,
 				struct mbuf *mb, unsigned lostc,
 				bool new_source,
@@ -633,12 +633,12 @@ static void stream_recv_handler(const struct rtp_header *hdr,
 	MAGIC_CHECK(a);
 
 	if (!a->aur)
-		return;
+		return EINVAL;
 
 	if (new_source)
 		aurecv_reset(a->aur);
 
-	aurecv_receive(a->aur, hdr, extv, extc, mb, lostc);
+	return aurecv_receive(a->aur, hdr, extv, extc, mb, lostc);
 }
 
 
--- a/src/aureceiver.c
+++ b/src/aureceiver.c
@@ -274,22 +274,23 @@ void aurecv_reset(struct audio_recv *ar)
 	mtx_unlock(ar->mtx);
 }
 
-/* Handle incoming stream data from the network */
-void aurecv_receive(struct audio_recv *ar, const struct rtp_header *hdr,
+
+int  aurecv_receive(struct audio_recv *ar, const struct rtp_header *hdr,
 		    struct rtpext *extv, size_t extc,
 		    struct mbuf *mb, unsigned lostc)
 {
 	bool discard = false;
 	int wrap;
+	int err = 0;
 	(void) lostc;
 
 	if (!mb)
-		return;
+		return EINVAL;
 
 	mtx_lock(ar->mtx);
 	if (hdr->pt != ar->pt) {
 		mtx_unlock(ar->mtx);
-		return;
+		return EAGAIN;
 	}
 
 
@@ -341,10 +342,11 @@ void aurecv_receive(struct audio_recv *ar, const struct rtp_header *hdr,
 /*        if (lostc)*/
 /*                (void)aurecv_stream_decode(ar, hdr, mb, lostc, drop);*/
 
-	(void)aurecv_stream_decode(ar, hdr, mb, 0);
+	err = aurecv_stream_decode(ar, hdr, mb, 0);
 
 out:
 	mtx_unlock(ar->mtx);
+	return err;
 }
 
 
--- a/src/core.h
+++ b/src/core.h
@@ -128,7 +128,7 @@ void aurecv_flush(struct audio_recv *ar);
 void aurecv_set_extmap(struct audio_recv *ar, uint8_t aulevel);
 int  aurecv_set_module(struct audio_recv *ar, const char *module);
 int  aurecv_set_device(struct audio_recv *ar, const char *device);
-void aurecv_receive(struct audio_recv *ar, const struct rtp_header *hdr,
+int  aurecv_receive(struct audio_recv *ar, const struct rtp_header *hdr,
 		    struct rtpext *extv, size_t extc,
 		    struct mbuf *mb, unsigned lostc);
 void aurecv_reset(struct audio_recv *ar);
@@ -302,7 +302,7 @@ struct rtp_header;
 
 enum {STREAM_PRESZ = 4+12}; /* same as RTP_HEADER_SIZE */
 
-typedef void (stream_rtp_h)(const struct rtp_header *hdr,
+typedef int (stream_rtp_h)(const struct rtp_header *hdr,
 			    struct rtpext *extv, size_t extc,
 			    struct mbuf *mb, unsigned lostc, bool new_source,
 			    void *arg);
--- a/src/rtprecv.c
+++ b/src/rtprecv.c
@@ -260,7 +260,7 @@ static int lostcalc(struct rtp_receiver *rx, uint16_t seq)
 }
 
 
-static void handle_rtp(struct rtp_receiver *rx, const struct rtp_header *hdr,
+static int handle_rtp(struct rtp_receiver *rx, const struct rtp_header *hdr,
 		      struct mbuf *mb, unsigned lostc)
 {
 	struct rtpext extv[8];
@@ -286,7 +286,7 @@ static void handle_rtp(struct rtp_receiver *rx, const struct rtp_header *hdr,
 			warning("rtp_receiver: corrupt rtp packet,"
 				" not enough space for rtpext of %zu bytes\n",
 				ext_len);
-			return;
+			return EPROTO;
 		}
 
 		mb->pos = mb->pos - ext_len;
@@ -298,7 +298,7 @@ static void handle_rtp(struct rtp_receiver *rx, const struct rtp_header *hdr,
 			if (err) {
 				warning("rtp_receiver: rtpext_decode failed "
 					"(%m)\n", err);
-				return;
+				return err;
 			}
 		}
 
@@ -313,9 +313,7 @@ handler:
 
 	bool new_source = rtprecv_consume_ssrc_change(rx, NULL);
 
-	rx->rtph(hdr, extv, extc, mb, lostc, new_source, rx->arg);
-
-	return;
+	return rx->rtph(hdr, extv, extc, mb, lostc, new_source, rx->arg);
 }
 
 
@@ -374,8 +372,10 @@ static void decode_frames(struct rtp_receiver *rx)
 
 		lostc = lostcalc(rx, hdr.seq);
 
-		handle_rtp(rx, &hdr, mb, lostc > 0 ? lostc : 0);
+		err = handle_rtp(rx, &hdr, mb, lostc > 0 ? lostc : 0);
 		mem_deref(mb);
+		if (err == ENODEV)
+			return;
 	} while (--n);
 
 	delay = jbuf_next_play(rx->jbuf);


cspiel1:
After this is fixed -> next issue: The call is closed locally but not signaled to the peer with CANCEL or BYE.

We should discuss:
- What is the wanted behavior if a video error occurs?
  - Proceed the call without video?
  - Or, terminate the call?
  - If we want to terminate the call, should the baresip core signal the termination to the peer? Or the application, e.g. menu?
  - Do we want a BEVENT_VIDEO_ERROR? There is a BEVENT_AUDIO_ERROR already.
- Should the behavior for an audio error be the same?

alfredh:
it should be fine to add a new event called e.g.

```BEVENT_VIDEO_ERROR```

Then this would be symmetrical with the audio error.

If the call should be closed when video error happened, is not so important.
You can decide if it should be closed or not.

**Comment by cspiel1** [comment]
> After this is fixed -> next issue: The call is closed locally but not signaled to the peer with CANCEL or BYE.

Looking closer. Could not reproduce this. There is a `SIP/2.0 486 Busy Here` which is sent from `termwait()` in `re/src/sipsess/sess.c`. And this is called from the `destructor()` of the `sipsess`. Seems to work always.

**Comment by alfredh** [comment]
would it be possible to see the backtrace of the case that is failing in the first place ?
perhaps the module code can be rewritten to not access the call object after calling the handler.

Using `tmr_start` to destroy an object is not 100% guaranteed to work,
for example if the timer is pending and the application is exiting, the timer will not be run.

**Comment by cspiel1** [comment]
Sorry, I forgot to mention the fact that the use after free happens in the `decode_frames()` loop in `rtprecv.c`.

- `handle_rtp()` calls `rx->rtph() = stream_recv_handler()` in `video.c`, which calls the `video_error_handler()`
- The use after free is at the beginning of the loop `jbuf_get(rx->jbuf, ...)` because `rx` was already freed. Only if `n` is still positive.

If we need it, I could reproduce this and paste the back trace.

> Using tmr_start to destroy an object is not 100% guaranteed to work,
for example if the timer is pending and the application is exiting, the timer will not be run.

What else can we do? Should we add an `err` field to `struct call` which is processed when it is save?

**Comment by cspiel1** [comment]
> What else can we do? Should we add an err field to struct call which is processed when it is save?

Or a `bool deref;`? I'll try to add the check of this flag into `decode_frames()` for an early exit.

Edit:
Since `void stream_rtp_h()` is in the internal `core.h` API it would also be an option to add an err return value.

**Comment by cspiel1** [comment]
I dropped: 0b042e67da9633c546ad2768cf0ff8eded71a2ae

Instead added err return value to `int stream_rtp_h()` and an early exit to `decode_frames()`

**Comment by cspiel1** [comment]
This one is ready.

I found another failure in `decode_frames()` which needs an extra PR. In the audio stream: If `jbuf_get()` returns EAGAIN, then the RTP packet is dropped in `aureceiver.c` function `aurecv_stream_decode()`. This fix has to wait until this one is merged.

File: src/call.c [hunk 1]
```diff
@@ -419,6 +419,9 @@ static void video_error_handler(int err, const char *str, void *arg)
 
 	warning("call: video device error: %m (%s)\n", err, str);
 
+	bevent_call_emit(BEVENT_VIDEO_ERROR, call,
+			 "%d,%s", err, str);
+
 	call_stream_stop(call);
 	call_event_handler(call, CALL_EVENT_CLOSED, "%s", str);
```
alfredh, line 423: Should we send both the old and the new event ?

sreimers:
> This one is ready.
> 
> I found another failure in `decode_frames()` which needs an extra PR. In the audio stream: If `jbuf_get()` returns EAGAIN, then the RTP packet is dropped in `aureceiver.c` function `aurecv_stream_decode()`. This fix has to wait until this one is merged.

Maybe we can remove the ignore handling completely, I think it drops "good" audio packets currently, if `jbuf_get` is called a bit late.

https://github.com/baresip/baresip/pull/3770

cspiel1:
> would it be possible to see the backtrace of the case that is failing in the first place ?

Sorry, didn't recognize that you requested the bt already.

From main branch. Suppose that suddenly the display handler returns ENODEV! Can be simulated by:
```patch
--- a/modules/x11/x11.c
+++ b/modules/x11/x11.c
@@ -309,6 +309,8 @@ static int display(struct vidisp_st *st, const char *title,
        int err = 0;
        (void)timestamp;
 
+       return ENODEV;
+
        if (!st->disp)
                return ENODEV;
```

```
Thread 1 "baresip" received signal SIGSEGV, Segmentation fault.
0x00007ffff7f7f528 in jbuf_get (jb=0xb5b5b5b5b5b5b5b5, hdr=0x7fffffffd788, mem=0x7fffffffd780) at /home/cspiel/src/baresip/src/jbuf.c:683
683		mtx_lock(jb->lock);
(gdb) bt
#0  0x00007ffff7f7f528 in jbuf_get (jb=0xb5b5b5b5b5b5b5b5, hdr=0x7fffffffd788, mem=0x7fffffffd780) at /home/cspiel/src/baresip/src/jbuf.c:683
#1  0x00007ffff7f8aa3b in decode_frames (rx=0x555555710720) at /home/cspiel/src/baresip/src/rtprecv.c:372
#2  0x00007ffff7f893dd in decode_tmr (arg=0x555555710720) at /home/cspiel/src/baresip/src/rtprecv.c:330
#3  0x00007ffff7ed6b88 in tmr_poll (tmrl=0x55555555b8e0) at /home/cspiel/src/re/src/tmr/tmr.c:166
#4  0x00007ffff7eb4961 in re_main (signalh=0x555555556c60 <signal_handler>) at /home/cspiel/src/re/src/main/main.c:1085
#5  0x0000555555556b16 in main (argc=1, argv=0x7fffffffdbd8) at /home/cspiel/src/baresip/src/main.c:329
```

cspiel1:
> Maybe we can remove the ignore handling completely, I think it drops "good" audio packets currently, if `jbuf_get` is called a bit late.
> 
> #3770

Not completely. It helps to ignore inbound DTMF. I'll comment in your draft.

src/video.c
```diff
@@ -904,7 +904,7 @@ static void stream_recv_handler(const struct rtp_header *hdr,
 	if (lostc)
 		request_picture_update(&v->vrx);
 
-	(void)video_stream_decode(&v->vrx, hdr, mb);
+	return video_stream_decode(&v->vrx, hdr, mb);
```
sreimers, line 905: We should test if only fatal errors reported by video modules. Since a invalid video frame should not stop the stream.

sreimers: Ok just checked, this would break video streams on protocol/frame errors. Ignoring return code here is intended, to wait for next key frame.

https://github.com/baresip/baresip/blob/0d6e332fb697fcf3825dc65f4f997278ff956848/modules/avcodec/decode.c#L331-L339

alfredh:
could it be an option to run the timer with a fixed interval ?

```diff
diff --git a/src/rtprecv.c b/src/rtprecv.c
index bce5521f..2e8b8adc 100644
--- a/src/rtprecv.c
+++ b/src/rtprecv.c
@@ -360,7 +360,6 @@ static void decode_frames(struct rtp_receiver *rx)
        struct rtp_header hdr;
        void *mb;
        int lostc;
-       int32_t delay;
        int err;
 
        if (!rx || !rx->jbuf)
@@ -368,6 +367,8 @@ static void decode_frames(struct rtp_receiver *rx)
 
        uint32_t n = jbuf_packets(rx->jbuf);
 
+       tmr_start(&rx->tmr_decode, 4, decode_tmr, rx);
+
        do {
                err = jbuf_get(rx->jbuf, &hdr, &mb);
                if (err && err != EAGAIN)
@@ -382,12 +383,6 @@ static void decode_frames(struct rtp_receiver *rx)
                if (err && err != EAGAIN)
                        break;
        } while (n--);
-
-       delay = jbuf_next_play(rx->jbuf);
-       if (delay < 0)
-               delay = 10; /* Fallback time */
-
-       tmr_start(&rx->tmr_decode, delay, decode_tmr, rx);
 }
 
 
```


The timer system is not accurate, and the timer will fire at the given delay
or a bit later.

sreimers:
> The timer system is not accurate, and the timer will fire at the given delay
> or a bit later.

This is not a big problem. aubuf handle these short periods of late packets easily. And late tmr is mostly due to a other slow callback/processing within main thread and since there is always a new packet to handle, the event loop wake ups regularly, too. so without a dedicated thread I see no benefits for a constant tmr interval.

src/video.c
cspiel1: We could handle decoder errors generally as to be ignored and display errors as to stop processing immediately. E.g. ENODEV could stop the call.

 If we keep the `mem_deref(call)` for `CALL_EVENT_CLOSED` and allow that in the `v->errh` a `CALL_EVENT_CLOSED` is emitted then after a call to `v->errh(...)` no access to the heap is allowed.
 
Thus we have to apply some special knowledge about upper SW layers in `rtprecv.c` in order to solve the SEGV. I'll make a commit.

cspiel1:
I rebased to main. Added commit: 98be3f9188a429315cac540085662fd887ef0585

Edit:
If you prefer we could first proceed here: https://github.com/baresip/baresip/pull/3770

include/baresip.h
```diff
@@ -918,6 +918,7 @@ enum bevent_ev {
 	BEVENT_VU_TX,
 	BEVENT_VU_RX,
 	BEVENT_AUDIO_ERROR,
+	BEVENT_VIDEO_ERROR,
```
alfredh: It would be cleaner if adding BEVENT_VIDEO_ERROR and usage of it
was a separate commit...

src/call.c
cspiel1: The reason for this PR was not to add a new event. It should solve the SEGV. I'll remove the new event in this PR.

include/baresip.h
cspiel1: The purpose of this PR was not to add a new event. It should solve the SEGV. I dropped the new event for now.

cspiel1:
I'll resolve the conflicts.

src/video.c
cspiel1: > Thus we have to apply some special knowledge about upper SW layers in rtprecv.c in order to solve the SEGV.

This is not good. Do you have a better idea?

alfredh:
one option is to check the return value in video.c where disph() is called:

```c
	if (vrx->vd && vrx->vd->disph && vrx->vidisp)
		err = vrx->vd->disph(vrx->vidisp, v->peer, frame,
				     pkt.timestamp);

	frame_filt = mem_deref(frame_filt);
	if (err == ENODEV) {
		warning("video: video-display was closed\n");
		vrx->vidisp = mem_deref(vrx->vidisp);
		vrx->vd = NULL;

		mtx_unlock(&vrx->lock);

		if (v->errh) {
			v->errh(err, "display closed", v->arg);
		}

		return err;
	}
```

if the video display is closed you could mark the video-stream as terminated,
and close the call.

In other projects I have used a boolean flag to indicate that a state is terminated,
which can be checked in loops etc:

```c
struct state {
    bool terminated;
};


void terminate(struct state *st)
{
    st->terminated = true;
}

void loop(struct st *st)
{
    for (;;) {
        /* something ... */

        if (st->terminated)
            break;
    }
}
```

sreimers:
Yes maybe something like this could work (untested):

```patch
diff --git a/src/rtprecv.c b/src/rtprecv.c
index 79a7e856..c343c418 100644
--- a/src/rtprecv.c
+++ b/src/rtprecv.c
@@ -52,6 +52,7 @@ struct rtp_receiver {
 	int pt_tel;                    /**< Payload type for tel event       */
 	uint32_t srate;                /**< Receiver Samplerate              */
 	struct tmr tmr_decode;         /**< Decode Timer                     */
+	RE_ATOMIC bool *terminated;    /**< Destroy flag of decode_frames()  */
 };
 
 
@@ -355,6 +356,7 @@ static void rtprecv_resync(struct rtp_receiver *rx,
 static void decode_frames(struct rtp_receiver *rx)
 {
 	struct rtp_header hdr;
+	RE_ATOMIC bool *terminated;
 	void *mb;
 	int lostc;
 	int32_t delay;
@@ -365,6 +367,10 @@ static void decode_frames(struct rtp_receiver *rx)
 
 	uint32_t n = 1;
 
+	/* handle_rtp() may lead to a CALL_CLOSED event, which destroys the
+	 * call and thus this RTP receiver. The flag outlives rx. */
+	terminated = mem_ref(rx->terminated);
+
 	do {
 		err = jbuf_get(rx->jbuf, &hdr, &mb);
 		if (err == EAGAIN)
@@ -376,6 +382,11 @@ static void decode_frames(struct rtp_receiver *rx)
 
 		handle_rtp(rx, &hdr, mb, lostc > 0 ? lostc : 0);
 		mem_deref(mb);
+
+		/* After CALL_CLOSED no data of the call object may be
+		 * accessed anymore -- including rx itself */
+		if (re_atomic_acq(terminated))
+			goto out;
 	} while (--n);
 
 	delay = jbuf_next_play(rx->jbuf);
@@ -383,6 +394,9 @@ static void decode_frames(struct rtp_receiver *rx)
 		delay = 10; /* Fallback time */
 
 	tmr_start(&rx->tmr_decode, delay, decode_tmr, rx);
+
+ out:
+	mem_deref(terminated);
 }
 
 
@@ -722,6 +736,10 @@ static void destructor(void *arg)
 {
 	struct rtp_receiver *rx = arg;
 
+	/* Tell a decode_frames() loop below us on the stack to bail out */
+	if (rx->terminated)
+		re_atomic_rls_set(rx->terminated, true);
+
 	if (re_atomic_rlx(&rx->run)) {
 		rtprecv_enable(rx, false);
 		re_atomic_rlx_set(&rx->run, false);
@@ -740,6 +758,7 @@ static void destructor(void *arg)
 	mem_deref(rx->mtx);
 	mem_deref(rx->jbuf);
 	mem_deref(rx->cname);
+	mem_deref(rx->terminated);
 }
 
 
@@ -769,6 +788,12 @@ int rtprecv_alloc(struct rtp_receiver **rxp,
 	rx->pt     = -1;
 	rx->pt_tel = -1;
 
+	rx->terminated = mem_zalloc(sizeof(*rx->terminated), NULL);
+	if (!rx->terminated) {
+		err = ENOMEM;
+		goto out;
+	}
+
 	err  = str_dup(&rx->name, name);
 	err |= mutex_alloc(&rx->mtx);
 	if (err)

```

cspiel1:
Sorry, currently I have no time to proceed here. On Tuesday next week my 2 week vacation starts. I could try your suggestion in the end of August. Or you create another PR and close this one.

alfredh:
We can close this one and create a new PR when you are back from vacation ...

Enjoy your Holidays!

cspiel1:
Thank you!

PR #3770, opened 2026-07-13
core: cleanup stream_rtp_h ignore handling

sreimers:
Since the new jbuf is already aware of clock timing, there should no need for dropping a packet anymore. A jbuf_get EAGAIN should always be decoded.

### DIFF
--- a/src/aureceiver.c
+++ b/src/aureceiver.c
@@ -213,7 +213,7 @@ static int aurecv_push_aubuf(struct audio_recv *ar, const struct auframe *af)
 
 static int aurecv_stream_decode(struct audio_recv *ar,
 				const struct rtp_header *hdr,
-				struct mbuf *mb, unsigned lostc, bool drop)
+				struct mbuf *mb, unsigned lostc)
 {
 	struct auframe af;
 	size_t sampc = ar->sampvsz / aufmt_sample_size(ar->fmt);
@@ -256,11 +256,6 @@ static int aurecv_stream_decode(struct audio_recv *ar,
 	auframe_init(&af, ar->fmt, ar->sampv, sampc, ac->srate, ac->ch);
 	af.timestamp = ((uint64_t) hdr->ts) * AUDIO_TIMEBASE / ac->crate;
 
-	if (drop) {
-		aubuf_drop_auframe(ar->aubuf, &af);
-		goto out;
-	}
-
 	err = aurecv_process_decfilt(ar, &af);
 	if (err)
 		goto out;
@@ -286,10 +281,9 @@ void aurecv_reset(struct audio_recv *ar)
 /* Handle incoming stream data from the network */
 void aurecv_receive(struct audio_recv *ar, const struct rtp_header *hdr,
 		    struct rtpext *extv, size_t extc,
-		    struct mbuf *mb, unsigned lostc, bool *ignore)
+		    struct mbuf *mb, unsigned lostc)
 {
 	bool discard = false;
-	bool drop = *ignore;
 	int wrap;
 	(void) lostc;
 
@@ -299,11 +293,9 @@ void aurecv_receive(struct audio_recv *ar, const struct rtp_header *hdr,
 	mtx_lock(ar->mtx);
 	if (hdr->pt != ar->pt) {
 		mtx_unlock(ar->mtx);
-		*ignore = true;
 		return;
 	}
 
-	*ignore = false;
 
 	/* RFC 5285 -- A General Mechanism for RTP Header Extensions */
 	const struct rtpext *ext = rtpext_find(extv, extc, ar->extmap_aulevel);
@@ -353,7 +345,7 @@ void aurecv_receive(struct audio_recv *ar, const struct rtp_header *hdr,
 /*        if (lostc)*/
 /*                (void)aurecv_stream_decode(ar, hdr, mb, lostc, drop);*/
 
-	(void)aurecv_stream_decode(ar, hdr, mb, 0, drop);
+	(void)aurecv_stream_decode(ar, hdr, mb, 0);
 
 out:
 	mtx_unlock(ar->mtx);
--- a/src/rtprecv.c
+++ b/src/rtprecv.c
@@ -260,7 +260,7 @@ static int lostcalc(struct rtp_receiver *rx, uint16_t seq)
 }
 
 
-static int handle_rtp(struct rtp_receiver *rx, const struct rtp_header *hdr,
+static void handle_rtp(struct rtp_receiver *rx, const struct rtp_header *hdr,
 		      struct mbuf *mb, unsigned lostc)
 {
 	struct rtpext extv[8];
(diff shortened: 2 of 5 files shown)


src/aureceiver.c
```diff
@@ -256,11 +256,6 @@ static int aurecv_stream_decode(struct audio_recv *ar,
 	auframe_init(&af, ar->fmt, ar->sampv, sampc, ac->srate, ac->ch);
 	af.timestamp = ((uint64_t) hdr->ts) * AUDIO_TIMEBASE / ac->crate;
 
-	if (drop) {
-		aubuf_drop_auframe(ar->aubuf, &af);
```
cspiel1, line 260: I think this breaks ajb. We can drop ajb from aubuf now. Because the adaptive jbuf is a replacement.

So for us this would be okay.

src/rtprecv.c
```diff
@@ -370,16 +367,17 @@ static void decode_frames(struct rtp_receiver *rx)
...
 
-		err = handle_rtp(rx, &hdr, mb, lostc > 0 ? lostc : 0,
-				 err == EAGAIN);
+		err = handle_rtp(rx, &hdr, mb, lostc > 0 ? lostc : 0);
 		mem_deref(mb);
 
-		if (err && err != EAGAIN)
+		if (err)
 			break;
 	} while (n--);
```
cspiel1: I think if `jbuf_get()` returns zero the loop should also be left here  (after processing the last packet).

src/rtprecv.c
```diff
@@ -314,9 +313,7 @@ static int handle_rtp(struct rtp_receiver *rx, const struct rtp_header *hdr,
 
 	bool new_source = rtprecv_consume_ssrc_change(rx, NULL);
 
-	rx->rtph(hdr, extv, extc, mb, lostc, new_source, &ignore, rx->arg);
-	if (ignore)
-		return EAGAIN;
```
cspiel1, line 319: I think this also breaks ajb. See: https://github.com/baresip/baresip/pull/3770/changes#r3572431942

src/rtprecv.c
```diff
@@ -370,16 +367,17 @@ static void decode_frames(struct rtp_receiver *rx)
 
 	do {
 		err = jbuf_get(rx->jbuf, &hdr, &mb);
-		if (err && err != EAGAIN)
+		if (err == EAGAIN)
+			++n;
```
cspiel1, line 371: `n` is the complete number of packets in jbuf. This can't be incremented here.

src/rtprecv.c
```diff
@@ -492,7 +490,7 @@ void rtprecv_decode(const struct sa *src, const struct rtp_header *hdr,
 		}
 	}
 	else {
-		(void)handle_rtp(rx, hdr, mb, 0, false);
+		(void)handle_rtp(rx, hdr, mb, 0);
```
cspiel1, line 490: We should test both branches with inbound DTMF.
- with jbuf
- without jbuf

The telev events should not be forwarded to the audio filters and aubuf.

src/rtprecv.c
```diff
@@ -261,11 +261,10 @@ static int lostcalc(struct rtp_receiver *rx, uint16_t seq)
 
 
 static int handle_rtp(struct rtp_receiver *rx, const struct rtp_header *hdr,
-		      struct mbuf *mb, unsigned lostc, bool drop)
```
cspiel1, line 264: Can we drop the return value? Or should we stop the stream if an error occurs in the filters or in the decoder?

This question influences also the `decode_frames()` loop.

src/aureceiver.c
sreimers: I think it's already broken, since `EAGAIN` has a slightly different meaning now. So yes we can maybe remove ajb.

src/rtprecv.c
sreimers: `n` is only a hint (since `jbuf_get` returns `ENOENT` if no packet ready), but I improved the loop with last commit, too.

src/rtprecv.c
sreimers: I think there is a bug for async handling (`rtp_rxmode = thread`) at first look:

https://github.com/baresip/baresip/blob/0d6e332fb697fcf3825dc65f4f997278ff956848/src/rtprecv.c#L477-L498

https://github.com/baresip/baresip/blob/0d6e332fb697fcf3825dc65f4f997278ff956848/src/audio.c#L595-L619

`stream_pt_handler()` (`rx->pth`) returns `ENODATA` (`telephone-event`) if called directly but  `pass_pt_work()` async path returns always success (does not forward err) and packet is forwarded to `jbuf_put()` or `handle_rtp()`. But looks unrelated to this PR.

src/rtprecv.c
sreimers: Good point. I don't think we should stop the stream, since a video decoder for example can produce temporary errors and request new full frame by peer. I removed the err handling.

src/rtprecv.c
sreimers: Maybe a fix could look like this:

```diff
--- a/src/rtprecv.c
+++ b/src/rtprecv.c
@@ -473,6 +473,9 @@ void rtprecv_decode(const struct sa *src, const struct rtp_header *hdr,
                err = pass_pt_work(rx, hdr->pt, mb);
                if (err)
                        return;
+
+               if (hdr->pt == rx->pt_tel)
+                       return;
        }

        if (rx->jbuf) {
```

src/rtprecv.c
cspiel1: Looks good and short now.

src/rtprecv.c
cspiel1: Yes, this could work. Next PR!

src/rtprecv.c
```diff
@@ -374,12 +374,9 @@ static void decode_frames(struct rtp_receiver *rx)
 
 		lostc = lostcalc(rx, hdr.seq);
 
-		err = handle_rtp(rx, &hdr, mb, lostc > 0 ? lostc : 0);
+		handle_rtp(rx, &hdr, mb, lostc > 0 ? lostc : 0);
```
cspiel1, line 377: In this PR https://github.com/baresip/baresip/pull/3764 maybe we will have to add again the return value. It is possible that `handle_rtp()` leads to  CALL_CLOSED and to `mem_deref` of the call and `rx`. Then we need a simple check for an early exit.

src/rtprecv.c
cspiel1: Thanks! But sorry, I forgot that we maybe will need the return value for the early exit. See https://github.com/baresip/baresip/pull/3770#discussion_r3576435935 !

src/rtprecv.c
cspiel1: Here: https://github.com/baresip/baresip/pull/3773

src/aureceiver.c
alfredh: It would be great to remove AJB, it would also simplify the code and simplify the config.

cspiel1:
This looks good for me. Is there a reason why this is still a draft?

PR #3771, opened 2026-07-14
l16: honor account/peer ptime for low-rate modes

allesrebel:
The codec-table ptime pins mono 16kHz L16 to 20ms packetization: audio_encoder_set() applies a codec-specific ptime after both the account ptime and the peer's SDP ptime have been processed, so neither an accounts ;ptime= parameter nor a remote a=ptime request can select 10ms TX framing for this codec.

Set the table ptime to 0 (no codec-specific preference), the convention sample-based codecs such as g711 already use. Behavior is unchanged for anyone not setting a ptime: the stream still initializes from the global 20ms default.


modules/l16/l16.c
```diff
@@ -82,7 +82,7 @@ static struct aucodec l16v[NR_CODECS] = {
 {LE_INIT,    0, "L16", 48000, 48000, 1, 1, 10, 0, 0, encode, 0, decode, 0,0,0},
 {LE_INIT, "11", "L16", 44100, 44100, 1, 1, 10, 0, 0, encode, 0, decode, 0,0,0},
 {LE_INIT,    0, "L16", 32000, 32000, 1, 1, 20, 0, 0, encode, 0, decode, 0,0,0},
-{LE_INIT,    0, "L16", 16000, 16000, 1, 1, 20, 0, 0, encode, 0, decode, 0,0,0},
+{LE_INIT,    0, "L16", 16000, 16000, 1, 1,  0, 0, 0, encode, 0, decode, 0,0,0},
```
alfredh, line 85: is there any reason that not all codecs should have ptime=0 ?

Why does only that codec need to have ptime=0 ?

hme-arebel: From what I can tell, the rationale for pTimes configured for the other bit rates were driven 1500 Byte MTU size. 
For bit rates and channels up to 32K mono, the default 20 ms fits. We can apply the suggested change to leave the pTime choice to the user and keep the graceful fallback. I'll test this and confirm before committing up the change. (4 additional changes)

For 44.1kHz mono and above, the 20ms default isn't safe. (would exceed a 1500 byte frame)
Zeroing out these items in the codec list could cause some deployments to fail if they relied on this behavior (such as over the internet).

I'll leave it to you to decide if it's a good call to base on 1500 MTU size. 

Regarding why L16? And not other codecs
It can support any pTime requested by downstream plug ins with no issue. Since it's effectively 16bit raw PCM audio, there's no impact from different duration audio being sent over the wire. Zeroing out these smaller values allows users to actually bump up pTime if they wanted. The other codecs, there may be other factors that force pTime to specific values. Like framed codecs being constrained to multiples of their frame duration. I can look into these if you like - but probably a different PR.

allesrebel: Confirmed both peer/account pTime is honored for the zeroed ptime bitrates.

Added doc update + test update to prevent regression

test/l16.c
```diff
@@ -0,0 +1,55 @@
...
+#include <baresip.h>
+#include "test.h"
+
+
+#define DEBUG_MODULE "l16"
+#define DEBUG_LEVEL 5
+#include <re_dbg.h>
+
+
+int test_l16_ptime(void)
```
alfredh: Would it be possible to add L16 to the existing call testcases instead?

In general I prefer integration/end-to-end tests over unit-tests, as they cover
more code with less test-code.

allesrebel: No problem, removed and replaced with call level testcases.

alfredh:
what is the current logic regarding selecting ptime for sending audio packets ?
the value can come from three sources: ptime in accounts, ptime in SDP, ptime in codec.

which value gets priority ?

allesrebel:
right now the priority order is:
codec > peer SDP > account > global

global ptime is loaded on creation of account struct (via account_alloc)

the logic is use codec list's ptime if non-zero - regardless what any requested ptime (via audio_encoder_set)

if codec ptime == 0, peer's requested ptime is used (via audio_sdp_attr_decode).

if the peer does not have a preference, the account config is used (decoded during sip_params_decode) or fallback to global.

test/call.c
```diff
@@ -1185,6 +1185,98 @@ int test_call_aulevel(void)
...
+			    ";ptime=%u;audio_codecs=L16/%u/%u",
+			    aptime, srate, ch);
+	else
+		re_snprintf(prm, sizeof(prm),
+			    ";audio_codecs=L16/%u/%u", srate, ch);
+
+	fixture_init_prm(f, prm);
+
+	cancel_rule_new(BEVENT_CALL_RTPESTAB, f->b.ua, 1, 0, 1);
+	cancel_rule_and(BEVENT_CALL_RTPESTAB, f->a.ua, 0, 0, 1);
```
cspiel1, line 1219: Should we also check if some audio frames arrive at both sides?

```c
	cr->prm = "auframe";
	cr->n_auframe = 3;
```

Example: `test_100rel_audio_base()`

Note, this checks if the audio frames reach the audio player (`auplay`).

hme-arebel: This test intentionally focuses on bidirectional wire packetization - 
Since audio delivery to auplay is already covered by existing call tests, my hunch is we shouldn't. But if you want, I'm happy to add it in

modules/l16/l16.c
alfredh: I am not sure about this ...

It does not make sense that one codec module has a mixture of
ptime values 0 or non-0. it would be more clean if all ptime values
are either 0, or set to a good value.

Perhaps the ptime value should be calculated from the MTU value.

hme-arebel: If the module had only 0 values - the danger comes if either peer or account don't specify a good ptime.  (and the fall back to the global is used). However, we allow the user to make the choice - which I think is powerful.

If the module only had specific pTime values based on MTU, then we remove user configurability at lower bitrates. For example, in the extreme case 8kHz mono can do up to 91 ms pTimes for 1500 MTU frames; significantly limiting user configurability. But the out of box experience has a higher likelyhood of working in general.

For me, the mixture presents a good tradeoff.
But I lean towards having zeros for maximum user control - since this codec module is flexible. Maybe as a follow on to this, we can look into a way to make the higher bitrate codecs to also safely honor ptimes. 

Perhaps a way to fall back gracefully to a pTime that works when peer or account doesn't specify. Something like active probing with increased packet sizes to detect MTU size of link (or use a value configured by user). Using this MTU size as the basis for the pTimes we call back to (instead of the account.c global). 
Another approach could be flagging the user with a warning / error when  using the default value at these higher rates. Would probably produce bug reports, but be very easy to spot/triage. 

What's your preference - happy to go down any path here

alfredh:
how about changing the codec ptime from 20ms to 10ms ?

something like this:


```c
/* See RFC 3551 */
static struct aucodec l16v[NR_CODECS] = {
{LE_INIT,    0, "L16", 48000, 48000, 2, 2,  4, 0, 0, encode, 0, decode, 0,0,0},
{LE_INIT, "10", "L16", 44100, 44100, 2, 2,  4, 0, 0, encode, 0, decode, 0,0,0},
{LE_INIT,    0, "L16", 32000, 32000, 2, 2, 10, 0, 0, encode, 0, decode, 0,0,0},
{LE_INIT,    0, "L16", 16000, 16000, 2, 2, 10, 0, 0, encode, 0, decode, 0,0,0},
{LE_INIT,    0, "L16",  8000,  8000, 2, 2, 10, 0, 0, encode, 0, decode, 0,0,0},
{LE_INIT,    0, "L16", 48000, 48000, 1, 1, 10, 0, 0, encode, 0, decode, 0,0,0},
{LE_INIT, "11", "L16", 44100, 44100, 1, 1, 10, 0, 0, encode, 0, decode, 0,0,0},
{LE_INIT,    0, "L16", 32000, 32000, 1, 1, 10, 0, 0, encode, 0, decode, 0,0,0},
{LE_INIT,    0, "L16", 16000, 16000, 1, 1, 10, 0, 0, encode, 0, decode, 0,0,0},
{LE_INIT,    0, "L16",  8000,  8000, 1, 1, 10, 0, 0, encode, 0, decode, 0,0,0},
};
```


would that solve your issue, and also keep RTP packets under 1500 bytes ?

allesrebel:
> how about changing the codec ptime from 20ms to 10ms ?
> 
> something like this:
> 
> ```c
> /* See RFC 3551 */
> static struct aucodec l16v[NR_CODECS] = {
> {LE_INIT,    0, "L16", 48000, 48000, 2, 2,  4, 0, 0, encode, 0, decode, 0,0,0},
> {LE_INIT, "10", "L16", 44100, 44100, 2, 2,  4, 0, 0, encode, 0, decode, 0,0,0},
> {LE_INIT,    0, "L16", 32000, 32000, 2, 2, 10, 0, 0, encode, 0, decode, 0,0,0},
> {LE_INIT,    0, "L16", 16000, 16000, 2, 2, 10, 0, 0, encode, 0, decode, 0,0,0},
> {LE_INIT,    0, "L16",  8000,  8000, 2, 2, 10, 0, 0, encode, 0, decode, 0,0,0},
> {LE_INIT,    0, "L16", 48000, 48000, 1, 1, 10, 0, 0, encode, 0, decode, 0,0,0},
> {LE_INIT, "11", "L16", 44100, 44100, 1, 1, 10, 0, 0, encode, 0, decode, 0,0,0},
> {LE_INIT,    0, "L16", 32000, 32000, 1, 1, 10, 0, 0, encode, 0, decode, 0,0,0},
> {LE_INIT,    0, "L16", 16000, 16000, 1, 1, 10, 0, 0, encode, 0, decode, 0,0,0},
> {LE_INIT,    0, "L16",  8000,  8000, 1, 1, 10, 0, 0, encode, 0, decode, 0,0,0},
> };
> ```
> 
> would that solve your issue, and also keep RTP packets under 1500 bytes ?

While it would solve my issue for my use case - it could cause others to have ptime issues.

For example, suppose another user wants 8kHz mono at 30ms ptime and they would always get 10ms even if they configured both account and peer.

alfredh:
hi, sorry about the slow response ...

I think the current ptime logic is a bit weird, perhaps it is time to update it.

It looks like the codec specific ptime is more correctly defined as "maximum ptime".
For example the L16 codec with ptime=4 could also use ptime=1,2,3

As an experiment we could redefine the meaning of this codec-ptime and update the code
to use the value as an upper limit.

Please review and test this patch, let me know what you think:


```patch
index fcf3672c..3a190162 100644
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1282,7 +1282,7 @@ struct aucodec {
 	uint32_t crate;             /* RTP Clock rate   */
 	uint8_t ch;
 	uint8_t pch;                /* RTP packet channels */
-	uint32_t ptime;             /* Packet time in [ms] (optional) */
+	uint32_t ptime_max;         /* Max packet time in [ms] (optional) */
 	const char *fmtp;
 	auenc_update_h *encupdh;
 	auenc_encode_h *ench;
diff --git a/src/audio.c b/src/audio.c
index aa5f3807..1886ee50 100644
--- a/src/audio.c
+++ b/src/audio.c
@@ -744,9 +744,6 @@ int audio_alloc(struct audio **ap, struct list *streaml,
 
 		struct aucodec *ac = le->data;
 
-		if (ac->ptime)
-			minptime = min(minptime, ac->ptime);
-
 		err = add_audio_codec(stream_sdpmedia(a->strm), ac);
 		if (err)
 			goto out;
@@ -1359,8 +1356,13 @@ int audio_encoder_set(struct audio *a, const struct aucodec *ac,
 	telev_set_srate(a->telev, ac->crate);
 
 	/* use a codec-specific ptime */
-	if (ac->ptime)
-		tx->ptime = ac->ptime;
+	if (ac->ptime_max && tx->ptime > ac->ptime_max) {
+
+		re_printf(".... encoder: codec '%s' new ptime: %u -> %u\n",
+			  ac->name, tx->ptime, ac->ptime_max);
+
+		tx->ptime = ac->ptime_max;
+	}
 
 	return err;
 }
@@ -1543,41 +1545,9 @@ static bool extmap_handler(const char *name, const char *value, void *arg)
 
 void audio_sdp_attr_decode(struct audio *a)
 {
-	const char *attr;
-
 	if (!a)
 		return;
 
-	/* This is probably only meaningful for audio data, but
-	   may be used with other media types if it makes sense. */
-	attr = sdp_media_rattr(stream_sdpmedia(a->strm), "ptime");
-	if (attr) {
-		struct autx *tx = &a->tx;
-		uint32_t ptime_tx = atoi(attr);
-
-		if (ptime_tx && ptime_tx != a->tx.ptime
-		    && ptime_tx <= MAX_PTIME) {
-
-			info("audio: peer changed ptime_tx %ums -> %ums\n",
-			     a->tx.ptime, ptime_tx);
-
-			tx->ptime = ptime_tx;
-
-			if (tx->ac) {
-				size_t sz;
-
-				sz = aufmt_sample_size(tx->src_fmt);
-
-				tx->psize = sz * au_calc_nsamp(tx->ac->srate,
-							    tx->ac->ch,
-							    ptime_tx);
-			}
-
-			sdp_media_set_lattr(stream_sdpmedia(a->strm), true,
-					    "ptime", "%u", ptime_tx);
-		}
-	}
-
 	/* Client-to-Mixer Audio Level Indication */
 	if (a->cfg.level) {
 		sdp_media_rattr_apply(stream_sdpmedia(a->strm),
```

allesrebel:
No worries at all! Thank you for looking at this

Regarding the idea - I love it - treating the pTime as a ceiling / "maximum ptime" would solve mine and most of the potential issues other's may encounter. 

Regarding the patch
I disagree with the removal of the 'ptime' parsing from audio_sdp_attr_decode. Being a good peer to others is a good thing! If they have a preference, we should try to honor it (bounded by max ptime). 

I checked every in-tree user of aucodec::ptime, only other places where positional initialization weren't used and named fields were was in modules/aac/aac.c and modules/aptx/aptx.c. I'll update these references. 

I'll update the table entries to make it clear how exactly they are calculated and from what MTU value.

alfredh:
We can keep the name `aucodec::ptime` unchanged, just update the semantics.

Here is an updated patch:

```patch
index fcf3672c..05f02ec1 100644
--- a/include/baresip.h
+++ b/include/baresip.h
@@ -1282,7 +1282,7 @@ struct aucodec {
 	uint32_t crate;             /* RTP Clock rate   */
 	uint8_t ch;
 	uint8_t pch;                /* RTP packet channels */
-	uint32_t ptime;             /* Packet time in [ms] (optional) */
+	uint32_t ptime;             /* Max packet time in [ms] (optional) */
 	const char *fmtp;
 	auenc_update_h *encupdh;
 	auenc_encode_h *ench;
diff --git a/src/audio.c b/src/audio.c
index aa5f3807..525380b0 100644
--- a/src/audio.c
+++ b/src/audio.c
@@ -744,9 +744,6 @@ int audio_alloc(struct audio **ap, struct list *streaml,
 
 		struct aucodec *ac = le->data;
 
-		if (ac->ptime)
-			minptime = min(minptime, ac->ptime);
-
 		err = add_audio_codec(stream_sdpmedia(a->strm), ac);
 		if (err)
 			goto out;
@@ -1359,8 +1356,13 @@ int audio_encoder_set(struct audio *a, const struct aucodec *ac,
 	telev_set_srate(a->telev, ac->crate);
 
 	/* use a codec-specific ptime */
-	if (ac->ptime)
+	if (ac->ptime && tx->ptime > ac->ptime) {
+
+		re_printf(".... encoder: codec '%s' new ptime: %u -> %u\n",
+			  ac->name, tx->ptime, ac->ptime);
+
 		tx->ptime = ac->ptime;
+	}
 
 	return err;
 }
```

if you agree with these changes, could you please include this patch in your PR ?

allesrebel:
Will do, should be pushing up shortly after getting everything tested and regressed
Thank you again

allesrebel:
This turned out to be a bit of an adventure - so please feel free to let me know if I should PR the last few commits separately due to the larger surface area. 

While regression testing and looking at the code paths that could execute them - AI review flagged a issue with re-INVITEs. Validated it was a real problem, the scenario where caps would be retained between codec changes to uncapped codecs. (L16 -> PCMU for example)

To address this issue, implemented a method to calculate ptime and resize buffers to handle this.

Added integration tests like before

alfredh:
the windows job has a new warning:

```
list: insert_after: le linked to 0x20a15501ea0
```

it means that the list-element is already linked to another list.

is it related to this PR or something else ?

allesrebel:
Will be root causing - I'm assuming it's something in the PR until root caused elsewhere. Should be fixing this up today if that's the case

alfredh:
Many thanks for your contributions and improvements :)

src/audio.c
```diff
@@ -424,6 +427,15 @@ static void poll_aubuf_tx(struct audio *a)
...
 
+	if (aubuf_cur_size(tx->aubuf) < tx->psize) {
+		underrun = ++tx->stats.aubuf_underrun;
+		tx->ts_ext += tx->ptime * tx->ac->crate / 1000;
+		mtx_unlock(tx->mtx);
+		debug("audio: thread: tx aubuf underrun (total %llu)\n",
+		      underrun);
+		return;
+	}
+
```
alfredh: would it be possible to move this block and related code to a separate PR ?

We have had some regressions in the last years, and the audio pipeline is a sensitive area.
It is easier to bisect git commits if changes are small ...

allesrebel:
> Many thanks for your contributions and improvements :)

Thank me when it's merged ;)

src/audio.c
allesrebel: Will do - will extract this out!

allesrebel:
> the windows job has a new warning:
> 
> ```
> list: insert_after: le linked to 0x20a15501ea0
> ```
> 
> it means that the list-element is already linked to another list.
> 
> is it related to this PR or something else ?

Was able to find what probably caused this - will push up a fix

With a bit of bisection and splicing - it looks like this is due the application exceeding POOL_FRAMES (inside re) - causing reallocs (and a flush). This shows up on the windows CI (and linux too!) if you mess with TX_AUBUF_MAX_PACKETS.

The CI issue may have been caused by mem_pool_flush() in libre (re/src/mem/mem_pool.c:266) rebuilding the pool's free list without first clearing pool->freel. Any slot that was already free could get linked twice, closing the list into a cycle. (needs to be validated) Let's see what happens after the next commit gets tested on the CI.

In short - yes my fault, should have thought harder about the 30

sreimers:
> With a bit of bisection and splicing - it looks like this is due the application exceeding POOL_FRAMES (inside re) - causing reallocs (and a flush)

Can you retest, there was a possible problem with flush and should be [fixed](https://github.com/baresip/re/commit/8a520ea410fec9256a1eb0cbc074108d442595f8) now

src/audio.c
hme-arebel: removed - restored the original upstream in b37f26e6199b0bd5506fc1df9bd13d0cbf7fdee4

Will do a separate PR for the changes (if they are even needed)

src/audio.c
```diff
@@ -897,7 +902,10 @@ static int tx_thread(void *arg)
 			      " (total %llu)\n", tx->stats.aubuf_underrun);
 		}
 
-		ts += tx->ptime;
+		mtx_lock(tx->mtx);
+		ptime = tx->ptime;
+		mtx_unlock(tx->mtx);
+		ts += ptime;
```
alfredh: there is perhaps a bit too many mutex lock/unlock blocks inside this thread.
perhaps it is more efficient to only lock/unlock once in the beginning, and copy
the needed variables there ?

src/audio.c
```diff
@@ -1294,6 +1303,41 @@ bool audio_started(const struct audio *a)
 }
 
 
+static int audio_tx_apply_ptime(struct audio *a, const struct aucodec *ac)
```
alfredh: this function only returns "0" perhaps consider to make it return `void` instead ?

alfredh:
I have done some basic testing, seems to still work. In general the PR looks good now.

I will let the other maintainers take a look also, before merging ...

sreimers:
I think we should split this PR up, the core audio changes and keep here only the related to l16.

src/audio.c
allesrebel: Agreed - I like that pattern - applying

src/audio.c
allesrebel: Yeah, changing now - was originally to catch resize errors - should have changed this function to void after consolidating to a single resize

allesrebel:
> I think we should split this PR up, the core audio changes and keep here only the related to l16.

Agreed, I'll split between the earlier codec table changes which relied on the older ptime behavior (the fall back behavior described above in the priority order) and on the newer behavior that treats suggested ptime as a cap.

src/audio.c
allesrebel: included in 9fd4beeb47e4dd44aa24fcf6428215d35f127d15

src/audio.c
allesrebel: included in f7303c16fdbd460542b3aa23470ce5210a639a7d

modules/l16/l16.c
allesrebel: Will be addressed with the feature on changing the behavior of ptime to be a cap

cspiel1:
Please add notes how your other PRs relate to this one! Is this one replaced by them? If so then please close it with a comment!

allesrebel:
No problem! If there was one to reject / close, it would probably be this PR. The functionality in this PR is the same as #3802 + #3805 to the end user. This is a lower risk PR if those other two are delayed - since nothing in core audio is touched at all. 

Happy to close, just let me know if the other route is preferred

This PR (#3771) - modifies the codec table for L16 to allow the low rate codecs in the L16 list to allow accept pTimes from peers/offers. No audio or behavior changes to the rest of the codebase. 

PR #3802 -> Changes the core system ptime behavior. pTime would be treated as a maximum, and allow peers to ask for different packetization times and get them if they are below the cap. 
PR #3805 -> depends on #3802 being merged first. Updates the L16 codec table to define pTimes derived from 1500 MTU sizes explicitly. Compatible with #3771, simply extends to all items on the list, rather than just the low rate codecs.

PR #3773, opened 2026-07-16
rtprecv: avoid processing of telev events

cspiel1:
In case of `rtp_rxmode thread` telev events had been forwarded to the
decoder pipeline. This commit adds an early exit for telev events

### DIFF
--- a/src/rtprecv.c
+++ b/src/rtprecv.c
@@ -344,7 +344,7 @@ static void rtprecv_resync(struct rtp_receiver *rx,
 	rx->pseq = (uint16_t)(hdr->seq - 1);
 	rx->pseq_set = true;
 	rx->pt = -1;
-	rx->pt_tel = 0;
+	rx->pt_tel = -1;
 	rx->ssrc_changed = true;
 	mtx_unlock(rx->mtx);
 }
@@ -397,7 +397,7 @@ static bool rtprecv_filter_pt(struct rtp_receiver *rx,
 	bool handle;
 
 	handle = hdr->pt != rx->pt;
-	if (rx->pt_tel)
+	if (rx->pt_tel != -1)
 		handle |= hdr->pt == rx->pt_tel;
 
 	if (!handle)
@@ -772,6 +772,7 @@ int rtprecv_alloc(struct rtp_receiver **rxp,
 	rx->arg    = arg;
 	rx->pseq   = -1;
 	rx->pt     = -1;
+	rx->pt_tel = -1;
 
 	err  = str_dup(&rx->name, name);
 	err |= mutex_alloc(&rx->mtx);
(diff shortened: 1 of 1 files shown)


src/rtprecv.c
```diff
@@ -478,6 +478,9 @@ void rtprecv_decode(const struct sa *src, const struct rtp_header *hdr,
 		err = pass_pt_work(rx, hdr->pt, mb);
 		if (err)
 			return;
+
+		if (hdr->pt == rx->pt_tel)
+			return;
```
alfredh, line 483: does it mean that incoming TELEV skip the Jitter buffer ?

The jitter buffer is also doing re-ordering, so if the incoming TELEV is received
out of order, the events will be presented in the wrong order.

cspiel1: Good point! Then the TELEV processing should be moved to `handle_rtp()`. In case there is a jbuf this will be called after `jbuf_get()`.

I am not sure if we should do this in this PR.

cspiel1: I tried it and seems to be a not so big change. It works and should be save.

Edit: Or not? A leak check test fails.
https://github.com/baresip/baresip/actions/runs/29500271169/job/87627177596

cspiel1: Not sure what's going wrong with the last commit. I have to revert it for now. We should address this in another PR.

When we merge and squash this let's delete the commit message of the reverted commit!

cspiel1:
There is still a timeout at the mediaenc test. Not sure if this test https://github.com/baresip/baresip/actions/runs/29722179899/job/88288373351?pr=3773 has something to do with this PR.

I am not able to reproduce this on my dev host. Okay, it's a bit slow. `test_call_mediaenc (rx thread)` takes ~3 seconds on my dev host with valgrind.

Edit: I think the initialization of `pt_tel` is missing, because zero is a valid payload type.

maximilianfridrich:
Payload types can change on re-INVITES, but once `pt_tel` is initialized, it is never updated. So if the payload type of the audio stream changes to what used to be a telephone-event, then no more audio would get decoded. Probably quite unlikely, but still possible. If we want this to be as robust as possible, `pt_tel` should be updated accordingly.

cspiel1:
> Payload types can change on re-INVITES, but once `pt_tel` is initialized, it is never updated. So if the payload type of the audio stream changes to what used to be a telephone-event, then no more audio would get decoded. Probably quite unlikely, but still possible. If we want this to be as robust as possible, `pt_tel` should be updated accordingly.

Changing the codec and re-using the pt of the previous codec for telev would be very theoretical. The solution in `rtprecv_filter_pt()` would make the code somehow complicated. So let's keep it like it is!

PR #3775, opened 2026-07-16
audio: increase RTP timestamp also for underruns

cspiel1:
If the audio source misses to deliver a frame in time the tx thread
looses one RTP packet slot and has to increase the time stamp. This
avoids a jump in the skew value.

Note: We observed jumps of the skew value in the TX stream of our devices, if pulseaudio missed to send a frame. One to three audio frames are absent if the pulseaudio stream is moved from one device to another. This is finally not so dramatic if the timestamps fit to the following packets. With this fix the skew keeps low and at the receiver it will be not more than a short crack.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -894,6 +894,9 @@ static int tx_thread(void *arg)
 		}
 		else {
 			++tx->stats.aubuf_underrun;
+			mtx_lock(tx->mtx);
+			tx->ts_ext += (tx->ptime * tx->ac->crate / 1000);
+			mtx_unlock(tx->mtx);
 
 			debug("audio: thread: tx aubuf underrun"
 			      " (total %llu)\n", tx->stats.aubuf_underrun);


cspiel1:
@maximilianfridrich and me discussed if the `ts_ext` could be decremented again if `aubuf` reaches some level over its minimum configured and it was incremented earlier because of an underrun. This could avoid increasing latency by one additional ptime slot if the audio source jitters.

Currently we don't observe a jitter case with pulseaudio. This PR so far only addresses the loss of an audio frame.

What do you think about decrementing `ts_ext` later if two audio frames arrive in one ptime slot?

sreimers:
> What do you think about decrementing `ts_ext` later if two audio frames arrive in one ptime slot?

 Not sure if I understand this correctly but `ts_ext` is the base for rtp timestamp and should never decrement:

https://www.rfc-editor.org/info/rfc3550/
>  timestamp: 32 bits
      The timestamp reflects the sampling instant of the first octet in
      the RTP data packet.  The sampling instant MUST be derived from a
      clock that increments monotonically and linearly in time to allow
      synchronization and jitter calculations

cspiel1:
> Not sure if I understand this correctly but ts_ext is the base for rtp timestamp and should never decrement

I meant decrement after increment. But, for now this is not important. This PR solves the jump in skew if an audio frame is missing. At the receiver the skew jump leads to chopped up sound for 10 seconds, until the skew adjustment solves this.

With this commit the skew stays nearly at zero when the pulseaudio stream is moved from one audio source to another, e.g. from internal microphone to handset microphone.

alfredh:
If the pulseaudio source is moved from one device to another, you can reset the SSRC or the timestamp.
This will trigger the RTP receiver to reset its jitter buffer.

The audio frame structure has a timestamp field which may be used.

cspiel1:
> If the pulseaudio source is moved from one device to another, you can reset the SSRC or the timestamp. This will trigger the RTP receiver to reset its jitter buffer.
> 
> The audio frame structure has a timestamp field which may be used.

baresip doesn't know that this happens. Only the pulseaudio daemon knows this. E.g. This could be triggered with `pactl move-source-output`. It should be only an example. There are a lot reasons why an audio source could miss to send some frames.

Issue #2988, opened 2024-04-10
support webrtc data-channel


stoneinchina:
Can this library support the data channel function of webrtc

Issue #870, opened 2019-12-10
Using libre for WebRTC, datachannels support


chertov:
Hi! I am trying to use libre webrtc components in my C application. libre is awesome, i was build it for armv5 and i can run it on hisilicon chip for ip cams with 64Mb ram.
We have only 16Mb spi flash for linux and 32-256Mb ram on iot devices like ip cams.
libnice has dependencies on glib and needs additional code, libwebrtc from google needs C++ runtime and it can't take raw h264 nal stream, pion webrtc doesn't support to build with tinyGo at the moment..
I think to use libre for iot communicate via WebRTC is a good choice!
My build for armv5: libre.so (only webrtc + rtmp) 1.9Mb, librem.so 177kb, baresip 663kb without strip and compression
Maybe we can get a bit small size

I can't find datachannels in libre. Do you have any plans to support them?

alfredh:
What is the name of the Hisilicon chip that you are using ?

when compiled for OpenWRT the filesize is a bit smaller than yours:

https://openwrt.org/packages/pkgdata_lede17_1/libre
https://openwrt.org/packages/pkgdata_lede17_1/baresip

libre: 129kB
baresip: 206kB

have a look at the openwrt toolchain for how it is compiled:

https://github.com/openwrt/telephony/tree/master/net/baresip

can you share details about your build environment ?

when compiling for embedded targets you should use this flag:

`$ make RELEASE=yes`

Have a look at this WebRTC demo which is a complete
Web server with signaling and media:

https://github.com/alfredh/baresip-webrtc

Have a look at this comparison table:

https://github.com/alfredh/baresip/wiki/WebRTC-features

Data-channels is not supported, patch to baresip is welcome.

nonibytes:
Having data channels would be great as that would make it possible to use libre on low footprint devices as a replacement for WebRTC.

PR #3789, opened 2026-08-01
peerconn: add WebRTC data channels support

obinnaokechukwu:
## Summary

Add WebRTC data-channel support to peer connections and SIP calls, including
data-only and media-plus-data BUNDLE sessions. Data channels can be negotiated
in an initial offer/answer or added to an established call by re-INVITE.

This PR depends on the libre transport and protocol support in
[baresip/re#1603](https://github.com/baresip/re/pull/1603).

Related requests: [#870](https://github.com/baresip/baresip/issues/870), [#2988](https://github.com/baresip/baresip/issues/2988)

## Motivation

Data channels have been requested as the remaining WebRTC capability needed by
small cameras, IoT devices, and applications that want message transport
without taking on a larger WebRTC runtime. Baresip already has ICE, DTLS-SRTP,
BUNDLE, and a peer-connection API; this change completes that path with an
optional, message-oriented channel API while preserving existing SIP and RTP
behavior.

## Changes

- Add public APIs for creating, accepting, observing, sending on, and closing
  data channels.
- Introduce shared ICE and DTLS transport generations for BUNDLE. A new
  generation is prepared separately and then published atomically, or rolled
  back without disturbing the active media path.
- Support data-only and media-plus-data peer connections, DCEP and negotiated
  channels, text and binary messages, ordered and partial-reliable delivery,
  backpressure, and clean stream reset.
- Negotiate the application m-line in SIP calls during initial offer/answer
  and re-INVITE, including rejection and rollback without corrupting stable
  audio or video state.
- Add an opt-in relay-only ICE policy to the WebRTC demonstration server for
  deployments that require the selected path to traverse TURN.

## Compatibility

Data channels are built only when the compatible libre data-channel companion
is available. Disabled builds retain stub APIs with explicit `ENOTSUP`
behavior and do not require usrsctp.

RTP-only calls keep their existing transport path. In particular, non-muxed
RTP calls do not acquire a data transport, and rejecting an application m-line
does not reject otherwise valid media.

This adds public data-channel and media-transport APIs and therefore advances
the baresip ABI SOVERSION once, from 28 to 29.

## Verification

- Complete baresip tests with data channels enabled and disabled, including
  main-loop and receive-thread modes
- Shared and static builds and static dependency-closure checks
- Peer-connection and SIP-call tests for data-only, media-plus-data, late
  addition, rejection, rollback, restart, message ordering, partial
  reliability, backpressure, and callback-time close/destruction
- Existing non-muxed RTP, ICE, DTLS-SRTP, BUNDLE, call, audio, and video
  regression coverage
- ASan/UBSan lifecycle runs

Repository-local unit and loopback tests cannot prove that baresip interoperates
with independent WebRTC implementations. The compatibility suite at
[obinnaokechukwu/baresip-datachannel-compliance](https://github.com/obinnaokechukwu/baresip-datachannel-compliance)
therefore communicates with Chrome/Chromium and Firefox, aiortc in Python, and
Pion in Go. It exercises data-only and media-plus-data sessions, TURN relay
selection, deterministic message manifests, and impaired network paths through
baresip's public signaling surface and real wire protocols.

The suite lives in a separate repository so it cannot depend on baresip or
libre private APIs, fixtures, or build state. It records exact source and peer
versions, runtime artifacts, commands, SDP, manifests, statistics, and logs so
the compatibility results can be reproduced and diagnosed.


obinnaokechukwu:
Hi @sreimers, can you let me know the reason for closing the pr?

PR #3800, opened 2026-08-10
Feature: Send Custom X-Headers on Outbound Calls

ajrice6713:
Thank you for considering this feature. We use Baresip for testing internal sip proxies, our proxies expect certain static `x-` headers from calls across different scenarios. This allows us to create an account per scenario, essentially and pass those headers to the proxy so that we can test end to end call flows via the CLI. 

Please let me know if you have any questions/feedback! Happy to make any necessary changes. 

## Summary

Adds support for configuring static custom SIP headers per account via a new `x_headers` addr-param. Headers are applied to every outbound call made from that account.

### Format

```
# ~/.baresip/accounts

;x_headers=Header-Name:value,Other-Header:value
```

Multiple headers are comma-separated; name and value are separated by `:`.

### Example

```
# ~/.baresip/accounts

<sip:user@example.com>;auth_pass=secret;x_headers=X-My-Header:foo,X-Other:bar
```

### With `/uaaddheader`

Account-level headers (`x_headers`) are applied first. Runtime headers added via `/uaaddheader` are appended on top without overwriting them, so both sets are sent. This allows the new per-account headers to coexist with headers set at runtime.

### Changes

- `src/core.h` — adds `struct list custom_hdrs` to `struct account`; declares `call_append_custom_hdrs`
- `src/account.c` — parses `x_headers` param in `account_alloc` using `csl_parse` (consistent with `audio_codecs` parsing); flushes the list in the destructor; prints in `account_debug`; exposes `account_custom_hdrs()` accessor
- `src/call.c` — adds `call_append_custom_hdrs`, a non-flushing variant of `call_set_custom_hdrs` for merging header lists
- `src/ua.c` — `ua_call_alloc` applies account headers first, then appends UA-level runtime headers
- `docs/examples/accounts` — documents the new param and adds a usage example
- `test/account.c` — verifies `x_headers` is parsed correctly from account string
- `test/call.c` — `test_call_account_custom_headers` verifies headers appear in the SIP INVITE on the wire

### DIFF
--- a/src/ua.c
+++ b/src/ua.c
@@ -958,8 +958,11 @@ int ua_call_alloc(struct call **callp, struct ua *ua,
 	if (err)
 		return err;
 
+	if (!list_isempty(&ua->acc->custom_hdrs))
+		call_set_custom_hdrs(*callp, &ua->acc->custom_hdrs);
+
 	if (!list_isempty(&ua->custom_hdrs))
-		call_set_custom_hdrs(*callp, &ua->custom_hdrs);
+		call_append_custom_hdrs(*callp, &ua->custom_hdrs);
 
 	call_set_handlers(*callp, NULL, call_dtmf_handler, ua);
 
--- a/src/ua.c
+++ b/src/ua.c
@@ -958,11 +958,8 @@ int ua_call_alloc(struct call **callp, struct ua *ua,
 	if (err)
 		return err;
 
-	if (!list_isempty(&ua->acc->custom_hdrs))
-		call_set_custom_hdrs(*callp, &ua->acc->custom_hdrs);
-
 	if (!list_isempty(&ua->custom_hdrs))
-		call_append_custom_hdrs(*callp, &ua->custom_hdrs);
+		call_set_custom_hdrs(*callp, &ua->custom_hdrs);
 
 	call_set_handlers(*callp, NULL, call_dtmf_handler, ua);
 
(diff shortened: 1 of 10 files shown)


noahg1 (approved):

alfredh:
There is already code for custom headers for both register and calling, for example:

```c
/** Defines a SIP User Agent object */
struct ua {
	MAGIC_DECL                   /**< Magic number for struct ua         */
	struct le le;                /**< Linked list element                */
	struct account *acc;         /**< Account Parameters                 */
	struct list regl;            /**< List of Register clients           */
	struct list calls;           /**< List of active calls (struct call) */
	struct pl extensionv[8];     /**< Vector of SIP extensions           */
	size_t    extensionc;        /**< Number of SIP extensions           */
	char *cuser;                 /**< SIP Contact username               */
	char *pub_gruu;              /**< SIP Public GRUU                    */
	enum presence_status pstat;  /**< Presence Status                    */
	struct list hdr_filter;      /**< Filter for incoming headers        */
	struct list custom_hdrs;     /**< List of outgoing headers           */
	char *ansval;                /**< SIP auto answer value              */
	struct sa dst;               /**< Current destination address        */
};
```


is there something special about this PR to add another way to handle custom headers,
or can the existing code be reused ?

ajrice6713:
Hey @alfredh - thanks for looking! 

yes - i missed that. I think i could remove `custom_hdrs()` from `account` as well as `account_custom_hdrs` and simply copy account headers into `ua->custom_hdrs` during `ua_alloc` - If that sounds good, I will update the PR!

alfredh:
The `struct account` should be matching what is in the `~/.baresip/accounts` file.

It should probably be enough to keep a list of custom headers either in `struct ua`
or `struct account`, while the latter is slightly more correct.

ajrice6713:
Hey @alfredh - please see the latest commit addressing the PR feedback.

src/ua.c
```diff
@@ -1201,6 +1201,23 @@ static int ua_cuser_gen(struct ua *ua)
...
+	struct le *le;
+
+	LIST_FOREACH(&ua->acc->custom_hdrs, le) {
+		const struct sip_hdr *h = le->data;
+		char *buf = NULL;
+
+		if (re_sdprintf(&buf, "%r", &h->name))
+			return;
+
+		custom_hdrs_add(&ua->custom_hdrs, buf, "%r", &h->val);
```
cspiel1: Still a redundant list in `struct ua`. Isn't it enough to have the list in `struct account` like @alfredh suggests?

ajrice6713: Good catch 😓 Thanks for the patience - updating now.

cspiel1: Looks good now.

PR #3806, opened 2026-08-23
echo: refactor call_event_handler closed with bevent

sreimers:
Avoids overriding `call_event_handler`, as this interferes with other modules (like `menu`) and prevents `BEVENT_CLOSED` from being handled correctly.

### DIFF
--- a/modules/echo/echo.c
+++ b/modules/echo/echo.c
@@ -32,26 +32,6 @@ static void destructor(void *arg)
 }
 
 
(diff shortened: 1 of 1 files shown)


alfredh:
I am planning to release v4.11.0 on Tuesday 25. Aug

Should we include this fix in the release ?

sreimers:
It's a last minute change but fixes a use after free crash within menu, so I think we should include this.

alfredh:
thanks for the fix !

PR #3809, opened 2026-09-01
jbuf: safer jbuf_set_id

cspiel1:
Increasing the reference counter of a struct pl that lies on the stack lead
to an illegal memory access.

```c
	char id[32];
	re_snprintf(id, sizeof(id), "mcreceiver-%d", prio);
	struct pl pl_id = PL_INIT;
	pl_set_str(&pl_id, id);
	jbuf_set_id(mcreceiver->jbuf, &pl_id);
```

Now the setter accepts a const char string
which is safer and simplifies also the application code

```c
	char id[32];
	re_snprintf(id, sizeof(id), "mcreceiver-%d", prio);
	err = jbuf_set_id(mcreceiver->jbuf, id);
	if (err)
		goto out;
```

### DIFF
--- a/src/jbuf.c
+++ b/src/jbuf.c
@@ -60,7 +60,7 @@ struct packet {
  * sequence number.
  */
 struct jbuf {
-	struct pl *id;       /**< Jitter buffer Identifier                   */
+	char *id;            /**< Jitter buffer Identifier                   */
 	struct rtp_sock *gnack_rtp; /**< Generic NACK RTP Socket             */
 	struct list pooll;   /**< List of free packets in pool               */
 	struct list packetl; /**< List of buffered packets                   */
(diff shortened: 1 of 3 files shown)


src/jbuf.c
```diff
@@ -266,14 +266,16 @@ void jbuf_set_srate(struct jbuf *jb, uint32_t srate)
...
+int jbuf_set_id(struct jbuf *jb, const char *id)
 {
 	if (!jb)
-		return;
+		return EINVAL;
 
+	int err;
 	mtx_lock(jb->lock);
-	jb->id = mem_ref(id);
+	err = str_dup(&jb->id, id);
```
sreimers: This breaks the `re_trace_event` API, `pl_alloc_str` could be a option.

PR #3813, opened 2026-09-03
audio: serialize encoder configuration in audio_encoder_set

allesrebel:
This code deals with a race that gets triggered by a mid-call codec switch.
After a mid-call codec switch, the TX thread keeps running while the main thread rewrites the fields it reads. The locks ensure this switch is safe.

During this codec switch, the following audio_start then re-enters the source-allocation block and writes psize, ausrc_prm, and aubuf_maxsz

Meanwhile tx_thread reads psize under the lock at src/audio.c:877, and poll_aubuf_tx reads psize and ausrc_prm under the lock at src/audio.c:427. 

Note: buils on this PR https://github.com/baresip/baresip/pull/3810

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -1314,13 +1314,12 @@ bool audio_started(const struct audio *a)
 int audio_encoder_set(struct audio *a, const struct aucodec *ac,
 		      int pt_tx, const char *params)
 {
-	struct autx *tx;
 	int err = 0;
 
 	if (!a || !ac)
 		return EINVAL;
 
-	tx = &a->tx;
+	struct autx *tx = &a->tx;
 
 	if (ac != tx->ac) {
 		info("audio: Set audio encoder: %s %uHz %dch\n",
@@ -1333,8 +1332,19 @@ int audio_encoder_set(struct audio *a, const struct aucodec *ac,
 			aubuf_flush(tx->aubuf);
 		}
 
+		mtx_lock(tx->mtx);
 		tx->enc = mem_deref(tx->enc);
 		tx->ac = ac;
+		if (ac->encupdh) {
+			struct auenc_param prm = { .bitrate = 0 };
+
+			err = ac->encupdh(&tx->enc, ac, &prm, params);
+		}
+		mtx_unlock(tx->mtx);
+		if (err) {
+			warning("audio: alloc encoder: %m\n", err);
+			return err;
+		}
 
 		if (!list_isempty(baresip_aufiltl())) {
 			err = aufilt_setup(a, baresip_aufiltl());
@@ -1342,13 +1352,12 @@ int audio_encoder_set(struct audio *a, const struct aucodec *ac,
 				return err;
 		}
 	}
+	else if (ac->encupdh) {
+		struct auenc_param prm = { .bitrate = 0 };
 
-	if (ac->encupdh) {
-		struct auenc_param prm;
-
-		prm.bitrate = 0;        /* auto */
-
+		mtx_lock(tx->mtx);
 		err = ac->encupdh(&tx->enc, ac, &prm, params);
+		mtx_unlock(tx->mtx);
 		if (err) {
 			warning("audio: alloc encoder: %m\n", err);
 			return err;


src/audio.c
```diff
@@ -1334,7 +1336,9 @@ int audio_encoder_set(struct audio *a, const struct aucodec *ac,
 		}
 
 		tx->enc = mem_deref(tx->enc);
+		mtx_lock(tx->mtx);
 		tx->ac = ac;
+		mtx_unlock(tx->mtx);
```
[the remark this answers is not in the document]
allesrebel: A fair point - added commit a1a2afb7b98511ca8c194912501e4d6aa51372f1
To make TX thread always see a consistent (ac, enc) pair

src/audio.c
```diff
@@ -344,8 +344,10 @@ static void encode_rtp_send(struct audio *a, struct autx *tx,
 
 	len = mbuf_get_space(tx->mb);
 
+	mtx_lock(tx->mtx);
 	err = tx->ac->ench(tx->enc, &marker, mbuf_buf(tx->mb), &len,
 			   af->fmt, af->sampv, af->sampc);
+	mtx_unlock(tx->mtx);
```
alfredh, line 352: the encode function could potentially take a long time to execute, or indirectly
call back to the audio structure, potentially causing a deadlock.

is it better to make a stack copy of the needed variables (protected with mutex),
and then use those variables to call `ench` ?

allesrebel: Agreed! good catch

We'll grab all the information while protected, then use with ench ( in fd5f7fb03259b83f7e423b6ece7530538a64d776 )

Regarding the mem_ref - AI flagged that it could be a use-after-free since codec changes cause main thread to free via audio_encoder_set. It's possible that in the TX thread, we already loaded the pointer and will use it.

src/audio.c
```diff
@@ -1066,13 +1068,15 @@ static int start_source(struct autx *tx, struct audio *a, struct list *ausrcl)
...
-
 		sz = aufmt_sample_size(tx->src_fmt);
 
 		psize_alloc = sz * au_calc_nsamp(prm.srate, prm.ch, prm.ptime);
+
+		mtx_lock(tx->mtx);
+		tx->ausrc_prm = prm;
 		tx->psize = psize_alloc;
 		tx->aubuf_maxsz = tx->psize * 30;
+		mtx_unlock(tx->mtx);
```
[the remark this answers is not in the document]
allesrebel: This is correct 
ausrc_read_handler() will snapshot aubuf_maxsz inside the lock section that already snapshots psize, and the overrun check uses the local copy (75e5e6b7)

src/audio.c
```diff
@@ -1333,8 +1337,10 @@ int audio_encoder_set(struct audio *a, const struct aucodec *ac,
 			aubuf_flush(tx->aubuf);
 		}
 
+		mtx_lock(tx->mtx);
 		tx->enc = mem_deref(tx->enc);
 		tx->ac = ac;
+		mtx_unlock(tx->mtx);
```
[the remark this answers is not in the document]
allesrebel: Addressed in 2e7b5920.
encode_rtp_send now snapshots the pair in a lock

src/audio.c
```diff
@@ -1484,7 +1501,9 @@ void audio_mute(struct audio *a, bool muted)
 	if (!a)
 		return;
 
+	mtx_lock(a->tx.mtx);
 	a->tx.muted = muted;
+	mtx_unlock(a->tx.mtx);
```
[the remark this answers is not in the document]
allesrebel: I believe these calls occur on the main thread - meaning there's no race unless there's a way to call mute outside the thread (maybe via library call?).

It's only 2 lines, but there maybe other spots

src/audio.c
```diff
@@ -296,6 +296,8 @@ static void encode_rtp_send(struct audio *a, struct autx *tx,
 	size_t ext_len = 0;
 	uint32_t ts_delta = 0;
 	bool marker = tx->marker;
+	const struct aucodec *ac;
+	struct auenc_state *enc;
```
alfredh, line 300: could these two lines be moved down to the next block, so that
the operation is declaration + assignment ?

src/audio.c
```diff
@@ -344,8 +346,14 @@ static void encode_rtp_send(struct audio *a, struct autx *tx,
 
 	len = mbuf_get_space(tx->mb);
 
-	err = tx->ac->ench(tx->enc, &marker, mbuf_buf(tx->mb), &len,
-			   af->fmt, af->sampv, af->sampc);
+	mtx_lock(tx->mtx);
+	ac  = tx->ac;
+	enc = mem_ref(tx->enc);
```
alfredh, line 351: I dont think mem_ref() is needed here, should be sufficient with
a simple assignment `enc = tx->enc` ...

src/audio.c
alfredh: I dont think we should fix things until it becomes a problem.
We have good run time checks with Sanitizer and valgrind that will catch
these types of errors.

The AI seems to be slightly pessimistic and reporting problems that are not real problems...

src/audio.c
allesrebel: No problem, moved in 85e19ac8d2139b68c7bfa517ff53950128bd1884

src/audio.c
allesrebel: I agree with your approach - I'll let AI run loose and see if it's a real problem with a valid use case first (will PR, if it makes sense)

Dropped the mem_ref in 85e19ac8d2139b68c7bfa517ff53950128bd1884

src/audio.c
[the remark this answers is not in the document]
allesrebel: Not going to 'harden' anything else until the use cases are ones that matter / tested for

src/audio.c
allesrebel: Yeah, very pessimistic - I'll do separate PR if there's a valid and useful application of baresip that exercises the issues AI flagged. I'll stay away from hardening for AI's sake

alfredh:
thanks for your efforts. It looks like the PR is now going in the right direction.

If possible, it would also be nice to split up the PR into more logical blocks.
Perhaps this is a pragmatic approach, but the PR is touching very sensitive code
that is used by many people.

src/audio.c
```diff
@@ -344,8 +344,13 @@ static void encode_rtp_send(struct audio *a, struct autx *tx,
 
 	len = mbuf_get_space(tx->mb);
 
-	err = tx->ac->ench(tx->enc, &marker, mbuf_buf(tx->mb), &len,
-			   af->fmt, af->sampv, af->sampc);
+	mtx_lock(tx->mtx);
+	const struct aucodec *ac = tx->ac;
+	struct auenc_state *enc  = tx->enc;
+	mtx_unlock(tx->mtx);
```
alfredh: I think my previous comment was wrong, apologies for that.

When the `ac` pointer is copied from `tx->ac` there is a chance that this pointer can be
deleted just after the mutex is unlocked.

Copying values (e.g. int, char, ...) is thread safe, but copying pointers is not thread-safe.

Here is a safe paradigm that we have used in the past:

https://github.com/baresip/baresip/blob/main/modules/audiounit/player.c#L60

src/audio.c
```diff
@@ -530,20 +535,21 @@ static void ausrc_read_handler(struct auframe *af, void *arg)
...
 	bool muted = tx->muted;
 	size_t psize = tx->psize;
+	size_t aubuf_maxsz = tx->aubuf_maxsz;
 	mtx_unlock(tx->mtx);
 
 	if (fmt != af->fmt) {
 		warning("audio: ausrc format mismatch:"
 			" expected=%d(%s), actual=%d(%s)\n",
-			fmt, aufmt_name(tx->src_fmt),
+			fmt, aufmt_name(fmt),
```
alfredh: this change is unrelated, but good to have. Could you make a separate PR for this ?

src/audio.c
alfredh: consider also to make a separate PR for this lock/unlock change

src/audio.c
allesrebel: Absolutely

src/audio.c
allesrebel: Sure - no problem

src/audio.c
allesrebel: No worries at all, thank you for the feedback!

Yeah, I'm with you on the copying and using values being thread safe. 
I'll add in the guards as proposed by player.c to check the pointers before using them

PR #3814, opened 2026-09-04
vidsrc: remove vidsrc_alloc() -- unused

alfredh:


sreimers:
This is used by vidloop https://github.com/baresip/baresip-apps/blob/main/modules/vidloop/vidloop.c#L672

Issue #3788, opened 2026-07-29
Aureceiver: post decoder filters overwrite the decoder sample format


SnowMaster12:
## Description
When `audec_format` differs from `auplay_format` and the `auconv` module is enabled, the decoder sample format stored in `audio_recv` is overwritten with the post-filter playback format.

With G.711, the first PCMU frame is decoded as `S16LE` and then converted to `FLOAT` by `auconv`. After the converted frame is pushed to the audio buffer, `aurecv_push_aubuf()` assigns the resulting `FLOAT` format back to `ar->fmt.`

The following PCMU packets are therefore passed to the G.711 decoder requesting `FLOAT` output. Since the G.711 decoder only supports `S16LE`, it returns `ENOTSUP`.

##Environment
- Baresip: 4.4.0
- Platform: iOS
- Audio module: `audiounit`
- Audio codec: PCMU

## Actual behavior

After receiving PCMU audio, the following warning is repeatedly generated:
audio_recv: PCMU codec decode 160 bytes: Operation not supported [45]
Received audio is consequently distorted or discarded.

## Root cause
`ar->fmt `is defined and initialized as the decoder sample format:

enum aufmt fmt;
ar->fmt = cfg->dec_fmt;

It is then passed to every decoder invocation:

```C
err = ac->dech(
        ar->dec,
        ar->fmt,
        ar->sampv,
        &sampc,
        marker,
        mbuf_buf(mb),
        mbuf_get_left(mb)
);
```
However, after the decoding filters process the frame, aurecv_push_aubuf() overwrites it:

```C
ar->srate = af->srate;
ar->ch    = af->ch;
ar->fmt   = af->fmt;
```
The auconv decoding filter changes af->fmt from S16LE to the configured playback format, which is FLOAT in this case.

Therefore, after processing the first frame:

ar->fmt: S16LE → FLOAT

The next G.711 decode operation receives AUFMT_FLOAT and returns ENOTSUP.

## Expected behavior
The decoder format stored in `ar->fmt` should remain equal to `cfg->dec_fmt` for the entire lifetime of the decoder.
The post-filter format should only be used for the audio buffer, playback and latency calculations.

sreimers:
> * Baresip: 4.4.0

This version is quite old, please test with latest release/main.

SnowMaster12:
> > * Baresip: 4.4.0
> 
> This version is quite old, please test with latest release/main.

hi @sreimers 
[...]

PR #3817, opened 2026-09-09
auresamp: allocate output buffer before format conversion

SnowMaster12:
## **Description**

Allocate `sampv` before converting resampled S16LE audio to a non-S16LE output format.

When the input format is already `AUFMT_S16LE`, `sampv` is not allocated by the input conversion path. If the configured output format differs from `AUFMT_S16LE`, `auconv_from_s16()` can therefore be called with a NULL destination buffer.

Ensure that sampv is allocated before performing the output format conversion.

Additionally, keep the decoder sample format in `aureceiver` unchanged after processing the receive audio filters.

Previously, the filtered frame format was assigned back to `ar->fmt`. Since `ar->fmt` is also used as the format passed to the decoder, a filter changing the frame format (for example from `S16LE` to `FLOAT`) caused subsequent decode operations to use the playback format instead of the configured decoder format.

The filtered frame format is now used only where appropriate for the audio buffer calculations, while `ar->fmt` remains the decoder format.

This also prevents incorrect decoder buffer sizing and the resulting memory corruption when the decoder and playback formats have different sample sizes.

Related to #3788.


alfredh:
hi @cspiel1 could you please take a look at this one ?

src/aureceiver.c
```diff
@@ -195,9 +195,8 @@ static int aurecv_push_aubuf(struct audio_recv *ar, const struct auframe *af)
 
 	ar->srate = af->srate;
 	ar->ch    = af->ch;
```
cspiel1, line 197: Should we also avoid writing back `srate` and `ch`?

src/aureceiver.c
```diff
@@ -195,9 +195,8 @@ static int aurecv_push_aubuf(struct audio_recv *ar, const struct auframe *af)
...
-	ar->fmt   = af->fmt;
 
-	bpms = (uint64_t)ar->srate * ar->ch * aufmt_sample_size(ar->fmt) /
+	bpms = (uint64_t)ar->srate * ar->ch * aufmt_sample_size(af->fmt) /
 	       1000;
```
cspiel1, line 200: Not related to this PR. But, I just saw that some fields of `ar->stats` are only valid for the DEBUG build and others are used for RELEASE and DEBUG. This is only a design weakness.

src/aureceiver.c
cspiel1: Looked closer:
- `ar->srate` and `ar->ch` are used to compute the latency in the `ar->aubuf`. They got the wrong title (or Doxygen). Let's drop them! Instead the `ar->auplay_prm` fields can be used.
- Right the `ar->fmt`, the "Decoder sample format" should not be changed.
- Let's also update these Doxygen lines and move them up to the other "Decoder ..." fields!
```c
	void *sampv;                  /**< Decoder sample buffer              */
	size_t sampvsz;               /**< Decoder sample buffer size         */
```

src/aureceiver.c
alfredh: it would be nice if the code matches the structure members:

```c
#ifndef RELEASE
	int32_t d, da;
	uint64_t t;
	t = tmr_jiffies_usec();
	if (ar->t) {
		d = (int32_t) (int64_t) ((t - ar->t) - ar->ptime);
		da = abs(d);
		ar->stats.dmax = max(ar->stats.dmax, da);
		ar->stats.jitter += (da - ar->stats.jitter) / JITTER_EMA_COEFF;
	}

	ar->t = t;
#endif
```

in this case the structure members used in debug build could also
be put under an `#ifndef RELEASE` check.

cspiel1: PR for that: https://github.com/baresip/baresip/pull/3826

src/aureceiver.c
cspiel1: PR for that: https://github.com/baresip/baresip/pull/3827
Note, it includes already the `fmt` fix.

After that this PR can be rebased and should contain only the `modules/auresamp` changes.

cspiel1: In the meanwhile also head issues with the selftests and opened this PR https://github.com/baresip/baresip/pull/3837 realizing the relation to this one just a minute ago. Thank's!

sreimers:
replaced by https://github.com/baresip/baresip/pull/3837

PR #3822, opened 2026-09-09
audio: thread-safe TX source parameters and aubuf_maxsz

allesrebel:
Split from https://github.com/baresip/baresip/pull/3813

Adds concurrency guards to another set of parameters for later ptime calculations. 

tx->ausrc_prm, tx->psize, and tx->aubuf_maxsz with tx->mtx in start_source(), and snapshot tx->aubuf_maxsz under tx->mtx in ausrc_read_handler() to prevent data races and inconsistent buffer limits.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -530,6 +530,7 @@ static void ausrc_read_handler(struct auframe *af, void *arg)
 	enum aufmt fmt = tx->src_fmt;
 	bool muted = tx->muted;
 	size_t psize = tx->psize;
+	size_t aubuf_maxsz = tx->aubuf_maxsz;
 	mtx_unlock(tx->mtx);
 
 	if (fmt != af->fmt) {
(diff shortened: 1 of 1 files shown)


sreimers:
Unsure about these mutex changes, is it possible to stop audio tx thread and restart after change? Changing codecs, ptime or params while running is a really rare case and should maybe avoided.

alfredh:
it should be possible to stop and restart the transmit thread, if the encoder is changed:

```c
static void stop_tx(struct autx *tx, struct audio *a)
{
	if (!tx || !a)
		return;

	stream_enable_tx(a->strm, false);
	if (re_atomic_rlx(&tx->thr.run)) {
		re_atomic_rlx_set(&tx->thr.run, false);
		thrd_join(tx->thr.tid, NULL);
	}

	/* audio source must be stopped first */
	tx->ausrc = mem_deref(tx->ausrc);
	tx->aubuf = mem_deref(tx->aubuf);

	list_flush(&tx->filtl);
}
```

if a re-INVITE is sent/received and leading to codec change, many other parameters such as jitter buffer,
stats etc. has to be reset.

allesrebel:
Ah that seems like a better approach, it gets rid of a bunch of guarding
I'll give it a shot with the ptime changes and report back

PR #3823, opened 2026-09-09
audio: snapshot encoder state in encode_rtp_send

allesrebel:
Split from https://github.com/baresip/baresip/pull/3813

Handles guarding of:
tx->ausrc_prm, tx->psize, tx->aubuf_maxsz

Which are used in a draft PR for ptime behavior changes


alfredh:
as mentioned in another PR, it would be better to stop and start the transmit thread,
if the encoder is changed.

In this case the transmit thread can run without checking any locks.

PR #3824, opened 2026-09-10
rtprecv: fix teardown race and send-only RTCP

tomioe:
### Description change:

1. Keep callback dispatch in threaded mode until the receive thread has been joined. This prevents late callbacks from running inline while the receiver is being destroyed.

1. Continue processing RTCP feedback when RTP reception is disabled so send-only streams can react to control messages such as PLI.

Building the project with this patch yields no additional warnings, and `ccheck` emits:

```
$ ./test/ccheck.py src/rtprecv.c
Statistics:
~~~~~~~~~~~
Number of files processed:     c: 1  cpp: 0  h: 0  mk: 0  m4: 0  py: 0  m: 0  s: 0  java: 0  php: 0
Number of lines with errors:   0
```

Full test suite passes as well. 

### Note

I know very little about this project and it's internal structure. However, I have been using it as the workhorse in a headless SIP client that was AI-generated.

After deploying this headless client in my staging environment, I noticed that video streams were not working properly. My SIP server was issuing PLI packets that seemed to go nowhere. The GUI SIP Client I was using before (LinPhone) seemed to work fine. A bit of back and forth, and `gpt-5.6-sol-max` produced this patch. Tested with an H.264 send-only call endpoint. Before this change, three received PLIs produced no new IDR frames. Afterward, all three produced SPS/PPS/IDR recovery units on the next encoded frame.  If you think this patch is AI-garbage, please close it.


sreimers:
Can you try this patch:

```c
diff --git a/src/rtprecv.c b/src/rtprecv.c
index c685d092..681bff4a 100644
--- a/src/rtprecv.c
+++ b/src/rtprecv.c
@@ -503,11 +503,6 @@ void rtprecv_handle_rtcp(const struct sa *src, struct rtcp_msg *msg,

        MAGIC_CHECK(rx);
        mtx_lock(rx->mtx);
-       if (!rx->enabled) {
-               mtx_unlock(rx->mtx);
-               return;
-       }
-
        rx->ts_last = tmr_jiffies();
        mtx_unlock(rx->mtx);
```

I think it should enough to remove the enabled check for rtcp handling.

tomioe:
Works wonderfully, the result is the same and your solution is more elegant. I tested it by first reverting the change in `rtprecv.c`'s `rtprecv_handle_rtcp` back to the original upstream (3ebb22), observing PLI didn't yield any keyframe information. Then implemented your change, ran `ccheck` and full test-suite. Testing against my headless client's H.264 suite showed that the 3/3 PLIs produced ordered SPS/PPS/IDR responses in 67.222, 66.966, and 66.843 ms. More detailed results here:
[validation.md](https://github.com/user-attachments/files/32074664/validation.md)
[baseline-media.json](https://github.com/user-attachments/files/32074696/baseline-media.json)
[suggestion-media.json](https://github.com/user-attachments/files/32074703/suggestion-media.json)



Can create the commit for your change in this PR if needed. 

Also, thanks for this - I was a bit nervous about submitting this PR to begin with, since it embodies what many would consider "AI Slop". But the change seemed simple and beneficial enough...

cspiel1:
You can close this PR and open a new one which references this one.

PR #3825, opened 2026-09-11
rtprecv: tiny cleanup for startup

cspiel1:
- **rtprecv: rtp_estab tiny cleanup**
- **stream: avoid decode in main thread if rx thread is activated**


alfredh:
there are some errors from the testcode ...

could you also rebase to main HEAD please ?

PR #3826, opened 2026-09-11
aureceiver: stats cleanup

cspiel1:


cspiel1:
MAC ci fails with
```
Run brew install codec2 \
Warning: No available formula with the name "fdk-aac".
==> Searching for similarly named formulae and casks...
Error: No formulae or casks found for fdk-aac.
```

alfredh:
it looks like the package `fdk-aac` has been removed from Homebrew ...
Could you try to remove it from the CI scripts ?

PR #3827, opened 2026-09-11
aureceiver: fix decoder sample format and cleanup

cspiel1:
- Sample format of decoder and auplay can differ because a filter can
  convert it. This commit fixes the decoders sample format if the audio
  play sample format differs.
- This commit also includes cleanup of struct audio_recv that should
  avoid redundancy and confusions where fields belong to.


alfredh:
please rebase to main HEAD to get the CI fixes ...

alfredh:
it looks like the test is not working for some reason ...

cspiel1:
> it looks like the test is not working for some reason ...

Yes, I found a bug in the `auresamp` module. The fix follows soon. I'll check also the tests.

cspiel1:
Separated the auresamp fix: https://github.com/baresip/baresip/pull/3837

cspiel1:
> > it looks like the test is not working for some reason ...
> 
> Yes, I found a bug in the `auresamp` module. The fix follows soon. I'll check also the tests.

To be correct. I encountered the same bug like earlier @SnowMaster12 but didn't realize the connection. https://github.com/baresip/baresip/pull/3817

alfredh:
Thanks for the fix -- this patch looks good to me :)

SnowMaster12:
> > > it looks like the test is not working for some reason ...
> > 
> > 
> > Yes, I found a bug in the `auresamp` module. The fix follows soon. I'll check also the tests.
> 
> To be correct. I encountered the same bug like earlier @SnowMaster12 but didn't realize the connection. #3817

Thanks for the mention, and sorry for not replying earlier, i wasn't able to.

This is the same issue i was addressing in #3817.
I had included both changes there because the `aureceiver` format handling and the `auresamp` buffer allocation issue were closely related and part of the same audio format conversion path.

Glad to see the `auresamp` fix has now been merged through #3837, while the `aureceiver` format handling is covered by this PR.

PR #3832, opened 2026-09-13
cmake: use variable for libdir libbaresip.pc.in

### LINKED ISSUE #3830: packaging/libbaresip.pc.in: Hardcoded libdir
Hello. Gentoo requires the libdir in the .pc file to reflect the correct libdir which is lib64 and not lib. https://bugs.gentoo.org/show_bug.cgi?id=981529

Is it possible to update packaging/libbaresip.pc.in to use lib/lib64 appropriately?
I tested using `libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@` in libbaresip.pc.in which seems to work but this is not something I am familiar with. 

Thank you.

fictitiousexistence:
Use @CMAKE_INSTALL_LIBDIR@ instead of hardcoding lib.   

Closes #3830


sreimers:
Thanks!

